Issue #17, opened 2021-03-07
Support multiple candidate selection


bdarcus:
In `helm-bibtex`, when presented with a list of candidates, one can do the following:

- hit `control-space` to mark multiple candidates
- hit tab to access action options, or return to run default

![Screenshot from 2021-03-07 02-54-08](https://user-images.githubusercontent.com/1134/110233233-a62ab680-7ef0-11eb-8e4e-883fc976c923.png)

In my view, this is an almost ideal UX for selecting and acting on multiple candidates (though arguably the selected candidates should then be grouped at the top of the list).

That will then pass multiple keys to the relevant functions, and for the insert-at-point commands, those functions take care of the formatting; in some cases they are comma-separated inline, and in others (like inserting BibTeX entries), they are separated by newlines.

So how to do this here?

The obvious answer is to change the read function to use `completing-read-multiple`.

But the selectrum UI support for multiple selection is not really optimized for examples like this with very long candidates. Same for icomplete-vertical, which uses a similar UI.

I did, however, submit an [enhancement request](https://github.com/raxod502/selectrum/issues/489) to selectrum to support this case. No code yet, but I'm optimistic we will see it. 

So it would be good to resolve this on our end before this happens, so we can test and contribute feedback.

Seems we would also need to get `embark` to support multiple (see [comment](https://github.com/tmalsburg/helm-bibtex/pull/355#issuecomment-784258088)).

PR #18, opened 2021-03-07
Change to completing-read-multiple

bdarcus:
The bibtex-completion functions all take multiple keys as input, and
some of them (notably the "input" ones) lose functionality without
support for this.

Hence, this changes the core read function to use
'completing-read-multiple'.

In order to do that, it also sets the `crm-separator` to ampersand, since 
default comma won't work for these data.

Addresses #17.

see https://github.com/oantolin/embark/issues/176#issuecomment-794597585

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -67,19 +67,22 @@
   (bibtex-actions-read-backend))
 
 (defun bibtex-actions--completing-read ()
-  "Read bibtex-completion entries for completion using 'completing-read'."
+  "Read bibtex-completion entries for completion using 'completing-read-multiple'."
+  ;; Note: the candidate string we use is very long, which is a bit awkward when
+  ;; dealing with TAB completion style multi selection interfaces."
   (bibtex-completion-init)
-  (when-let ((candidates (bibtex-actions--get-candidates))
+  (when-let ((crm-separator "\\s-*&\\s-*")
+             (candidates (bibtex-actions--get-candidates))
              (chosen
-              (completing-read
+              (completing-read-multiple
                "BibTeX entries: "
                (lambda (string predicate action)
                  (if (eq action 'metadata)
                      '(metadata
-                       ;; (annotation-function . bibtex-completion--annotation)
+                       ;; TODO (annotation-function . bibtex-completion--annotation)
                        (category . bibtex))
                    (complete-with-action action candidates string predicate))))))
-    (cdr (assoc chosen candidates))))
+    (mapcar (lambda (choice) (cdr (assoc choice candidates))) chosen)))
 
 (defun bibtex-actions--get-candidates ()
   "Return all keys from 'bibtex-completion-candidates'."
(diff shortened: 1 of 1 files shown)


bibtex-actions.el
```diff
@@ -67,16 +67,17 @@
...
                "BibTeX entries: "
+               ;; FIX getting a "wrong number of arguments" error when interactive
                (lambda (string predicate action)
                  (if (eq action 'metadata)
                      '(metadata
-                       ;; (annotation-function . bibtex-completion--annotation)
+                       ;; TODO (annotation-function . bibtex-completion--annotation)
                        (category . bibtex))
                    (complete-with-action action candidates string predicate))))))
     (cdr (assoc chosen candidates))))
```
bdarcus, line 83: I think the problem is here, which is assuming a single key.

Yevgnen:
Hi, at a quick glance, without UI changes of `selectrum`, I thinks it's a bit hard to achieve (better UE). 

I see that `selectrum-completing-read-multiple` overrides `completing-read-multiple`.

I've thought maybe we could insert the bibtex keys (which are much shorter) as selected items but `selectrum` seems insert the candidate (which is the formated citation) directly with no possible transformation. 

Personally, selecting multiple items in minibuffer is not that convenient for me. I prefer to show only 7~10 items in minibuffer due to a limited display. I used to use `ivy-mark` to selecting multiple items but I rarely used it. Whether to keep the selected items in the top is also another UE issue. Operating multiple items in a separate buffer like `ivy-occur` or `embark-collect` seems much better for me. I see that `helm-bibtex` seems use a large window to display window to display bibtex items, I think it works closely like working in a separate buffer. But the difference is, one need to keep the 'input session' for selecting items when using `helm-bibtex`, but working in a separator need not.

Maybe I'm wrong since I'm not a very experience user in `ivy` and `helm`. 

How do you think?

bdarcus:
I agree the current multiple select UI in selectrum is not well suited to this.

But the idea the selectrum dev suggested in the feature request I submitted would be a huge improvement.

The changes to the function I pushed last night, which are almost working, were suggested by the embark author. He thinks it should work fine with embark, and so also the ivy-occur-like embark-collect.

bibtex-actions.el
```diff
@@ -67,19 +67,19 @@
...
                "BibTeX entries: "
                (lambda (string predicate action)
                  (if (eq action 'metadata)
                      '(metadata
-                       ;; (annotation-function . bibtex-completion--annotation)
+                       ;; TODO (annotation-function . bibtex-completion--annotation)
                        (category . bibtex))
                    (complete-with-action action candidates string predicate))))))
-    (cdr (assoc chosen candidates))))
+    (mapcar (lambda (choice) (cdr (assoc choice candidates))) chosen)))
```
bdarcus, line 82: So with changes last night, if I understand right, the read function is getting fed nil results?

So, `bibtex-actions-insert-key` results in a comma-separated list of empty results, and some of the other insert commands complain about wrong type.

And here's what the read function returns:

``` elisp
ELISP> (bibtex-actions--completing-read)
(nil nil nil nil nil)
```

That makes me think there's something wrong with this line, but I can't see any problem.

@oantolin?

oantolin: I think the problem is simply that you are using the default action of `crm-separator`, which is a comma (well, a regexp that matches optional spaces, then a comma, then optional spaces), while most formatted bibtex entries also contain a comma, so `completing-read-multiple` gets confused. You should pick a separator that is unlikely to occur in the formatted bibtex entries. I noticed the bibtex file I was testing with had no ampersands so I set `crm-separator` to `"\\s-*&\\s-*"` and it seemed to work then.

bdarcus: Ah, that makes sense!

I believe in BibTeX you have to escape ampersands with a backslash.

How do I set the `crm-separator`?

bdarcus: Ah, think I figured it out.

Only thing is, I have to manually add the separator.

bdarcus:
This actually now should work. You just have to use the ampersand separator instead of comma.

bdarcus:
Commands now work.

Embark integration doesn't generate errors, but also doesn't generate correct output with multiple selections.

bdarcus:
I merged this. We can test further and see if we can figure out the embark piece.

PR #37, opened 2021-03-17
feat(ui): add affixation function

### LINKED ISSUE #33: Add affixation function to support prefixes and suffixes for candidate display
See https://github.com/bdarcus/bibtex-actions/discussions/28#discussion-3267692 and https://github.com/minad/marginalia/issues/50#issuecomment-798896705.

Emacs 28 introduces affixation support, which we should be able to exploit for a more functional and elegant UI/UX.

To do that, we need an affixation function, or even a separate package, that adds color-differentiated note/pdf/link icons to the prefix where those items are associated with the candidate, and (maybe user-configurable) secondary metadata to suffix. 

Has to be gorgeous, functional, and work well on dark and light themes.

The idea is this would enhance visual scanning of the candidate list (and, TBH, improve the aesthetics).

Example mockup:

Screenshot from 2021-03-13 15-04-00

This will address how to filter on these: https://github.com/tmalsburg/helm-bibtex/issues/363.

## Details

The `bibtex-completion-candidates` function includes the metadata one can use to construct the affixes, and I assume we want to use `all-the-icons` for the actual icons.

bdarcus:
The basics of this works; here's a screenshot with the default prefix display.

Screenshot from 2021-03-20 16-24-13

This has no dependencies, but adds the following:

1. prefix that holds symbols or (optionally) icons to indicate presence of pdfs, notes, links; icon configuration example and screenshot included in the README
2. to enable 1, I added ability to filter on these too, using `has:pdf` etc
3. a defcustom for suffix display that works exactly the same as the main bibtex-completion display formats variable

Still TODO:

4. suffix metadata function that hacks `bibtex-completion-format-entry` to instead generate the suffix (is that possible?). I'm leaning towards leaving this for a separate PR, as I don't really know how to do what I want, and it could take some work.

Screenshot with `all-the-icons`:

Screenshot from 2021-03-19 18-59-51

Closes #33.


bibtex-actions.el
```diff
@@ -92,6 +92,12 @@
      candidate (1- (frame-width)))) ; allow this to be configurable?
     (cdr (assoc "=key=" candidate)))))
 
+(defun bibtex-actions--affixation ()
+  (cl-loop
+   for candidate in (bibtex-actions--get-candidates)
+   collect
+   (candidate "ABC" "XYZ")))
```
bdarcus, line 99: I'm trying to get started on this, but am stuck at the what this affixation function should be returning.

bibtex-actions.el
```diff
@@ -92,6 +92,12 @@
      candidate (1- (frame-width)))) ; allow this to be configurable?
     (cdr (assoc "=key=" candidate)))))
 
+(defun bibtex-actions--affixation ()
+  (cl-loop
+   for candidate in (bibtex-actions--get-candidates)
+   collect
+   (car (candidate) "ABC" "XYZ")))
+
```
bdarcus, line 100: I'm trying to get started on this, but am stuck at the beginning: what this affixation function should be returning.

bibtex-actions.el
```diff
@@ -93,6 +93,12 @@
      candidate (1- (frame-width)))) ; allow this to be configurable?
     (cdr (assoc "=key=" candidate)))))
 
+(defun bibtex-actions--affixation ()
+  (cl-loop
+   for candidate in (bibtex-actions--get-candidates)
+   collect
+   ((car candidate) "ABC" "XYZ")))
+
```
bdarcus, line 101: Trying to get this minimal version working first, but no dice.

mohkale:
Just a heads up that on terminals the icons will either be unavailable or in my case truncated because the all-the-icons glyphs aren't monospaced. I've been getting around the latter by padding them by a single space with the same face so their always at least 2 characters wide when written.

Screenshot_20210319_232423

Not sure whether you have any intention to support tty alternatives. For what it's worth you can get something working by applying this diff:

```diff
--- /home/mohkale/.config/dotfiles/prog/editors/emacs/lisp/straight/repos/bibtex-actions/bibtex-actions.el
+++ #<buffer bibtex-actions.el>
@@ -110,9 +110,11 @@
    collect
    (let ((pdf
           ;; FIX: why doesn't this work????!!!!
-          (if (string-match bibtex-completion-pdf-symbol candidate)
-              (all-the-icons-icon-for-file "foo.pdf" :face 'all-the-icons-lred)
-            (all-the-icons-icon-for-file "foo.pdf" :face 'all-the-icons-off)))
+          (propertize (concat (all-the-icons-icon-for-file "foo.pdf") " ")
+                      'face
+                      (if (string-match bibtex-completion-pdf-symbol candidate)
+                          'all-the-icons-lred
+                        'all-the-icons-off)))
          (link
           (if "link"
               (all-the-icons-faicon "external-link-square" :v-adjust 0.02)
```

Screenshot_20210319_233056


Frankly I'd prefer if there was a configuration option for this, something like instead of asking you to support my weird situation here:

```lisp
(defcustom bibtex-actions-icon
  `((pdf . ("P" . " "))
    (note . ("N" . " "))
    (link . ("L" . " ")))
  "Configuration alist specifying which icon to pick for a bib entry.")

(car (alist-get 'pdf bibtex-actions-icon)) ;; Icon for a bib entry with existing PDF
(cdr (alist-get 'pdf bibtex-actions-icon)) ;; Icon for a bib entry with missing PDF
```

bdarcus:
Yeah, I definitely think there should be a config option, and that default this should probably be off, since the current solution does similar with plain text.

bdarcus:
> I'd prefer if there was a configuration option for this

I added two defcustoms to handle this, but there are a few issues with how I've set this up; one to turn on the "rich" ui, and the other to define the separator between the icons.

Any insight into why "rich-ui" isn't now activating when I set `(setq bibtex-actions-rich-ui t)`?

bdarcus:
On this:
> ```lisp
> (defcustom bibtex-actions-icon
>   `((pdf . ("P" . " "))
>     (note . ("N" . " "))
>     (link . ("L" . " ")))
>   "Configuration alist specifying which icon to pick for a bib entry.")
> 
> (car (alist-get 'pdf bibtex-actions-icon)) ;; Icon for a bib entry with existing PDF
> (cdr (alist-get 'pdf bibtex-actions-icon)) ;; Icon for a bib entry with missing PDF
> ```

Do you think this is necessary beyond the defcustom I just added?

mohkale:
@bdarcus

> Do you think this is necessary beyond the defcustoms I just added; at least at this point?

Well I'm pretty sure there'll still be a truncation issue for me. The problem is icons take up more than one character width and when there's just a space following the icon my terminal lets the icon grow into that space. But when you've got something like: `(concat ICON-WITH-A-FACE " ")` emacs will write out `FACE-COLOR+ICON+RESET_COLOR+" "` and so that extra escape sequence for resetting the color causes my (and I presume many other lazy) terminal/s to not see the space the icon could grow into and instead truncate it.

I hope I've explained that well enough. My issue wasn't the seperator, it was the icon having a different face from the seperator. 

End of the day I still think a defcustom like I've described above is the easiest way to fix this... but if you don't want to fix this that's okay. I'll probably just advise it like I've had to do a bunch of different things.

bdarcus:
I'm not opposed to it; just trying to understand. Also my elisp isn't great, so I was wondering if it might be easier to have a separate PR to assess that. But I'm not set on that; was just a question.

So to be clear, you do want the affixes; just not using the icon fonts? In your setup, you would just use the standard characters for the prefix content, as with your example code?

If yes, how would that be different for you than just using the `bibtex-completion-pdf-symbol` etc. variables and [display config variable](https://github.com/tmalsburg/helm-bibtex#customize-layout-of-search-results), which can already do this? Like:

``` elisp
  (setq bibtex-completion-display-formats
        '((t . "${=has-pdf=:1}${=has-note=:1}    ${author:24}   ${title:64}   ${year:4}"))))
```

Do the affixes provide some advantage over that?

Again: just trying to understand.

mohkale:
I was planning to use icons... just didn't think it would be pretty defining the `defcustom` above with it. So for example this would use the same configuration for pdf as you have setup now:

```lisp
(defcustom bibtex-actions-icon
  `((pdf . (,(all-the-icons-icon-for-file "foo.pdf" :face 'all-the-icons-dred) .
            ,(all-the-icons-icon-for-file "foo.pdf" :face 'bibtex-actions-icon-dim)))
    (note . ("N" . " "))
    (link . ("L" . " ")))
  "Configuration alist specifying which icon to pick for a bib entry.")
```

bdarcus:
So I guess ideally with this approach we would not require `all-the-icons` at all, and just include that as a config option in the README?

I do like the idea of limiting the dependencies as much as possible.

mohkale:
In practice you could do that I guess. There's no real dependency on all-the-icons with what I'm suggesting.

bdarcus:
> In practice you could do that I guess.

What would be the better alternative?

Having all-the-icons default?

In any case, I just pushed the change.

bdarcus:
What's the right `:type` for that defcustom?

mohkale:
> What would be the better alternative?

Hmmmm... I'm not sure. I'd say all-the-icons is widespread enough that most emacs users have it. On the other hand some people use [nerd-icons](https://github.com/twlz0ne/nerd-icons.el) instead and there's no guarantee `all-the-icons` will always be used/available.

I'd say I'm okay with removing the dependency on all-the-icons. Maybe include a section in the README explaining how to get 

> What's the right :type for that defcustom?

No idea. I've never used the customize interface so I rarely specify `:type`. You could put `sexp` if you're lazy (like me). Or you can put [alist](https://www.gnu.org/software/emacs/manual/html_node/elisp/Composite-Types.html#Composite-Types) (something like `(alist :key-type symbol :value-type (cons string string))`, but you'll have to experiment).

bdarcus:
> I'd say I'm okay with removing the dependency on all-the-icons. Maybe include a section in the README explaining how to get it to look like the screenshot with all-the-icons.

Yeah, that's what I was thinking.

I think there's value in having this as lightweight as possible. Indeed, when @mtreca started this discussion with an issue and PR at helm-bibtex, he started with requiring selectrum, consult, and marginalia, but @minad made the strong case none of those dependencies were needed.

The same would seem to make sense for the icons.

With the current code, the affixation function will be ignored pre-Emacs 28, and we just have the single dependency, with tons of room for config flexibility.

That'll be the plan then, unless someone else weighs in with a strong opinion, and good argument, in favor of the alternative.

Aside from that, I think the only other thing to resolve here is that suffix function.

PS - I took a stab at the type on the defcustom. Seems to work.

bdarcus:
Hmm ... maybe defaults should be the bibtex-completion symbols.

minad:
I didn't follow the discussion here, but I suggest you don't introduce a dependency to an icon package, since there are multiple icon packages to choose from. Personally I am not very fond of all-the-icons for example. I don't have it installed and dislike it if a package pulls it in as a superfluous dependency.

bibtex-actions.el
```diff
@@ -38,10 +38,59 @@
...
+entry has a notes file. The \"author\" field is expanded to
+either the author names or, if the entry has no author field, the
+editor names."
+  :group 'bibtex-actions
+  :type '(alist :key-type symbol :value-type string))
+
+(defcustom bibtex-actions-icon
+  `((pdf .      (,bibtex-completion-pdf-symbol . " "))
+    (note .     (,bibtex-completion-notes-symbol . " "))
+    (link .     ("L" . " ")))
```
bdarcus, line 79: @mohkale if/when you have time, I could use help fixing a few minor issues I have with my elisp. I will flag them with line comments. Feel free to reply with the "suggesting" feature so I can click to commit.

First problem, I want to use these existing variables for the symbols, but it's not working.

Aside: I will submit a PR to add the link symbol to bibtex-completion.

bibtex-actions.el
```diff
@@ -73,25 +126,55 @@
                (lambda (string predicate action)
                  (if (eq action 'metadata)
                      '(metadata
-                       ;; TODO (affixation-function . bibtex-actions--affixation)
+                       (where (,bibtex-actions-rich-ui)
```
bdarcus, line 129: This is the other problematic line, @mohkale; probably the same problem.

I want this affixation function to run by default, but to allow a user to turn it off if they prefer.

bibtex-actions.el
```diff
@@ -38,10 +38,59 @@
...
 (defvar crm-separator)
 
+(defcustom bibtex-actions-rich-ui t
+  "Adds prefix symbols or icons and secondary metadata for suffix UI.
+Affixation was first introduced in Emacs 28, and will be ignored
+in previous versions."
+  :group 'bibtex-actions
+  :type 'boolean)
+
+(defcustom bibtex-actions-suffix-display-formats
```
bdarcus, line 54: Finally, this is more an FYI: I added this config option for the suffix; copied from the main display defcustom for bibtex-completion.

bibtex-actions.el
```diff
@@ -73,25 +126,55 @@
                (lambda (string predicate action)
                  (if (eq action 'metadata)
                      '(metadata
-                       ;; TODO (affixation-function . bibtex-actions--affixation)
+                       (when bibtex-actions-rich-ui
```
bdarcus, line 129: This is the other problematic line; maybe the same problem.

I want this affixation function to run by default, but to allow a user to turn it off if they prefer.

bibtex-actions.el
mohkale: I don't see any issue here. You just need to define the symbols before you try to apply them. eg:

```elisp
(defvar bibtex-completion-pdf-symbol "P")
(defvar bibtex-completion-notes-symbol "N")

(defcustom bibtex-actions-icon
  `((pdf .      (,bibtex-completion-pdf-symbol . " "))
    (note .     (,bibtex-completion-notes-symbol . " "))
    (link .     ("L" . " ")))
  "Configuration alist specifying which symbol or icon to pick for a bib entry."
  :group 'bibtex-actions
  :type '(alist :key-type string
                :value-type (choice (string :tag "Icon"))))
```

S.N. If you're unsure about how an evaluation will end up you can run `M-x eval-last-sexp` and it'll evaluate it and then show you what the result would be. So for example

```
  `((pdf .      (,bibtex-completion-pdf-symbol . " "))
    (note .     (,bibtex-completion-notes-symbol . " "))
    (link .     ("L" . " ")))
```

With your cursor on the last paren in this expression you can run `eval-last-sexp` and it'll show you the result.

bibtex-actions.el
mohkale: oh, Umm you probably wanted to evaluate that second sub-expression instead of inserting it as a quoted list. You can do either:

```lisp
`(metadata
  ,(when bibtex-actions-rich-ui
     '(affixation-function . bibtex-actions--affixation)))
```

or

```lisp
(list 'metadata
      (when bibtex-actions-rich-ui
        '(affixation-function . bibtex-actions--affixation)))
```

If you're unsure about the difference please ask :-).

bibtex-actions.el
bdarcus: > You just need to define the symbols before you try to apply them

Those variables are already defined in `bibtex-completion`; why I'm using them.

> If you're unsure about how an evaluation will end up you can run M-x eval-last-sexp and it'll evaluate it and then show you what the result would be.

That should be helpful; thanks!

bibtex-actions.el
bdarcus: Yeah, I'm totally lost with the '`, stuff!

Everything was working fine before I introduced that single "when" line. I don't understand why fixing it would require syntax tweaks across multiple lines.

bibtex-actions.el
bdarcus: Interesting. I had to restart emacs to get that to evaluate correctly, which it does now.

bibtex-actions.el
bdarcus: Not sure why, but all fixed now.

And I did submit a PR for bibtex-completion; will remove the link symbol variable here if that's merged.

bibtex-actions.el
```diff
@@ -72,26 +133,56 @@
...
+          (if (string-match "has:note" candidate)
+                  (car (cdr (assoc 'note bibtex-actions-icon)))
+                (cdr (cdr (assoc 'note bibtex-actions-icon))))))
+   (list candidate (concat
+                    (s-join bibtex-actions-icon-separator
+                            (list pdf note link))"	")"	XYZ"))))
+
+(defun bibtex-actions--make-suffix (entry)
+  "Create the formatted ENTRY suffix string for the 'rich-ui'."
+    ;;TODO make use of 'bibtex-completion' so as simple and flexible as possible.
```
bdarcus, line 184: Is there a way to advise, or otherwise hack, rather than recreate that code, [bibtex-completion-format-entry](https://github.com/tmalsburg/helm-bibtex/blob/ca09076c3d6e51cc4ffe208c8176fccf3710fcc6/bibtex-completion.el#L851) to instead generate the suffix?

mohkale: What're you trying to make `bibtex-actions--make-suffix` do? Not include a pdf, note and link symbol? If so I think the easiest approach would be to modify `bibtex-completion-display-formats` so that it doesn't contain those formats.

bdarcus: > What're you trying to make `bibtex-actions--make-suffix` do? Not include a pdf, note and link symbol? If so I think the easiest approach would be to modify `bibtex-completion-display-formats` so that it doesn't contain those formats.

No; I'm not worried about that.

The idea I had for the suffix was it would be for secondary metadata related to the candidate.

So main display is primary (author, date, title, for example), and suffix is things perhaps like tags, type, etc.

I was assuming people would want to be able to configure it, just like they can now for the main display using `bibtex-completion-display-formats`.

What you ask seems to suggest using that variable (and function) to format the suffix, but formatting authors, titles, dates can be messy. I'd really rather leave that to `bibtex-completion` if possible.

mohkale: Ah, I see. I don't think advice would work very well here. Honestly I don't really see the need for a suffix, you can probably push all of that into main. I'd say the best approach would be to explain your intentions to the bibtex maintainer and ask if he'd be willing to change to get the behaviour you need. Otherwise, yeah, re-implementation is the only approach I can think of.

bdarcus: Yeah, what I was thinking.

I'll hold off awhile on asking, as I know he's really busy, and as you note, it's not critical.

Can you confirm this works as expected on your end when you get a chance, including the actual commands (open and such)?

I'm seeing some inconsistent behavior on my end, but I suspect that's not representative.

mohkale: I will when I get a chance but not soon I'm afraid.

bdarcus: I did some more testing and have now merged this. Thanks for your input and help!

Issue #39, opened 2021-03-22
multi-select broke?


bdarcus:
Selection of multiple candidates broke (returns empty results), as a result of [this change](https://github.com/bdarcus/bibtex-actions/commit/a5429097a843b27df1f0a9320db22f47d3581f8d). I guess I didn't test it enough.

The first result below is with a single selection, and the other with two.

``` elisp
ELISP> (bibtex-actions-read)
("sassen2000")

ELISP> (bibtex-actions-read)
(nil nil)
```

I've tried various tweaks, including (both of [these](https://github.com/bdarcus/bibtex-actions/pull/37#discussion_r598152756)), but it's still not fixed, and I can't see the problem. 

Any ideas @mohkale?

EDIT: progress. I think I've established the problem is with --get-candidates.

bdarcus:
@clemera - before I dig into this further, did you change anything with crm in the past few days? I see your latest commit fixed a crm-related issue, but I'm running that commit.

PR #40, opened 2021-03-22
fix for #39

bdarcus:
Something I did with #37 broke crm multi-selection of candidates.

I have narrowed it down to the changes I made to `bibtex-actions--get-candidates`. It's probably something obvious, but I can't see it ATM!

This code now works correctly, but removes the feature I wanted from the offending code: add the pdf/link/note strings to the candidate.

EDIT: an idea: it would be the whitespace I ended up introducing at the end of the candidate strings?

cc @mohkale

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -126,18 +126,24 @@ may be indicated with the same icon but a different face."
              collect (cdr (assoc choice candidates)))))
 
 (defun bibtex-actions--get-candidates ()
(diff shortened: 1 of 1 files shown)


bdarcus:
> EDIT: an idea: it would be the whitespace I ended up introducing at the end of the candidate strings?

Need to test a bit more, but this appears to have been the problem. Latest commit should resolve it.

PR #45, opened 2021-03-23
feature(ui): add suffix 

### LINKED ISSUE #44: add suffix UI
To accompany the recent prefix addition, add a configurable suffix.

bdarcus:
Add suffix UI, using the same configuration method as for the main display formatting.

Also add annotation-function, to provide the same UI pre-Emacs 28..

To enable this, I have adapted some code from bibtex-completion. I will 
that code, or greatly simplify it, if and when this PR is merged:

https://github.com/tmalsburg/helm-bibtex/pull/367

Closes #44

------

Here's what it looks like now, with these two PRs, and configured to use `all-the-icons`:

Screenshot from 2021-03-26 11-04-25

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -82,7 +82,8 @@ may be indicated with the same icon but a different face."
   (setq bibtex-completion-display-formats
         '((t . "${author:20}   ${title:48}   ${year:4}")))
   (setq bibtex-completion-display-formats-suffix
-        '((t . "${=type=}"))))
+        '((t . "          ${=key=:15}    ${=type=:12}    ${tags:30}"))))
+
 
 ;;; Keymap
 
@@ -135,7 +136,7 @@ key associated with each one."
           (citekey (bibtex-completion-get-value "=key=" candidate))
           (add (s-trim-right (s-join " " (list pdf note link))))
           ; TODO: add separation between target and suffix
-          (suffix  (bibtex-completion-format-entry candidate (1- (frame-width))
+          (suffix (bibtex-completion-format-entry candidate (1- (frame-width))
                         bibtex-completion-display-formats-suffix-internal)))
    (cons
     ;; Here use one string for display, and the other for search.
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -124,9 +124,8 @@ may be indicated with the same icon but a different face."
                (lambda (string predicate action)
                  (if (eq action 'metadata)
                      `(metadata
-                       ; TODO add annotation-function?
-                       ,(when bibtex-actions-rich-ui
-                          '(affixation-function . bibtex-actions--affixation))
+                       (affixation-function . bibtex-actions--affixation)
+                       (annotation-function . bibtex-actions--annotation)
                        (category . bibtex))
                    (complete-with-action action candidates string predicate))))))
     (cl-loop for choice in chosen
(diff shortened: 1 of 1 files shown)


bdarcus:
FYI, @mohkale, I figured out a simple solution to suffixes. 

I think, next step, and what would make it worth merging, is having the content driven by the alist variable, just as with `bibtex-completion-display-formats`. The difference is this would be a smaller set of variables, all simple strings.

``` elisp
(defcustom bibtex-actions-suffix-display-formats
  '((t . "${=type=:7} ${tags:24}")))
```

mtreca:
Looking good!

On an unrelated note, I tested your package but face highlighting does not seem to work with selectrum:

2021-03-19-180546_1920x1080_scrot

Should I open a separate issue?

bdarcus:
@mtreca - do you have any insight into what's actually going on with that?

I test with two different setups: emacs -Q-based selectrum test files with solarized installed, and my standard doom setup, and that all seems to work fine.

bibtex-actions.el
```diff
@@ -136,15 +132,22 @@ key associated with each one."
...
    (cons
     ;; Here use one string for display, and the other for search.
     ;; The candidate string we use is very long, which is a bit awkward
     ;; when using TAB-completion style multi selection interfaces.
     (propertize
-     (s-append add (car candidate)) 'display (bibtex-completion-format-entry
-     candidate (1- (frame-width))))
-    (cdr (assoc "=key=" candidate))))))
+     (s-append add (car candidate))
+     'display (bibtex-completion-format-entry candidate (1- (frame-width)))
```
bdarcus, line 146: I don't really understand widths in emacs; is this right?

mtreca: That's what was being used in the original package no? LGTM

mtreca:
I am pretty sure it has to do with the newly added `rich-ui` option. Will test without to see if anything changes.

bibtex-actions.el
bdarcus: > That's what was being used in the original package no?

Yes, but I just don't understand it, so wasn't sure if it mattered if we added a suffix with the same. Apparently not.

> LGTM

It's ready to merge, pending Titus merging the bibtex-completion PR.

bdarcus:
One of the nice side effects of using affixes is the target display string can be much more compact, which makes the current selectrum multi-select UI work better.

bibtex-actions.el
```diff
@@ -136,15 +133,22 @@ key associated with each one."
    (let* ((pdf (if (assoc "=has-pdf=" (cdr candidate)) " has:pdf"))
           (note (if (assoc "=has-note=" (cdr candidate)) "has:note"))
           (link (if (assoc "doi" (cdr candidate)) "has:link"))
-          (add (s-trim-right (s-join " " (list pdf note link)))))
+          (citekey (bibtex-completion-get-value "=key=" candidate))
+          (add (s-trim-right (s-join " " (list pdf note link))))
+          ; TODO: add separation between target and suffix
+          (suffix (bibtex-completion-format-entry candidate (1- (frame-width))
```
bdarcus, line 139: This, plus the defvar, are the only pieces of code dependent on the bibtex-completion PR. So could comment out, and add a fixed placeholder for now.

bibtex-actions.el
```diff
@@ -136,15 +133,22 @@ key associated with each one."
...
+                        bibtex-completion-display-formats-suffix)))
```
bdarcus, line 140: This and the template setq are the only pieces of code dependent on the bibtex-completion PR. Could comment out and add a fixed placeholder for now.

bdarcus: > This and the template setq are the only pieces of code dependent on the bibtex-completion PR. Could comment out and add a fixed placeholder for now.

Latest commit implements this. I'll wait a bit to see if Titus responds.

bdarcus: The other alternative is to just add the two functions we need to do this ourselves, and then do this, so all of it is configurable.

``` elisp
(defcustom bibtex-actions-display-templates
  ; prefix will only work with rich ui, and on Emacs 28
  ; 'target' is the default entry display string
  '((prefix . '((t. "${=has-pdf=:1} ${=has-note=:1}  ")))
    (target . '((t . "${author:20}   ${title:48}   ${year:4}")))
    (suffix . '((t. "          ${=key=:15}    ${=type=:12}    ${tags:*}"))))
    "Configures display formatting for 'target', 'prefix' and 'suffix'.
The latter two are optional, but are used in the 'rich ui'."
    :group 'bibtex-actions
    :type  '(alist :key-type symbol :value-type function))
```

bdarcus:
OK, I changed the branch to not rely on the bibtex-completion PR. I'll take another look in the morning, and may merge it at that point.

PR #51, opened 2021-04-01
refactor: use display-format for cands

### LINKED ISSUE #47: No selection highlighting
As mentioned in #45, selectrum's syntax highlighting is not shown when using `bibtex-action` functions. Looking at the README, there doesn't seem to be any highlighting either (except an current line background highlight).

2021-03-19-180546_1920x1080_scrot

I originally thought that changing `bibtex-completion-display-formats` was responsible (through `bibtex-actions-rich-ui`), but setting it to nil doesn't change anything.

The other major change from my implementation (which still displays faces properly) is the use of `propertize` in `bibtex-actions-get-candidates`, which I suspect resets pre-existing face settings in the completion minibuffer.

bdarcus:
This simplifies the code and user config story, adds matching on 
search substrings, and gives wider compatibility.

Specifically, it:

1. removes the annotation function and affixation suffix
2. adds "suffix" candidate string with the 'bibtex-actions'suffix face 

The 'bibtex-actions--get-candidates' function no longer uses 
'bibtex-completion-candidates' for the candidate strings, and instead 
uses 'bibtex-actions--format-entry', so that if the user explicitly 
configures the UI, they can customize what will be searched.

Closes #47.


bibtex-actions.el
```diff
@@ -149,12 +149,12 @@ key associated with each one."
...
     ;; when using TAB-completion style multi selection interfaces.
     (propertize
-     (s-append add (car candidate))
-     'display
      (bibtex-actions--format-entry
       candidate
       (1- (frame-width))
       bibtex-actions-display-template)
+     ;'invisible
+     ;(s-append add (car candidate))
```
bdarcus, line 157: I just need to figure out to get this to work to merge this.

bdarcus:
One followup question @mtreca:

> I think I agree with minad when stating that this trade-off is not great for the end user who will expect highlighting to work similarly everywhere, including bibtex-actions.

So we're faced with a choice here then:

Do I remove affixation suffix and annotation, and instead just have that configurable content be a different face on the candidate?

E.g. do you think we want matching on the "suffix" content too?

I think in the end it probably doesn't make a difference, since it should look the same. So in that case, the answer to my question should be "yes." We'd then just use affixation prefixes for as they are now.

bdarcus:
> Do I remove affixation suffix and annotation, and instead just have that configurable content be a different face on the candidate?

OK, latest commit I pushed makes good progress on this (thanks again @minad).

Screenshot from 2021-04-01 13-36-13

So to recap I:

1. switched to using display-format for the candidate
2. added some metadata to an invisible string, to allow search for notes and PDFS using ":note" and such.
3. turned off annotation and affixation, and simply put that content in a separate face, which is the same as I was using for annotation and affixation

Upshot is we retain the search and display power and configurability, but now almost everything will have match highlighting.

The only remaining issue has to do with alignment of the "suffix" section gets screwed in some cases.

Is this preferable @mtreca?

And if you have any more thoughts @minad?

bdarcus:
There is one downside of this change, however: the candidate strings when selecting multiple candidates are now really long again, which makes enhancements to selectrum on this even more important.

bdarcus:
Hmm ... also kind of weird the selected candidate highlighting doesn't extend to the edge of the frame.

PR #57, opened 2021-04-04
change keymap to use simple menus

bdarcus:
Embark has support for this, which means description strings for the `embark-act` menu, rather than command names. 

Two issues.

First, this should work, but does not with which-key:

Screenshot from 2021-04-03 20-28-10

This might just need to wait on this which-key PR: https://github.com/justbur/emacs-which-key/pull/282.

Second, need to hide the embark commands, or (better) put them in a submenu.

See https://github.com/oantolin/embark/issues/197


bdarcus:
Going to close this, as I don't think it's worth doing this until the above features are available in which-key and embark.

PR #62, opened 2021-04-07
remove bibtex-completion-init call

### LINKED ISSUE #41: keep UI updated when changes to bib file(s) and pdf/note availability
Here's what we want:

If a user modifies specified bib files or note or pdf directory, the UI should update; e.g. the `bibtex-actions-refresh` command should auto-run.

I couldn't figure out how to elegantly add this (I don't want to use the `bibtex-completion` approach to this), but see this wiki section for some ideas.

https://github.com/bdarcus/bibtex-actions/wiki/Configuration#proactive-reloading-of-library

I think ideally a user could add a "watcher" function to their config, and have it handle this seemlessly, but that it would not be on by default.

It might even be a separate file, as with straight.

A PR is welcome.

https://github.com/tmalsburg/helm-bibtex/issues/369

bdarcus:
This initially started based on a misunderstanding of the `candidates` variable in the `bibtex-actions-read` function.

Given that the candidates are refreshed each time the function is run, instead go the opposite direction: remove any of this file-notify-based logic.

Can always revisit later if need be.

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -105,7 +105,6 @@ may be indicated with the same icon but a different face."
 ;;; Completion functions
 (defun bibtex-actions-read ()
   "Read bibtex-completion entries for completion using 'completing-read-multiple'."
-  (bibtex-actions--init)
   (when-let ((crm-separator "\\s-*&\\s-*")
              (candidates (bibtex-actions--get-candidates))
              (chosen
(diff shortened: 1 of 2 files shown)


bdarcus:
Initial test returns this:

``` elisp
Error running timer: (wrong-number-of-arguments (((path . "~/org/bib/newer.bib") (watch-these nil nil "~/org/bib/newer.bib") t) nil (bibtex-completion-candidates)) 1)
```

bdarcus:
Fixed the one problem; now there's this:

``` elisp
Error running timer: (wrong-number-of-arguments (((path . "~/org/bib/newer.bib") (watch-these "~/org/bib/newer.bib") t) nil (bibtex-completion-candidates)) 1)
```

bibtex-actions.el
```diff
@@ -185,6 +185,41 @@ key associated with each one."
...
+    ;; Check that all specified files or directories exist and add
+    ;; watches for automatic reloading of the bibliography when a file
+    ;; or directory is changed:
+    (mapc
+     (lambda (path)
+       (if (f-exists? path)
+           (let ((watch-descriptor
+                  (file-notify-add-watch
+                   path '(change)
+                   (lambda (_event) (bibtex-completion-candidates)))))
```
bdarcus, line 217: I don't get why this can't just be:

```suggestion
                   (bibtex-completion-candidates))))
```

bdarcus:
If you have time, could you take a look at this @mohkale?

I think there's just some subtle thing wrong with it.

bdarcus:
I'm going to close this in favor of simply documenting it on the wiki.

https://github.com/bdarcus/bibtex-actions/wiki/Configuration#expand-proactive-loading-of-library-to-pdfs-and-notes

If anyone is interested in this, please test that solution, and revise if necessary.

PR #70, opened 2021-04-10
[WIP] make caching work correctly

### LINKED ISSUE #69: add caching of bib candidates
Thanks for this nice package!  I gave it a try and it worked smoothly for me.

I have a relatively large BibTeX file with over 1,000 entries (and 10,000 lines) and I have been using `ivy-bibtex`.  The `bibtex-ctions` takes a noticeable time to read/load the BibTeX file every time, while `ivy-bibtex` takes a noticeable time only the first time (assuming that the BibTeX file has not been changed).  I guess it is because `ivy-bibtex` somehow caches the BibTeX file that has been read the first time.  Thus, IMHO, it would be great if `bibtex-actions` contains a similar caching mechanism.

_Originally posted by @wenjie2wang in https://github.com/bdarcus/bibtex-actions/discussions/68_

bdarcus:
The bibtex-completion caching is based on `bibtex-completion-candidates`.

Seems like it should be working here, but according to #69 is not.

Edit: actually, not true (see below for details).

----

### Notes

Here's how `ivy-bibtex` does it:

``` elisp
(defun ivy-bibtex (&optional arg local-bib)
  "Search BibTeX entries using ivy.
With a prefix ARG the cache is invalidated and the bibliography
reread.
If LOCAL-BIB is non-nil, display that the BibTeX entries are read
from the local bibliography.  This is set internally by
`ivy-bibtex-with-local-bibliography'."
  (interactive "P")
  (when arg
    (bibtex-completion-clear-cache))
```

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -84,9 +84,6 @@ may be indicated with the same icon but a different face."
   :group 'bibtex-actions
   :type 'string)
 
-(defvar bibtex-actions--cache nil
-  "A cache that stores the results of 'bibtex-actions--get-candidates'.")
-
 ;;; Keymap
 
 (defvar bibtex-actions-map
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -106,7 +106,7 @@ may be indicated with the same icon but a different face."
 (defun bibtex-actions-read ()
   "Read bibtex-completion entries for completion using 'completing-read-multiple'."
   (when-let ((crm-separator "\\s-*&\\s-*")
-             (candidates (bibtex-actions--get-candidates))
+             (candidates (bibtex-completion-candidates))
              (chosen
               (completing-read-multiple
                "BibTeX entries: "
@@ -161,11 +161,11 @@ This propertizes the candidate for display and search."
   (cl-loop
    for entry in cands
    collect
-   ;; TODO change get-candidetes to accept a single CAND as input
-   (setcar entry 'bibtex-actions--transform-candidate)))
+   ;; TODO the key logic
+   (setcar entry (bibtex-actions--transform-candidate entry))))
 
-;; Add advice to replace candidate strings
 (advice-add 'bibtex-completion-candidates :filter-return #'bibtex-actions--advice)
+(advice-remove 'bibtex-completion-candidates #'bibtex-actions--advice)
 
 (defun bibtex-actions--affixation (cands)
   "Add affixation prefix to CANDS."
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -123,35 +123,30 @@ may be indicated with the same icon but a different face."
              ;; Collect citation keys of selected candidate(s).
              collect (cdr (assoc choice candidates)))))
 
-(defun bibtex-actions--get-candidates ()
-  "Transform candidates from 'bibtex-completion-candidates'.
-This both propertizes the candidates for display, and grabs the
-key associated with each one."
+(defun bibtex-actions--transform-candidate (candidate)
+  "Transform CANDIDATE.
+This propertizes the candidate for display and search."
   ;; TODO refactor to 'add-advice' to use 'setcar' to replace
   ;; 'bibtex-completion-candidates' car with this string.
-  (cl-loop
-   for candidate in (bibtex-completion-candidates)
-   collect
-   (let* ((pdf (if (assoc "=has-pdf=" (cdr candidate)) " has:pdf"))
-          (note (if (assoc "=has-note=" (cdr candidate)) "has:note"))
-          (link (if (or (assoc "doi" (cdr candidate))
-                        (assoc "url" (cdr candidate))) "has:link"))
-          (citekey (bibtex-completion-get-value "=key=" candidate))
-          (candidate-main
-           (bibtex-actions--format-entry
-            candidate
-            (1- (frame-width))
-            bibtex-actions-template))
-          (candidate-suffix
-           (bibtex-actions--format-entry
-            candidate
-            (1- (frame-width))
-            bibtex-actions-template-suffix))
-          ;; We display this content already using symbols; here we add back
-          ;; text to allow it to be searched, and citekey to ensure uniqueness
-          ;; of the candidate.
-          (candidate-hidden (s-join " " (list pdf note link citekey))))
-   (cons
+  (let* ((pdf (if (assoc "=has-pdf=" (cdr candidate)) " has:pdf"))
+         (note (if (assoc "=has-note=" (cdr candidate)) "has:note"))
+         (link (if (or (assoc "doi" (cdr candidate))
+                       (assoc "url" (cdr candidate))) "has:link"))
+         (citekey (bibtex-completion-get-value "=key=" candidate))
+         (candidate-main
+          (bibtex-actions--format-entry
+           candidate
+           (1- (frame-width))
+           bibtex-actions-template))
+         (candidate-suffix
+          (bibtex-actions--format-entry
+           candidate
+           (1- (frame-width))
+           bibtex-actions-template-suffix))
+         ;; We display this content already using symbols; here we add back
+         ;; text to allow it to be searched, and citekey to ensure uniqueness
+         ;; of the candidate.
+         (candidate-hidden (s-join " " (list pdf note link citekey))))
     ;; If we don't trim the trailing whitespace, 'completing-read-multiple' will
     ;; get confused when there are multiple selected candidates.
     (s-trim-right
@@ -162,8 +157,7 @@ key associated with each one."
       ;;   3. the 'candidate-hidden' variable to be hidden
       (propertize candidate-main) " "
       (propertize candidate-suffix 'face 'bibtex-actions-suffix) " "
-      (propertize candidate-hidden 'invisible t)))
-    citekey))))
+      (propertize candidate-hidden 'invisible t)))))
 
 (defun bibtex-actions--advice (cands)
   "The advice function to replace CANDS strings."
@@ -171,7 +165,7 @@ key associated with each one."
    for entry in cands
    collect
    ;; TODO change get-candidetes to accept a single CAND as input
-   (setcar entry 'bibtex-actions--get-candidates)))
+   (setcar entry 'bibtex-actions--transform-candidate)))
 
 ;; Add advice to replace candidate strings
 (advice-add 'bibtex-completion-candidates :filter-return #'bibtex-actions--advice)
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -84,6 +84,9 @@ may be indicated with the same icon but a different face."
   :group 'bibtex-actions
   :type 'string)
 
+(defvar bibtex-actions--cache nil
+  "A cache that stores the results of 'bibtex-actions--get-candidates'.")
+
 ;;; Keymap
 
 (defvar bibtex-actions-map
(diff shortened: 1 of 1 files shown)


bibtex-actions.el
```diff
@@ -84,6 +84,9 @@ may be indicated with the same icon but a different face."
   :group 'bibtex-actions
   :type 'string)
 
+(defvar bibtex-actions--cache nil
+  "A cache that stores the results of 'bibtex-actions--get-candidates'.")
+
```
bdarcus, line 89: If I use `add-advice` per below, this won't be necessary.

mohkale:
@bdarcus 

I'm slightly confused about what you're trying to do here? You reference the `ivy-bibtex` cache but wasn't that just the cache of all bibtex entries (loaded from your various `*.bib` files). If your caching the results of `--get-candidates` is there something wrong with just using the history parameter of `completing-read`.

bdarcus:
> @bdarcus
> 
> I'm slightly confused about what you're trying to do here? You reference the `ivy-bibtex` cache but wasn't that just the cache of all bibtex entries (loaded from your various `*.bib` files). If your caching the results of `--get-candidates` ...

To recap, currently, there is no caching, which is not a problem for smaller or mid-size files, but is on large ones per #69.

>  is there something wrong with just using the history parameter of `completing-read`.

IDK; hence me pinging you :-)

Bottomline: what's the best way to enable this?

How would I use the history, for example?

mohkale:
> To recap, currently, there is no caching, which is not a problem for smaller or mid-size files, but is on large ones per #69.

oh, I see, From the looks of things `bibtex-completion` has it's own caching mechanism that it uses automatically when you call `bibtex-completion-candidates`. Part of that is loading the entire file into a buffer, creating a hash/checksum and then offloading it which I presume is what's causing the slowdown described in issue #69. If so I'm not sure what improvements you could bring in this package.

They do mention 

> The bibtex-ctions takes a noticeable time to read/load the BibTeX file every time, while ivy-bibtex takes a noticeable time only the first time

So their might be some configuration issue in your package that invalidates the cache every time. That sounds like a bug.

> How would I use the history, for example?

If you look at the docstring of `completing-read-multiple` it has a HIST parameter (which is described in the docstring of `completing-read`). Essentially whenever you select or complete a candidate it saves that candidate to a variable and you can quickly re-enter it with `previous-history-element` and `next-history-element`.

Although personally I'm using prescient so I rarely need history. In retrospect I don't think that's what you're talking about.

mohkale:
@bdarcus 

Also please @ mention me in any messages you try to send me. I generally don't subscribe to PRs/issues because I get a lot of notifications like that.

bdarcus:
@mohkale 

> Bibtex-completion has it's own caching mechanism that it uses automatically when you call bibtex-completion-candidates.

Right. They already figured this out; I just need to hook it up to `bibtex-actions` correctly.

> So their might be some configuration issue in your package that invalidates the cache every time. That sounds like a bug.

Yeah, you might be right.

At first I was thinking the problem was I was using a different set of candidates.

But thinking again, that doesn't really make any sense, given that `--get-candidates` runs of `bibtex-completion-candidates`.

Unless there's something subtle I'm missing on calling that function.

mohkale:
> Yeah, you might be right.

I just tried running the debugger on it and it didn't look like the cache was invalidated. I'd maybe try to reach out to the author behind #69 and ask him to try and reproduce with `emacs -Q`. 

Also mention that if he's editing the bibtex files in emacs then `bibtex-completion` should automatically reload the file when you save it.

bdarcus:
@mohkale done; thanks!

I'll close this then. Even if there's a problem (which I don't think there is), the direction I was going was the wrong one.

bdarcus:
Actually, @mohkale, to go back to the beginning for me:

Maybe the issue is not `bibtex-completion-candidates`, but rather the fact that `bibtex-actions--get-candidates` is not cached, and so runs every time that `bibtex-actions-read` is called.

That could explain the discrepancy here.

mohkale:
@bdarcus 

> Maybe the issue is not bibtex-completion-candidates, but rather the fact that bibtex-actions--get-candidates is not cached, and so runs every time that bibtex-actions-read is called

I don't think that's relevent. I mean all `bibtex-actions--get-candidates` does is fetch the already cached bibtex entries and format them in a way that's OK to show.

Have you managed to reproduce the caching error described in #69, until you can I doubt there's much you can do to fix this.

bdarcus:
> I mean all `bibtex-actions--get-candidates` does is fetch the already cached bibtex entries and format them in a way that's OK to show.

Yeah, but I've benchmarked it, and it's a lot slower than `bibtex-completion-candidates.`

> Have you managed to reproduce the caching error described in #69.

I just tested his file, and every time I load a command, there's a pause, but it's certainly less than a second. 

I'll ask him to clarify.

mohkale:
@bdarcus 

> Yeah, but I've benchmarked it, and it's a lot slower than bibtex-completion-candidates.

Hmmm... You do seem to have a call to `(- 1 (frame-width))` which you repeat for every candidate in the cache. Maybe try moving that out to a local `let` clause instead of recalculating it every time. Subtle things like that can add a performance cost (try using the builtin profiler if you're still not sure).


> I just tested his file, and every time I load a command, there's a pause, but it's certainly less than a second.

The pause could literally be because emacs is loading the file into a temp-buffer every time the candidates are generated (even from cache). I've noticed similair delays with my 20,000 line emacs config. Although if that's the issue he/she should raise it with `bibtex-completion` instead of this package (although because he/she says it doesn't happen with ivy-bibtex I'm honestly not sure).

bdarcus:
@mohkale 

Here's the benchmarks:

``` elisp
ELISP> (benchmark-run 100 (bibtex-completion-candidates))
(1.39453827 6 0.5778311429999974)

ELISP> (benchmark-run 100 (bibtex-actions--get-candidates))
(59.037539040999995 208 18.455108738000003)
```

> Hmmm... You do seem to have a call to `(- 1 (frame-width))` which you repeat for every candidate in the cache. Maybe try moving that out to a local `let` clause instead of recalculating it every time.

With that change, which I'd actually had in main, already, it shaves the total time down to 35 seconds.

mohkale:
@bdarcus 

That is a pretty dramatic increase, you're best bet is to try the profiler I'd say.

bdarcus:
@mohkale 
 
> That is a pretty dramatic increase, you're best bet is to try the profiler I'd say.

This is what it tells me, based on running one of the commands:

Screenshot from 2021-04-10 12-10-14

Doesn't seem very helpful to me. Do you see anything significant there?

I wonder if it's because that function also grabs the key for each candidate?

mohkale:
@bdarcus 

You can expand the profiler sections to get more detail (see the plus). Pretty sure it's bound to TAB by default but you should be able to get as detailed a breakdown of what emacs is doing as possible.

bdarcus:
@mohkale - OIC now. This is obviously much more helpful!

What do you make of this?

Screenshot from 2021-04-10 13-22-46

mohkale:
@bdarcus 

Hmmm...  looks like most of the CPU is spent on format-entry, which is beyond our control. I doubt this is as big an issue as we're making it out to be.

I'd suggest maybe switching back to ivy-bibtex and profiling that as well but my guess is there shouldn't be much a performance difference between this package and that. You ran the benchmarks before on just getting the candidates from `bibtex-completion-candidates` and then on your package formatting them but to get a good comparison you should compare how ivy-bibtex formats them as well.

If you'd like you could probably improve the performance of `bibtex-completion-format-entry` but I'm not sure whether the benefit outweighs the time investment.

Regardless I'm confident this issue can be closed.

bdarcus:
@mohkale 

> Hmmm... looks like most of the CPU is spent on format-entry, which is beyond our control. 

That's my interpretation too; not surprising to me.

> I'd suggest maybe switching back to ivy-bibtex and profiling that as well but my guess is there shouldn't be much a performance difference between this package and that. You ran the benchmarks before on just getting the candidates from `bibtex-completion-candidates` and then on your package formatting them but to get a good comparison you should compare how ivy-bibtex formats them as well.

Here's `ivy-bibtex`:

Screenshot from 2021-04-10 13-43-03

I don't really understand how the display formatters in those packages work from the performance standpoint. They have to use format-entry there too.

> Regardless I'm confident this issue can be closed.

So you don't think I should look into working around somehow that get-candidates is always loaded?

mohkale:
@bdarcus 

> get-candidates is always loaded

You mean every call to get-candidates invalidates the cache. You should fix that, if it actually does that. For me it doesn't so I'm not sure how to help there.

bdarcus:
@mohkale - I just mean every time `bibtex-actions-read` is called it will run `bibtex-actions-get-candidates`.

mohkale:
@bdarcus 

> I just mean every time bibtex-actions-read is called it will run bibtex-actions-get-candidates.

Yeah, I don't see any issue with that. You need to get a list of candidates and it's better to rely on bibtex-completion rather than creating your own caching logic IMO.

bdarcus:
@mohkale 
 
> > I just mean every time bibtex-actions-read is called it will run bibtex-actions-get-candidates.
> 
> Yeah, I don't see any issue with that. You need to get a list of candidates and it's better to rely on bibtex-completion rather than creating your own caching logic IMO.

No, I don't want to do that. What I was contemplating, however, was using `advice-add` to change the candidate strings (the car for each entry) of `bibtex-completion-candidates` itself, and then using that.

bibtex-actions.el
```diff
@@ -159,8 +154,18 @@ key associated with each one."
...
-    citekey))))
+      (propertize candidate-hidden 'invisible t)))))
+
+(defun bibtex-actions--advice (cands)
+  "The advice function to replace CANDS strings."
+  (cl-loop
+   for entry in cands
+   collect
+   ;; TODO the key logic
+   (setcar entry (bibtex-actions--transform-candidate entry))))
```
bdarcus, line 165: Like, why can't I do something like this, @mohkale?

Doesn't work, but ...

mohkale: I don't see why you can't

```lisp
(advice-add #'bibtex-completion-candidates :filter-return #'bibtex-actions--advice)
```

but I don't see why you would. There's no benefit to doing this compared to just using 

```lisp
(cl-loop
   for entry in (bibtex-completion-candidates)
   collect
   ;; TODO the key logic
   (setcar entry (bibtex-actions--transform-candidate entry)))
```

And worse your altering the return format of a function in another library which means any other libraries aside from yours which also uses bibtex-completion might break.

bdarcus: >  There's no benefit to doing this compared to just using ...

... what do you mean by your example @mohkale? Where would that `cl-loop` be used?

> And worse your altering the return format of a function in another library which means any other libraries aside from yours which also uses bibtex-completion might break.

Ah, hadn't thought about that. I guess that feature is better for local configs.

OK, then; bad idea :-)

mohkale: > what do you mean by your example @mohkale? Where would that cl-loop be used?

Wherever you have a call to `bibtex-completion-candidates` but want whatever your advice does to be applied to it before returning.

bdarcus: @mohkale -  the only place I would call that function would be in the "candidates" let variable on `bibtex-actions-read`, and the only reason to make the mod would be so I can let bibtex-completion deal with caching of the candidates.

If you get a chance, if in fact you think that would work, could you fill out the example code please? I'm not understanding exactly how that would work.

bdarcus:
OK, closing this again.

FYI, @mohkale, I've [asked](https://github.com/bdarcus/bibtex-actions/issues/69#issuecomment-817232480) the bibtex-completion author if there's room for adding a hook there so I can just use a different function here to define the candidate strings. From my incomplete understanding, that would be the ideal solution.

bdarcus:
And on further thought, @mohkale , I decided to just add a cache. Problem solved, with very little code.

https://github.com/bdarcus/bibtex-actions/commit/edb568ae9177026487d8b3aa649e7e7878512c53

PR #71, opened 2021-04-11
refactor: add candidates cache

### LINKED ISSUE #69: add caching of bib candidates

bdarcus:
Refactor so that  `bibtex-actions-read` pulls candidates from a 
cache: `bibtex-actions--candidates-cache`.

Add:

- `bibtex-actions--candidates-cache` variable
- `bibtex-actions-refresh` interactive command, to update cache

Refactor:

-  `bibtex-actions--get-candidates` renamed to `bibtex-actions--format-candidates`
-  `bibtex-actions--get-candidates` now grabs the candidates from the cache

fixes #69

-----

@wenjie2wang - I do think it would also be good to add the hook to bibtex-completion, but you want to give this a try and let me know?

Turns out it was easier than I thought to add this.

But I haven't fully tested it and seen if it has any unintended consequences.

It does cache the candidates, however, so should address #69.

So if you could test it further and let me know, that'd be great.

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -186,11 +186,11 @@ key associated with each one."
   "Store the candidates list.")
 
 (defun bibtex-actions--get-candidates ()
-    "Get the cached candidates.
+  "Get the cached candidates.
 If the cache is nil, this will load the cache."
-    (if (not bibtex-actions--candidates-cache)
-     (bibtex-actions--reload-candidates-cache))
-    bibtex-actions--candidates-cache)
+  (if (not bibtex-actions--candidates-cache)
+      (bibtex-actions--reload-candidates-cache))
+  bibtex-actions--candidates-cache)
 
 (defun bibtex-actions--reload-candidates-cache ()
   "Reload the candidates cache."
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -191,7 +191,8 @@ key associated with each one."
   "Store the candidates list.")
 
 (defun bibtex-actions--get-candidates ()
-    "Load bib files and cache them."
+    "Get the cached candidates.
+If the cache is nil, this will load the cache."
     (if (not bibtex-actions--candidates-cache)
      (bibtex-actions--reload-candidates-cache))
     bibtex-actions--candidates-cache)
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -189,11 +189,12 @@ key associated with each one."
   "Get the cached candidates.
 If the cache is nil, this will load the cache."
   (if (not bibtex-actions--candidates-cache)
-      (bibtex-actions--reload-candidates-cache))
+      (bibtex-actions-refresh))
   bibtex-actions--candidates-cache)
 
-(defun bibtex-actions--reload-candidates-cache ()
+(defun bibtex-actions-refresh ()
   "Reload the candidates cache."
+  (interactive)
   (setq bibtex-actions--candidates-cache
         (bibtex-actions--format-candidates)))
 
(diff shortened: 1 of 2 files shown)


bibtex-actions.el
```diff
@@ -187,6 +187,19 @@ key associated with each one."
...
+(defvar bibtex-actions--candidates-cache nil
+  "Store the candidates list.")
+
+(defun bibtex-actions--get-candidates ()
+    "Load bib files and cache them."
+    (if (not bibtex-actions--candidates-cache)
+     (setq bibtex-actions--candidates-cache (bibtex-actions--format-candidates)))
+    bibtex-actions--candidates-cache)
+
+(defun bibtex-actions--clear-candidates-cache ()
```
bdarcus, line 199: I'm a little unsure on this. Do we want this function, or should it rather be a "reload-candidates-cache" function?

wenjie2wang:
Thanks, @bdarcus !  I tried this PR and I found that it worked great.  There is no more pause after the first run due to the added candidate cache.

bdarcus:
> Thanks, @bdarcus ! I tried this PR and I found that it worked great. There is no more pause after the first run due to the added candidate cache.

Great!

Can you review the code for the basic logic and confirm that it makes sense to you; that it shouldn't cause any problem when I merge it? It's very simple, but I think should work.

wenjie2wang:
How will the cache get invalidated?  I removed a few BibTeX entries and I still got the cached results.

bdarcus:
> How will the cache get invalidated? I removed a few BibTeX entries and I still got the cached results.

The idea is one would have to run `bibtex-actions--reload-candidates-cache` (hmm, maybe I should make that a public function?), either manually, or via a `filenotify` process as mentioned here:

https://github.com/bdarcus/bibtex-actions/wiki/Configuration#expand-proactive-loading-of-library-to-pdfs-and-notes

But that would need to include the bib files, since while bibtex-completion already handles this, this PR does not.

I omitted it because I'm thinking it's too complicated (see [this recent still open thread](https://github.com/tmalsburg/helm-bibtex/issues/369)), and gives users more flexibility.

WDYT?

Also, maybe you can test the config I list on the wiki for your bib file to confirm it works for you?

bdarcus:
Updated the wiki section to address here:

https://github.com/bdarcus/bibtex-actions/wiki/Configuration#proactive-reloading-of-library

bdarcus:
Latest commit:

- changes that reload function to the `bibtex-actions-refresh` interactive command
- adds section to the README (just need to confirm those `file-notify` configs are right)

wenjie2wang:
> Latest commit:
> 
>     * changes that reload function to the `bibtex-actions-refresh` interactive command
> 
>     * adds section to the README (just need to confirm those `file-notify` configs are right; may not be for the bib one if there is more than one bib file?)

I followed the README and added `file-notify` configs.  The cache got invalidated correctly.

bdarcus:
Merged; thanks much!

Let me know of course if you have other ideas, etc.

bdarcus:
PS - I'm glad I did this. It's nice to have the instant loading!

PR #76, opened 2021-04-12
refactor: add optional parameter to refresh command, update docs

### LINKED ISSUE #73: add watch functions/commands?
Not sure about this, but we could add a convenience function/command to add and remove file-notify watches?

I wouldn't want to have any hard-coded assumptions though.

For sake of possible illustration, maybe something like:

``` elisp
(bibtex-actions-add-watch)
(bibtex-actions-rm-watch)
```

... and an accompanying defcustom?

``` elisp
(setq bibtex-actions-watch (list bibtex-completion-bibliography bibtex-completion-notes-path))
```

Or alternately, we can out some of this on the wiki.

cc @wenjie2wang

bdarcus:
Update `bibtex-actions-refresh` to allow it be invoked as a callback on `file-notify-add-watch` without requiring a lambda.

Also update the README to incorporate the change.

Relates to #73.

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -192,7 +192,7 @@ If the cache is nil, this will load the cache."
       (bibtex-actions-refresh))
   bibtex-actions--candidates-cache)
 
-(defun bibtex-actions-refresh ()
+(defun bibtex-actions-refresh (&optional _event)
   "Reload the candidates cache."
   (interactive)
   (setq bibtex-actions--candidates-cache
--- a/README.md
+++ b/README.md
@@ -113,25 +113,23 @@ The `bibtex-actions-refresh` command will reload the cache, and you can call thi
 You can also add `bibtex-completion`-style proactive loading by using `filenotify` something like this:
 
 ``` emacs-lisp
-;; Of course, you could also use `bibtex-complation-bibliography` here, but would need 
+;; Of course, you could also use `bibtex-completion-bibliography` here, but would need 
 ;; to adapt this if you specify multiple files.
-(file-notify-add-watch "/path/to/file.bib"
-                       '(change)
-                       (lambda (event) (bibtex-actions-refresh)))
+(file-notify-add-watch 
+  "/path/to/file.bib" '(change) 'bibtex-actions-refresh)
 ```
 
 You can also extend this to do the same thing for your PDF files, or notes:
 
 ``` emacs-lisp
-(file-notify-add-watch bibtex-completion-library-path
-                       '(change)
-                       (lambda (event) (bibtex-actions-refresh)))
+(file-notify-add-watch 
+  bibtex-completion-library-path '(change) 'bibtex-actions-refresh)
 
-(file-notify-add-watch bibtex-completion-note-path
-                       '(change)
-                       (lambda (event) (bibtex-actions-refresh)))
-
+(file-notify-add-watch 
+  bibtex-completion-note-path '(change) 'bibtex-actions-refresh)
 ```
+
+For additional configuration options on this, see [the wiki](https://github.com/bdarcus/bibtex-actions/wiki/Configuration#automating-path-watches).
 
 ## Usage
 


bdarcus:
@wenjie2wang - couple of things on this, if you have time and interest:

1. I have updated the "refresh" command to remove the lambda, and so also changed the README instructions.
2. I have added the two "watch" commands (but I'm not sure they work correctly ATM). WDYT of this idea?

bibtex-actions.el
```diff
@@ -264,6 +275,29 @@ TEMPLATE."
               (- width (cddr format)))
             0 ?\s)))))))
 
+;;; Convenience commands
+
+(defun bibtex-actions-add-watches ()
+  "Add watches using 'file-notify-add-watch'.
+For the paths specified in 'bibtex-actions-watch-list' will add a
+watch that runs 'bibtex-actions-refresh' upon a change."
+  ;; TODO return message of success or error
```
bdarcus, line 284: I don't know how to do this!

wenjie2wang:
Thanks, @bdarcus !

> I have updated the "refresh" command to remove the lambda, and so also changed the README instructions.

It is great to have a simplified interface.

> I have added the two "watch" commands (but I'm not sure they work correctly ATM). WDYT of this idea? I'm still not sure.

I can help test them after work (I wish I knew more about Elisp and help you more on it).  IMHO, the current `file-notify-add-watch` approach works well.  As a user, I do not mind adding those configurations to my `.emacs.d` as long as it is documented in the README.

bdarcus:
> As a user, I do not mind adding those configurations to my `.emacs.d` as long as it is documented in the README.

Probably if in doubt, I shouldn't include the commands at this point then.

> I can help test them after work.

OK, thanks.

It could be that I remove the commands, etc., and simply update the refresh command and README.

Let me know WYT after you have a chance to test.

bdarcus:
Another option, if we get it working, is to include an example in the README like what I'm trying to do here:

https://github.com/bdarcus/.doom.d/commit/f4b67c087e9294d94017b8268ea05ff9132e92fd

wenjie2wang:
Hi @bdarcus 

1. The `bibtex-actions-refresh` worked well. 
2. I also tried `bibtex-actions-watch-list` as follows:
```emacs-lisp
(setq bibtex-actions-watch-list
      (list bibtex-completion-bibliography
            bibtex-completion-library-path))
(mapc (lambda (x)
        (file-notify-add-watch x '(change)
                               'bibtex-actions-refresh))
      (-flatten bibtex-actions-watch-list))
```
It also worked.

However, it seems that whenever I add/remove entries to/from the bib file, the `bibtex-completion-candidates` is called twice (I checked the `*Message* buffer`), which is probably unnecessary.  I guess that the additional call is triggered by the `file-notify-add-watch` set in the `bibtex-completion.el`.  Therefore, I looked into the `bibtex-completion.el` and tried to remove the extra file watcher as follows:
```emacs-lisp
;; Remove current watch-descriptors for bibliography files:
(mapc (lambda (watch-descriptor)
        (file-notify-rm-watch watch-descriptor))
      bibtex-completion-file-watch-descriptors)
(setq bibtex-completion-file-watch-descriptors nil)
```
This code chunk was directly taken from the `bibtex-completion.el` and worked for me.  I am not sure if you observed the same issue.  If so, I believe it would be reasonable to keep only one file watch (given that the caching by `bibtex-completion-candidates` somehow does not work in this case).

bdarcus:
> I guess that the additional call is triggered by the `file-notify-add-watch` set in the `bibtex-completion.el`.

Yeah, we call two different callback on the same paths.

But note: in bibtex-actions, I no longer use the bibtex-completion-init function, which is what sets those up.

So that should only happen if you use both bibtex-actions and ivy/helm-bibtex.

So in the end, you think once I remove the new commands, this is ready to merge?

wenjie2wang:
> So that should only happen if you use both bibtex-actions and ivy/helm-bibtex.

You are right.  If I only use bibtex-actions, there will not be such an issue.

> So in the end, you think once I remove the new commands, this is ready to merge?

Yes, this PR looks good to me.  Thanks!

bdarcus:
Hmm ... I think maybe the refresh command needs to clear the bibtex-completion cache?

bdarcus:
Oops, sorry @wenjie2wang, I accidentally deleted your earlier comment with the code example you used from bibtex-completion.

I had want to reply to note since that code is now on the wiki, we (and others) can experiment with how to improve it.

What I merged ended super simple then: just makes the setup simpler.

PR #77, opened 2021-04-13
refactor: move template processing

bdarcus:
This fixes so that the template processing function is only called once,
rather than every time the format-entry function runs.


bdarcus:
Sigh ... merged this too soon.

I think I want to add a new function while fixing this: `bibtex-actions-format-candidate`.

bdarcus:
Sigh ... merged this too soon.

I think I want to add a new function while fixing this: `bibtex-actions-format-candidate`.

PR #84, opened 2021-04-14
fix: field truncate-to-width on format-entry

### LINKED ISSUE #67: format-entry widths
#66 fixed one problem, but introduced another minor one.

The default suffix template has `${tags:*}`, which should extend the last column to the end.

But it's truncated much earlier (if you see below, it seems like the tags are truncated at about 27 characters).

Screenshot from 2021-04-13 12-41-19

Must be because we no longer assume a single template.

Here's what I see in `ielm`:

``` elisp
ELISP> (bibtex-actions--process-display-formats bibtex-actions-template-suffix)
((t "          ${=key=:15}    ${=type=:12}    ${tags:*}" . 45))
ELISP> bibtex-completion-display-formats-internal
((t "${author:36} ${title:*} ${year:4} ${=has-pdf=:1}${=has-note=:1} ${=type=:7}" . 53))
```

bdarcus:
fix #67

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -262,20 +262,18 @@ TEMPLATE."
          (if (not field-width)
              field-value
            (setq field-width (string-to-number field-width))
-           (truncate-string-to-width
+           (bibtex-actions--truncate-field
             field-value
-            ;; When converted to number, "*" is 0.
-            (if (> field-width 0)
-                field-width
-              ;; This is the line that specifies the width where the field width
-              ;; is specified as "*".
-              (- width (cddr format)))
-            0 ?\s)))))))
+            field-width
+            width)))))))
 
-(defun bibtex-actions--truncate (str width)
-  "Truncate STR to WIDTH."
-  ; placeholder ATM
-  (truncate-string-to-width str width 0 32 ?\s))
+(defun bibtex-actions--truncate-field (str field-width template-width)
+  "Truncate STR to FIELD-WIDTH.
+If the value of WIDTH is 0, will fill the TEMPLATE-WIDTH."
+  (let ((width (if (> field-width 0)
+                   field-width
+                 template-width)))
+    (truncate-string-to-width str width 0 32 ?\s)))
 
 ;;; Command wrappers for bibtex-completion functions
 
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -264,8 +264,11 @@ TEMPLATE."
            (setq field-width (string-to-number field-width))
            (truncate-string-to-width
             field-value
+            ;; When converted to number, "*" is 0.
             (if (> field-width 0)
                 field-width
+              ;; This is the line that specifies the width where the field width
+              ;; is specified as "*".
               (- width (cddr format)))
             0 ?\s)))))))
 
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -263,6 +263,8 @@ TEMPLATE."
              field-value
            (setq field-width (string-to-number field-width))
              (let ((width (if (> field-width 0)
+                              ;; If user specifies field width of "*", use
+                              ;; WIDTH; else use the explicit 'field-width'.
                               field-width
                             width)))
                (truncate-string-to-width field-value width 0 ?\s))))))))
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -47,7 +47,7 @@
   :group 'bibtex-actions)
 
 (defcustom bibtex-actions-template
-  '((t . "${author:20}   ${title:48}   ${year:4}"))
+  '((t . "${author:20}   ${title:*}   ${year:4}"))
   "Configures formatting for the BibTeX entry.
 When combined with the suffix, the same string is used for
 display and for search."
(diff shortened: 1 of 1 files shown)


bibtex-actions.el
```diff
@@ -266,8 +274,8 @@ TEMPLATE."
             field-value
             (if (> field-width 0)
                 field-width
-              (- width (cddr format)))
-            0 ?\s)))))))
+              (- final-width width)
```
bdarcus, line 277: This seems to be the line that causes the problem, but this is not the solution.

bdarcus:
@nobiot - I think I fixed this, but don't entirely understand how! It included borrowing some code from the marginalia truncate function, and creating my own.

nobiot:
@bdarcus 

I guess why this is a bit hard to understand is in this:

`(truncate-string-to-width str width 0 32 ?\s)`

"32" in it is not an integer but the same as "?\s", [a space char](https://www.gnu.org/software/emacs/manual/html_node/elisp/Basic-Char-Syntax.html). So it is the same as this:

`(truncate-string-to-width str width 0 ?\s ?\s)`

I think you can omit the last optional argument, "ellipsis".

`(truncate-string-to-width str width 0 ?\s)`

Then it reads more easily, does it not?

It returns the string in the "width", either chopping too long a string or padding too short a string with spaces added at the end.

bdarcus:
@nobiot - OIC. Yes, I'll obviously remove that 32. 

Thank you!

But doesn't explain why this works, but the code it replaces doesn't. Is it obvious to you?

nobiot:
Sorry, I see your question now...
hmm...

The code is a bit hard to read now in the diff mode in GH. 

Intuitively without examination, it seems that the old way chops off ending spaces for suffix.
The new way doesn't.

OLD: `(- width (cddr format))`
NEW: `template-width` (coming from the `width` value above).

I thought that it was simply that OLD was wrong, and the NEW is correct for that case?
I cannot see how `width` is calculated, but I assume it is done different from `(- width (cddr format))`.
Is it simply that subtracting `(cddr format)` from `width` was unnecessarily?

nobiot:
So... What I seem to have implied is...
Can we simply replace `(- width (cddr format))` with `width` in the original form?

It seems to be working (?)
What do you think?

```
(truncate-string-to-width
            field-value
            (if (> field-width 0)
                field-width
              width)
            0 ?\s)
```

image

The tags is deliberately very long -- see the bib entry I created below.
```
@online{zotero-748,
  title = {Logitech Announces Carbon Labeling},
  url = {https://www.yahoo.com/lifestyle/logitech-announces-carbon-labeling-120514350.html},
  urldate = {2021-03-19},
  abstract = {Logitech has announced a new labeling system to inform consumers about the environmental impact of its products, which will be rolled out for its entire catalogue by 2025.  The computer peripherals specialist plans to introduce the new labels, which will initially feature on its gaming products, by the end of the year.  At the same time, Logitech will also provide access to the methodology and protocol applied to calculate carbon footprints on its website.},
  file = {C\:\\Users\\nobiot\\Zotero\\storage\\6A5EASRH\\logitech-announces-carbon-labeling-120514350.html},
  langid = {american},
  tags = {tags corporate information that is too long to parse here next}
}
```

bdarcus:
@nobiot - I actually spoke too soon. 

Just to be sure this was working as expected, I added a "*" to the main template so that there's one on each. 

If this is working correctly, the table should extend to the end of the frame, but not beyond. It now extends beyond.

I'll experiment a bit to see if I can fix, and respond to your replies

Hopefully this is straightforward.

nobiot:
Ah, good catch.

bdarcus:
Maybe I just need tell people to only use one "*" in total :-)

bdarcus:
> I cannot see how width is calculated in the new way from the diff, but I assume it is done different from (- width (cddr format)).
> Is it simply that subtracting (cddr format) from width was unnecessarily?

Maybe that's indeed the case.

> Can we simply replace `(- width (cddr format))` with `width` in the [original form](https://github.com/bdarcus/bibtex-actions/blob/main/bibtex-actions.el#L265-L270)?

The separate function isn't really needed, since it's only used in the one place; right?

I thought I had tried that in the original version, and that it didn't work.

Hence my confusion.

But it's possible I misremembered.

bdarcus:
> The separate function isn't really needed, since it's only used in the one place; right?
> 
> I thought I had tried that in the original version, and that it didn't work.

Just pushed that change, and it seems to work.

But it won't work right if there's more than one total "*". I think I'll call that good enough, and put a note about it in the README.

Does that sound reasonable to you @nobiot?

bdarcus:
> I haven't tested it, but it may be that ivy-bibtex and helm-bibtex have the same limitation.

Yes, it does!

Closing this then.

bdarcus:
> Just pushed that change, and it seems to work.

Interestingly, works correctly on selectrum, but not on vertico (table extends beyond frame).

But vertico is really new and under active development, so may not be my fault!

nobiot:
> But it won't work right if there's more than one total "*". I think I'll call that good enough, and put a note about it in the README.

> I haven't tested it, but it may be that ivy-bibtex and helm-bibtex have the same limitation.

> Does that sound reasonable to you @nobiot?

Works for me :) I am only a very light user of Bibtex but I like the approach this package has taken (because I have moved away from Ivy and Helm). 

I use Vertico but didn't even notice the problem. I don't use tags in my .bib file and even struggled to get them to show (with additional search field variable from `bibtex-completion`). So... for this weak usage, I have had no issue :)

Issue #91, opened 2021-04-15
Add completion-at-point


bdarcus:
## Overview

Perhaps at-point completion would be more efficient for the more narrow task of citation and/or key insertion when writing, particularly for the [forthcoming new org-mode citation support](https://lists.gnu.org/archive/html/emacs-orgmode/2021-04/msg00227.html), something like you see in the Zettlr markdown editor?

![Screenshot from 2021-04-22 04-07-16](https://user-images.githubusercontent.com/1134/115712598-f647c600-a342-11eb-9401-5aebfb56c4bf.png)

Don't really know anything about this, but thought of it because of this new package.

https://github.com/minad/corfu

And there's, of course, `company`.

If the feature makes sense, I may experiment with it. But a PR would be even better!

EDIT: see [this comment](https://www.reddit.com/r/emacs/comments/msx6s1/elpa_corfu_completion_overlay_region_function/guyney5?utm_source=share&utm_medium=web2x&context=3) for limitations and possibilities, in particular the note you can make it possible move from at-point to the minibuffer.

## Possible implementation approach

If I understand how the API works, we'd probably want something like the following functions:

Edited based on discussion in comments.

1. `bibtex-actions-complete-key-at-point` (as with Zettlr screenshot above; the easiest and most general, and will cover 90% or so of need, and work with both org and pandoc syntax)
2. `bibtex-actions-complete-citation-at-point` (specifically for new org cite support, which will have more complex syntax and features; see #108)

## Org-cite syntax

Here's the new syntax:

``` 
 [cite/style:common prefix ;prefix @key suffix; ... ; common suffix]
```
Goal is to make citation entry really smooth and efficient, then.

So I'm thinking user experience would be something like:

- type `[c` and that can auto-complete to `[cite`
- type `/` and presents pop-up options of styles, completes that, and places point after this `[cite/style:`
- type `@` and list of citation keys pops up for selection (this seems the only piece that would be necessary)
- etc.

minad:
I guess many people would like such a functionality. If you have Corfu installed it is pretty simple to get something running. However this won't work with Company since as far as I know, Company does not provide a completion in region function. Company only supports completion at point functions.

~~~ elisp
(defun bibtex-complete ()
  (interactive)
  (pcase (preceding-char)
    (?@ (completion-in-region (- (point) 1) (point) '("@key1" "@key2" "@key3" "@key4")))
[...]

PR #113, opened 2021-04-22
Add org-cite function

### LINKED ISSUE #108: Add support for new org cite
While it's not done or merged yet, this thread suggests it likely will be "soon" (the syntax is basically done, as is a draft API).

https://lists.gnu.org/archive/html/emacs-orgmode/2021-04/msg00430.html

When it is, citation insertion would need to be adapted to take advantage of it, ideally in bibtex-completion. I don't believe  the `bibtex-completion-insert-citation` function even needs to be changed to do this, given it has customization capabilities; see:

https://github.com/tmalsburg/helm-bibtex#format-of-citations

#113 is adding this.

I'd expect there would be overlap with `org-ref` too.

It seems, however, there will be opportunities here too, regardless of what code ends up where. 

See, for example, #91, which may we the place to start.

PRs welcome, as I'm not sure I have the time or skill to do this ATM.

Aside: this project is under more active current development than bibtex-completion et al. So at times it makes sense to add code here that might later be transferred over to the former later. We use the same license, so that won't be any impediment.

bdarcus:
## Introduction

It looks like `org-cite` is close to merging.

https://lists.gnu.org/archive/html/emacs-orgmode/2021-04/msg00790.html

So this implements a function to format that syntax.

You can use this init file to test this with the org-cite branch code:

https://gist.github.com/bdarcus/2645f99363fc47ddab2aae24c5d9e66c

Here's what the style prompt looks like:

Screenshot from 2021-05-06 09-00-34

... with results when inserting into the `new-org-cite` org branch:

Screenshot from 2021-05-03 13-50-04

## Details

Bibtex-completion has flexible citation formatting functions.

Here is the default currently:

``` elisp
(defcustom bibtex-completion-format-citation-functions
  '((org-mode      . bibtex-completion-format-citation-ebib)
```

So this will hopefully replace `bibtex-completion-format-citation-ebib` as default.

I adapted this from the pandoc function, compared to which I have:

1. changed "prenote" to "prefix" to match `org-cite` language
2. removed "postnote" (because this is mostly used for things like page numbers, which are specific to an individual cite item; users will have to add this manually after insertion)
3. added `org-cite`-specific doc string and examples
4. defined a default style, which is ignored, and others, which are configurable, and which result in `[cite/text:@doe19;@doe20;@doe21]`.

## Issues

In general, I'm a little uncertain on the UX, but I have decided to keep this simple, and have users manually edit the inserted citation for more complex cases.

This is in part related to #91. It might be nice to support this with `completion-at-point`.

### Affixes

`Org-cite` has two-levels:

- one for the citation as a whole
- another for each individual citation within that.

I elected only to initially support prompting for the first, and only for the prefix.

### Styles

The styles currently overlap with those in the `oc-basic` POC, which may end up the fallback.

Regardless, I think merging this may have to wait on andras-simonyi/citeproc-org#16.

Closes #108.

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -64,7 +64,7 @@ display and for search."
 
 (defvar bibtex-actions-org-cite-commands
   ;; TODO make defcustom?
-  '("text" "year" "title" "author" "locators" "nocite" "plain")
+  '("text" "year" "title" "author" "locators" "nocite")
     "Org citation commands, and command descriptions.")
 
 (defcustom bibtex-actions-link-symbol "🔗"
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -62,9 +62,9 @@ display and for search."
     :group 'bibtex-actions
     :type  '(alist :key-type symbol :value-type function))
 
-(defcustom bibtex-actions-org-cite-commands
-  '("text" "year" "title" "author" "locators" "nocite")
-  "Org citation commands, and command descriptions."
+(defcustom bibtex-actions-org-cite-styles
+  '("text" "noauthor" "title" "author" "locators" "nocite")
+  "Org citation styles."
   :group 'bibtex-actions
   :type '(repeat string))
 
@@ -163,7 +163,7 @@ specific to the item, rather than the citation as a whole.
 
   [cite: see ;@key pp23-24]"
   (let* ((prefix  (if bibtex-completion-cite-prompt-for-optional-arguments (read-from-minibuffer "Prefix: ") ""))
-         (styles bibtex-actions-org-cite-commands)
+         (styles bibtex-actions-org-cite-styles)
          (style  (if bibtex-completion-cite-prompt-for-optional-arguments
                      (ido-completing-read "Style: " styles nil nil nil nil "default")))
          (prefix  (if (string= "" prefix)  "" (concat prefix  " ;")))
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -62,10 +62,11 @@ display and for search."
     :group 'bibtex-actions
     :type  '(alist :key-type symbol :value-type function))
 
-(defvar bibtex-actions-org-cite-commands
-  ;; TODO make defcustom?
+(defcustom bibtex-actions-org-cite-commands
   '("text" "year" "title" "author" "locators" "nocite")
-    "Org citation commands, and command descriptions.")
+  "Org citation commands, and command descriptions."
+  :group 'bibtex-actions
+  :type '(repeat string))
 
 (defcustom bibtex-actions-link-symbol "🔗"
   "Symbol to indicate a DOI or URL link is available for a publication.
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -62,6 +62,11 @@ display and for search."
     :group 'bibtex-actions
     :type  '(alist :key-type symbol :value-type function))
 
+(defvar bibtex-actions-org-cite-commands
+  ;; TODO make defcustom?
+  '("text" "year" "title" "author" "locators" "nocite" "plain")
+    "Org citation commands, and command descriptions.")
+
 (defcustom bibtex-actions-link-symbol "🔗"
   "Symbol to indicate a DOI or URL link is available for a publication.
 This should be a single character."
@@ -84,6 +89,14 @@ may be indicated with the same icon but a different face."
   :group 'bibtex-actions
   :type 'string)
 
+(setq bibtex-completion-format-citation-functions
+  '((org-mode      . bibtex-actions--format-citation-org)
+    (latex-mode    . bibtex-completion-format-citation-cite)
+    (markdown-mode . bibtex-completion-format-citation-pandoc-citeproc)
+    (python-mode   . bibtex-completion-format-citation-sphinxcontrib-bibtex)
+    (rst-mode      . bibtex-completion-format-citation-sphinxcontrib-bibtex)
+    (default       . bibtex-completion-format-citation-pandoc-citeproc)))
+
 ;;; Keymap
 
 (defvar bibtex-actions-map
@@ -102,6 +115,40 @@ may be indicated with the same icon but a different face."
     (define-key map (kbd "r") 'bibtex-actions-refresh)
     map)
   "Keymap for 'bibtex-actions'.")
+
+;;; Org-cite citation function
+
+(defun bibtex-actions--format-citation-org (keys)
+  "Format org-cite citations for the entries in KEYS.
+
+The full citation syntax is:
+
+  [cite/style:common prefix ;prefix @key suffix; ... ; common suffix]
+
+Everything is optional, except the brackets, 'cite' and the colon.
+Also the citation must contain at least a key.
+So its minimal form is:
+
+  [cite:@key]
+
+Note that the prefix here is for the citation as a whole; if you
+need to add one specific to an individual citation item, you
+would need to add that after inserting.
+
+  [cite:see ;item prefix @key]
+
+The same is true for suffixes like page numbers, which are
+specific to the item, rather than the citation as a whole.
+
+  [cite: see ;@key pp23-24]"
+  (let* ((prefix  (if bibtex-completion-cite-prompt-for-optional-arguments (read-from-minibuffer "Prefix: ") ""))
+         (styles bibtex-actions-org-cite-commands)
+         (style  (if bibtex-completion-cite-prompt-for-optional-arguments
+                     (ido-completing-read "Style: " styles nil nil nil nil "default")))
+         (prefix  (if (string= "" prefix)  "" (concat prefix  " ;")))
+         (style  (if (string-equal style "default") "" (concat "/" style))))
+    ;; this is derived from the pandoc syntax function, but this has two-levels of affixes
+    (format "[cite%s:%s%s]" style prefix (s-join ";" (--map (concat "@" it) keys)))))
 
 ;;; Completion functions
 (defun bibtex-actions-read ()


bibtex-actions.el
```diff
@@ -103,6 +103,31 @@ may be indicated with the same icon but a different face."
...
+  [cite:@key]
+
+... or its 'suppress author' variant:
+
+  [cite:-@key]"
+  (let* ((prenote  (if bibtex-completion-cite-prompt-for-optional-arguments (read-from-minibuffer "Prenote: ") ""))
+         (postnote (if bibtex-completion-cite-prompt-for-optional-arguments (read-from-minibuffer "Postnote: ") ""))
+         (prenote  (if (string= "" prenote)  "" (concat prenote  " ")))
+         (postnote (if (string= "" postnote) "" (concat ", " postnote))))
+    ;; this is derived from the pandoc syntax function, but this has two-levels of affixes
```
bdarcus, line 128: UX question: does it make sense to prompt users for affixes on each cite, or is that just going to be annoying?

bdarcus:
Here's the completing-read UI for the commands now.

Screenshot from 2021-04-22 18-42-17

bibtex-actions.el
```diff
@@ -103,6 +103,47 @@ may be indicated with the same icon but a different face."
...
+    (format "[cite%s:%s%s%s]" command prenote (s-join ";" (--map (concat "@" it) keys)) postnote)))
+
+(defun bibtex-actions--citation-commands-completion-table (commands)
+  "Create the completion table for COMMANDS."
+  (lambda (string pred action)
+    (if (eq action 'metadata)
+        `(metadata
+          ;; better here to use x-group?
+          (display-sort-function . ,#'identity)
+          (annotation-function . (mapcar (lambda (command) (cdr command) commands)))))
```
bdarcus, line 144: Not currently working.

bibtex-actions.el
```diff
@@ -103,6 +109,53 @@ may be indicated with the same icon but a different face."
...
+    (format "[cite%s:%s%s]" command prefix (s-join ";" (--map (concat "@" it) keys)))))
+
+(defun bibtex-actions--citation-commands-read (commands default)
+  "Prompt user for citation command/style from COMMANDS, and define DEFAULT."
+  (completing-read
+   "Style: "
+   (lambda (string pred action)
+     (if (eq action 'metadata)
+         '(metadata
+           (affixation-function . (mapcar
```
bdarcus, line 153: Not currently working.

denismaier:
Let me comment on the current user interface:

First: I think it's totally reasonable to let users edit the
source. In that case you might even want to remove the prefix prompt as
this might confuse some users.

But, if you want to add more support here, what about this:
- user calls bibtex-actions-insert-citation
- minibuffer opens
- ask for a style
- user selects one (!) item, presses enter
- prompt for prefix and suffix for this item
- minibuffer shows the list of items again
- repeat
- Finish the whole process with C-RET

If you think the last part sucks, that could be a distinct
multi-citation function.
So for single item citations, it's:

- user calls bibtex-actions-insert--single-citation
- minibuffer opens
- ask for a style
- user selects item, presses enter
- prompt for prefix and suffix for this item

Global affixes are by far less common then prefixes and suffixes on
individual items in a citation, so I think such an approach could make
sense. Of course, you could still add an option to add global affixes,
perhaps with a prefix function, or a users selects the style with C-RET
or so. That could give you:

- user calls bibtex-actions-insert-citation-with-global-affixes
- minibuffer opens
- ask for a style
- prompt for global prefix
- user selects one (!) item, presses enter
- prompt for prefix and suffix for this item
- minibuffer shows the list of items again
- repeat
- Finish the whole process with C-RET
- prompt for global suffix

bdarcus:
Thanks for the detailed feedback @denismaier!

I think what you suggest is worth experimenting with; hard to know for sure in the abstract.

On this, just to clarify:

> Global affixes are by far less common then prefixes and suffixes on
> individual items in a citation

It's probably confusing, but I was thinking about this for a specific case: "see" references.

I think in that case, they are global; no?

E.g. (for more on foo, see Doe, 2019; Roberts, 2020).

denismaier:
> I think in that case, they are global; no?
> 
> E.g. `(for more on foo, see Doe, 2019; Roberts, 2020)`.
> 
> Edit: and these are pretty common.

Correct. The problem with them is just that you don't really support for that in most backends. Biblatex has support for this. Citeproc-org/citeproc-el, already?

But if you use natbib, relying on a global prefix might give you headaches. OTOH, resolving these issues might be the job of the export framework.

bdarcus:
> Edit: for export to natbib, can just prepend the global prefix to the first cite prefix?

Actually, while you raise this, we should figure out if we want to provide some feedback to Nicolas on this.

Right now, oc-natbib ignores the global prefix. Should it?

Maybe you can think about that and incorporate it into whatever you might post on the org list?

denismaier:
> Edit: for export to natbib, can just prepend the global prefix to the first cite prefix?

That should be possible, yes.

> Right now, oc-natbib ignores the global prefix. Should it?

No, merging would be better.

denismaier:
FWIW, flexible input is as easy as this:

```lisp
(defun bibtex-actions--insert-cite-fragment ()
 "Insert a cite fragment"
  (interactive)
  (let* ((styles bibtex-actions-org-cite-commands)
         (style  (if bibtex-completion-cite-prompt-for-optional-arguments
                     (ido-completing-read "Style: " styles nil nil nil nil "default")))
         (style  (if (string-equal style "default") "" (concat "/" style))))
    (insert (format "[cite%s: " style))))

(defun bibtex-completion-format-citation-default (keys)
  "Default formatter for keys, separates multiple keys in KEYS with commas."
  (s-join ", " (--map (concat "@" it) keys)))
```

- call bibtex-actions--insert-cite-fragment
- choose style
- start typing; global prefix
- add the first citation key with bibtex-actions-insert-cite-key
- add prefix and suffix manually
- repeat
- add global suffix
- close the brackets

denismaier:
> That would be more efficient; no?

Sure. That was just a quick and dirty poor man's solution.

bdarcus:
Thanks for proposing it; it does demonstrate the design challenges I was thinking about.

Ideally, I really want an approach that works really well for both single-cite citations and multi-cite; that provides as little friction as possible for writers.

bdarcus:
> #91 can be used to do a lot of this.

I started #132 to address this.

bdarcus:
> ... and that's it. Anything else is added inline?

I'm a little unclear of dividing line between org completion, and third-party.

In any case, seems for inline there really two capf actions:

1. the citation; e.g. type `[c` and cite completes, with a pop-up for the style.
2. insert key

denismaier:
Just a thought: what's the performance hit of these autocompletion frameworks?

bdarcus:
> Just a thought: what's the performance hit of these autocompletion frameworks?

Likely insignificant.

The capf function I'm playing with would use the same cached completion candidates as the other commands, so should be very performant.

bdarcus:
I updated this to use the "noauthor" instead of "year" style now.

denismaier:
Shouldn't there be a space after the colon?

bibtex-actions.el
```diff
@@ -122,6 +137,40 @@ manager like Zotero or JabRef."
...
+
+  [cite: see ;@key pp23-24]"
+  (let* ((prefix  (if bibtex-completion-cite-prompt-for-optional-arguments (read-from-minibuffer "Prefix: ") ""))
+         (styles bibtex-actions-org-cite-styles)
+         (style  (if bibtex-completion-cite-prompt-for-optional-arguments
+                     (ido-completing-read "Style: " styles nil nil nil nil "default")))
+         (prefix  (if (string= "" prefix)  "" (concat prefix  " ;")))
+         (style  (if (string-equal style "default") "" (concat "/" style))))
+    ;; this is derived from the pandoc syntax function, but this has two-levels of affixes
+    (format "[cite%s:%s%s]" style prefix (s-join ";" (--map (concat "@" it) keys)))))
```
denismaier, line 172: suggested change:
```elisp
(format "[cite%s: %s%s]" style prefix (s-join ";" (--map (concat "@" it) keys)))))
```

WDYT?

bdarcus:
> Shouldn't there be a space after the colon?

Can you attach the comment to the specific line in question? I'm not sure what you're referring to ;-)

denismaier:
> Can you attach the comment to the specific line in question? I'm not sure what you're referring to ;-)

I've just added an inline comment.

denismaier:
Hmm, there's no space in Nicholas's original post describing the syntax. But I think it's more readable with a space. (And it doesn't seem to affect to export.

bibtex-actions.el
bdarcus: I updated your comment to use the github "suggested" feature; note the difference in the source. You can access it by clicking the left most icon on the toolbar.

But on the substance, I don't believe so. Wouldn't that result in `[cite: @doe]`?

bdarcus:
> Hmm, there's no space in Nicholas's original post describing the syntax. But I think it's more readable with a space. (And it doesn't seem to affect to export.

Right; the delimiters are online colons and semi-colons. The current behavior is correct, and I think what you're proposing could actually screw things up if there is a prefix.

denismaier:
> Right; the delimiters are only colons and semi-colons. The current behavior is correct, and I think what you're proposing could actually screw things up if there is a prefix.

Could be, yes. Maybe we should check with Nicolas once things are a bit more stable. (Not a high priority issue of course...)

bdarcus:
> Just a thought: what's the performance hit of these autocompletion frameworks?

FYI, @denismaier, I've made enough progress on #91 to confirm performance will not be an issue.

Issue #102, opened 2021-04-16
test: update test.bib to better test UI


bdarcus:
Given we now have a nice testing script, seems like we should revise the test.bib file etc to better test out the UI; things like:

- tags
- PDFs/notes

PR #118, opened 2021-04-28
Add filtering on open commands

bdarcus:
@apc - here's the start on #116.

It allows `bibtex-actions-read` to be called like this:

``` elisp
(bibtex-actions-read :initial "has:pdf")
```

So I changed the read function to include the `:initial` parameter, and also added the "initial-value" arguments to the relevant open commands.

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -105,7 +105,7 @@ may be indicated with the same icon but a different face."
 
 ;;; Completion functions
 (cl-defun bibtex-actions-read (&optional &key initial)
-  "Read bibtex-completion entries with INITIAL options.
+  "Read bibtex-completion entries with INITIAL option.
 
 This provides a wrapper around 'completing-read-multiple', with
 the following optional arguments:
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -128,8 +128,8 @@ the following optional arguments:
              ;; Collect citation keys of selected candidate(s).
              collect (cdr (assoc choice candidates)))))
 
-(cl-defun bibtex-actions--filter (str)
-  "A predicate function to filter candidates not containing STR."
+(defun bibtex-actions--filter (candidate &optional str)
+  "A predicate function to filter a CANDIDATE not containing STR."
   ;; want to use 'get-property' to grab the 'hidden' string
   ;; or to generalize, may need more
   ;; along with 'seq-filter' or '-keep' or '-filter'
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -286,7 +286,7 @@ TEMPLATE."
 Opens the PDF(s) associated with the KEYS.  If multiple PDFs are
 found, ask for the one to open using ‘completing-read’.  If no
 PDF is found, try to open a URL or DOI in the browser instead."
-  (interactive (list (bibtex-actions-read :initial "has:link has:pdf")))
+  (interactive (list (bibtex-actions-read :initial "has:link has:pdf ")))
   (bibtex-completion-open-any keys))
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


apc:
> Edit: on reflection, I think the initial value approach is better. This is because, when using embark, you can change the action command on the fly, so would want to have quick access to the full candidate list, without running another command.
> 
> But you can certainly experiment and report back ;-)

I agree that it would be better to have quick access to the full candidate list at all times, and if the initial value approach is the only one that allows for that, then that's the way to go!

Although, for what it's worth, I got the sense from [this post](https://mail.gnu.org/archive/html/help-gnu-emacs/2018-05/msg00068.html) by Stefan Monnier that initial inputs are considered bad practice (I linked to this in the other thread, so you may have already seen this). This may not be true (I don't know), and even if it is, you may reasonably decide that it's the right approach here anyway (of course!). I just thought it was interesting that some at least seem to think it best to avoid the initial input approach. 

I'll checkout this branch and play around with it. Will report back if I find anything worth mentioning. 

Thanks again!

bdarcus:
> Although, for what it's worth, I got the sense from [this post](https://mail.gnu.org/archive/html/help-gnu-emacs/2018-05/msg00068.html) by Stefan Monnier that initial inputs are considered bad practice (I linked to this in the other thread, so you may have already seen this). This may not be true (I don't know), and even if it is, you may reasonably decide that it's the right approach here anyway (of course!).

I think I disagree with him for this case. Initial and default are for different purposes, and using the latter for this case wouldn't work. Default may be appropriate, it seems to me, if you have a small number of candidates.

The test will really be the UX though; how well it works, intuitive it feels, etc.

bdarcus:
Here's what an annotation function, which also takes a single candidate as input, looks like:

https://github.com/bdarcus/bibtex-actions/blob/a095d4328b3024d187c65530b990e8440643540b/bibtex-actions.el#L143

bdarcus:
My suggestion, @apc, is you test/get working the "initial" approach first. It may even work now.

If you think something about that is a deal-breaker, we can figure out the issue with the filter function.

apc:
It doesn't seem to be working. Even though `bibtex-actions--candidates-cache` is non-nil, calling `bibtex-actions-insert-citation` or `bibtex-actions-open-pdf` doesn't show any entries. 

image

I'll look around and see if I can figure out anything more helpful to say.

bdarcus:
I probably just did something stupid with argument order or something.

Neither of these work, but both should I think:

``` elisp
(bibtex-actions-read :initial "has:pdf")
(bibtex-actions-read)
```

apc:
This seems to work: 

image

bdarcus:
> There's a missing closing parenthesis on line 125 and an extra closing parenthesis on line 126. Also, I believe the arguments after the completion candidates should be `filter nil initial nil nil nil` and not `filter initial nil nil nil`.

Ah, lisp :-)

Can you add that as a ["suggested" line comment](http://haacked.com/archive/2019/06/03/suggested-changes/) please, so I can just click to commit?

bibtex-actions.el
```diff
@@ -116,11 +122,19 @@ may be indicated with the same icon but a different face."
                      `(metadata
                        (affixation-function . bibtex-actions--affixation)
                        (category . bibtex))
-                   (complete-with-action action candidates string predicate))))))
+                   (complete-with-action action candidates string predicate))
+                 filter initial nil nil nil))))
```
apc, line 126:
```suggestion
                   (complete-with-action action candidates string predicate)))
                 filter nil initial nil nil nil)))
```

apc (commented):
Is this what you need?

bdarcus:
> Is this what you need?

Yes; thank you; committed!

So it now works?

WDYT about the UX?

apc:
So far so good! It's simple enough: you see the "has:pdf" string, which you can then delete if you so desire, and you get the list of entries with PDFs, which is what would make sense to see as options when calling `bibtex-actions-open-pdf`. And the same with "has:link" and "has:note". And it seems to me the strings should be those, rather than the shorthands `:pdf` or `:p`, since those may not work in all setups. So: I like!

bdarcus:
I should be able to test it myself later today. But sounds like we have our simple solution.

I do think it makes sense to also add the filter function while we're at it, but we'll have to see how easy that is to get working!

bdarcus:
@apc - on discussion thread, I asked if ivy or helm-bibtex supported arbitary saved search, and they indeed do:

https://github.com/tmalsburg/helm-bibtex#predefined-searches

bdarcus:
@apc - I removed the filter function, and decided to deal with the question of predefined searches in #119.

I tried to test this, but I think I've run into a bug with my Emacs or something.

Could you please test this again and confirm it all works; that I didn't break something with my latest commit?

apc:
I just pulled the latest changes and restarted my Emacs. All seems to work fine on my end. What in particular were you having issues with?

bdarcus:
> What in particular were you having issues with?

I can't even run `emacs -Q` (the test `run.sh` script)! I get some weird error, that I'm not sure where it comes from.

But I did just test on a VM with Emacs 27, which reminds me we need a better test.bib file in the repo (per #102), as it doesn't have any links, notes, pdfs!

But testing did lead me to change one little thing: I just added a space after the initial input, so that the user doesn't have to add that.

Thanks; I'll merge this now.

Issue #119, opened 2021-04-28
Add predefined searches


bdarcus:
Along with #118, we could support arbitrary pre-defined searches also, as supported in `ivy-bibtex` and `helm-bibtex`.

https://github.com/tmalsburg/helm-bibtex#predefined-searches

How best to do this here, given we have no common entry point command?

We might need a generic `bibtex-actions` command that can be wrapped like this; so:

``` elisp
(defun my/bib-mypubs ()
   (bibtex-actions :initial "Doe"))
```

... or simply tell users to wrap their default action command. 

But if we did this, I guess I'd need an optional initial parameter on all commands.

Any other ideas?

Though I've tried to avoid macros, maybe this would be a good candidate?

OTOH, I'm not clear how useful this feature is, given how easy it is to run a search. The example of searching for your own pubs is not a very compelling one in my view.

Maybe could be useful for things like project-related searches though?

bdarcus:
@oantolin, @minad, @mohkale - if you have a few moments, any thoughts?

cc @apc

apc:
Could you have something like a prefix argument that, when passed to either of the functions, prompts the user to pick from a list of predefined searches?

bdarcus:
> Could you have something like a prefix argument that, when passed to either of the functions, prompts the user to pick from a list of predefined searches?

That would seem better from a UX perspective; no?

oantolin:
Do the various commands that serve as entry points share the same history? If not, they should. Then you can just prepopulate the history with the predefined searches, and people can use a command to search history with completion such as `consult-history`.

bdarcus:
> Do the various commands that serve as entry points share the same history? If not, they should. Then you can just prepopulate the history with the predefined searches, and people can use a command to search history with completion such as `consult-history`.

One of the problems I keep running up against is lack of examples of how to use these CR features. I have to dig through stackexchange and such to find nuggets.

Any pointers of where I can look for good examples of how to work with history per your (sounds like excellent) suggestion?

Maybe consult itself?

bdarcus:
OIC - adding history is really straightforward.

I'll open a PR.

PR #120, opened 2021-04-29
add bibtex-actions-history

bdarcus:
This adds history support.

It also may address #119, per suggestion from @oantolin in https://github.com/bdarcus/bibtex-actions/issues/119#issuecomment-828810012.

> ... you can just prepopulate the history with the predefined searches, and people can use a command to search history with completion such as `consult-history`.

Two questions, @oantolin:

First, to "prepopulate the history", Is it really so simple as the below?

``` elisp
(setq bibtex-actions-history '("analysis" "chem"))
```

Second, how to activate the history?

If I run `consult-history`, for example, in any random buffer, I get a "no history configured" for that mode error.

Would this be something where I would just need to turn it on, for say, `text-mode`?? If yes, how? And in this package, or in a suggested README config?

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -223,10 +223,14 @@ If the cache is nil, this will load the cache."
   (setq bibtex-actions--candidates-cache
         (bibtex-actions--format-candidates)))
 
+;;;###autoload
 (defun bibtex-actions-insert-preset ()
   "Prompt for and insert a predefined search."
   (interactive)
-  (when-let ((search (completing-read "Preset: " bibtex-actions-presets)))
+  (unless (minibufferp)
+    (user-error "Command can only be used in minibuffer"))
+  (when-let ((enable-recursive-minibuffers t)
+             (search (completing-read "Preset: " bibtex-actions-presets)))
     (insert search)))
 
 ;;; Formatting functions
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -306,7 +306,7 @@ TEMPLATE."
 Opens the PDF(s) associated with the KEYS.  If multiple PDFs are
 found, ask for the one to open using ‘completing-read’.  If no
 PDF is found, try to open a URL or DOI in the browser instead."
-  (interactive (list (bibtex-actions-read :initial "has:link has:pdf ")))
+  (interactive (list (bibtex-actions-read :initial "has:link\|has:pdf ")))
   (bibtex-completion-open-any keys))
 
 ;;;###autoload
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -133,7 +133,7 @@ candidate list"
                        (affixation-function . bibtex-actions--affixation)
                        (category . bibtex))
                    (complete-with-action action candidates string predicate)))
-                 nil nil initial 'bibtex-actions-history 'bibtex-actions-presets nil)))
+                 nil nil initial 'bibtex-actions-history bibtex-actions-presets nil)))
     (cl-loop for choice in chosen
              ;; Collect citation keys of selected candidate(s).
              collect (cdr (assoc choice candidates)))))
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -84,6 +84,9 @@ may be indicated with the same icon but a different face."
   :group 'bibtex-actions
   :type 'string)
 
+(defvar bibtex-actions-history nil
+  "Search history for `bibtex-actions'.")
+
 ;;; Keymap
 
 (defvar bibtex-actions-map
@@ -123,7 +126,7 @@ candidate list"
                        (affixation-function . bibtex-actions--affixation)
                        (category . bibtex))
                    (complete-with-action action candidates string predicate)))
-                 nil nil initial nil nil nil)))
+                 nil nil initial 'bibtex-actions-history nil nil)))
     (cl-loop for choice in chosen
              ;; Collect citation keys of selected candidate(s).
              collect (cdr (assoc choice candidates)))))
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -84,8 +84,13 @@ may be indicated with the same icon but a different face."
   :group 'bibtex-actions
   :type 'string)
 
+;;; History, including future history list.
+
 (defvar bibtex-actions-history nil
   "Search history for `bibtex-actions'.")
+
+(defcustom bibtex-actions-presets nil
+  "List of predefined searches.")
 
 ;;; Keymap
 
@@ -126,7 +131,7 @@ candidate list"
                        (affixation-function . bibtex-actions--affixation)
                        (category . bibtex))
                    (complete-with-action action candidates string predicate)))
-                 nil nil initial 'bibtex-actions-history nil nil)))
+                 nil nil initial 'bibtex-actions-history 'bibtex-actions-presets nil)))
     (cl-loop for choice in chosen
              ;; Collect citation keys of selected candidate(s).
              collect (cdr (assoc choice candidates)))))
@@ -215,6 +220,12 @@ If the cache is nil, this will load the cache."
   (interactive)
   (setq bibtex-actions--candidates-cache
         (bibtex-actions--format-candidates)))
+
+(defun bibtex-actions-insert-preset ()
+  "Prompt for and insert a predefined search."
+  (interactive)
+  (when-let ((search (completing-read "Preset: " bibtex-actions-presets)))
+    (insert search)))
 
 ;;; Formatting functions
 ;;  NOTE this section will be removed, or dramatically simplified, if and
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -90,7 +90,9 @@ may be indicated with the same icon but a different face."
   "Search history for `bibtex-actions'.")
 
 (defcustom bibtex-actions-presets nil
-  "List of predefined searches.")
+  "List of predefined searches."
+  :group 'bibtex-actions
+  :type '(repeat string))
 
 ;;; Keymap
 
--- a/README.org
+++ b/README.org
@@ -138,7 +138,7 @@ Here's how to configure it to use =all-the-icons=:
 
 *** History and Predefined searches
 
-=Bibtex-actions= has functionality similar to the [[(https://github.com/tmalsburg/helm-bibtex#p][predefined search]] in =helm-bibtex= and =ivy-bibtex=, but with a different implementation.
+=Bibtex-actions= has functionality similar to the [[(https://github.com/tmalsburg/helm-bibtex#p][predefined search]] functionality in =helm-bibtex= and =ivy-bibtex=, but with a different implementation.
 Rather than create a new command with the search terms as argument, you just set the =bibtex-actions-presets= variable, and add the strings you want to access:
 
 #+begin_src emacs-lisp
--- a/README.org
+++ b/README.org
@@ -50,10 +50,16 @@ Bibtex-actions is available for installation from [[https://melpa.org][MELPA]].
 To setup bibtex-actions using =use-package=, you can simply do:
 
 #+BEGIN_SRC emacs-lisp
-  (use-package bibtex-actions
-    :config
-    ;; Make the 'bibtex-actions' bindings available from `embark-act'.
-    (add-to-list 'embark-keymap-alist '(bibtex . bibtex-actions-map)))
+(use-package bibtex-actions
+  :bind (("C-c b" . bibtex-actions-insert-citation)
+         :map minibuffer-local-map
+         ("M-b" . bibtex-actions-insert-preset))
+  :after embark
+  :config
+  ;; Make the 'bibtex-actions' bindings available from `embark-act'.
+  (add-to-list 'embark-keymap-alist '(bibtex . bibtex-actions-map))
+  :custom
+  (bibtex-completion-bibliography '("~/bib/references.bib")))
 #+END_SRC
 
 Since most of the command logic resides in bibtex-completion, that is where to look for different [[https://github.com/tmalsburg/helm-bibtex#basic-configuration-recommended][configuration options]].
--- a/README.org
+++ b/README.org
@@ -136,6 +136,20 @@ Here's how to configure it to use =all-the-icons=:
        :group 'all-the-icons-faces)
 #+END_SRC
 
+*** History and Predefined searches
+
+=Bibtex-actions= has functionality similar to the [predefined search](https://github.com/tmalsburg/helm-bibtex#predefined-searches) feature in `helm-bibtex` and `ivy-bibtex`, but with a different implementation.
+Rather than create a new command with the search terms as argument, you just set the =bibtex-actions-presets= variable, and add the strings you want to access:
+
+#+begin_src emacs-lisp
+(setq bibtex-actions-presets '("one search string" "another search string"))
+#+end_src
+
+Internally, this is using the minibuffer "future history" functionality available in =completing-read=, which you access in =selectrum= by using =M-n= from the prompt.
+
+=Bibtex-actions= also preserves the history of your selections (see caveat below about multiple candidate selection though), also accessible in =selectrum=, but by using =M-p=.
+You can save this history across sessions by adding =bibtex-actions-history= to =savehist-additional-variables=.
+
 *** Proactive reloading of library
     :PROPERTIES:
     :CUSTOM_ID: proactive-reloading-of-library
--- a/README.org
+++ b/README.org
@@ -138,7 +138,7 @@ Here's how to configure it to use =all-the-icons=:
 
 *** History and Predefined searches
 
-=Bibtex-actions= has functionality similar to the [[(https://github.com/tmalsburg/helm-bibtex#p][predefined search]] functionality in =helm-bibtex= and =ivy-bibtex=, but with a different implementation.
+=Bibtex-actions= has functionality similar to the [[https://github.com/tmalsburg/helm-bibtex#p][predefined search]] functionality in =helm-bibtex= and =ivy-bibtex=, but with a different implementation.
 Rather than create a new command with the search terms as argument, you just set the =bibtex-actions-presets= variable, and add the strings you want to access:
 
 #+begin_src emacs-lisp
--- a/README.org
+++ b/README.org
@@ -81,6 +81,7 @@ One of the beauties of the new suite of completing-read packages is the flexibil
 For =bibtex-actions= to work correctly, and as you may have come to expect in `helm-bibtex` or `ivy-bibtex`, you need to install and configure either prescient or orderless.
 You can read more about this at the [[https://github.com/raxod502/selectrum#usage][selectrum README]], but here's an example using orderless with its [[https://github.com/oantolin/orderless#style-dispatchers][style dispatchers]].
 
+
 #+CAPTION: orderless matching
 [[file:images/orderless.png]]
 
--- a/README.org
+++ b/README.org
@@ -157,9 +157,12 @@ Rather than create a new command with the search terms as argument, you just set
 (setq bibtex-actions-presets '("one search string" "another search string"))
 #+end_src
 
-Internally, this is using the minibuffer "future history" functionality available in =completing-read=, which you access in your completion UI by using =M-n= from the prompt.
+You then have two ways to access these strings from the completion prompt:
 
-=Bibtex-actions= also preserves the history of your selections (see caveat below about multiple candidate selection though), also accessible in your completion UI, but by using =M-p=.
+1. by using =M-n= from the prompt, which will cycle through the strings
+2. by calling =bibtex-actions-insert-preset= with a keybinding, and then selecting the string
+
+=Bibtex-actions= also preserves the history of your selections (see caveat below about multiple candidate selection though), which are also accessible in your completion UI, but by using =M-p=.
 You can save this history across sessions by adding =bibtex-actions-history= to =savehist-additional-variables=.
 
 *** Proactive reloading of library
--- a/README.org
+++ b/README.org
@@ -138,7 +138,7 @@ Here's how to configure it to use =all-the-icons=:
 
 *** History and Predefined searches
 
-=Bibtex-actions= has functionality similar to the [predefined search](https://github.com/tmalsburg/helm-bibtex#predefined-searches) feature in `helm-bibtex` and `ivy-bibtex`, but with a different implementation.
+=Bibtex-actions= has functionality similar to the [[(https://github.com/tmalsburg/helm-bibtex#p][predefined search]] in =helm-bibtex= and =ivy-bibtex=, but with a different implementation.
 Rather than create a new command with the search terms as argument, you just set the =bibtex-actions-presets= variable, and add the strings you want to access:
 
 #+begin_src emacs-lisp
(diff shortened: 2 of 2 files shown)


oantolin:
> First, to "prepopulate the history", Is it really so simple as the below?
> 
> ```emacs-lisp
> (setq bibtex-actions-history '("analysis" "chem"))
> ```

Yes, that's all you need to do!

> Second, how to activate the history?
> 
> If I run `consult-history`, for example, in any random buffer, I get a "no history configured" for that mode error.
> 
> Would this be something where I would just need to turn it on, for say, `text-mode`?? If yes, how? And in this package, or in a suggested README config?

History is only active at certain times, when you are in the minibuffer or at a shell prompt, a REPL prompt, etc. It's not active in regular text buffers. I'm sure you have used history before; when it is active you can cycle through history items using the up and down arrow keys or `M-p` and `M-n`. The `consult-history` command lets you _search_ through history using completion rather than moving through it one item at a time.

For this package, the history we are talking about would be active in the minibuffer when you run one of these commands that prompts you for a reference.

bdarcus:
> For this package, the history we are talking about would be active in the minibuffer when you run one of these commands that prompts you for a reference.

As in, I'm done?

Well, I may want to add a defcustom for predefined search strings?

But otherwise, completing-read, etc will handle the details?

oantolin:
Yes, I'd say you are done.

(I may have explained more than necessary in my previous comment because I found it very odd that you expected `consult-history` to _do something_ in a text buffer.)

You may not need a defcustom for the predefined searches, maybe just an example of storing them in the history variable in the README would be enough.

But also, I'm not so sure my suggestion of prepopulating the history is really such a good solution. As you use the commands the history grows and makes it harder to find your predefined searches. Maybe they should be keep separate. As a low-tech solution you could do something like this:

```elisp
(defcustom bibtex-actions-presets nil
  "List of predefined searches")

(defun bibtex-actions-insert-preset ()
  "Prompt for and insert a predefined search."
  (interactive)
  (when-let ((search (completing-read "Preset: " bibtex-actions-presets)))
    (delete-minibuffer-contents)
    (insert search)))
```

And just tell users to bind that command to some key in `minibuffer-local-map`. The usage would be as follows: you run any of the bibtex-actions commands, when you realize you wanted a predefined search you run the above command, then you can select (using completion or the arrow keys or whatever) one of the predefined searches, the command then erases whatever you had already typed at the minibuffer prompt and replaces it with the chosen predefined search.

bdarcus:
> I'm not so sure my suggestion of prepopulating the history is really such a good solution. As you use the commands the history grows and makes it harder to find your predefined searches. 

Yeah, I was wondering about that :-)

> Maybe they should be keep separate. As a low-tech solution you could do something like this:
> 
> ```emacs-lisp
> (defcustom bibtex-actions-presets nil
>   "List of predefined searches")
> 
> (defun bibtex-actions-insert-preset ()
>   "Prompt for and insert a predefined search."
>   (interactive)
>   (when-let ((search (completing-read "Preset: " bibtex-actions-presets)))
>     (delete-minibuffer-contents)
>     (insert search)))
> ```
> 
> And just tell users to bind that command to some key in `minibuffer-local-map`. The usage would be as follows: you run any of the bibtex-actions commands, when you realize you wanted a predefined search you run the above command, then you can select (using completion or the arrow keys or whatever) one of the predefined searches, the command then erases whatever you had already typed at the minibuffer prompt and replaces it with the chosen predefined search.

That seems better indeed, and low-tech seems appropriate here.

So I could include those two bits of code, and we could add an example of how to configure that on the wiki.

@apc - care to experiment with this idea?

Thanks @oantolin!

apc:
Awesome! Will have a look. May not have time today, but will get back to you tomorrow or so.

apc:
Quick check: is it by design that when navigating through the history I see the entire string with all the spaces in between that, I imagine, are there only for formatting purposes? 

image

Also, is it by design that repeatedly pressing `M-n` takes me through the entire list of entries, even entries that I haven't searched for?

bdarcus:
> Quick check: is it by design that when navigating through the history I see the entire string with all the spaces in between that, I imagine, are there only for formatting purposes?

Is that inserting the candidate string, rather than the search that led to that?

BTW, I added that code to the branch.

oantolin:
I'll let @bdarcus say whether it is by design, that is whether it is intended, but I can confirm that is how the normal Emacs history mechanism works.

First, the history items are full selected candidates from previous runs, not just the bit you typed before selecting a candidate (say, with the arrow keys and RET) and are displayed just as they would be when you selected them the first time, containing all the special formatting present.

Second, `M-n` moves you forward through history, but when it reaches the end of the history it starts moving through "future history" (see the bit about future history in `(info "(emacs)Minibuffer History")`) . The future history can be manipulated by commands that call completing-read and contains all the candidates by default. A nice use of future history you may be familiar with, is that if there is a file name at point in some buffer you are editing and you call `find-file`, it puts the file name as the first element of the future history.

oantolin:
Come to think about it, @bdarcus, maybe the *future* history is a good place to put the predefined searches. The history grows with use but the future history is whatever the developer puts there.

oantolin:
Putting the predefined in searches in the future history can be done in addition to any other approach you might also want to implement.

bdarcus:
> I'll let @bdarcus say whether it is by design, that is whether it is intended in this case, but I can confirm that is how the normal Emacs history mechanism works.

So for example, if there are multiple candidates that a user selected, this would insert them all in that CRM prompt?

Therefore, it's not the history of what you see in the list of candidates, but of what you select among them?

I didn't realize that.

I think this may just be another place where the default CRM UI in these tools just does not play well with the design of these bibtex-completion frontends, which rely on long candidate strings. 

I need to think how well this would work, even with the positive (for these cases) changes contemplated in https://github.com/raxod502/selectrum/issues/489.

This isn't how the predefined search approach in `ivy-bibtex` and `helm-bibtex` work, but that's not necessarily a problem. Will just have to think about this a little more.

Thoughts @apc?

bdarcus:
> I need to think how well this would work, even with the positive (for these cases) changes contemplated in [raxod502/selectrum#489](https://github.com/raxod502/selectrum/issues/489).

I _think_ if the completion UI (in this case `selectrum`, but it works similarly in `icomplete-vertical` and `vertico`) allowed for a more `helm`-like selection and indication of candidates, this would be a non-issue.

apc:
> > I'll let @bdarcus say whether it is by design, that is whether it is intended in this case, but I can confirm that is how the normal Emacs history mechanism works.
> 
> So for example, if there are multiple candidates that a user selected, this would insert them all in that CRM prompt?
> 
> Therefore, it's not the history of what you see in the list of candidates, but of what you select among them?
> 
> I didn't realize that.
> 
> I think this may just be another place where the default CRM UI in these tools just does not play well with the design of these bibtex-completion frontends, which rely on long candidate strings.
> 
> I need to think how well this would work, even with the positive (for these cases) changes contemplated in [raxod502/selectrum#489](https://github.com/raxod502/selectrum/issues/489).
> 
> This isn't how the predefined search approach in `ivy-bibtex` and `helm-bibtex` work, but that's not necessarily a problem. Will just have to think about this a little more.
> 
> Thoughts @apc?
> 

Yes, I think there's something a bit unfortunate about having the entire candidate string populate the prompt either when navigating through history or when selecting multiple entries to pass to one of the `bibtex-actions-` functions. I read the thread @bdarcus just referred to in the previous comment, and it does look as if the idea of having selected candidates stack at the top (and of showing the history at the top, perhaps) would be a nice solution. Another, perhaps easier-to-implement solution would be to at least remove all the extra spaces that are there for alignment purposes, so you would show a much more compact string  when going through the history or when selecting multiple candidates. 

Incidentally, I no longer see how to combine the saved queries as just part of the history, since presumably the queries wouldn't correspond to a single candidate. Earlier I thought that the history would be populated by search strings, in which case you would be able to save helpful searches that way. But if it is populated with individual candidates, I'm not sure how this would help e.g. save a search for all entries having some word in the title. @oantolin were you thinking about the saved searches differently than I am?

oantolin:
The things automatically stored by Emacs in the history are full chosen crm-separated candidate sets. The things manually put in the future history by the developer are arbitrary strings, you can put any sensible query you want there.

bdarcus:
Alright, time to play with future history.

This seems a good post:

https://engineering.collbox.co/post/working-faster-in-emacs-by-reading-the-future/

bdarcus:
There's discussion of future history, with example, also on the consult issue tracker:

https://github.com/minad/consult/issues/96

bdarcus:
Here's a minimal example, @apc, that allows to see how it would work:

``` elisp
(completing-read "Select: " (list "one text" "two text") nil nil nil nil '("one" "two" "three"))
```

The future history list is in the "default" slot.

apc:
Neat! Will play around with it next time I have a chance.

apc:
I keep getting this `Wrong type argument: char-or-string-p` error when calling `bibtex-actions-insert-citation` and others, whenever I try to navigate through the history: 

image

Also, calling `bibtex-actions-insert-citation` gives me this

image

But I assume you want the value of `bibtex-actions-presets` that shows there, no?

bibtex-actions.el
```diff
@@ -123,7 +133,7 @@ candidate list"
                        (affixation-function . bibtex-actions--affixation)
                        (category . bibtex))
                    (complete-with-action action candidates string predicate)))
-                 nil nil initial nil nil nil)))
+                 nil nil initial 'bibtex-actions-history 'bibtex-actions-presets nil)))
```
bdarcus, line 136: > I keep getting this `Wrong type argument: char-or-string-p` error when calling `bibtex-actions-insert-citation` and others, whenever I try to navigate through the history

Yeah, I was just going to report the same. Maybe there's some little lispy thing wrong with this line, though I don't see what.

I thought maybe the type on the defcustom is wrong, but I don't believe it is.

bdarcus: > Just remove the quote from `'bibtex-actions-presets`, you need to pass a list of string, not a symbol naming a variable containing the list. (It's different for the history, there you need to pass the name of the history variable, so completing read can modify it.)

Thank you for the explanation; those little details keep tripping me up.

Fix pushed, and now works.

I only tested it really briefly, but first impression is I like it!

Let me know WYT @apc.

I guess the only remaining for question for me beyond the general UX one is whether to adapt that `bibtex-actions-insert-preset` function to this (it won't be persistent ATM; right), or just keep it simple for now and remove it, and tell users to add it to their config manually.

oantolin: > (it won't be persistent ATM; right)

What do you mean?

bdarcus: The history variable I know/assume will get saved by savehist.

Is that also true of the future history? I was assuming not for some reason.

oantolin: The future history is sort of created fresh for every call to `completing-read`, by whatever expression you, the developer, use as the DEF argument. Here you simply pass the value of the variable `bibtex-actions-presets`, so it's a matter of how that variable gets its value. I was assuming that users wouldn't need this persisted since they would actually just assign that variable a value in their init file, a value consisting of the predefined searches they find useful.

bdarcus: > I was assuming that users wouldn't need this persisted since they would actually just assign that variable a value in their init file, a value consisting of the predefined searches they find useful.

Yes, sorry; that's what I meant.

In that case, there's no point for that function; right?

oantolin: I don't use savehist, so I can say from firsthand experience, but the docstring seems to suggest users need to add `'bibtex-actions-history` to `savehist-additional-variables` if they want it saved.

oantolin: (Why don't I use savehist?)

oantolin: > Yes, sorry; that's what I meant. 
> In that case, there's no point for that function; right?

You mean the `bibtex-actions-insert-preset` function? I think it depends on how many predefined searches users will have. The future history mechanism only lets you scroll through presets one at a time. If a user has lot of presets, they'd probably prefer a function like `bibtex-actions-insert-preset`. (Note that a command like `consult-history` only completes items from the actual history, not from the so-called "future history".)

bdarcus: > The future history mechanism only lets you scroll through presets one at a time. If a user has lot of presets ...

These two are in some tension though.

Seems you'd want to have less than 10 for it be worth using.

I'm leaning towards removing it, but will sleep on it.

apc:
Very nice! One question: do you want the first item in `bibtex-actions-presets` to show as the 'default' whenever one calls one of the `bibtex-actions-open` functions? It's not a big deal, but it would be cleaner IMHO if you didn't get saved searches suggesting themselves all the time. 

image


(Incidentally: did you consider renaming `bibtex-actions-open` with `bibtex-actions-open-link` or some such? It's a bit counterintuitive that the 'open *simpliciter*' function opens the link, as opposed to the entry, or the PDF. I always need to remind myself that `bibtex-actions-open` is the one that opens the url...)

bdarcus:
> Very nice! One question: do you want the first item in `bibtex-actions-presets` to show as the 'default' whenever one calls one of the `bibtex-actions-open` functions? It's not a big deal, but it would be cleaner IMHO if you didn't get saved searches suggesting themselves all the time.

Not following here. What's the issue with that?

Regardless, we should get it "right".

> (Incidentally: did you consider renaming `bibtex-actions-open` with `bibtex-actions-open-link` or some such? It's a bit counterintuitive that the 'open _simpliciter_' function opens the link, as opposed to the entry, or the PDF. I always need to remind myself that `bibtex-actions-open` is the one that opens the url...)

 `bibtex-actions-open` uses the `bibtex-completion-open-any` function, and was my attempt to simplify the name.

It actually is broader than links; if there is a PDF, it will prefer that, and fallback to link.

That's why.

I thought about something like "open-doc", but could never settle settle one that felt idea.

I am open to reconsidering some of those names, but would want to do this soon, so I can declare 1.0 and not look back ;-)

Feel free to open a new issue if you want to make specific suggestions.

apc:
> > Very nice! One question: do you want the first item in `bibtex-actions-presets` to show as the 'default' whenever one calls one of the `bibtex-actions-open` functions? It's not a big deal, but it would be cleaner IMHO if you didn't get saved searches suggesting themselves all the time.
> 
> Not following here. What's the issue with that?

Maybe I'm thinking about 'saved searches' in the wrong way: I was thinking more like things you might resort to in specific occasions, not as having any kind of starring role in my workflow. (I wouldn't want the first thing on my browser bookmarks to be what a new window or tab opens by default, say.) Of course, the first string in `bibtex-actions-preset` is 'default' only in that it is labeled as such in the prompt, and will be what you search for if you press `RET`, but there's no need to delete it or anything of the sort. So again, this is not a big deal!

> 
> Regardless, we should get it "right".
> 
> > (Incidentally: did you consider renaming `bibtex-actions-open` with `bibtex-actions-open-link` or some such? It's a bit counterintuitive that the 'open _simpliciter_' function opens the link, as opposed to the entry, or the PDF. I always need to remind myself that `bibtex-actions-open` is the one that opens the url...)
> 
> `bibtex-actions-open` uses the `bibtex-completion-open-any` function, and was my attempt to simplify the name.
> 
> It actually is broader than links; if there is a PDF, it will prefer that, and fallback to link.
> 

Oh, I see. I somehow missed that. 

That said, it now looks as if `bibtex-actions-open` is filtering out (by default) all entries that lack either a PDF or a link, but if the intent is to have that function open either PDF or link, this seems like a bug. (I do not know if you could have a string that amounts to something like `has:pdf` OR `has:link`. 

> That's why.
> 
> I thought about something like "open-doc", but could never settle settle one that felt idea.
> 
> I am open to reconsidering some of those names, but would want to do this soon, so I can declare 1.0 and not look back ;-)

Make sense! I'd be inclined to use `open-any` to avoid giving the impression that `bibtex-actions-open` has a privileged status. But this may be a somewhat idiosyncratic concern. 🤷‍♂️

> 
> Feel free to open a new issue if you want to make specific suggestions.

apc:
Also, FWIW (and take this with a huge grain of salt since I don't really have any experience thinking about UX or really programming of any sort for that matter) I sort of prefer the `bibtex-actions-insert-preset` approach to the `future history` approach. Not sure if it's just because it more cleanly separates history from 'bookmarks' or 'saved searches', or if it's because the idea of 'future history' seems, well, like an oxymoron. 🤷‍♂️

bdarcus:
> I sort of prefer the `bibtex-actions-insert-preset` approach to the `future history` approach. 

To clarify, you mean you prefer the full candidate set to be added, as with your first screenshots in this PR?

apc:
> > I sort of prefer the `bibtex-actions-insert-preset` approach to the `future history` approach.
> 
> To clarify, you mean you prefer the full candidate set to be added, as with your first screenshots in this PR?
> 

No, sorry. At the moment I take it there are two ways of accessing the saved searches: typing `M-n` from the mini buffer (after e.g. calling `bibtex-actions-open`) or by calling `bibtex-actions-insert-preset`, which prompts you then with a list of saved searches you can choose from. This is what it looks like on my end if I call `bibtex-actions-insert-preset` after calling `bibtex-actions-open` (note, though, that I modified the definition as suggested in [my previous comment](https://github.com/bdarcus/bibtex-actions/pull/120#issuecomment-829735683) since otherwise I got the `attempt to use minibuffer while in minibuffer` error): 

image

oantolin:
> I think you want to add `(enable-recursive-minibuffers t)` inside the scope of `when-let` in the definition of `bibtex-actions-insert-preset`

You are right, @apc, sorry for the oversight. I use `(setq enable-recursive-minibuffers t)` and forget that not everyone does!

> @oantolin - what's the syntax for has:pdf orhas:link in orderless?

orderless only matches against candidates strings. I don't know if you already put the information of whether an item has a link or pdf in your candidate strings, @bdarcus, but if you don't there is nothing orderless can do here.

If you do add the string "has:pdf" (or similar, like "#pdf") to items that have a pdf, then a user could use "has:pdf" to match them (of course that would also match items without a pdf but that happen to have "has:pdf" in the title, or were authored by "steve has:pdf", but one hopes that is unlikely). If an orderless user has a style dispatcher for `orderless-without-literal`, say like the one in the readme that uses `!`, they can match references without a pdf using "!has:pdf".

If you do add "has:pdf" or "#pdf" or some sort of marker to the candidates that have pdfs, but don't want this marker visible you can add the invisible property to the marker: even invisible text can be matched against.

bdarcus:
@oantolin -

> If you do add "has:pdf" or "#pdf" or some sort of marker to the candidates that have pdfs, but don't want this marker visible you can add the invisible property to the marker: even invisible text can be matched against.

Yeah, I already have this. Sorry; I should have explained more.

The problem @apc noticed is that I add this string to the initial value for the "open" command:

```
has:pdf has:link
```
The problem is that will only match candidates that have both, of course, when what we need is to match either/or; the sort of thing you might do with a traditional regexp like:

```
(has:pdf|has:link)
```

Can I do that with orderless?

bdarcus:
> I said I preferred the latter to the former. But I take that back.

Ok, so initially you were advocating to keep `bibtex-actions-insert-preset`, and now are thinking it's fine to ditch it?

I just pushed the change you suggested, and it works.

I'm not sure myself.

I guess we could just include it and see what a wider circle of users think.

bdarcus:
> The Emacs syntax uses `\|`, so you can use `has:pdf\|has:link`

The one variant I hadn't tried!

Awesome; thank you!

So if I have this style dispatcher setup, I'd want to remove it; right?

``` elisp
(defun first-initialism (pattern index _total)
  (if (= index 0) 'orderless-initialism))
```

And I probably want to tell people to use `orderless` (or `prescient`, which I believe also supports the regexp by default) in the README?

apc:
> > I said I preferred the latter to the former. But I take that back.
> 
> Ok, so initially you were advocating to keep `bibtex-actions-insert-preset`, and now are thinking it's fine to ditch it?
> 

I think there’s a place for both, especially in light of your point about combining predefined searches below. At first I couldn’t see a reason to have the future history solution other than it made use of stuff that was already there. But after playing around with it I see it’s quite a nice way to implement this feature too. 


> I just pushed the change you suggested, and it works.
> 
> I'm not sure myself.
> 
> I guess we could just include it and see what a wider circle of users think.

fair enough! 

> 
> EDIT: I did just note one advantage of the "insert" approach is you can include more than one preset in the prompt.
> 

Neat! 

> Also, the future history UI works exactly the same in `vertico`, and (I assume, though haven't tested) `icomplete-vertical`.

exactly the same as it does with order less/selectrum or exactly the same as `bibtex-actions- insert-preset`?

bdarcus:
> exactly the same as it does with order less/selectrum or exactly the same as `bibtex-actions- insert-preset`?

I mean the UX you get in selectrum is the same with the others, whether using future history, or `bibtex-actions-insert-preset`.

Not surprising, but helpful to have confirmed.

bdarcus:
Should we update the `use-package` example on the README to include a `bibtex-actions-preset` binding?

If yes, any suggestions?

apc:
> Should we update the `use-package` example on the README to include a `bibtex-actions-preset` binding?
> 
> If yes, any suggestions?

Good point. How about this? 

```elisp
(use-package bibtex-actions
  :bind (("C-c b" . bibtex-actions-insert-citation)
         :map minibuffer-local-map
         ("M-b" . bibtex-actions-insert-preset))
  :after embark
  :config
  ;; Make the 'bibtex-actions' bindings available from `embark-act'.
  (add-to-list 'embark-keymap-alist '(bibtex . bibtex-actions-map)))

```

A mnemonic could be `M-b` for 'bookmarked`? 

Also, if you want to include a bit in the `use-package` declaration specifying the path to the `bib` file, you can put this: 

```elisp
(use-package bibtex-actions
  :bind (("C-c b" . bibtex-actions-insert-citation)
         :map minibuffer-local-map
         ("M-b" . bibtex-actions-insert-preset))
  :after embark
  :config
  ;; Make the 'bibtex-actions' bindings available from `embark-act'.
  :custom 
  (bibtex-completion-bibliography '("~/bib/references.bib")))

```

bdarcus:
OK, I think this is RTM; your extended use-package example included @apc.

I was thinking I should add you two as co-authors on the commit.

If that sounds good, can you get me your email addresses, or just tell me it's OK to leave it out?

https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors#required-co-author-information

apc:
Either way is fine by me! My email is [my University](https://www.umass.edu)'s email address, using my GH username before the `@` sign.

bdarcus:
> Either way is fine by me! My email is [my University](https://www.umass.edu)'s email address, using my GH username before the `@` sign.

Full name or gh user name on the commit, @apc?

apc:
Don't really mind either way. gh user name is shorter! Full name is 'Alejandro Pérez Carballo'.

bibtex-actions.el
bdarcus: I decided to keep. Can see what feedback comes in.

bdarcus:
Done. Thanks much to you both.

I fixed the bug with open while I was at it, BTW.

oantolin:
> One question: do you want the first item in `bibtex-actions-presets` to show as the 'default' whenever one calls one of the `bibtex-actions-open` functions?

@apc, this is a Selectrum thing. Other completion UIs such as the default completion don't not aggressively promote the default value like that. In default completion the default is not shown unless you include it by hand in the prompt.

oantolin:
> I was thinking I should add you two as co-authors on the commit.

I also don't mind either being included or not; up to you, @bdarcus. My name and email can be found by doing `M-x describe-package RET embark RET`. 😛

PR #124, opened 2021-05-02
Add hook for force reloading

publicimageltd:
This adds a hook which is run when reloading is forced by pressing C-u.

Note that this hook is not run in a usual refresh, since recreating a bibliography file can take a while and thus should be explicitly requested. Here's my setup, using the function `pullbib-pull` to create the bibliography file on demand:

```emacs-lisp
  (setq bibtex-actions-force-refresh-hook  (list #'pullbib-pull))
```

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -86,10 +86,10 @@ may be indicated with the same icon but a different face."
 
 (defcustom bibtex-actions-force-refresh-hook nil
   "Hook run when user forces a (re-) building of the candidates cache.
-This hook is only called when the user explicitly requests that
-the cache has to be rebuilt. It is intended for 'heavy'
-operations such as a complete export of a Zotero library to a
-bibliography file."
+This hook is only called when the user explicitly requests the
+cache to be rebuilt.  It is intended for 'heavy' operations which
+recreate entire bibliography files using an external reference
+manager like Zotero or JabRef."
   :group 'bibtex-actions
   :type '(repeat function))
 
@@ -233,7 +233,8 @@ If FORCE-REBUILD-CACHE is t, force reloading the cache."
 ;;;###autoload
 (defun bibtex-actions-refresh (&optional force-rebuild-cache)
   "Reload the candidates cache.
-If called interactively with a prefix, also run the hook
+If called interactively with a prefix of if FORCE-REBUILD-CACHE
+is non-nil, also run the hook
 `bibtex-actions-before-refresh-hook'"
   (interactive "P")
   (when force-rebuild-cache
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -84,6 +84,15 @@ may be indicated with the same icon but a different face."
   :group 'bibtex-actions
   :type 'string)
 
+(defcustom bibtex-actions-force-refresh-hook nil
+  "Hook run when user forces a (re-) building of the candidates cache.
+This hook is only called when the user explicitly requests that
+the cache has to be rebuilt. It is intended for 'heavy'
+operations such as a complete export of a Zotero library to a
+bibliography file."
+  :group 'bibtex-actions
+  :type '(repeat function))
+
 ;;; History, including future history list.
 
 (defvar bibtex-actions-history nil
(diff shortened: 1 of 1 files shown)


bdarcus:
Looks good!

The CI has a few minor warnings you should resolve. You can see them by clicking the "files" tab on the PR.

bibtex-actions.el
```diff
@@ -84,6 +84,15 @@ may be indicated with the same icon but a different face."
   :group 'bibtex-actions
   :type 'string)
 
+(defcustom bibtex-actions-force-refresh-hook nil
+  "Hook run when user forces a (re-) building of the candidates cache.
+This hook is only called when the user explicitly requests that
+the cache has to be rebuilt. It is intended for 'heavy'
+operations such as a complete export of a Zotero library to a
```
bdarcus, line 91: Aside from addressing the warnings, while I use Zotero myself, could we have a more generic example?

Not a big deal if not ....

bdarcus (changes requested):

bdarcus:
A question just occurred to me:

This assumes the user explicitly calls this using the prefix argument.

But what if I want to automate this using notify and a script, with Emacs daemon?

bdarcus (approved):

publicimageltd:
> Or is this a non-issue; the hook is for user input situations, and this is still available?

Exactly, because you still can pass an explicit argument to `refresh` in order to run the hook without user interaction.

Thanks for merging!

publicimageltd:
Sorry, wrong button -- just wanted to comment :-)

bdarcus:
Done.

Probably should add something to the README?

publicimageltd:
> Probably should add something to the README?

I could do it, but I would prefer you  todo that, since I always have to edit my docs thrice until I am satisfied with them (language problem). I promise I'll include the README for my next PR. Hell, I did not think that it takes /so/ many steps to finish a PR. :-)

bdarcus:
No problem.

BTW, the docstring warnings from the CI; `checkdoc` will flag those locally.

bdarcus:
Is pullbib ready enough from your perspective to put as an example in the README, or on the wiki?

I'll test it later, as I do use Zotero and BBT.

publicimageltd:
Do test pullbib, please! It works for me. Only thing missing is the ability to add extra parameters to the query, e.g. "don't export notes". I think it's not necessary, so I am glad if someone else tests it.

bdarcus:
I saw you work with `org-roam`; do you make any effort to sync zotero notes with `org-roam`, or do you do all bib note-taking in the later?

The forthcoming PDF viewer and annotation support in Zotero is its first killer feature in a long time.

publicimageltd:
> I saw you work with org-roam; do you make any effort to sync 
> zotero notes with org-roam, or do you do all bib note-taking in 
> the later?

I only use org-roam; Zotero always felt too clumsy (yikes! mouse 
movements!).
 
> The forthcoming PDF viewer and annotation support in Zotero is 
> its first killer feature in a long time.

I did not know about that and just looked at it! Wow, that's 
really a change, finally! I guess I will not switch to Zotero, 
however, as long as it insists on using a mouse. Also, I love 
Emacs' flexibility: I can copy and paste, move to registers, 
quickly write a defun to collect stuff in a separate buffer, then 
save it, and and and...

I will rewrite my delve package for the new org roam version v2; 
but I am still looking for a new concept for the UI. I get more 
and more convinced that the completing-read-style of doing things 
is the Emacs way, so my current implementation, which is basically 
a list application, seems too static. If I have some time, I'll 
look at embark -- maybe making  snapshots of collection of 
org-roam pages, with some sensible actions to work with them, can 
be the killer feature I am looking for.

bdarcus:
> I get more and more convinced that the completing-read-style of doing things is the Emacs way ...

... even more so given that all of the innovation happening in this space the past year is being consolidate into patches to core Emacs, to enhance completing-read.

I learned a lot from those guys, BTW; and this package pretty much reflects their approach to these questions.

bdarcus:
PS - on org, if you've not seen it, you might want to look at #113, and links from there.

Will be particularly relevant for the `org-roam` ecosytem.

Issue #126, opened 2021-05-04
bibtex-actions-read returns (nil nil) when selecting item


publicimageltd:
I have encountered a strange bug: Certain bibliography items seem to cause `bibtex-actions-read` to hickup. It returns `(nil nil)`. Here's the `.bib` file:

```bibtex
@book{jorke_2003_Demokratie,
  title = {Demokratie als Erfahrung: John Dewey und die politische Philosophie der Gegenwart},
  shorttitle = {Demokratie als Erfahrung},
  author = {Jörke, Dirk},
  date = {2003},
  publisher = {{Westdt. Verl}},
  location = {{Wiesbaden}},
  url = {http://bvbr.bib-bvb.de:8991/F?func=service&doc_library=BVB01&doc_number=010344150&line_number=0001&func_code=DB_RECORDS&service_type=MEDIA},
  urldate = {2012-08-31},
  file = {/home/jv/Zotero/storage/3LUESP9S/Jörke - 2003 - Demokratie als Erfahrung John Dewey und die polit.pdf},
  langid = {german}
}

@article{joerissenAnthropologischeHinsichtenPragmatische2002,
  title = {Anthropologische Hinsichten, pragmatische Absichten},
  author = {Jörissen, Benjamin},
  date = {2002},
  pages = {24},
  file = {/home/jv/Zotero/storage/X7X3JUAD/Jörissen - 2002 - Anthropologische Hinsichten, pragmatische Absichte.pdf},
  langid = {german}
}
```

Selecting the first item yields the key (in a list); selecting the second item yields `(nil nil)`.

Trying to debug that behavior, I encountered another weird error. I wanted to instrument the function with edebug. Edebug, however, said the defun was malformed. So I added a fake variable to the argument list:
```emacs-lisp
(cl-defun bibtex-actions-read (&optional _fake_argument &key initial rebuild-cache)
  "Read bibtex-completion entries. ....")
```

Now debugging worked -- but the error was gone!

So it turned out that the error occurs only as long as I use the byte compiled version of the package. If I evaluate the function in the buffer directly, everything works fine. I get the error again if I explicitly force a reloading of the package with `(load "bibtex-actions")`.

So my guess is that the problem has to do with compilation.

I use GNU Emacs 27.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.27, cairo version 1.17.4) of 2021-03-26, and the latest bibtex-actions commit 149f9ae.

[...]

PR #127, opened 2021-05-08
Add compiler declaration of crm-separator

publicimageltd:
Declare `crm-separator` for byte compilation to avoid lexical/dynamic binding issue (see #126).

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -138,18 +138,18 @@ candidate list
 
 'REBUILD-CACHE' if t, forces rebuilding the cache before
 offering the selection candidates"
-  (when-let ((crm-separator "\\s-*&\\s-*")
-	     (candidates (bibtex-actions--get-candidates rebuild-cache))
-             (chosen
-              (completing-read-multiple
-               "BibTeX entries: "
-               (lambda (string predicate action)
-                 (if (eq action 'metadata)
-                     `(metadata
-                       (affixation-function . bibtex-actions--affixation)
-                       (category . bibtex))
-                   (complete-with-action action candidates string predicate)))
-                 nil nil initial 'bibtex-actions-history bibtex-actions-presets nil)))
+  (let* ((crm-separator "\\s-*&\\s-*")
+	 (candidates (bibtex-actions--get-candidates rebuild-cache))
+         (chosen
+          (completing-read-multiple
+           "BibTeX entries: "
+           (lambda (string predicate action)
+             (if (eq action 'metadata)
+                 `(metadata
+                   (affixation-function . bibtex-actions--affixation)
+                   (category . bibtex))
+               (complete-with-action action candidates string predicate)))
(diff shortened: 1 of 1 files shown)


bdarcus:
Thanks @publicimageltd!

Do you think we should "fix" the related `when-let` issue too that Daniel noted?

That was another piece of code, contributed by someone else in an earlier iteration, that I didn't really understand. It was before I added the crm-separator, and so was focused on the candidates variable.

bdarcus:
Like, if I understand what he's saying, this would have been correct (put `candidates` first):

``` elisp
  (when-let ((candidates (bibtex-actions--get-candidates rebuild-cache)
           (crm-separator "\\s-*&\\s-*")
```

... but maybe best to just change `when-let` to `let*`?

Do I understand that correctly?

bdarcus:
> ... but maybe best to just change `when-let` to `let*`?

I made this change. Let me know if you approve.

And do we need the defvar with this?

publicimageltd:
I do approve! Thanks for adding it. You might consider using `when-let*`, since it would stop processing if there are no candidates. In the current version, the empty candidate list would be passed on to `completing-read`. I don't know how `completing-read` behaves if you pass `nil` as candidates. On the other hand, it might be worth considering to explicitly throw an error if the candidates list is `nil,` since this is most likely an error in the configuration.

And yes, we still need the `defvar`, AFAIU.

bdarcus:
> You might consider using `when-let*`, since it would stop processing if there are no candidates. In the current version, the empty candidate list would be passed on to `completing-read`. 

I'm pretty sure that's why Omar suggested that originally.

So if I had swapped the order, that would have been correct?

> On the other hand, it might be worth considering to explicitly throw an error if the candidates list is `nil,` since this is most likely an error in the configuration.

That's kind of what I was thinking.

Comments on #128?

publicimageltd:
#128 - shouldn't the check be in `bibtex-actions-read`? Also, I would throw a `user-error` instead of just displaying a `message`. After all, the function can't continue doing what it is called for.

bdarcus:
> shouldn't the check be in `bibtex-actions-read`?

IDK; I put it there since that's the only place the `bibtex-completion-candidates` is called, and when successful, user gets the message from that which signals success.

Should I move it?

publicimageltd:
> Should I move it?

IDK either. Does `bibtex-completion-candidates` return a nil value if there are no candidates? In this case, the formatting functions (would just return `nil,` too, and thus propagate the lack of candidates down to the calling function. I'm not a programmer, but I try to keep these kind of things separate: A function which returns candidates returns candidates, or nil; and if I need candidates, it is the task of that other function to deal with the fact that there might be zero candidates. In other words, in my view, polling the candidates is distinct from doing something with it, and thus only the latter should throw an error. (We can imagine writing a function later that checks if there are candidates at all, which could use that behavior.) - All that, of course, assumes that the polling functions return nil if there are no candidates.

PR #128, opened 2021-05-08
add check candidates list is non-nil

bdarcus:


bdarcus:
I'm copying this over here, to keep track:

https://github.com/bdarcus/bibtex-actions/pull/127#issuecomment-835406635

@publicimageltd - on this:

> Does bibtex-completion-candidates return a nil value if there are no candidates? 

It returns a nil if the `bibtex-completion-bibliography` values is nil, but this error if the file path is wrong:

```
*** Eval error ***  Opening input file: No such file or directory, /tmp/foo.bib
```

Seems like this error handling should be addressed there; no?

publicimageltd:
I had spontaneously written a suggestion, but I somehow checked 
out the wrong branch and can't do a PR. Actually, the whole code 
is gone. So here's a suggestion, I could do a PR later, now I want 
to enjoy the sun:

 - In `bibtex-actions--format-candidates`, we turn candidates into 
 an explicit argument.
 
  - Before refreshing, we check if the files are valid.

```emacs-lisp
(defun bibtex-actions-refresh (&optional force-rebuild-cache) 
  "Reload the candidates cache. 
If called interactively with a prefix or if FORCE-REBUILD-CACHE is 
non-nil, also run the hook `bibtex-actions-before-refresh-hook'. 
Throw an error if the bibliography files are not readable." 
  (interactive "P") (when force-rebuild-cache 
  (let* ((files (bibtex-completion-normalize-bibliography 
  'bibtex)) 
         (files-exist-p (and files (seq-reduce (lambda (acc f) 
         (and acc (file-readable-p f))) 
                                               files t)))) 
    (if files-exist-p  
              (bibtex-actions--format-candidates 
              (bibtex-completion-candidates))) 
      (user-error "Missing or unreadable biblography files"))))
```

The motivation for this is that `bibtex-completion-candidates` 
does no error handling (but should do, btw). So the cleanest 
solution would to move these file checks to `bibtex-completion`.

publicimageltd:
I had spontaneously written a suggestion, but I somehow checked 
out the wrong branch and can't do a PR. Actually, the whole code 
is gone. So here's a suggestion, I could do a PR later, now I want 
to enjoy the sun:

 - In `bibtex-actions--format-candidates`, we turn candidates into 
 an explicit argument.
 
  - Before refreshing, we check if the files are valid.

```emacs-lisp

(defun bibtex-actions-refresh (&optional force-rebuild-cache) 
  "Reload the candidates cache. 
If called interactively with a prefix or if FORCE-REBUILD-CACHE is 
non-nil, also run the hook `bibtex-actions-before-refresh-hook'. 
Throw an error if the bibliography files are not readable." 
  (interactive "P") (when force-rebuild-cache 
  (let* ((files (bibtex-completion-normalize-bibliography ''bibtex)) 
         (files-exist-p (and files (seq-reduce (lambda (acc f) 
                                                        (and acc (file-readable-p f))) 
                                                        files t)))) 
    (if files-exist-p  
              (bibtex-actions--format-candidates 
              (bibtex-completion-candidates))) 
      (user-error "Missing or unreadable biblography files"))))

```

The motivation for this is that `bibtex-completion-candidates` 
does no error handling (but should do, btw). So the cleanest 
solution would to move these file checks to `bibtex-completion`.

bdarcus:
> The motivation for this is that `bibtex-completion-candidates`
> does no error handling (but should do, btw).

Maybe one of us should submit a PR there?

It does seem like it should be in `bibtex-completion` the more I think about it.

bdarcus:
Oh wait; I see the issue.

See here: https://github.com/tmalsburg/helm-bibtex/blob/master/bibtex-completion.el#L412.

So `bibtex-completion` DOES have a check; it just happens on `bibtex-completion-init`, which I bypass because I don't want or not the rest of it.

So yeah, I'll look at adding something like that.

Issue #121, opened 2021-04-30
Preparing for 1.0; last call for command or function name changes


bdarcus:
@apc raised a question about the wisdom of the `bibtex-actions-open` command name, and I explained the reasoning in a followup here: https://github.com/bdarcus/bibtex-actions/pull/120#issuecomment-829724678.

I'm kind of agnostic, so would probably leave as is unless there were strong arguments in favor of changing.

But now is the time to raise the question, as I'm planning to tag 1.0 soon, which is me arbitrarily drawing a line-in-the-sand on command and function name changes. 

It doesn't mean we can't revisit later, but that I'd prefer not to.

To recap, we have the follow action verbs:

1. open
2. insert
3. add

... and then command names based on these, the potential confusing ones being:

- `bibtex-actions-open`
- `bibtex-actions-open-entry`

Do note that aside from the clarity of the commands, I was also shooting for conciseness, to accommodate `M-x`, and menu displays like `which-key` (though see [this comment]((defalias 'apc/bib-open #'bibtex-actions-open-entry))) when using `embark-act`.

Bottomline, this (with marginalia annotations from the docstrings) should be as elegant, clear and concise as possible:

![Screenshot from 2021-04-30 13-04-52](https://user-images.githubusercontent.com/1134/116729092-c046b980-a9b4-11eb-9c4a-55e253164c57.png)

So what do people think?

cc @mtreca @wenjie2wang

apc:
As I mentioned to you, I found the name `bibtex-actions-open` confusing as a name for that function. Especially so now that in calling the function you get the initial string `has:link\|has:pdf`, so that by default you don't see the entire list of entries. If anything, i think if you had asked me what `bibtex-actions-open` does, knowing only that this is like `helm-bibtex` and `ivy-bibtex` (packages advertised as designed to "allow you to search and manage your BibTeX bibliography"), I would have said that it would open the `BibTeX` entry. 

But here's a thought: could you have `bibtex-actions-open` be a `defcustom`, give it one of the `bibtex-actions-open` functions as its default, and allow folks to change it if they so desire? This would essentially give you something like the following functionality: 

- A function that is your primary entry point to your list of entries,
- that will perform an action of your choosing (by default) with the selected entries,
- but which (much like the other functions) always make it easy for you to use `embark` say to call other actions. 

You could even allow for customization of the 'initial' string for that function. 

[...]

PR #130, opened 2021-05-09
Customize initial input

### LINKED ISSUE #129: Make initial value for the different actions customizable
While debugging that `crm` problem, I encountered a more conceptual issue. Vanilla Emacs has no sophisticated completion system, it just offers you all strings which match the inital input from the beginning on (like `"^initial input`). Therefore, the initial input (like `"has:pdf"`) will not filter the list of available items. I had to delete the initial input in order to access the item. It only works if you have the appropriate completion style installed, in my case, that would be `orderless`. Assuming that one goal of that package is to provide completion with as little requirements as possible, it makes sense to remove that functional dependency.

So my suggestion is to introduce a variable to make the initial input customizable. Here's a draft:

```emacs-lisp

(defcustom bibtex-actions-initial-inputs 
  '((files . "has:pdf")
    (notes . "has:notes")
 ...)

(defcustom bibtex-actions-filter-with-initial-input t
  "Set this to nil of you use an 'unflexible' completion style"
....)

(defun bibtex-actions--get-initial-input (input-type)
 (when bibtex-actions-filter-with-initial-input
  (assoc input-type bibtex-actions-initial-input))

(defun bibtex-actions-open-notes (keys)
    (interactive (list (bibtex-actions-read :initial (bibtex-actions--get-initial-input 'notes)
   (bibtex-completion-edit-notes keys))
``` 

There would be further stuff to change, like `bibtex-actions--affixation`. And if we go that route, `:initial` could actually be changed to just pass a symbol, that would make the code more elegant.

What do you think?

publicimageltd:
Here's the PR for the customizable initial input. AFAIU, all other occurences of strings like "has:pdf" in the source code are used for actually inserting these strings in the candidates, so they should not be affected by that change.

Closes #129.

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -97,6 +97,30 @@ manager like Zotero or JabRef."
   :group 'bibtex-actions
   :type '(repeat function))
 
+(defcustom bibtex-actions-initial-inputs
+  '((pdf    . "has:pdf")
+    (note   . "has:note")
+    (link   . "has:link")
+    (source . "has:link\\|has:pdf"))
+  "Alist defining the initial input for some bibtex open actions.
+Given a flexible completion style, this will restrict the list of
+available candidates to those matching the initial input.
+
+The association key can be one of the symbols `pdf', `note',
+`link' or `source' and defines the input for the function
+`bibtex-action-open-pdf', `bibtex-action-open-link', etc. The
+associated value must be nil, meaning that there will be no
+initial input, or a string.
+
+To match entries with certain properties (e.g. files attached),
+you can use the following initial inputs: \"has:pdf\",
+\"has:link\" and \"has:note\"."
+  :group 'bibtex-actions
+  :type '(alist :key-type symbol
+                :value-type (choice string
+                                    (const :tag "No initial input" nil))))
+
+  
 ;;; History, including future history list.
 
 (defvar bibtex-actions-history nil
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -108,7 +108,7 @@ available candidates to those matching the initial input.
 
 The association key can be one of the symbols `pdf', `note',
 `link' or `source' and defines the input for the function
-`bibtex-action-open-pdf', `bibtex-action-open-link', etc. The
+`bibtex-action-open-pdf', `bibtex-action-open-link', etc.  The
 associated value must be nil, meaning that there will be no
 initial input, or a string.
 
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -99,7 +99,7 @@ manager like Zotero or JabRef."
 
 (defcustom bibtex-actions-initial-inputs
   '((pdf    . "has:pdf")
-    (note   . "has:note")
+    (note   . nil)
     (link   . "has:link")
     (source . "has:link\\|has:pdf"))
   "Alist defining the initial input for some bibtex open actions.
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -119,7 +119,6 @@ you can use the following initial inputs: \"has:pdf\",
   :type '(alist :key-type symbol
                 :value-type (choice string
                                     (const :tag "No initial input" nil))))
-
   
 ;;; History, including future history list.
 
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -157,13 +157,15 @@ you can use the following initial inputs: \"has:pdf\",
 This provides a wrapper around 'completing-read-multiple', with
 the following optional arguments:
 
-'INITIAL' provides the initial value, for pre-filtering the
-candidate list
+If 'INITIAL' matches one of the keys defined in
+`bibtex-actions-initial-inputs', use the associated initial
+input.
 
 'REBUILD-CACHE' if t, forces rebuilding the cache before
 offering the selection candidates"
   (let* ((crm-separator "\\s-*&\\s-*")
 	 (candidates (bibtex-actions--get-candidates rebuild-cache))
+         (initial-input (assoc-default initial bibtex-actions-initial-inputs))
          (chosen
           (completing-read-multiple
            "BibTeX entries: "
@@ -173,7 +175,12 @@ offering the selection candidates"
                    (affixation-function . bibtex-actions--affixation)
                    (category . bibtex))
                (complete-with-action action candidates string predicate)))
-           nil nil initial 'bibtex-actions-history bibtex-actions-presets nil)))
+           nil
+           nil
+           (and initial-input
+                (stringp initial-input)
+                (concat initial-input " "))
+           'bibtex-actions-history bibtex-actions-presets nil)))
     (cl-loop for choice in chosen
              ;; Collect citation keys of selected candidate(s).
              collect (cdr (assoc choice candidates)))))
@@ -354,7 +361,7 @@ Opens the PDF(s) associated with the KEYS.  If multiple PDFs are
 found, ask for the one to open using ‘completing-read’.  If no
 PDF is found, try to open a URL or DOI in the browser instead.
 With prefix, rebuild the cache before offering candidates."
-  (interactive (list (bibtex-actions-read :initial "has:link\\|has:pdf "
+  (interactive (list (bibtex-actions-read :initial 'source
 					  :rebuild-cache current-prefix-arg)))
   (bibtex-completion-open-any keys))
 
@@ -364,7 +371,7 @@ With prefix, rebuild the cache before offering candidates."
 If multiple PDFs are found, ask for the one to open using
 ‘completing-read’.
 With prefix, rebuild the cache before offering candidates."
-  (interactive (list (bibtex-actions-read :initial "has:pdf "
+  (interactive (list (bibtex-actions-read :initial 'pdf
 					  :rebuild-cache current-prefix-arg)))
   (bibtex-completion-open-pdf keys))
 
@@ -372,7 +379,7 @@ With prefix, rebuild the cache before offering candidates."
 (defun bibtex-actions-open-link (keys)
   "Open URL or DOI link associated with the KEYS in a browser.
 With prefix, rebuild the cache before offering candidates."
-  (interactive (list (bibtex-actions-read :initial "has:link "
+  (interactive (list (bibtex-actions-read :initial 'link
 					  :rebuild-cache current-prefix-arg)))
  (bibtex-completion-open-url-or-doi keys))
 
@@ -415,7 +422,7 @@ With prefix, rebuild the cache before offering candidates."
 (defun bibtex-actions-open-notes (keys)
   "Open notes associated with the KEYS.
 With prefix, rebuild the cache before offering candidates."
-  (interactive (list (bibtex-actions-read :initial "has:note "
+  (interactive (list (bibtex-actions-read :initial 'note
 					  :rebuild-cache current-prefix-arg)))
  (bibtex-completion-edit-notes keys))
 
--- a/README.org
+++ b/README.org
@@ -154,7 +154,10 @@ Here's how to configure it to use =all-the-icons=:
        :group 'all-the-icons-faces)
 #+END_SRC
 
-*** History and Predefined searches
+*** History and predefined searches
+    :PROPERTIES:
+    :CUSTOM_ID: history-and-predefined-searches
+    :END:
 
 =Bibtex-actions= has functionality similar to the [[https://github.com/tmalsburg/helm-bibtex#p][predefined search]] functionality in =helm-bibtex= and =ivy-bibtex=, but with a different implementation.
 Rather than create a new command with the search terms as argument, you just set the =bibtex-actions-presets= variable, and add the strings you want to access:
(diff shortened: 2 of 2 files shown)


bibtex-actions.el
```diff
@@ -97,6 +97,29 @@ manager like Zotero or JabRef."
   :group 'bibtex-actions
   :type '(repeat function))
 
+(defcustom bibtex-actions-initial-inputs
+  '((pdf    . "has:pdf")
+    (note   . "has:note")
+    (link   . "has:link")
+    (source . "has:link\\|has:pdf"))
```
bdarcus, line 104:
```suggestion
  '((pdf    . "has:pdf ")
    (note   . "has:note ")
    (link   . "has:link ")
    (source . "has:link\\|has:pdf "))
```

Right? So point is separated from initial input with a space?

bdarcus (commented):
I added one minor suggestion (which you can commit by just clicking to accept it), and added a link to the issue to your post.

But the bigger question is I'm not clear on the use case, given my response to #129.

Can you clarify?

bibtex-actions.el
publicimageltd: Space is added in `bibtex-actions-read`, which see.

As to the use case, I thought the idea is to just make it customizable anyways. It's Emacs, after all. :smile: 

For me, at least, there _is_ actually a good use case: I use `bibtex-action-open-notes` with the fallback mechanism provided by `bibtex-completion`, which creates a note file if there is none. So the initial input is useless, since for me, `bibtex-action-open-notes` is supposed to open _or_ create a note with one command (which is very handy, btw). Another use case would be the ability to define what "source" actually means (as discussed in #121 ).

But TBH, I cannot imagine many more use cases right now. :shrug: So it's up to the maintainer!

bdarcus: > Space is added in `bibtex-actions-read`, which see.

Oops! I just hid the comment then.

> As to the use case, I thought the idea is to just make it customizable anyways. It's Emacs, after all. 

Well, no; customization for the sake of it would be overkill.

But when there's need, that's another matter.

> For me, at least, there _is_ actually a good use case: I use `bibtex-action-open-notes` with the fallback mechanism provided by `bibtex-completion`, which creates a note file if there is none. 

Ooh, that's actually one I care about too; I had totally forgot about that!

This does raise a larger design question, which is what @apc and I were going back-and-forth on in #118.

We settled on the initial-input approach, because the goal was filtering, but to make it easy for the user to effectively turn off or modify the filtering mid-stream; in this case to delete the initial-input.

This not only because of cases like this, but also because with embark, you may open your library with one action in mind, but change your mind, and so need access to a wider list of candidates.

_**So the big question: did we make the right decision? And if yes, is this the right addition to this?**_

I'm inclined to say yes on both counts, but curious what you and @apc think.

> So the initial input is useless, since for me, `bibtex-action-open-notes` is supposed to open _or_ create a note with one command (which is very handy, btw). Another use case would be the ability to define what "source" actually means (as discussed in #121 ).
> 
> But TBH, I cannot imagine many more use cases right now.  So it's up to the maintainer!

No, that's a good enough reason, per above.

bdarcus:
It does occur to me, though, that default value for notes should be nil.

This is largely because they are different than the others, for the reason you note.

apc:
I'm inclined to agree with @publicimageltd that setting the default value of 'notes' to `nil` might be confusing. I think it'd be best to have all `bibtex-actions-open-` commands treated the same way. 

I think it might be helpful to emphasize that the user may want to set a 'default' entry point to her list of references (I essentially treat `bibtex-actions-insert-citations` that way, and it's the one thing I've bound in my global keymap so far). Depending on use-case, this may be `bibtex-actions-open-pdf` or `bibtex-actions-open-entry` etc., but it would make sense to have one's main entry point something that gives you access to one's entire list of references. For then the user can pull the list of references from her `.bib` files and, using `embark`, call `bibtex-actions-open-notes` if the user wants to just add a note to an entry without any notes, say. In other words, I think it'd be good to think of the commands `bibtex-actions-open-` as functioning slightly differently depending on the context. When the user calls `bibtex-actions-open-`, it makes sense to be presented with a list of things for which the `open` action makes sense. When the user calls the 'default' entry point, where plausibly she will know that the entry in question does not have a note associated with it, and then calls `bibtex-actions-open-`, the prefilters aren't relevant anymore. 

So ideally, you would have a single command for each of `pdf`, `notes`, and `link`, that would (1) add a corresponding pdf, note, or even link to KEY when called with input KEY and (2) list entries with associated pdf, notes, or link, when called without any input key. (This makes me wonder whether there's a way of merging the functionality of `bibtex-actions-add-pdf-to-library` into that of  `bibtex-actions-open-pdf`...)

bdarcus:
> I'm inclined to agree with @publicimageltd that setting the default value of 'notes' to `nil` might be confusing. I think it'd be best to have all `bibtex-actions-open-` commands treated the same way.

To be clear, we all agree on this, for the reasons @publicimageltd outlined.
 
> So ideally, you would have a single command for each of `pdf`, `notes`, and `link`, that would (1) add a corresponding pdf, note, or even link to KEY when called with input KEY and (2) list entries with associated pdf, notes, or link, when called without any input key. 

Before I wrap my head around this, I just want to make sure we're all on the same page WRT to notes.

Are you aware of this @apc?

https://github.com/tmalsburg/helm-bibtex#notes

This is what @publicimageltd is referring to.

I actually don't remember whether the add new notes if none exist functionality is built in to bibtex-completion, or it's automatically added when you use `org-roam-bibtex`.

But in any case, the config a user has to do it minimal.

I think our question for this PR is twofold:

1. which initial-value options need to be configurable. Clearly notes, but the others?
2. how should they be configured; alist vs simple boolean, etc.

apc:
> > I'm inclined to agree with @publicimageltd that setting the default value of 'notes' to `nil` might be confusing. I think it'd be best to have all `bibtex-actions-open-` commands treated the same way.
> 
> To be clear, we all agree on this, for the reasons @publicimageltd outlined.
> 
> > So ideally, you would have a single command for each of `pdf`, `notes`, and `link`, that would (1) add a corresponding pdf, note, or even link to KEY when called with input KEY and (2) list entries with associated pdf, notes, or link, when called without any input key.
> 
> Before I wrap my head around this, I just want to make sure we're all on the same page WRT to notes.
> 
> Are you aware of this @apc?
> 
> https://github.com/tmalsburg/helm-bibtex#notes
> 
> This is what @publicimageltd is referring to.

I'm not super clear on what you mean by 'this' here. Is it the option to work with a single note file or with multiple notes files? Or is it the functionality to create a note if none exists? I was aware of both, but I took @publicimageltd to be mostly talking about the latter. But maybe I'm misunderstanding something here? 

> I actually don't remember whether the "add new notes if none exist" functionality is built in to bibtex-completion, or it's automatically added when you use `org-roam-bibtex`.

No, it is built into `bibtex-completion`. See [here](https://github.com/tmalsburg/helm-bibtex/blob/9f6ea920a49457d85096caa0e61f086a42b2908e/bibtex-completion.el#L1394). 

> 
> But in any case, the config a user has to do it minimal.
> 
> I do agree it's not on by default, though, which is why it makes sense to filter by default.

What is not on by default? Sorry, I think I'm only half following here. Maybe I should be looking more carefully at the PR...

> 
> I think our question for this PR is twofold:
> 
> 1. which initial-value options need to be configurable. Clearly notes, but the others?
> 2. how should they be configured; alist with strings vs simple boolean, etc.

I see little harm in having the most general functionality available to users—let anyone customize what, if anything, they want as filters for the relevant actions, but give 'sane' defaults. I cannot think of a reason to use anything other than `nil' or one of the original values as initial filters, but I imagine most people will not need to tweak the defaults, so I say go for a "if you want to mess with these defaults, here, you can do whatever you want" approach. ;)

apc:
> > So ideally, you would have a single command for each of `pdf`, `notes`, and `link`, that would (1) add a corresponding pdf, note, or even link to KEY when called with input KEY and (2) list entries with associated pdf, notes, or link, when called without any input key.
> 
> Before I wrap my head around this, I just want to make sure we're all on the same page WRT to notes.
> 

Oh, and @bdarcus I didn't mean to suggest any changes. I was just thinking that there were two options for how to think of the filtering etc. 

1. You could have two commands for each of `pdf`, `note`, and `link`: one that adds and one that opens. This would essentially amount to adding an analogue of `bibtex-actions-add-pdf-to-library` for each of `note` and `link`. Filtering would make sense by default for the `open` command and not for the `add` command. 
2. Have a single 'open or edit/add if none exist' command for each of `note`, `link`, and `pdf`, and think of two different uses for each of them: when called with a KEY argument (what you effectively do, AFAIU, when you call these commands using `embark` from a selection in the minibuffer), the filtering mechanism is irrelevant and thus need not be modified; when called without a KEY argument, the filtering makes sense, and thus should be left as is (by default).  

I was suggesting that the better approach is 2. (At first, I thought this could, but need not, involve making `bibtex-actions-add-pdf-to-library` obsolete, and having `bibtex-actions-open-pdf` have a similar 'adding-if-none-present' fallback mechanism. But I think that's a mistake, since you may want more than one pdf file associated with a given entry, and that possibility is already taken into account by `bibtex-completion`.)

apc:
> > > But in any case, the config a user has to do it minimal.
> > > I do agree it's not on by default, though, which is why it makes sense to filter by default.
> > 
> > 
> > What is not on by default? Sorry, I think I'm only half following here. Maybe I should be looking more carefully at the PR...
> 
> Sorry. I complain about non-specific referents with my students all the time!

😆 

> So not only is it included, but it's on my default?

I think so, yes.

> 
> If that's the case, then I go back to my earlier point: I don't think we should filter by default.

I ultimately think it's fine either way. But, FWIW, even taking the `bibtex-completion` defaults into account, I think once you add reasonable filtering defaults to some, it makes sense to add them to all. It's simpler and you don't need to explain why you're handling notes differently. Especially since you *already* have a way of viewing your entire list of entries, without filtering, with `bibtex-actions-insert-citations` or `-open-entry`, and depending on how you end up handling #121, you will also have the option of having `bibtex-actions-open` give you the entire list of entries. 

That said: perhaps my mistake here is thinking that how you get to the list is essentially irrelevant because of embark, and not properly keeping in mind that someone may just want to create a note by default from the list of entries by simply pressing `RET` rather than having to do so via embark...

bdarcus:
Let me just explain why I initially added `has:pdf` and such to begin with, which is what all of this builds on.

I was consistently annoyed when wanting to open a document (pdf or link) that I didn't know which entries that action would return a result on.

So I wanted to filter on presence of difference related items, and also for the UI to reflect that.

But you, @apc, pointed out the next step was to incorporate this into the relevant commands.

But `bibtex-actions-open-notes` will never return nil, so it seems wrong to me to filter those, when the user can do that manually; in my case at least simply by doing `:n`.

apc:
I see. That makes sense. 

I was probably thinking about things differently. Since you already have indicators that will tell you which entries have pdfs, say, it never occurred to me you would call `bibtex-actions-open-pdf` on an entry you know doesn't have a pdf. The way I was thinking about it, it's only if you knew about the fallback mechanism that it made sense to call `bibtex-actions-open-notes` on an entry which the list of candidates tells you has no note to open. 

I guess there are two ways of thinking about what's best to do as a default: have it so that there's as little friction as possible for `helm-bibtex` and `ivy-bibtex` users to transition to using `bibtex-actions` *or* have it so that there's as little friction as possible for new users. It is of course hard to put oneself in the position of a new user here, but my hunch was that a new user would think there was something wrong with the package if she gets filtering with `-open-pdf` and `-open-link` but not with `-open-notes`. But I don't feel very confident about any of this, so... ¯\\\_(ツ)\_/¯

publicimageltd:
I'm kind of lost in this discussion now, to be honest. So what about utmost explicitness?

`bibtex-actions-open-or-create-notes` with no initial input and explicit reliance on the fallback mechanism;

`bibtex-actions-open-xxxx` with prefiltering and predefined action (open pdf, open notes file, scratch your head, whatever);

`bibtex-actions-open-source` with no filtering and a predefined action what to do with the result, but _no_ creation of new files whatsoever.

bdarcus:
Actually, here's a start of possible README content:

------------

By default, `bibtex-actions` will pre-filter entries for the following commands.

1. `bibtex-actions-open`: pre-narrows the list to those which have associated pdf or links
2. `bibtex-actions-open-link`: pre-narrows the list to those which have associated links
3. `bibtex-actions-open-pdf`: -pre-narrows the list to those which have associated pdf(s)

That is, upon running the command, an `initial-input` value will be inserted to narrow the results. You can also delete that if you prefer to see the full list of candidates.

If you want to modify those values, or remove them entirely, you can set `bibtex-actions-initial-inputs` like so; in this case turning off pre-filtering for `bibtex-actions-open-pdf`:

``` elisp
(setq bibtex-actions-initial-inputs
  '((pdf    . nil)
    (note   . nil)
    (link   . "has:link")
    (source . "has:link\\|has:pdf"))
```

bibtex-actions.el
```diff
@@ -97,6 +97,29 @@ manager like Zotero or JabRef."
   :group 'bibtex-actions
   :type '(repeat function))
 
+(defcustom bibtex-actions-initial-inputs
+  '((pdf    . "has:pdf")
+    (note   . "has:note")
```
bdarcus, line 102:
```suggestion
    (note   . nil)
```

PR #131, opened 2021-05-11
Clarify docstrings, possible name changes

### LINKED ISSUE #121: Preparing for 1.0; last call for command or function name changes

bdarcus:
Not convinced changing command names is worth the hassle at this point.

Regardless, clarifying the docstrings will be helpful and necessary, so this begins (and may end) with that.

Closes #121.

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -127,8 +127,8 @@ manager like Zotero or JabRef."
   "Keymap for 'bibtex-actions'.")
 
 ;;; Completion functions
-(cl-defun bibtex-actions-read (&optional &key initial rebuild-cache)
-  "Read bibtex-completion entries.
+(cl-defun bibtex-actions-read (&optional &key prompt initial rebuild-cache)
+  "Read bibtex-completion entries using PROMPT.
 
 This provides a wrapper around 'completing-read-multiple', with
 the following optional arguments:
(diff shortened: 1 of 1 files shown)


bdarcus:
#222 supercedes this

PR #132, opened 2021-05-12
Add bibtex-actions-complete-key-at-point

bdarcus:
Add a complete-at-point function, to complement `bibtex-actions-insert-citation`, targeting `org-cite`.

Some users may find this more efficient, in some cases. And we use the same candidates.

Basic completion works, with this the UI I'm looking for ...

Screenshot from 2021-06-03 10-45-42

Current status is outlined here: https://github.com/bdarcus/bibtex-actions/pull/132#discussion_r656325631.

Aside: I did make a recommendation to include this in `org-cite` directly, though I imagine that may not get support. My idea was just to allow the candidates list to be configurable, so that using it with `bibtex-actions` would just involve something like:

``` elisp
(setq org-cite-capf-candidates (bibtex-actions--get-candidates))
```

See #91.

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -234,12 +234,34 @@ If FORCE-REBUILD-CACHE is t, force reloading the cache."
     (bibtex-actions-refresh force-rebuild-cache))
   bibtex-actions--candidates-cache)
 
+(defun bibtex-actions-complete-key-at-point-org-cite ()
+    "Complete org-cite citation key at point.
+
+When inserting '@' in a buffer the capf UI will present user with
+a list of entries, from which they can narrow against a string
+which includes title, author, etc., and then select one.  This
+function will then return the key 'key', resulting in '@key' at
+point."
+  (when (eq ?@ (char-before))
+    (let* ((candidates (bibtex-actions--get-candidates))
+           ;; set both begin and end to point so we use capf UI to narrow and
+           ;; select
+           (begin (point))
+           (end (point)))
+      (list begin end candidates
+            :exit-function #'bibtex-actions--complete-key-at-point-exit-fn))))
+
+(defun bibtex-actions--complete-key-at-point-exit-fn (&optional str _status)
+  "Exit function to return key from a completion STR."
+    (let ((candidates (bibtex-actions--get-candidates)))
+      (delete-char (- (length str)))
+      (insert (cdr (assoc str candidates)))))
+
 ;;;###autoload
 (defun bibtex-actions-refresh (&optional force-rebuild-cache)
   "Reload the candidates cache.
 If called interactively with a prefix or if FORCE-REBUILD-CACHE
-is non-nil, also run the hook
-`bibtex-actions-before-refresh-hook'"
+is non-nil, also run the `bibtex-actions-before-refresh-hook' hook."
   (interactive "P")
   (when force-rebuild-cache
     (run-hooks 'bibtex-actions-force-refresh-hook))
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -234,28 +234,31 @@ If FORCE-REBUILD-CACHE is t, force reloading the cache."
     (bibtex-actions-refresh force-rebuild-cache))
   bibtex-actions--candidates-cache)
 
-(defun bibtex-actions-complete-key-at-point-org-cite ()
-    "Complete org-cite citation key at point.
+(defun bibtex-actions-complete-key-at-point ()
+    "Complete org-cite or pandoc citation key at point.
 
-When inserting '@' in a buffer the capf UI will present user with
-a list of entries, from which they can narrow against a string
-which includes title, author, etc., and then select one.  This
-function will then return the key 'key', resulting in '@key' at
-point."
-  (when (eq ?@ (char-before))
-    (let* ((candidates (bibtex-actions--get-candidates))
-           ;; set both begin and end to point so we use capf UI to narrow and
-           ;; select
-           (begin (point))
-           (end (point)))
-      (list begin end candidates
-            :exit-function #'bibtex-actions--complete-key-at-point-exit-fn))))
+When inserting '@' in a buffer the capf UI will present a list of
+entries, from which the user can narrow against a string which
+includes title, author, etc., and then select one. This function
+will then return the key 'key', resulting in '@key' at point.
 
-(defun bibtex-actions--complete-key-at-point-exit-fn (&optional str _status)
-  "Exit function to return key from a completion STR."
-    (let ((candidates (bibtex-actions--get-candidates)))
-      (delete-char (- (length str)))
-      (insert (cdr (assoc str candidates)))))
+Supports the pandoc and 'org-cite' key syntax, in either
+'org-mode' or 'markdown-mode'."
+    ; FIX
+    (when ;(and (or (eq major-mode 'org-mode)
+         ;         (eq major-mode 'markdown-mode))
+               (eq ?@ (char-before))
+      (let* ((candidates (bibtex-actions--get-candidates))
+             ;; set both begin and end to point so we use capf UI to narrow and
+             ;; select
+             (begin (point))
+             (end (point)))
+        (list begin end candidates
+              :exit-function
+              (lambda (str _status)
+                ;; take completion str and replace with key
+                (delete-char (- (length str)))
+                (insert (cdr (assoc str candidates))))))))
 
 ;;;###autoload
 (defun bibtex-actions-refresh (&optional force-rebuild-cache)


bibtex-actions.el
```diff
@@ -234,6 +234,36 @@ If FORCE-REBUILD-CACHE is t, force reloading the cache."
...
+  ;; this on 'bibtex-completion-format-citation-functions'?
+  (when (string-equal "@" (string (preceding-char)))
+    ;; Use same candidates as 'bibtex-read'.
+    (let (candidates (bibtex-actions--get-candidates))
+      (list
+       ;; TODO I adapted this from @minad's completion-in-region example, but
+       ;; not sure if it's correct.
+       (- (point) 1) (point)
+       ;; TODO need to return the cdr of the candidate; not the car.
+       candidates :exclusive 'no))))
```
bdarcus, line 265: Could you please help me with this @minad?

The candidate car is the search string, and the cdr is the key I need to insert.

minad: @bdarcus What is the question here exactly? 

In case you are looking for an example capf, you can try this:

~~~ elisp
(defun test-capf ()
  (list
   (save-excursion
     (backward-word)
     (point))
   (point)
   (lambda (str pred action)
     (message "CALLED %S %S" str action)
     (complete-with-action action '("a1" "a2" "a3" "a4"
                                    "b1" "b2" "b3" "b4"
                                    "c1" "c2" "c3" "c4")
                           str pred))))

(add-to-list 'completion-at-point-functions #'test-capf)
~~~

minad: Btw. the PR code is unreadable here on the gh web, due this gh action noise. I cannot see the wood for the trees.

bdarcus: > Btw. the PR code is unreadable here on the gh web, due this gh action noise

Sorry about that. Fixed.

I'll take a look at your example. Thanks.

bdarcus: @minad 

> In case you are looking for an example capf, you can try this:

I finally got back to this, and managed to get your example working with corfu!

So if I understand right, I probably need a new function to create the completion table, that I can call from both this function and from the completing-read one. Correct?

And how do I get it to activate with `@`, and for the result at point to be `@key` (as it would be in org-mode and with pandoc; latex does not include the `@`, but I'll worry about that later)?

The completion table has keys, without the `@`.

bdarcus: OK, I adapted your example, @minad, to mirror my case. 

The completion UI now works as I want ...

Screenshot from 2021-06-02 10-50-30

... but I'm stuck on one last thing: how do I get the cdr of the candidate instead of the car?

``` elisp
(defun test-capf ()
  "My capf."
  (when (looking-back "@[a-zA-Z]+")
    (list
     (save-excursion
       (backward-word)
       (point))
     (point)
     (lambda (str pred action)
       (let ((candidates '(("ay one" . "a1")
                           ("ay two" . "a2")
                           ("ay three" . "a3")
                           ("ay four" . "a4"))))
       (message "CALLED %S %S" str action)
       (complete-with-action action candidates str pred))))))
```

bdarcus:
You know anything about capf @publicimageltd?

I'm stuck.

publicimageltd:
Not at all, sorry. Don't even know how to activate completion manually; I kind of gave up understanding all of that once company mode worked in lisp mode, and that's all completion I ever needed. You might take a look at how `org roam` does it; they also provide completion for links.

bdarcus:
That's a good point. I also asked just now on stackexchange.

I think this will work with company (via company-capf), BTW; that's certainly the intention.

bdarcus:
Here's org-roam's, which I don't understand initially!

https://github.com/org-roam/org-roam/blob/f754160402ccc12b6e0e8bc6b607d7deef5c5b60/org-roam-link.el#L276

bibtex-actions.el
```diff
@@ -234,12 +234,55 @@ If FORCE-REBUILD-CACHE is t, force reloading the cache."
...
+
+(defun bibtex-actions-capf ()
+  "My test capf; alt."
+  (when (looking-back "@[a-zA-Z0-9]*" 5)
+    (let* ((candidates (bibtex-actions--get-candidates))
+           (begin (save-excursion
+                    (backward-word)
+                    (point)))
+           (end (point)))
+           (cdr
```
bdarcus, line 274: This is my attempt to grab the cdr from the chosen candidate, but it doesn't work (I get a "no match").

denismaier:
I'm not sure, but some may find it annoying if that pops up when using `@` symbols in regular text. (Don't know how often that happens, but anyway.) Is there a way to restrict that behavior to `@`s inside a citation? (In pandoc that's obviously more tricky since keys don't have to be inside brackets.) I'd prefer to just call a function via a custom keybinding.

bdarcus:
> I'm not sure, but some may find it annoying if that pops up when using `@` symbols in regular text. (Don't know how often that happens, but anyway.) Is there a way to restrict that behavior to `@`s inside a citation? (In pandoc that's obviously more tricky since keys don't have to be inside brackets.) I'd prefer to just call a function via a custom keybinding.

I thought about this, and hadn't come to any conclusion.

1. yes, in some ways better to constrain for the more strict org-cite syntax (and later latex); but as you note, you lose pandoc
2. with `corfu` at least, default is one has to hit `TAB` to activate; I believe one can configure the same in `company`

So I'm really not sure WTD.

:shrug:

bibtex-actions.el
```diff
@@ -234,12 +234,50 @@ If FORCE-REBUILD-CACHE is t, force reloading the cache."
...
+       (point)
+       (mapcar #'cdr candidates) :exclusive 'no))))
+
+(defun bibtex-actions-capf ()
+  "My test capf; alt."
+  (when (looking-back "@[a-zA-Z0-9]*" 5)
+    (let* ((candidates (bibtex-actions--get-candidates))
+           ;; FIX
+           (cand (match-string-no-properties 1))
+           (begin (save-excursion (backward-word) (point)))
```
bdarcus, line 272: This may be wrong now, but I don't actually understand what "save-excursion" does here.

Either way, though, the string can be more than a word.

bibtex-actions.el
```diff
@@ -234,12 +234,50 @@ If FORCE-REBUILD-CACHE is t, force reloading the cache."
...
+         (point))
+       (point)
+       (mapcar #'cdr candidates) :exclusive 'no))))
+
+(defun bibtex-actions-capf ()
+  "My test capf; alt."
+  (when (looking-back "@[a-zA-Z0-9]*" 5)
+    (let* ((candidates (bibtex-actions--get-candidates))
+           ;; FIX
+           (cand (match-string-no-properties 1))
```
bdarcus, line 271: I need some way to save the selected candidate string, so that I can use it to look up the associated citekey.

bibtex-actions.el
publicimageltd: `save-excursion` simply keeps point where it is. The expression moves one word backwards, returns point, and then restores point to wherever it was previously. If you want to move more than one word, you should write your own regexp to match whatever you want (via `looking-back`).

denismaier: If my understanding of  `save-excursion` is correct, this is mostly used for moving around in a buffer performing some action, such as replacing stuff, without driving the user insane, right?

bibtex-actions.el
publicimageltd: I'm not sure I understand what you need (or what doesn't) work, but it seems like you should simply call `match-string` directly after `looking-back`. To return what you have found at via `looking-back`, use `(match-string-no-properties 0)`

```emacs-lisp
(when (looking-back regex 5)
  (let* ((match (match-string-no-properties 0))
           (candidates (bibtex-actions--get-candidates))
;;           .....assoc....
   )))
```

bdarcus: Right now, I get "no match" when invoking this.

Earlier, I got correct UI (per screenshot), but the returned string is the full candidate, rather than its key, which is stored in the cdr.

In the CR function, we search the candidates list, store the selected ones, and then iterate through those to return the list of keys.

Here's there's just one key though.

So I am looking for something similar.

My problem so far is there's more than one piece that confuses me!

bdarcus: Here's the initial CR (before multiple) function, for example:

``` elisp
(defun bibtex-actions--completing-read ()
  "Read bibtex-completion entries for completion using 'completing-read'."
  (bibtex-completion-init)
  (when-let ((candidates (bibtex-actions--get-candidates))
             ;; store selected candidate
             (chosen
              (completing-read
               "BibTeX entries: "
                 (if (eq action 'metadata)
                     '(metadata
                       ;; (annotation-function . bibtex-completion--annotation)
                       (category . bibtex))
                   (complete-with-action action candidates string predicate))))))
    ;; return selected candidate cdr (key)
    (cdr (assoc chosen candidates))))
```

bdarcus: Added `bibtex-actions-insert-key-2" alternative, which was attempting to mirrror the CR function.

"No match."

publicimageltd: Sorry, I'm out for now, I don't have the time to test it on my machine, and I really don't know this capf stuff...

bdarcus: No problem; thanks for the help!

bibtex-actions.el
```diff
@@ -234,12 +234,31 @@ If FORCE-REBUILD-CACHE is t, force reloading the cache."
...
+    "Complete org-cite citation key at point.
+
+When inserting '@' in a buffer the capf UI will present user with
+a list of entries, from which they can narrow against a string
+which includes title, author, etc., and then select one.  This
+function will then return the key 'key', resulting in '@key' at
+point."
+    ;; FIX exit-function is wrong; results in "no match"
+    ;; TODO don't need a regex for org-cite; what to use instead?
+  (when (and (eq major-mode 'org-mode) (looking-back "@[a-zA-Z]+" 5))
```
bdarcus, line 247: @ilupin - since you figured out the org-cite code for your at-point, do you know how I would adapt that to here?

I added the org-mode check of course.

bdarcus:
Main post now updated to reflect the current status.

I think this is close to working, but there's some little detail I'm not understanding.

bdarcus:
I just pushed a commit that gets the basics of at-point key insertion working.

But there are a few rough edges/issues to resolve:

1. I commented out the conditional to constrain when this is active, because it was too strict earlier. Need to add it back, such that it works.
2. If you type only `@`, you don't get any results. Not sure if this matters though.
3. I also commented out the named exit function; not sure this is needed or not.
4. need to add back "exclusive nil"

I think that's it.

This, combined with #142, will be a nice boost in features in preparation for org-cite!

Any thoughts on this, particularly 1, @ilupin?

bibtex-actions.el
```diff
@@ -234,12 +234,35 @@ If FORCE-REBUILD-CACHE is t, force reloading the cache."
...
 
+(defun bibtex-actions-complete-key-at-point-org-cite ()
+    "Complete org-cite citation key at point.
+
+When inserting '@' in a buffer the capf UI will present user with
+a list of entries, from which they can narrow against a string
+which includes title, author, etc., and then select one.  This
+function will then return the key 'key', resulting in '@key' at
+point."
+    ;; FIX exit-function is wrong; results in "no match"
```
bdarcus, line 245: I forgot to remove this comment. This is only true if the below lines are uncommented.

bibtex-actions.el
```diff
@@ -234,12 +234,32 @@ If FORCE-REBUILD-CACHE is t, force reloading the cache."
...
 
+(defun bibtex-actions-complete-key-at-point-org-cite ()
+    "Complete org-cite citation key at point.
+
+When inserting '@' in a buffer the capf UI will present user with
+a list of entries, from which they can narrow against a string
+which includes title, author, etc., and then select one.  This
+function will then return the key 'key', resulting in '@key' at
+point."
+  (when (looking-back "@[a-zA-Z0-9]*" 5))
```
bdarcus, line 245: See https://github.com/bdarcus/bibtex-actions/pull/132#issuecomment-865062449

bibtex-actions.el
```diff
@@ -242,12 +242,12 @@ a list of entries, from which they can narrow against a string
 which includes title, author, etc., and then select one.  This
 function will then return the key 'key', resulting in '@key' at
 point."
-  (when (looking-back "@[a-zA-Z0-9]*" 5))
+  (when (eq ?@ (char-before))
```
bdarcus, line 245: @minad - can you please help me with this?

I think this is actually correct, as in it appears to activate (I get a message that confirms the bibliography is parsed), but it returns no results.

I basically want a user to be able to type only `@`, and have the full list of candidates presented by `corfu`, as they are in `vertico` or `selectrum`, and then to narrow from there.

I can revert the change, but that just doesn't seem right from a UX POV.

Perhaps there's something else I need to add here to do that? Maybe `:predicate`?

minad: Maybe @ is part of the input string, such that this will filter out the candidates?

bdarcus: > Maybe @ is part of the input prefix (as determined by the start end position of the completion), such that this will filter out the candidates?

Does it make sense to you that the previous version worked as expected?

bdarcus: > And if yes, what is the appropriate way to set the positions in that case? Set both begin and end to point?

This works @minad; thanks for the hint!

bdarcus: If in the end, we need to support a few variations on this, do you recommend separate functions for each, or one capf?

1. pandoc markdown; syntax is `[@key]` or even the bare `@key`; so rather loose
1. this PR is currently for the forthcoming org-cite, to be merged in org-mode "real soon"; syntax is similar to 1, but tighter `[cite:@key]`.
3. latex; lower priority for me personally, and the syntax is too chaotic to easily constrain

bdarcus:
Last thing I'd like to figure out before merging this (we can enhance it later).

The current code works, but is unconstrained. So if you type "@" in any mode, it will activate.

But if I add this simple "and" condition to it, it stops activating.

``` elisp
    (when (and (or (eq major-mode 'org-mode)
                   (eq major-mode 'markdown-mode))
               (eq ?@ (char-before)))
```
Any ideas?

bdarcus:
I'm tired of looking at this PR, so merged without the tighter conditions. Will fix that separately, along with the documentation.

PR #139, opened 2021-06-06
Add bibtex-actions-get-key-org-cite

bdarcus:
Allows `bibtex-completions-at-point` to work with `org-cite`.

Are you OK with me adding this on your behalf, @ilupin?

Though it might be better to see if you can get it accepted to `bibtex-completion`?

cc @tmalsburg

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -40,6 +40,7 @@
 (require 'bibtex-completion)
 
 (declare-function org-element-context "org-element")
+(declare-function org-element-property "org-element")
 
 ;;; Declare variables for byte compiler
 
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -38,6 +38,9 @@
 ;;; Code:
 
 (require 'bibtex-completion)
+(require 'org-element)
+
+(declare-function org-element-context "org-element")
 
 ;;; Declare variables for byte compiler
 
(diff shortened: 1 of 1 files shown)


ilupin:
Yes. For `bibtex-completion`, it can wait until `org-cite` is merged.

bibtex-actions.el
```diff
@@ -38,6 +38,9 @@
 ;;; Code:
 
 (require 'bibtex-completion)
+(require 'org-element)
```
ilupin, line 41: This line can be deleted because `org-element-context` has been declared below

bdarcus: I edited your suggestion to use the "suggested" feature, so I could click to commit.

But, as you can see from the CI, I get an error without the require. I think that's why I added it to begin with.

Seems I need to add it back?

bibtex-actions.el
```diff
@@ -39,6 +39,8 @@
 
 (require 'bibtex-completion)
 
+(declare-function org-element-context "org-element")
+
```
ilupin, line 43:
```suggestion
(declare-function org-element-property "org-element")

```

Declaring `org-element-property`, which is also used below, should eliminate the warning

bdarcus:
BTW, I just realized the `oc-basic` has a "follow" (at-point) function tied to `M-x org-open-at-point`.

By default it opens the bibtex entry, but I've asked to make it configurable, so we could just do something like:

```
(setq org-cite-basic-open-fn bibtex-completion-open)
```

Issue #142, opened 2021-06-10
Add keymap, embark instructions, etc for buffer (at-point) actions


bdarcus:
This can be a killer feature in general, and when coupled with `org-cite`, and would feature to include for the 1.0 release.

In effect it would provide a super flexible kind of embarkified micro-org-ref (emphasis on "micro" though); or at least a framework for that to evolve.

What do we need?

1. a keymap specific to this case; maybe `bibtex-actions-buffer-map`? It would obviously include all the "open" commands; not sure what else.
2. README content for this in configuration and usage, in part to promote it (once org-cite is merged, I'll want to mention support for it prominently at the top, and include a screenshot), in part to make it easy for people to get working

Do we need any actual code here, or just code examples in the README? 

For example, @publicimageltd mentioned [here](https://github.com/bdarcus/bibtex-actions/pull/137#issuecomment-857649435): 

> ... would be cool, of course, to have some kind of automatic plug in, e.g. with eval-after-load. 

Not sure the idea here; would it apply to embark-minibuffer too? As in, rather than document the config in the README, we'd set it up for people if they have embark installed?

Also, now understanding all this, and recognizing the limitations of current CRM UIs, **I think we need an option, maybe default, to run at-point functions directly, without further user interaction**.

I'll be traveling for a couple of weeks starting tomorrow (though will still have my computer and access; just maybe not as much), so thought I'd just start this here, while ideas are fresh. If someone wants to start a PR, please feel free.

See also #141 and #140.

cc @ilupin

publicimageltd:
Good idea! I am not sure I understand all questions or possibilities listed above. So here's my vision:
 - provide a keymap for all "open" actions, which can then be used for embark
 - provide some code which would hook these actions in embark, combined with the target finder, once embark is (also) loaded (the `with-eval-after-load` idea)
 - Provide some functions or functionality to call all open functions also directly "on point". This could be done either via "initial input" in the current functions, or by providing specific functions which then use the initial input.

Is this what we are talking about? I am currently also more offline, RL is using up my time; but if there is a clear roadmap, I'd like to contribute some of this logic. Or at least I'd continue commenting stuff without even updating my local git repo :smile:

bdarcus:
Sorry; I sometimes get a bit wordy as I'm trying to sort things out in my head, and am a bit distracted. But ...

> Is this what we are talking about? 

But yes @publicimageltd, that sounds perfect to me!

One little amendment:

[...]

PR #143, opened 2021-06-12
Add `embark` support for citation keys at point

ilupin:
This PR enhances `embark` support for `bibtex-actions-at-point`, which is inspired by the discussion in #142 and the [code](https://github.com/bdarcus/bibtex-actions/pull/137#issuecomment-857456998) presented by @publicimageltd.

If `embark` is installed, `bibtex-actions-at-point` will call `embark-dwim` (the default) or `embark-act`, depending on the option `bibtex-actions-embark-dwim`, when one or more citation keys are found at point, where `embark-dwim` runs the default action and `embark-act` prompts the user for an action in `bibtex-actions-embark-map`. If no key is found, `bibtex-actions-read` is used for choosing entries, and functions in `bibtex-actions-map` can still be called via `embark-act`.

This PR uses `bibtex-actions-get-key-org-cite` and `bibtex-completion-key-at-point` to find citation keys, and should work with `org-cite` and latex-mode/bibtex-mode buffers.

<del>BTW, `embark-target-finders` requires the TARGET at point being a string, while `bibtex-completion-*` only accept a list of citation keys as input. Therefore, `bibtex-completion-*` cannot be used in `bibtex-actions-embark-map` directly, and `bibtex-actions-embark-*` does the argument conversion.</del>

### DIFF
--- a/README.org
+++ b/README.org
@@ -306,7 +306,7 @@ From there, you can run different actions on the candidates at will, rather than
 
 =M-x bibtex-actions-at-point= will run the default action on citation keys found at point directly.
 
-By setting =bibtex-actions-embark-dwim= to nil, other actions in =bibtex-actions-buffer-map= will be prompted if =embark= is installed. Press =RET= will run the default action.
+If =embark= is installed, setting =bibtex-actions-embark-dwim= to nil will prompt for other actions in =bibtex-actions-buffer-map= . From there, pressing =RET= will run the default action.
 
 If no citation key is found, all items will be shown in the minibuffer for selection. This behavior can be disabled by setting =bibtex-actions-at-point-fallback= to nil.
 
--- a/README.org
+++ b/README.org
@@ -233,6 +233,25 @@ You can also extend this to do the same thing for your PDF files, or notes:
 
 For additional configuration options on this, see [[https://github.com/bdarcus/bibtex-actions/wiki/Configuration#automating-path-watches][the wiki]].
 
+*** Finding citation keys at point
+    :PROPERTIES:
+    :CUSTOM_ID: finding-citation-keys-at-point
+    :END:
+
+=bibtex-actions-at-point= can find citation keys at point in org-mode buffer, latex-mode buffer, etc. To add support for other major modes or citation syntax, you can write a function (below is an example for =org-cite=) and add it to =bibtex-completion-key-at-point-functions=.
+
+#+begin_src emacs-lisp
+(defun bibtex-actions-get-key-org-cite ()
+  "Return key at point for org-cite citation-reference."
+  (when-let (((eq major-mode 'org-mode))
+             (elt (org-element-context)))
+    (pcase (org-element-type elt)
+      ('citation-reference
+       (org-element-property :key elt))
+      ('citation
+       (org-cite-get-references elt t)))))
+#+end_src
+
 ** Usage
    :PROPERTIES:
    :CUSTOM_ID: usage
@@ -279,6 +298,17 @@ From there, you can run the same options discussed above using =embark-act= (whi
 
 So, for example, say you are working on a paper. You hold the complete super-set of items you are interested in citing at some point in that buffer.
 From there, you can run different actions on the candidates at will, rather than search individually for each item you want to cite.
+
+*** Use =bibtex-actions-at-point=
+    :PROPERTIES:
+    :CUSTOM_ID: use-bibtex-actions-at-point
+    :END:
+
+=M-x bibtex-actions-at-point= will run the default action on citation keys found at point directly.
+
+By setting =bibtex-actions-embark-dwim= to nil, other actions in =bibtex-actions-buffer-map= will be prompted if =embark= is installed. Press =RET= will run the default action.
+
+If no citation key is found, all items will be shown in the minibuffer for selection. This behavior can be disabled by setting =bibtex-actions-at-point-fallback= to nil.
 
 ** Comparisons
    :PROPERTIES:
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -41,10 +41,17 @@
 
 (declare-function org-element-context "org-element")
 (declare-function org-element-property "org-element")
+(declare-function org-element-type "org-element")
+(declare-function org-cite-get-references "org-cite")
+(declare-function embark-act "embark")
 
 ;;; Declare variables for byte compiler
 
 (defvar crm-separator)
+(defvar embark-keymap-alist)
+(defvar embark-target-finders)
+(defvar embark-general-map)
+(defvar embark-meta-map)
 
 ;;; Variables
 
@@ -143,7 +150,23 @@ you can use the following initial inputs: \"has:pdf\",
   "The default action for the `bibtex-actions-at-point' command."
   :group 'bibtex-actions
   :type 'function)
-    
+
+(defcustom bibtex-actions-at-point-fallback 'prompt
+  "Fallback action for `bibtex-actions-at-point'.
+The action is used when no citation key is found at point.
+`prompt' means choosing entries via `bibtex-actions-read' and nil
+means no action."
+  :group 'bibtex-actions
+  :type '(choice (const :tag "Prompt" 'prompt)
+                 (const :tag "Ignore" nil)))
+
+(defcustom bibtex-actions-embark-dwim t
+  "Whether to run the default action on citation keys at point.
+If non-nil, call the default embark action through `embark-dwim'.
+If nil, prompt the user for an action through `embark-act'."
+  :group 'bibtex-actions
+  :type 'boolean)
+
 ;;; History, including future history list.
 
 (defvar bibtex-actions-history nil
@@ -417,10 +440,39 @@ TEMPLATE."
 ;; something equivalent.
 (defun bibtex-actions-get-key-org-cite ()
   "Return key at point for org-cite citation-reference."
-  (when (eq major-mode 'org-mode)
-    (let ((elt (org-element-context)))
-      (if (eq (car elt) 'citation-reference)
-          (org-element-property :key elt)))))
+  (when-let (((eq major-mode 'org-mode))
+             (elt (org-element-context)))
+    (pcase (org-element-type elt)
+      ('citation-reference
+       (org-element-property :key elt))
+      ('citation
+       (org-cite-get-references elt t)))))
+
+;;; Embark
+
+(defun bibtex-actions-citation-key-at-point ()
+  "Return citation keys at point as a list for `embark'."
+  (when-let ((key (or (bibtex-actions-get-key-org-cite)
+                      (bibtex-completion-key-at-point))))
+    (cons 'citation-key (if (listp key) key (list key)))))
+
+(defvar bibtex-actions-buffer-map
+  (let ((map (make-sparse-keymap)))
+    (define-key map (kbd "t") 'bibtex-actions-add-pdf-attachment)
+    (define-key map (kbd "a") 'bibtex-actions-add-pdf-to-library)
+    (define-key map (kbd "o") 'bibtex-actions-open)
+    (define-key map (kbd "e") 'bibtex-actions-open-entry)
+    (define-key map (kbd "l") 'bibtex-actions-open-link)
+    (define-key map (kbd "n") 'bibtex-actions-open-notes)
+    (define-key map (kbd "p") 'bibtex-actions-open-pdf)
+    (define-key map (kbd "RET") 'bibtex-actions-run-default-action)
+    map)
+  "Keymap for Embark citation-key actions.
+
+This keymap is used by `bibtex-actions-at-point' internally.  It
+should not be added to `embark-keymap-alist' because interactive
+commands (i.e. bibtex-actions-*) currently do not work well with
+`embark-act' outside the minibuffer.")
 
 ;;; Command wrappers for bibtex-completion functions
 
@@ -512,15 +564,44 @@ With prefix, rebuild the cache before offering candidates."
   (interactive (list (bibtex-actions-read :rebuild-cache current-prefix-arg)))
   (bibtex-completion-add-pdf-to-library keys))
 
+(defun bibtex-actions-run-default-action (keys)
+  "Run the default action `bibtex-actions-default-action' on KEYS."
+  (funcall bibtex-actions-default-action keys))
+
 ;;;###autoload
 (defun bibtex-actions-at-point (&optional arg)
-  "Search BibTeX entries and call `bibtex-actions-default-action'.
-With prefix ARG, rebuild the cache before offering candidates.
-The bibtex key found by `bibtex-completion-key-at-point' is used
-as the initial input."
+  "Run the default action on citation keys found at point.
+If no citation key is found, target entries can be chosen
+interactively when `bibtex-actions-at-point-fallback' is non-nil.
+With prefix ARG, rebuild the cache before offering candidates."
   (interactive "P")
-  (funcall bibtex-actions-default-action
-           (bibtex-actions-read :initial 'point :rebuild-cache arg)))
+  (if (fboundp 'embark-dwim)
+      (let ((embark-keymap-alist '((bibtex . bibtex-actions-map))))
+        (condition-case err
+            (let ((embark-target-finders '(bibtex-actions-citation-key-at-point))
+                  (embark-keymap-alist '((citation-key . bibtex-actions-buffer-map)))
+                  ;; `embark-general-map' is useless in this case
+                  ;; because we are not in the minibuffer and the
+                  ;; TARGET is not string
+                  (embark-general-map embark-meta-map))
+              ;; This is a workaround: unless in the minibuffer,
+              ;; embark cannot pass TARGET to action when the action
+              ;; is an interactive command
+              (cl-letf (((symbol-function 'embark--act)
+                         (lambda (action target _) (funcall action target))))
+                (if bibtex-actions-embark-dwim
+                    (embark-dwim)
+                  (embark-act))))
+          (user-error
+           (when (and (string-equal (error-message-string err) "No target found")
+                      bibtex-actions-at-point-fallback)
+             (bibtex-actions-run-default-action
+              (bibtex-actions-read :rebuild-cache arg))))))
+    (if-let ((keys (bibtex-actions-citation-key-at-point)))
+        (funcall bibtex-actions-default-action keys)
+      (when bibtex-actions-at-point-fallback
+        (bibtex-actions-run-default-action
+         (bibtex-actions-read :rebuild-cache arg))))))
 
 (provide 'bibtex-actions)
 ;;; bibtex-actions.el ends here
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -585,7 +585,10 @@ With prefix ARG, rebuild the cache before offering candidates."
               ;; embark cannot pass TARGET to action when the action
               ;; is an interactive command
               (cl-letf (((symbol-function 'embark--act)
-                         (lambda (action target _) (funcall action target))))
+                         (lambda (action target _)
+                           (if (where-is-internal action (list bibtex-actions-buffer-map))
+                               (funcall action target)
+                             (command-execute action)))))
                 (if bibtex-actions-embark-dwim
                     (embark-dwim)
                   (embark-act))))
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -125,14 +125,13 @@ manager like Zotero or JabRef."
   '((pdf    . "has:pdf")
     (note   . nil)
     (link   . "has:link")
-    (source . "has:link\\|has:pdf")
-    (point  . bibtex-completion-key-at-point))
+    (source . "has:link\\|has:pdf"))
   "Alist defining the initial input for some bibtex open actions.
 Given a flexible completion style, this will restrict the list of
 available candidates to those matching the initial input.
 
 The association key can be one of the symbols `pdf', `note',
-`link', `source' or `point' and defines the input for the function
+`link' or `source' and defines the input for the function
 `bibtex-action-open-pdf', `bibtex-action-open-link', etc.  The
 associated value must be nil, meaning that there will be no
 initial input, or a string.
@@ -143,8 +142,7 @@ you can use the following initial inputs: \"has:pdf\",
   :group 'bibtex-actions
   :type '(alist :key-type symbol
                 :value-type (choice string
-                                    (const :tag "No initial input" nil)
-                                    (const :tag "Key at point" bibtex-completion-key-at-point))))
+                                    (const :tag "No initial input" nil))))
 
 (defcustom bibtex-actions-default-action 'bibtex-actions-open
   "The default action for the `bibtex-actions-at-point' command."
@@ -259,9 +257,8 @@ offering the selection candidates"
            nil
            nil
            (and initial-input
-                (if (stringp initial-input)
-                    (concat initial-input " ")
-                  (funcall initial-input)))
+                (stringp initial-input)
+                (concat initial-input " "))
            'bibtex-actions-history bibtex-actions-presets nil)))
     (cl-loop for choice in chosen
              ;; Collect citation keys of selected candidate(s).


bibtex-actions.el
```diff
@@ -417,10 +421,67 @@ TEMPLATE."
 ;; something equivalent.
 (defun bibtex-actions-get-key-org-cite ()
   "Return key at point for org-cite citation-reference."
-  (when (eq major-mode 'org-mode)
-    (let ((elt (org-element-context)))
-      (if (eq (car elt) 'citation-reference)
-          (org-element-property :key elt)))))
+  (when-let (((eq major-mode 'org-mode))
```
publicimageltd, line 424: I think that should be `when-let*`, since the second let binding depends on the buffer being in org mode, which is checked in the first binding.

bdarcus:
Any thoughts on the CI warning on "with-eval-after-load" @ilupin @publicimageltd?

Any general comments on this embark-focused PR @oantolin?

publicimageltd:
> Any thoughts on the CI warning on "with-eval-after-load" @ilupin @publicimageltd?

AFAIU it's a warning which we can ignore insofar we do know what we do. But I don't know offhand how to silence this warning.

bdarcus:
> > Any thoughts on the CI warning on "with-eval-after-load" @ilupin @publicimageltd?
> 
> AFAIU it's a warning which we can ignore insofar we do know what we do. But I don't know offhand how to silence this warning.

And are we comfortable ignoring it?

Just want to confirm :-)

BTW, I tried to test this, but I have some problems on my end (long story).

So a question. if in `org-mode`, I run `C-c C-o` on an `org-cite` citation, what's intended behavior?

bdarcus:
> BTW, `embark-target-finders` requires the TARGET at point being a string, while `bibtex-completion-*` only accept a list of citation keys as input. Therefore, `bibtex-completion-*` cannot be used in `bibtex-actions-embark-map` directly, and `bibtex-actions-embark-*` does the argument conversion.

@ilupin - are there any other alternatives you considered for this?

oantolin:
> BTW, embark-target-finders requires the TARGET at point being a string, while bibtex-completion-* only accept a list of citation keys as input. 

I haven't read any of this new code, but this sounded a little suspicious to me. Here's what I thought:

- The bibtex-completions-* functions take lists of citation key as input, but
- they are interactive commands that call completing-read-multiple; interactively they take a crm-separator-separated list of strings which gets split and parsed into a list; therefore
- The target finder can return a single string, which will be injected into the minibuffer prompt and "split" into a singleton list.

bdarcus:
> * the target finder can return a single string, which will be injected into the minibuffer prompt and "split" into a singleton list.

Thanks @oantolin - this sounds familiar from our discussions of embark and CRM.

If this indeed the case, and we could avoid having to add specialized functions to do this, would have a huge advantage:

We could then just have a standardized API for these functions; take a list of keys and do X. 

This would then allow mixing-and-matching these functions from different packages.

Right???

bibtex-actions.el
```diff
@@ -417,10 +421,67 @@ TEMPLATE."
...
+
+(defun bibtex-actions-citation-key-at-point ()
+  "Return citation keys at point as string for `embark'."
+  (when-let ((key (or (bibtex-actions-get-key-org-cite)
+                      (bibtex-completion-key-at-point))))
+    (cons 'citation-key (if (listp key) (string-join key ";") key))))
+
+(with-eval-after-load 'embark
+  (add-to-list 'embark-target-finders 'bibtex-actions-citation-key-at-point)
+  (add-to-list 'embark-keymap-alist '(citation-key . bibtex-actions-embark-map))
```
bdarcus, line 442: Minor thing, but does the singular "key" suggest it won't work on multi-cites, like:

``` latex
\cite{doe,smith}
```
``` org
[cite:@doe;@smith]
```

bibtex-actions.el
```diff
@@ -519,8 +580,11 @@ With prefix ARG, rebuild the cache before offering candidates.
 The bibtex key found by `bibtex-completion-key-at-point' is used
 as the initial input."
   (interactive "P")
-  (funcall bibtex-actions-default-action
-           (bibtex-actions-read :initial 'point :rebuild-cache arg)))
+  (if (and (fboundp 'embark-act)
+           (bibtex-actions-citation-key-at-point))
+      (embark-act)
+    (funcall bibtex-actions-default-action
+             (bibtex-actions-read :initial 'point :rebuild-cache arg))))
```
bdarcus, line 587: Given the conversation, do we still need/want the initial-value here? Couldn't we just pass the keys to `'bibtex-actions-default-action`?

More consistent, etc.

bibtex-actions.el
ilupin: `when-let` evaluates each binding in turn and stops if a binding value is nil, i.e. the second let binding is not evaluated in non org-mode buffers.

bibtex-actions.el
ilupin: `bibtex-actions-get-key-org-cite` supports multiple keys in org mode if the point is at the citation object, e.g.

```
[cite‸:@doe;@smith]
```

bibtex-actions.el
ilupin: This function has been changed as suggested.

ilupin:
> Any thoughts on the CI warning on "with-eval-after-load"

How about `(require 'embark nil t)`?

> So a question. if in org-mode, I run C-c C-o on an org-cite citation, what's intended behavior?

You need the following code to make `C-c C-o` work on citations.
```
 (org-cite-register-processor 'bibtex-actions
    :follow (lambda (_ _) (call-interactively 'bibtex-actions-at-point)))

(setq org-cite-follow-processor 'bibtex-actions)
```

ilupin:
> > BTW, embark-target-finders requires the TARGET at point being a string, while bibtex-completion-* only accept a list of citation keys as input.
> 
> I haven't read any of this new code, but this sounded a little suspicious to me. Here's what I thought would happen:
> 
> * The bibtex-completions-* functions take lists of citation key as input, but
> * they are interactive commands that call completing-read-multiple; interactively they take a crm-separator-separated list of strings which gets split and parsed into a list; therefore
> * the target finder can return a single string, which will be injected into the minibuffer prompt and "split" into a singleton list.
> 
> But as I said, I haven't looked at this carefully.

As @bdarcus said, we are dealing with in-buffer citation keys. They need to be passed to `bibtex-completion-open-pdf`, etc., as a list. 

`bibtex-actions-citation-key-at-point` returns `(citation-key . "key1;key2")` for multiple keys now, but I find it actually works if it returns `(citation-key . ("key1" "key2"))` and its CDR is directly passed to `bibtex-completion-open-pdf`. I don't know if it causes other problems.

bibtex-actions.el
bdarcus: I was just wondering about the name. But it's a minor thing, as I said.

bdarcus:
> > Any thoughts on the CI warning on "with-eval-after-load"
> 
> How about `(require 'embark nil t)`?

I'm not really sure myself. Does that just silence the warning, but the setup code then works if embark is present?

I was also wondering, will there be any problem with a user using different maps in their config?
 
> > So a question. if in org-mode, I run C-c C-o on an org-cite citation, what's intended behavior?
> 
> You need the following code to make `C-c C-o` work on citations.

While I can probably help you, could you please include README additions and modifications for all this, when the code is settled?

bibtex-actions.el
```diff
@@ -417,10 +421,67 @@ TEMPLATE."
...
+    (define-key map (kbd "o") 'bibtex-actions-embark-open)
+    (define-key map (kbd "e") 'bibtex-actions-embark-open-entry)
+    (define-key map (kbd "l") 'bibtex-actions-embark-open-link)
+    (define-key map (kbd "n") 'bibtex-actions-embark-open-notes)
+    (define-key map (kbd "p") 'bibtex-actions-embark-open-pdf)
+    (define-key map (kbd "RET") 'bibtex-actions-embark-open)
+    map)
+  "Keymap for Embark citation-key actions.")
+
+(defun bibtex-actions-embark-open (keys)
```
bdarcus, line 458: Next big question: why do we need these functions? 

Can you not just have a little function to split if a string, and leave alone otherwise, so that you can remove them entirely?

Aside from great maintenance burden and less flexibility, it results in really long names that are a problem in which-key, at least.

Screenshot from 2021-06-13 11-09-34

BTW, if any has a great shorter and more general (non-bibtex-specific) name for this package ... :-)

bibtex-actions.el
```diff
@@ -417,10 +421,67 @@ TEMPLATE."
...
+;;; Embark
+
+(defun bibtex-actions-citation-key-at-point ()
+  "Return citation keys at point as string for `embark'."
+  (when-let ((key (or (bibtex-actions-get-key-org-cite)
+                      (bibtex-completion-key-at-point))))
+    (cons 'citation-key (if (listp key) (string-join key ";") key))))
+
+(with-eval-after-load 'embark
+  (add-to-list 'embark-target-finders 'bibtex-actions-citation-key-at-point)
```
bdarcus, line 441: Another question (and it is a question; I'm not sure) for both @publicimageltd and @ilupin:

Right now, one effectively cannot get to the default action directly. One must open `embark`, and then hit `RET`.

And both `bibtex-actions-at-point` and `embark-act` do the exact same thing; call `embark-act`.

Is that the right UX?

bibtex-actions.el
publicimageltd: You are right! Expanding the macro shows it is using `let*` under the hood. Didn't know this; I always used `when-let*` because I took it to be similar to `let` vs. `let*`. Good to know, thanks!

publicimageltd:
> > > Any thoughts on the CI warning on "with-eval-after-load"
> > 
> > 
> > How about `(require 'embark nil t)`?

I wouldn't do that, because we do not want to interfere with the way the user wants embark to be loaded. The `require` statement forces to load it, which I consider to be a bad style for a package (users might not expect it to be loaded).

After looking at the info page for `with-eval-after-load`, I still think it's the right choice. It says we should set the variables of an external packages if we need to set them anyways. This is true for all variables declared via `defvar` and `defcustom`. But still, we need to add something to a list -- so we need `embark` to initialize the variable before we set it, which is exactly what can be accomplished with `with-eval-after-load`.

See this example code:
```emacs-lisp
(progn
  (setq my-var nil)
  (add-to-list 'my-var :element)
  (defvar my-var '(a b c))
  my-var)
;; -> yields (:element), but we would want (a b c :element)
```

bibtex-actions.el
```diff
@@ -41,10 +41,14 @@
 
 (declare-function org-element-context "org-element")
 (declare-function org-element-property "org-element")
+(declare-function org-element-type "org-element")
+(declare-function org-cite-get-references "org-cite")
```
bdarcus, line 45: If this function doesn't exist (because user is using an older version of org), that's no problem; right?

bibtex-actions.el
bdarcus: PS - see this new comment, and linked commit, from @minad; might be relevant?

https://github.com/hlissner/doom-emacs/pull/4664#pullrequestreview-682442636

bdarcus:
> After looking at the info page for `with-eval-after-load`, I still think it's the right choice

I don't have time ATM to look into it, but what about `(when (featurep`?

I just remembered that's an approach used a lot in Doom.

publicimageltd:
> > After looking at the info page for `with-eval-after-load`, I still think it's the right choice
> 
> I don't have time ATM to look into it, but what about `(when (featurep`?

`featurep` can only check the presence of a package at runtime, that is in our case, when `bibtex-actions` is loaded. So what happens when `embark` is loaded afterwards? Nothing, unless we explicitly bind our code to the event "package is loaded", which is what `with-eval-after-load` does.

Maybe we should ask at reddit or stackoverflow for good ways to silence the compiler warning and best practices. But as I said above, the case of adding elements to a variable which has to be set by the package itself is exactly one of those cases, in my eyes, for which `with-eval-after-load` is made. It makes perfect sense to me.

bdarcus:
> Maybe we should ask at reddit or stackoverflow for good ways to silence the compiler warning and best practices.

This post seems helpful?

https://emacs.stackexchange.com/a/51463

publicimageltd:
> > Maybe we should ask at reddit or stackoverflow for good ways to silence the compiler warning and best practices.
> 
> This post seems helpful?
> 
> https://emacs.stackexchange.com/a/51463

The code at stackexchange wraps some definitions into an `eval` as to trick the byte compiler, because the called code is a macro. We don't use a macro, just some vars which can be announced correctly, so that's another use case. 

Here's a good solution (I didn't try it, but it looks good): https://github.com/tarao/bundle-el/blob/original/eval-after-load-compile.el

The idea seems to be to "load" or "require" the feature only when the code is compiling. This seems to be the core part:
```emacs-lisp

(let (loaded)
  (eval-when-compile
    (setq loaded
          (condition-case ()
              ;; don't eval after-loads during compilation
              (let ((after-load-alist nil))
                (require 'embark))
            (error nil))))
  (with-eval-after-load 'embark
    ....))
```

There's some byte compile vodoo going on in the snipped; I didn't quite get why the macro calls byte-compile in the case it is compiled. That might be a trick my snipped above doesn't catch. So maybe we just copy paste this macro, give it credit where due, and if it works, then it works, right?

I am currently procrastinating, I should do my work instead of this stuff. But it is an interesting problem. :-)

bdarcus:
That seems clever, but OTOH, it's simpler to just require embark, or remove those lines and add to the example config in the README.

Whether to require embark or not is not only a technical question, I'm thinking. It also signals priorities and such.

I'm not sure what I think ATM.

Maybe once this is ready for review (the other issues are addressed) we can ask for feedback on the whole thing. @ilupin could add an explicit "REVIEW" comment to this piece.

bdarcus:
@ilupin 

> You need the following code to make `C-c C-o` work on citations.
> 
> ```
>  (org-cite-register-processor 'bibtex-actions
>     :follow (lambda (_ _) (call-interactively 'bibtex-actions-at-point)))
> 
> (setq org-cite-follow-processor 'bibtex-actions)
> ```

So when org-cite is merged, we probably want to add the first piece to the package, and the second to the README?

publicimageltd:
Another solution would be to stuff all embark-related stuff into a function, and then offer a minor mode for embark. Initializiation could then look like this:

```emacs-lisp
(use-package bibtex-actions
 :config
 ;; for those who want to use it with embark:
 (use-package embark
   :config 
   (bibtex-actions-embark-minor-mode))) ;; or less verbose name
```

The `require` would then be in the minor mode definition:
```emacs-lisp
(define-minor-mode bibtex-actions-minor-mode
   :lighter ""
   :group 'bibtex-actions
   :global t
  (unless (require 'embark nil t)
     (user-error "Bibtex actions embark mode requires embark to be installed."))
  (if bibtex-actions-minor-mode
     ....setup...
  ....teardown...))
```

bdarcus:
> Another solution would be to stuff all embark-related initializations into a function, and then offer a minor mode for embark.

This is another detail I'm not super familiar with (from a developer POV), but seems worth considering.

Also occurs to me we might defer all this for a separate PR, if it holds up merging.

bdarcus:
@oantolin - on this:

> * the target finder can return a single string, which will be injected into the minibuffer prompt and "split" into a singleton list.

... I think we may have misunderstood you were saying here. Can you clarify?

Are you saying, for example, that the in-buffer at-point target-finder somehow gets run through, internally, completing-read? And since embark will convert the CRM string to a list, it should work here too, without additional code?

oantolin:
Oh, thanks for pinging me again, @bdarcus! It gives me a chance to correct myself. My earlier comment is _completing wrong_ because I confused two package names: bibtex-completion and bibtex-actions!

When I wrote that earlier comment I thought you wanted to use the functions from bibtex-actions (even though I mistakenly wrote bibtex-completions I was really thinking of bibtex-actions) as actions. Those functions are interactive commands, so when used as actions, Embark runs them interactively and inserts the target at the minibuffer prompt. So in that case there would be a completing-read call involved, the completing-read call present in the bibtex-actions commands.

But I was confused, you were really talking about bibtex-completion, and that's a different ball game because the functions there are not commands (i.e., the do not have an `(interactive ...)` specification). When given a function which is not a command as an action, what Embark does is simply to call the function (non-interactively) passing the target as an argument.

This is good news for this PR, I believe, because for non-interactive functions, Embark does not require the argument to be a string! I believe the functions in bibtex-completion take an argument which is a list of citation keys, right? The target finder can simply return a cons of the form `(cons 'citation-key list-of-keys)`. The cdr of the cons does not need to be a string, if you only plan to use the target with non-interactive functions.

bdarcus:
> But I was confused, you were really talking about bibtex-completion.

@oantolin - sorry, this is indeed confusing, and I guess we didn't help :-)

To recap:

- bibtex-actions interactive commands wrap bibtex-completion (non-interactive) functions
- @ilupin recently added a `bibtex-actions-at-point` command, which wraps `bibtex-completion-at-point`
- this PR, then, extends that to "embarify" at-point command access
- but for a reason I don't understand, @ilupin added new `bibtex-actions-embark-*` functions for these, I guess in part because of the string issue he mentioned

E.g. I personally think ideally we want to remove the `bibtex-actions-embark-*` functions, and instead call the bibtex-actions ones directly, passing them the keys extracted with the target finder.

I think we just weren't understand how best to do that. See, for example, https://github.com/bdarcus/bibtex-actions/pull/143#discussion_r650539822.

Does that make better sense now?

bibtex-actions.el
ilupin: I agree that `bibtex-actions-embark-*` functions should be removed finally.

> ... and then something like:

It doesn't seem to work as expected. The tricky part is the entries are always chosen via `completing-read` when `bibtex-actions-add-pdf-to-library` is interactively called. `embark` also interactively calls `bibtex-actions-add-pdf-to-library`, which means the `keys` found at point are not used by `bibtex-actions-add-pdf-to-library`. I need to find a proper way to handle this.

bibtex-actions.el
ilupin: I'm fine with adding a defcustom to make `embark-dwim` the default. It will be consistent with the case where `embark` is not installed.

bibtex-actions.el
bdarcus: I hadn't thought about the "add" commands. Could always defer those to a later PR.

bibtex-actions.el
ilupin: Yes. `bibtex-actions-get-key-org-cite` just returns nil if `org-cite` is not available.

bibtex-actions.el
ilupin: The problem is the same for `bibtex-actions-open-*` actually.

ilupin:
> While I can probably help you, could you please include README additions and modifications for all this, when the code is settled? Particularly the config details.

Of course. I will update README when there is no problem with the code.

> So when org-cite is merged to org master, we probably want to add the first piece to the package, and the second to the README?

Yes.

ilupin:
> I personally think ideally we want to remove the `bibtex-actions-embark-*` functions, and instead call the bibtex-actions ones directly, passing them the keys extracted with the target finder.
> 
> Edit: in other words, we want ideally to be able to use the same bibtex-actions interactive functions both with CRM, and at-point.

`bibtex-actions-embark-*` functions are removed in https://github.com/bdarcus/bibtex-actions/pull/143/commits/bb07e66ae8ced8bb0b362b97a9e2f99044b946c2, and `bibtex-completion-*` rather than `bibtex-actions-*` are used as embark actions. It is because the TARGET, i.e. citation keys, found in buffer cannot be passed to interactive commands such as `bibtex-actions-*`, as explained in https://github.com/bdarcus/bibtex-actions/pull/143#issuecomment-860988893.

bdarcus:
> > I personally think ideally we want to remove the `bibtex-actions-embark-*` functions, and instead call the bibtex-actions ones directly, passing them the keys extracted with the target finder.
> > Edit: in other words, we want ideally to be able to use the same bibtex-actions interactive functions both with CRM, and at-point.
> 
> `bibtex-actions-embark-*` functions are removed in [bb07e66](https://github.com/bdarcus/bibtex-actions/commit/bb07e66ae8ced8bb0b362b97a9e2f99044b946c2), and `bibtex-completion-*` rather than `bibtex-actions-*` are used as embark actions. It is because the TARGET, i.e. citation keys, found in buffer cannot be passed to interactive commands such as `bibtex-actions-*`, as explained in [#143 (comment)](https://github.com/bdarcus/bibtex-actions/pull/143#issuecomment-860988893).

Thanks @ilupin!

So we actually can't use the same interactive functions in both places.

Hmm ... somehow this just feels conceptually wrong. A user will see one list of actions when doing `embark-act` from the minibuffer, and another when running the same at point, but both sets of actions _do the exact same thing_.

Just want to confirm @oantolin: that's the upshot, and a necessary part of the embark design?

If so, I guess this goes back to figuring out how to get which-key to display labels instead of function names. That way the menus could at least be consistent.

But I think that's waiting in an unmerged which-key PR.

https://github.com/justbur/emacs-which-key/pull/282

bdarcus:
> If so, I guess this goes back to menus: figuring out how to get which-key to display labels instead of function names. That way the menus could at least be consistent.

We could add config info (probably on the README or the wiki), that makes use of this, to replace the command names in the menus with their descriptions:

https://github.com/justbur/emacs-which-key#custom-string-replacement-options

So rather than seeing "bibtex-completion-open-any" a user would see, for example, "Open (PDF or link)".

ilupin:
> > If so, I guess this goes back to menus: figuring out how to get which-key to display labels instead of function names. That way the menus could at least be consistent.... But I think that's waiting on an unmerged which-key PR.
> 
> For short-to-medium run, we could add config info (probably on the README or the wiki), that makes use of this, to replace the command names in the menus with their descriptions:

With the workaround in https://github.com/bdarcus/bibtex-actions/pull/143/commits/d11dc1dfa48b36cd4f4dffb18cf078b5d18a40ed, interactive commands `bibtex-actions-*` can be used in the `bibtex-actions-buffer-map`. Now `bibtex-actions` shows same embark action names in minibuffer and regular buffers.

bibtex-actions.el
```diff
@@ -417,10 +438,35 @@ TEMPLATE."
...
+
+;;; Embark
+
+(defun bibtex-actions-citation-key-at-point ()
+  "Return citation keys at point as a list for `embark'."
+  (when-let ((key (or (bibtex-actions-get-key-org-cite)
+                      (bibtex-completion-key-at-point))))
+    (cons 'citation-key (if (listp key) key (list key)))))
+
+(defvar bibtex-actions-embark-map
```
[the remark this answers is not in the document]
ilupin: Renamed in https://github.com/bdarcus/bibtex-actions/pull/143/commits/c232b73839cc88c9c471de5d74d2feb610b13dac.


> We could also rename the main map `bibtex-actions-minibuffer-map`?

I have no opinions.

bdarcus:
> The workaround, i.e. overriding `embark--act`, is only effective in `bibtex-actions-at-point` by design and doesn't affect any other code. Therefore, users cannot use `bibtex-actions-buffer-map` normally outside `bibtex-actions-at-point`, and the keymap should not be added to `embark-keymap-alist`.

Do you want to add a warning to that effect in the docstring?

bdarcus:
If you have time to test which-key support for simple menus, a new PR with request for just that.

https://github.com/justbur/emacs-which-key/pull/308

This should allow us to use #57 and not get "undefined" menu labels.

README.org
```diff
@@ -280,6 +299,17 @@ From there, you can run the same options discussed above using =embark-act= (whi
...
 From there, you can run different actions on the candidates at will, rather than search individually for each item you want to cite.
 
+*** Use =bibtex-actions-at-point=
+    :PROPERTIES:
+    :CUSTOM_ID: use-bibtex-actions-at-point
+    :END:
+
+=M-x bibtex-actions-at-point= will run the default action on citation keys found at point directly.
+
+By setting =bibtex-actions-embark-dwim= to nil, other actions in =bibtex-actions-buffer-map= will be prompted if =embark= is installed. Press =RET= will run the default action.
```
bdarcus, line 309:
```suggestion
If =embark= is installed, setting =bibtex-actions-embark-dwim= to nil will prompt for other actions in =bibtex-actions-buffer-map= . From there, pressing =RET= will run the default action.
```
The latter change I suggest is because I was confused by `RET` in this paragraph related to the same in the previous paragraph.

bdarcus:
@ilupin - could you update the first post with the commit message, so I can squash merge it, and use that?

I'm too tired ATM to edit it clearly!

bdarcus:
This looks RTM. 

Last call for feedback. I'll plan to merge later today (I'm in US West Coast ATM) or tomorrow.

bdarcus:
Thanks again @ilupin; this is a nice addition!

bdarcus:
@ilupin - back from travel, and so have more time to play with the new embark at-point functionality.

Are we missing something here to configure, or should it "just work" if I have embark configured?

https://github.com/bdarcus/bibtex-actions#basic

It currently doesn't appear to be working for me.

ilupin:
It doesn't need additional configuration.

> `embark-act` shows no `bibtex-actions` commands.

It is expected if you call `embark-act`, rather than `bibtex-actions-at-point`, on citation keys in buffer. `bibtex-actions-buffer-map` and `bibtex-actions-citation-key-at-point` is only used inside `bibtex-actions-at-point`.


> Edit: if I'm in a latex buffer, and run `bibtex-actions-at-point`, if it doesn't find the item, it now opens a warnings buffer. I thought it should instead open the the minibuffer?

Yes. What is the warning message?

bdarcus:
> It doesn't need additional configuration.

What I thought. Hmm ...
 
> > `embark-act` shows no `bibtex-actions` commands.
> 
> It is expected if you call `embark-act`, rather than `bibtex-actions-at-point`, on citation keys in buffer. 
> `bibtex-actions-buffer-map` and `bibtex-actions-citation-key-at-point` is only used inside `bibtex-actions-at-point`.

So how do you get the `embark-act` menu at-point?
 
> > Edit: if I'm in a latex buffer, and run `bibtex-actions-at-point`, if it doesn't find the item, it now opens a warnings buffer. I thought it should instead open the the minibuffer?
> 
> Yes. What is the warning message?

That the item isn't found; essentially duplicating the info in the modeline.

ilupin:
> So how do you get the `embark-act` menu at-point?

`(setq bibtex-actions-embark-dwim nil)`. It is described in README.

> Edit: and for now (until org-cite), is latex the only context where this will work?

No. There is a `bibtex-completion-get-key-org-bibtex` in `bibtex-completion-key-at-point-functions` for org-mode. It also works in `bibtex-mode`.

> That the item key isn't found. In this case, the key was wrong; if I correct it, `bibtex-actions-open` correctly runs.

I see. In this case, `bibtex-actions-citation-key-at-point` does find a key, but it has no means to check whether the key is valid or not. I think it is reasonable for `bibtex-actions-at-point` to pass the key to an action, resulting in a warning if the action fails though, rather than fallback to `completing-read`.

bdarcus:
> `(setq bibtex-actions-embark-dwim nil)`. It is described in README.

Is there a way we can have both though?

If the basic metaphor of `embark-act`, as described by Daniel Mendler, is right-click contextual menu, and `embark-dwim` is left-click, seems ideally both should be accessible at-point?

Like, hit `RET` for `embark-dwim` and some binding for `embark-act`?

That's what I was assuming with the question at least.

> I think it is reasonable for bibtex-actions-at-point to pass the key to an action, resulting in a warning if the action fails though, rather than fallback to completing-read.

OK.

ilupin:
> If the basic metaphor of `embark-dwim`, as described by Daniel Mendler, is left-click, and `embark-act` is right-click contextual menu, seems ideally both should be accessible at-point?
> 
> Like, hit `RET` (well, maybe not this, but ...) for `embark-dwim` and some binding for `embark-act`?

It is what `bibtex-actions-at-point` does when `bibtex-actions-embark-dwim` is nil.

In fact, when `bibtex-actions-embark-dwim` is t, call `bibtex-actions-at-point` == call `embark-dwim` == call the action binded to `RET` in `bibtex-actions-buffer-map`.

If `bibtex-actions-embark-dwim` is nil, call `bibtex-actions-at-point` == call `embark-act`. In this case, `RET` runs the action `embark-dwim` would call.

bdarcus:
Now that I committed the follow processor (I've opened a Doom PR to add OC support, so helpful to have that merged), back to this @ilupin:

So my big question: is there a way for us to have both `bibtex-actions-embark-dwim` t, AND be able to call `embark-act` from point and get our map options?

That would be ideal.

If yes, what's involved in adding that capability?

PR #153, opened 2021-06-26
Add start of org-cite constraints

bdarcus:
This doesn't currently work, but I need to add constraints so this capf doesn't conflict with others.

Also, there's a possibility completion support will be added to `org-cite`, which would simplify this.

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -52,6 +52,7 @@
 (defvar embark-target-finders)
 (defvar embark-general-map)
 (defvar embark-meta-map)
+(defvar org-element-citation-prefix-re)
 
 ;;; Variables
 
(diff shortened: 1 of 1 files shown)


bibtex-actions.el
```diff
@@ -355,8 +356,7 @@ will then return the key 'key', resulting in '@key' at point.
 Supports the pandoc and 'org-cite' key syntax, in either
 'org-mode' or 'markdown-mode'."
     ; FIX
-    (when (and (or (eq major-mode 'org-mode)
-                   (eq major-mode 'markdown-mode))
+    (when (and (looking-back org-element-citation-prefix-re (line-beginning-position))
```
bdarcus, line 359: With this line, or the two is replaces, this does not work.

With all removed, and only the below, it does work.

@ilupin since you've been able to contribute org-cite code, am I correct you've managed to get the branch working locally?

Can you see if you can figure this out please? I am thinking the problem is either a) my org-cite install, or b) something fundamental about the elisp here (it doesn't have to do with the capf-specific details).

Issue #135, opened 2021-05-24
Improve CRM UI


bdarcus:
As noted in the README, support for multiple candidate selection isn't great.

I do think the ideal long-term solution is for completion UIs to implement something similar to the helm approach to this.

But in the absence of that, I thought I'd point out this suggestion, which could be a reasonable shorter-term improvement.

https://github.com/minad/vertico/issues/1#issuecomment-817378284

The basic idea is to have a "short" representation of the candidate string that would be used in the CRM prompt.

Could also be useful for #91.

PR #155, opened 2021-06-30
Enhance CRM selection UI

bdarcus:
Addresses #135.

Basic idea is to hack the minibuffer to present shortened candidate strings in the CRM prompt, and so address the primary UI problem.

Issues:

1. how well does it actually work? there are currently some UI oddities (see below; after inserting multiple in vertico, there's no space or separator between them, and I can't "remove" them from the prompt). Seems one improvement would be adding a face to the selected candidates, so in this case they would have a light gray background, and some spacing between them.
2. it does work in Vertico; but Selectrum is problematic ATM (the candidates don't show up in the prompt, and cursor drops to the bottom of the screen)
3. but Vertico has other CRM UI issues (see below).

Screenshot from 2021-06-30 10-36-37

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -247,20 +247,38 @@ offering the selection candidates"
 	 (candidates (bibtex-actions--get-candidates rebuild-cache))
          (initial-input (assoc-default initial bibtex-actions-initial-inputs))
          (chosen
-          (completing-read-multiple
-           "BibTeX entries: "
-           (lambda (string predicate action)
-             (if (eq action 'metadata)
-                 `(metadata
-                   (affixation-function . bibtex-actions--affixation)
-                   (category . bibtex))
-               (complete-with-action action candidates string predicate)))
-           nil
-           nil
-           (and initial-input
-                (stringp initial-input)
-                (concat initial-input " "))
-           'bibtex-actions-history bibtex-actions-presets nil)))
+          (minibuffer-with-setup-hook
+              ;; configures the minibuffer to use a short string (the key) in the prompt
+              ;; FIX no gap between entries
+              (lambda ()
+                (add-hook 'after-change-functions
+                          (lambda (&rest _)
+                            (save-excursion
+                              (save-match-data
+                                (let ((pos (minibuffer-prompt-end)))
+                                  (goto-char pos)
+                                  (while (< pos (point-max))
+                                    (when (re-search-forward (concat crm-separator "\\|\\'") nil 'noerror)
+                                      (when-let (short (cdr (assoc (string-trim (buffer-substring-no-properties
+                                                                                 pos (match-beginning 0)))
+                                                                   candidates)))
+                                        (put-text-property pos (match-beginning 0) 'display short))
+                                      (setq pos (point))))))))
+                          nil 'local))
+            (completing-read-multiple
+             "Reference entries: "
+             (lambda (string predicate action)
+               (if (eq action 'metadata)
+                   `(metadata
+                     (affixation-function . bibtex-actions--affixation)
+                     (category . bibtex))
+                 (complete-with-action action candidates string predicate)))
+             nil
+             nil
+             (and initial-input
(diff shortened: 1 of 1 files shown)


bdarcus:
@minad - OK, got the basics of the idea working, but there are a few issues I need to resolve.

Do you expect this should work with Selectrum too? See my point 2 above.

I was assuming yes, given there's nothing specific to vertico.

bdarcus:
> Oh, I had hoped this works in Selectrum too. No idea why it would not

In general, there are just some UI eccentricities with this. 

It does "work", in the sense that I can select multiple in both vertico and selectrum, and the command will produce the correct output.

It's just with selectrum, it's like flying blind; I don't see the selected candidates, and the cursor drops to the bottom of the screen. Which would drive people crazy.

I guess we could call that a bug?

minad:
No idea, I would have to give it a test to see where the problem lies. I guess the problem is that Selectrum replaces the crm completion table entirely and does not use it. It uses its own dynamic completion function facility instead, since it does not implement dynamic completion table correctly.

bdarcus:
If you do end up testing it, note that I added a test `run.sh` script to the repo, which is derived from your selectrum test scripts.

bdarcus:
There is something that's a bit weird with that screenshot I added, @minad.

Why that syntax-highlighted line? Or rather, why is it colored like that?

Selectrum does the same thing, but I never see it, because it automatically adds the `crm-separator`. So I only see it if I remove the separator.

minad:
Yes, I've wondered about that too. Is this orderless highlighting?

bdarcus:
Yes.

minad:
Behind the `hunt-76mann-68` is the full bibtex string. `hunt-76mann-68` is only the display string. So orderless will still filter the candidates against the current input string as long as you haven't entered another separator.

bdarcus:
Ok, but note (and this is another issue specific to this PR):

That string in fact should represent two candidates. As in, I selected two
there.

So it must be matching that second one.

minad:
Okay, I don't understand this. Is `hunt-76mann-68` input or abbreviated candidate thanks to your crm enhancement. You may want to give the display string a different face to make this clear.

bdarcus:
I think I've been spoiled by CSS, as I look into emacs faces (no margin or padding attributes?). 

For now, I've just added a face with a box around it, so it's obvious.

Another issue: after adding candidates, I can't edit them. Is that expected? Is there any workaround?

bdarcus:
I went back to the minimal example: https://github.com/minad/vertico/issues/1#issuecomment-817378284, as I was trying to figure out a bug report.

That does work as expected in selectrum, so there's something about my code here that triggers it. I guess the main difference is I rely on multiple faces, etc.

minad:
> Another issue: after adding candidates, I can't edit them. Is that expected? Is there any workaround?

it should still be possible to delete them, but you may have to press backspace many many times. So you may want to add another special backspace command there too.

Given all these complications I am really not sure what the best approach is for this. I think it may be better to indeed write your own dynamic completion table, instead of this display/after-change-function trickery, which will not be robust. This is how I would approach this probably. I think I've mentioned this before in some other issue/discussion? However this will require the user to use a completion UI which is fully compliant (default completion, Vertico or Icomplete).

bdarcus:
Yes, you mentioned that, and I think you are right.

I'll close this then. It was worth a try.

Thank you!

bdarcus:
@minad - one question on custom dynamic completion tables:

Would  that give us the capability to enable the sort of ideas discussed on the selectrum issue tracker?

For example, even dispensing with the prompt entirely, and just using direct selection of multiple candidates a la helm?

minad:
No, this would still not be possible. But it may be possible to implement the short candidate idea from here in a cleaner way.

bdarcus:
And might that "cleaner way" be putting the key itself (the cdr of the
candidate; the thing ultimately returned) in the prompt after selection,
rather than doing the shortening, etc.?

Issue #160, opened 2021-07-04
Alternative crm idea


minad:
I thought maybe it is better to give up on crm and use something else based directly on `completing-read`. Try this:

~~~ elisp
(defun simple-crm (prompt candidates)
  (let ((selected-list)
        (selected-hash (make-hash-table :test #'equal))
        (items candidates))
    (while
        (let ((item
               (completing-read
                (format "%s (%s/%s): " prompt
                        (hash-table-count selected-hash)
                        (length candidates))
                (lambda (str pred action)
                  (if (eq action 'metadata)
                      `(metadata (display-sort-function . ,#'identity)
                                 (cycle-sort-function . ,#'identity)
                                 (group-function
                                  . ,(lambda (cand transform)
                                       (cond
                                        (transform cand)
                                        ((get-text-property 0 'simple-crm-selected cand) "Selected")
                                        (t "Not selected")))))
                    (complete-with-action action items str pred)))
                nil
                t
                nil
                nil
                "")))
          (unless (equal item "")
            (cond
             ((gethash item selected-hash)
              (remhash item selected-hash)
              (setq selected-list (delete item selected-list)))
             (t
              (puthash item t selected-hash)
              (setq selected-list (nconc selected-list
                                         (list (propertize item 'face 'region
                                                           'simple-crm-selected t))))))
            (setq items
                  (append selected-list
                          (seq-remove (lambda (x)
                                        (gethash x selected-hash))
                                      candidates)))
            t)))
    selected-list))

(simple-crm "Select multiple" '("first" "second" "third" "fourth"))
~~~

Fortunately there is also no need to implement a custom completion table. I wonder if such a functionality deserves to be provided by some package.

minad:
Instead of writing a separate `simple-crm` function here, one could also replicate the `completing-read-multiple` API and replace this using an advice. The question is then where to package this crm replacement. It could go to Consult since it is completion UI agnostic. What do you think?

bdarcus:
Yes; this is basically @clemera's original idea!!

Is there a way, or could you extend it, to then deselect?

To be clear, for my case, I don't imagine deselection would be a common need, but say I make a mistake, and need to remove a candidate.

A similar case could be the cc field for an email.

> It could go to Consult since it is completion UI agnostic. What do you think?

+1.

With this approach, then I could also rely on `consult--read`, and so get other benefits, like narrowing?

minad:
Deselect already works if you select the item again.

[...]

PR #163, opened 2021-07-05
New multi-selection UI

bdarcus:
Screenshot from 2021-07-06 14-32-22

This implements the ideas in #160, as a means to fix various problems with CRM for this package.

The key UI thing to realize is you don't really want to use `TAB` here; just `RET`.

Also, deselect works the same as select.

I added a defcustom for this, but I have set it to `t` here, as something like this UI is always what I had in mind, so if this doesn't work well, I may need to explore other alternatives (shorter candidate strings, avoiding CRM in general, focusing on Embark more for acting on multiple candidates, etc.).

I've also added some minimal documentation.

As always, you can test by running the test/run.sh script.

Note: the screenshot is with the `vertico-crm` prototype version of this.


bdarcus:
@minad - the big, and only significant, issue ATM is Embark doesn't work correctly.

The menu is accessible as expected, but whether one or multiple candidates, I get no results.

Is this because simple-crm is using CR, while I have existing code like this?

``` elisp
(interactive (list (bibtex-actions-read :rebuild-cache current-prefix-arg)))
```

andersjohansson:
Quickly tried it out and really like it! Hope the issues are solvable.

minad:
We could solve it by checking manually for the `embark--target` (if fboundp) in simple-crm/consult-completing-read-multiple. This is not ideal but at least ensures that Embark injection works. @oantolin Do you have a better idea?

minad:
@bdarcus Given that everyone seems to like the UI I may reconsider adding consult-completing-read-multiple to consult in the generic form with embark support. It has these few issues (flicker, double-RET, potential sorting problems) as we discussed but the cost/benefit ratio is not bad, in particular since it is compatible with different UIs.

oantolin:
> Do you have a better idea?

I don't know if this is better, but it's at least different: We could arrange for a target finder that if simple-crm is running returns a target that is a list of selected items, _not_ a string. And we could modify Embark so that if the target is a non-empty list, instead of injection the command gets called with the list as argument, whether or not the action is a command.

minad:
Well but this would require the command to take a list as argument. Furthermore I am not happy with the idea of list targets. When acting on the candidates in CRM you still act on single candidates (single strings). I would not overcomplicate things here - it is sufficient if we ensure that the single candidate injection works.

oantolin:
Well, without a special target finder for simple-crm, running embark-act from it will just pick up the single candidate for the current run of completing-read, no? So we still need someway to extract the selected candidates from simple-crm and it might as well be as a list (I guess the other option is as a single string with separators).

minad:
Okay, that's right. We would not pick up the selected items. If we would want to support this we would need a special target finder. But then I would prefer if the embark action is called once for each item in the list.

I only oppose adding another Embark calling convention for handling a list. Adding a special CRM target finder returning a list is okay.

minad:
Current `consult-completing-read-multiple` version: https://github.com/minad/consult/blob/43664f0ccafe2aa152a9a7c79f20844fdf43731e/consult.el#L2154-L2251

Does not support Embark yet, still have to figure out how to do this. Besides that, I think it works good enough.

bdarcus:
@mohkale @apc - can you give this a test and give us feedback?

@minad - is it also feasible to use the consult version as I do here, in addition to the advice approach?

minad:
When the consult version gets merged, my recommendation is to not add anything special here. Just use `completing-read-multiple` and be done :)

Take a look at the current version in https://github.com/minad/consult/pull/352. Main problem is Embark. This is certainly fixable but may require some Embark detection (gluing code).

bdarcus:
> When the consult version gets merged, my recommendation is to not add anything special here. Just use `completing-read-multiple` and be done :)

I guess one little issue remains: the filtering for pdf, etc.

Still the case that won't work? If yes, do I want to add a predicate for the pre-filtering?

minad:
> I guess one little issue remains: the filtering for pdf, etc.
> Still the case that (initial-input) won't work? If yes, do I want to add a predicate for the pre-filtering?

Initial input is completely treated as candidates. It is not treated as a filter string. Maybe that is not the best behavior. I could tweak it as follows:

1. Split the `initial-input` by `crm-separator`
2. Use the `butlast` items as selected candidates
3. Use the `last` item either as selected candidate or as filter string, depending on if the string is part of the candidate list

However the filter string should probably still be used only once, since this is more aligned with how `completing-read-multiple` by default works. Your `has:pdf` filter also does not work correctly with the default crm implementation, since it only applies to the first candidate. After pressing TAB the `has:pdf` filter is lost. Therefore I proposed to use a predicate instead or even filter the list of candidates before passing them to `completing-read-multiple`.

If you add your own implementation here, you can of course special case this somehow, but I think it would be better if you use a common CRM implementation, since this potentially leads to a more consistent UI.

Another option we could discuss is the following:

1. `consult-completing-read-multiple` treats initial-input completely as candidates as it does now
2. Add a variable `consult-crm-initial-input`, which you can let bind dynamically, such that it is always used as a filter.

bdarcus:
I'll defer to you on what's the best general approach for `consult-read-multiple`. 

But on this:

> Therefore I proposed to use a predicate instead or even filter the list of candidates before passing them to completing-read-multiple.

Is there some downside to using a predicate here?

I have a cache of all the candidates, where some commands need to get different subsets.

minad:
> Is there some downside to using a predicate here?
> I have a cache of all the candidates, where some commands need to get different subsets.

The predicate will be a bit more expensive since it filters the candidates every time. But the cost will probably be small. The downside of predicate/filter vs your current "has:pdf" approach is that the user cannot remove the filter. In contrast with "has:pdf" the user can delete this filter string in order to see all candidates. Is that important/desired?

bdarcus:
> The downside of predicate/filter vs your current "has:pdf" approach is that the user cannot remove the filter. In contrast with "has:pdf" the user can delete this filter string in order to see all candidates. Is that important/desired?

Yes, it is. That's why we used `initial-input` (and why @mohkale was experimenting with Consult narrowing).

I just don't know how important that flexibility is; it's possible not very.

If I don't need to make that choice now, I would probably leave things as is.

But the intial-input would need to consistently apply to all candidate selections for it to work as we expected. As you know, it doesn't in CRM.

bdarcus:
For those of you that might test this UI, my suggestion is you test @minad's branch instead, and provide feedback on that PR.

https://github.com/minad/consult/pull/352

Once that's merged, I'll probably drop the commit for the elisp file, and just make this a README change, with a recommendation to use `consult-completing-read-multiple`. Oh, and maybe update screenshots :-)

bdarcus:
So I dropped the initial commit here, and replaced with a test script modified to use `consult-completing-read-multiple`.

It doesn't currently work locally for me, but I think the problem is something other than what I committed.

Next I'll updated the docs.

PR #169, opened 2021-07-08
Allow bibtex-actions-read to return single candidate

### LINKED ISSUE #161: open-entry should use completing-read
Unless I'm missing something, this command only opens a single entry, so maybe should just use completing-read.

bdarcus:
This would add a `completing-read` mode for `bibtex-actions-read`, for commands that should only return a single candidate.

Fix #161.


bdarcus:
I experimented with this in #222, but decided to leave things as is for now (other than changing the name).

PR #171, opened 2021-07-08
fix embark at-point

### LINKED ISSUE #164: Make both embark-act and embark-dwim available at point
Now that I committed the follow processor (I've opened a Doom PR to add OC support, so helpful to have that merged), back to this @ilupin:

Currently, one has to chose for `bibtex-actions-at-point` to run either the default action, or to invoke `embark-act`.

Is there a way for us to have both `bibtex-actions-embark-dwim` t (for example using the OC follow processor where it's available via `org-open-at-point`), AND be able to call `embark-act` from point and get our map options?

That would be ideal. `embark-act` and `embark-dwim` are designed to be used together, after all.

It's also a bit confusing to me, and I suspect will be to other users, that when on an org-citation, calling `embark-act` gives a menu without contextual actions.

If yes, what's involved in adding that capability?

I don't fully understand this code, so ...

_Originally posted by @bdarcus in https://github.com/bdarcus/bibtex-actions/issues/143#issuecomment-872494249_

cc @publicimageltd

bdarcus:
This fixes the initial embark at-point integration.

With this PR, default behavior when point is on an `org-cite` citation or citation-reference:

- `M-x org-open-at-point` -> `embark-dwim` (the default action)
- `embark-act` presents bibtex-actions command options

Both `embark-act` and `embark-dwim` should also correctly work in `latex-mode`.

Please test and confirm this is all working as expected, including the minibuffer commands and `embark-act` behavior.

Also, I've updated the test script to show the `org-cite` capabilities.

Fix #164.

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -470,7 +470,7 @@ TEMPLATE."
 
 ;; Org-cite "follow" and "insert" processor
 
-(defun bibtex-actions-org-cite-insert (multiple)
+(defun bibtex-actions-org-cite-insert (&optional multiple)
   "Return a list of keys when MULTIPLE, or else a key string."
   (let ((references (bibtex-actions-read)))
     (if multiple
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -607,23 +607,25 @@ interactively when `bibtex-actions-at-point-fallback' is non-nil.
 With prefix ARG, rebuild the cache before offering candidates."
   (interactive "P")
   (if (fboundp 'embark-dwim)
-      (let ((embark-keymap-alist '((bibtex . bibtex-actions-map)
-                                   (citation-key . bibtex-actions-buffer-map))))
-        (condition-case err
-            (let ((embark-target-finders '(bibtex-actions-citation-key-at-point)))
-              (if bibtex-actions-embark-dwim
-                    (embark-dwim)
-                  (embark-act)))
-          (user-error
-           (when (and (string-equal (error-message-string err) "No target found")
-                      bibtex-actions-at-point-fallback)
-             (bibtex-actions-run-default-action
-              (bibtex-actions-read :rebuild-cache arg))))))
+      (condition-case err
+          (if bibtex-actions-embark-dwim
+              (embark-dwim)
+            (embark-act))
+        (user-error
+         (when (and (string-equal (error-message-string err) "No target found")
+                    bibtex-actions-at-point-fallback)
+           (bibtex-actions-run-default-action
+            (bibtex-actions-read :rebuild-cache arg)))))
     (if-let ((keys (bibtex-actions-citation-key-at-point)))
         (funcall bibtex-actions-default-action keys)
       (when bibtex-actions-at-point-fallback
         (bibtex-actions-run-default-action
          (bibtex-actions-read :rebuild-cache arg))))))
+
+(with-eval-after-load "embark"
+  (add-to-list 'embark-target-finders 'bibtex-actions-citation-key-at-point)
+  (add-to-list 'embark-keymap-alist '((bibtex . bibtex-actions-map)
+                                      (citation-key . bibtex-actions-buffer-map))))
 
 (provide 'bibtex-actions)
 ;;; bibtex-actions.el ends here
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -479,9 +479,9 @@ TEMPLATE."
 
 (when (require 'oc nil t)
   (org-cite-register-processor 'bibtex-actions
-    :insert (org-cite-make-insert-processor
-             (bibtex-actions-org-cite-insert t)
-             #'org-cite-basic--complete-style)
+    ;:insert (org-cite-make-insert-processor
+    ;         #'bibtex-actions-org-cite-insert
+    ;         #'org-cite-basic--complete-style)
     :follow (lambda (_datum _arg) (call-interactively 'bibtex-actions-at-point))))
 
 ;;; Embark
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -183,18 +183,18 @@ If nil, prompt the user for an action through `embark-act'."
 
 (defvar bibtex-actions-map
   (let ((map (make-sparse-keymap)))
-    (define-key map "t" '("add pdf attachment" . bibtex-actions-add-pdf-attachment))
-    (define-key map "a" '("add pdf to library" . bibtex-actions-add-pdf-to-library))
-    (define-key map "b" '("insert bibtex" . bibtex-actions-insert-bibtex))
-    (define-key map "c" '("insert citation" . bibtex-actions-insert-citation))
-    (define-key map "k" '("insert key" . bibtex-actions-insert-key))
-    (define-key map "f" '("insert reference" . bibtex-actions-insert-reference))
-    (define-key map "o" '("open source" . bibtex-actions-open))
-    (define-key map "e" '("open reference entry" . bibtex-actions-open-entry))
-    (define-key map "l" '("open source link" . bibtex-actions-open-link))
-    (define-key map "n" '("open notes" . bibtex-actions-open-notes))
-    (define-key map "p" '("open source pdf" . bibtex-actions-open-pdf))
-    (define-key map "r" '("refresh references" . bibtex-actions-refresh))
+    (define-key map (kbd "t") 'bibtex-actions-add-pdf-attachment)
+    (define-key map (kbd "a") 'bibtex-actions-add-pdf-to-library)
+    (define-key map (kbd "b") 'bibtex-actions-insert-bibtex)
+    (define-key map (kbd "c") 'bibtex-actions-insert-citation)
+    (define-key map (kbd "k") 'bibtex-actions-insert-key)
+    (define-key map (kbd "f") 'bibtex-actions-insert-reference)
+    (define-key map (kbd "o") 'bibtex-actions-open)
+    (define-key map (kbd "e") 'bibtex-actions-open-entry)
+    (define-key map (kbd "l") 'bibtex-actions-open-link)
+    (define-key map (kbd "n") 'bibtex-actions-open-notes)
+    (define-key map (kbd "p") 'bibtex-actions-open-pdf)
+    (define-key map (kbd "r") 'bibtex-actions-refresh)
     map)
   "Keymap for 'bibtex-actions'.")
 
@@ -265,7 +265,8 @@ offering the selection candidates"
            'bibtex-actions-history bibtex-actions-presets nil)))
     (cl-loop for choice in chosen
              ;; Collect citation keys of selected candidate(s).
-             collect (cdr (assoc choice candidates)))))
+             collect (cdr (or (assoc choice candidates)
+                              (rassoc choice candidates))))))
 
 (defun bibtex-actions--format-candidates ()
   "Transform candidates from 'bibtex-completion-candidates'.
@@ -485,26 +486,22 @@ TEMPLATE."
 
 ;;; Embark
 
-(cl-defun bibtex-actions-citation-key-at-point (&optional &key as-list)
-  "Return citation keys at point AS-LIST for `embark', or string."
-  (when-let* ((key (or (bibtex-actions-get-key-org-cite)
-                      (bibtex-completion-key-at-point)))
-              (crm-separator "\\s-*&\\s-*")
-              (key-list  (if (listp key) key (list key)))
-              ;; Currently, Embark only accepts a string. So we use a crm-like string to
-              ;; represent the list.
-              (return-value (if as-list key-list (string-join key-list "&"))))
-    (cons 'citation-key return-value)))
+(defun bibtex-actions-citation-key-at-point ()
+  "Return citation keys at point as a list for `embark'."
+  (when-let ((key (or (bibtex-actions-get-key-org-cite)
+                      (bibtex-completion-key-at-point))))
+    (cons 'citation-key (if (listp key) (string-join key " & ") key))))
 
 (defvar bibtex-actions-buffer-map
   (let ((map (make-sparse-keymap)))
-    (define-key map "t" '("add pdf attachment" . bibtex-actions-add-pdf-attachment))
-    (define-key map "a" '("add pdf to library" . bibtex-actions-add-pdf-to-library))
-    (define-key map "o" '("open source" . bibtex-actions-open))
-    (define-key map "e" '("open reference entry" . bibtex-actions-open-entry))
-    (define-key map "l" '("open source link" . bibtex-actions-open-link))
-    (define-key map "n" '("open notes" . bibtex-actions-open-notes))
-    (define-key map "p" '("open source pdf" . bibtex-actions-open-pdf))
+    (define-key map (kbd "t") 'bibtex-actions-add-pdf-attachment)
+    (define-key map (kbd "a") 'bibtex-actions-add-pdf-to-library)
+    (define-key map (kbd "o") 'bibtex-actions-open)
+    (define-key map (kbd "e") 'bibtex-actions-open-entry)
+    (define-key map (kbd "l") 'bibtex-actions-open-link)
+    (define-key map (kbd "n") 'bibtex-actions-open-notes)
+    (define-key map (kbd "p") 'bibtex-actions-open-pdf)
+    (define-key map (kbd "RET") 'bibtex-actions-run-default-action)
     map)
   "Keymap for Embark citation-key actions.")
 
@@ -602,10 +599,31 @@ With prefix, rebuild the cache before offering candidates."
   "Run the default action `bibtex-actions-default-action' on KEYS."
   (funcall bibtex-actions-default-action keys))
 
-(with-eval-after-load "embark"
-  (add-to-list 'embark-target-finders 'bibtex-actions-citation-key-at-point)
-  (add-to-list 'embark-keymap-alist '(bibtex . bibtex-actions-map))
-  (add-to-list 'embark-keymap-alist '(citation-key . bibtex-actions-buffer-map)))
+;;;###autoload
+(defun bibtex-actions-at-point (&optional arg)
+  "Run the default action on citation keys found at point.
+If no citation key is found, target entries can be chosen
+interactively when `bibtex-actions-at-point-fallback' is non-nil.
+With prefix ARG, rebuild the cache before offering candidates."
+  (interactive "P")
+  (if (fboundp 'embark-dwim)
+      (let ((embark-keymap-alist '((bibtex . bibtex-actions-map)
+                                   (citation-key . bibtex-actions-buffer-map))))
+        (condition-case err
+            (let ((embark-target-finders '(bibtex-actions-citation-key-at-point)))
+              (if bibtex-actions-embark-dwim
+                    (embark-dwim)
+                  (embark-act)))
+          (user-error
+           (when (and (string-equal (error-message-string err) "No target found")
+                      bibtex-actions-at-point-fallback)
+             (bibtex-actions-run-default-action
+              (bibtex-actions-read :rebuild-cache arg))))))
+    (if-let ((keys (bibtex-actions-citation-key-at-point)))
+        (funcall bibtex-actions-default-action keys)
+      (when bibtex-actions-at-point-fallback
+        (bibtex-actions-run-default-action
+         (bibtex-actions-read :rebuild-cache arg))))))
 
 (provide 'bibtex-actions)
 ;;; bibtex-actions.el ends here
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -485,22 +485,26 @@ TEMPLATE."
 
 ;;; Embark
 
-(defun bibtex-actions-citation-key-at-point ()
-  "Return citation keys at point as a list for `embark'."
-  (when-let ((key (or (bibtex-actions-get-key-org-cite)
-                      (bibtex-completion-key-at-point))))
-    (cons 'citation-key (if (listp key) key (list key)))))
+(cl-defun bibtex-actions-citation-key-at-point (&optional &key as-list)
+  "Return citation keys at point AS-LIST for `embark', or string."
+  (when-let* ((key (or (bibtex-actions-get-key-org-cite)
+                      (bibtex-completion-key-at-point)))
+              (crm-separator "\\s-*&\\s-*")
+              (key-list  (if (listp key) key (list key)))
+              ;; Currently, Embark only accepts a string. So we use a crm-like string to
+              ;; represent the list.
+              (return-value (if as-list key-list (string-join key-list "&"))))
+    (cons 'citation-key return-value)))
 
 (defvar bibtex-actions-buffer-map
   (let ((map (make-sparse-keymap)))
-    (define-key map "t" '("add pdf attachment" . bibtex-completion-add-PDF-attachment))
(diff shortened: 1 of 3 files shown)


bibtex-actions.el
```diff
@@ -38,6 +38,7 @@
 ;;; Code:
 
 (require 'bibtex-completion)
+(require 'embark)
```
minad, line 41: Why require embark? I recommend `with-eval-after-load` to register your package with Embark. I think your package can also be used without Embark, so the integration should be optional.

bibtex-actions.el
```diff
@@ -603,43 +604,9 @@ With prefix, rebuild the cache before offering candidates."
...
-             (bibtex-actions-run-default-action
-              (bibtex-actions-read :rebuild-cache arg))))))
-    (if-let ((keys (bibtex-actions-citation-key-at-point)))
-        (funcall bibtex-actions-default-action keys)
-      (when bibtex-actions-at-point-fallback
-        (bibtex-actions-run-default-action
-         (bibtex-actions-read :rebuild-cache arg))))))
+(add-to-list 'embark-target-finders 'bibtex-actions-citation-key-at-point)
+(add-to-list 'embark-keymap-alist '(bibtex . bibtex-actions-map))
+(add-to-list 'embark-keymap-alist '(citation-key . bibtex-actions-buffer-map))
```
minad, line 609: Wrap this by `with-eval-after-load`?

bibtex-actions.el
```diff
@@ -481,7 +482,7 @@ TEMPLATE."
     :insert (org-cite-make-insert-processor
              #'bibtex-actions-org-cite-insert
              #'org-cite-basic--complete-style)
-    :follow (lambda (_datum _arg) (call-interactively 'bibtex-actions-at-point))))
+    :follow (lambda (_datum _arg) (call-interactively 'embark-dwim))))
```
minad, line 485: Is it possible to detect Embark here and use another reasonable fallback if it is not available? The implementation below also used a fallback.

bdarcus: Done.

bdarcus: Yes, I can add that back.

But I wanted to strip this down to first get the basics working, which it's not currently.

bdarcus: Actually, I think that's why he added the at-point function; to allow people to invoke it without having embark installed.

But beyond the question you and Omar had, I was also wondering why it needed to be so complicated, with the work around comments.

bdarcus:
Can you take a look at what I've done here @ilupin?

`embark-dwim` and `embark-act` now correctly work at-point, with consistent menus in which-key (I haven't tested the default embark interface, but I believe it supports menu descriptions too).

I'm not quite sure how to simplify your at-point function, however. Suggestions?

Maybe better to start over with a much simpler one, since the workarounds are no longer necessary?

ilupin:
I think removing the workaround is enough. The main impact of this change to users is that they have to define two functions when adding an action.

```diff
diff --git a/bibtex-actions.el b/bibtex-actions.el
index 8cb2e88..11a8d61 100644
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -606,25 +606,17 @@ interactively when `bibtex-actions-at-point-fallback' is non-nil.
 With prefix ARG, rebuild the cache before offering candidates."
   (interactive "P")
   (if (fboundp 'embark-dwim)
-      (let ((embark-keymap-alist '((bibtex . bibtex-actions-map))))
+      (let ((embark-keymap-alist '((bibtex . bibtex-actions-map)
+                                   (citation-key . bibtex-actions-buffer-map))))
         (condition-case err
             (let ((embark-target-finders '(bibtex-actions-citation-key-at-point))
-                  (embark-keymap-alist '((citation-key . bibtex-actions-buffer-map)))
                   ;; `embark-general-map' is useless in this case
                   ;; because we are not in the minibuffer and the
                   ;; TARGET is not string
                   (embark-general-map embark-meta-map))
-              ;; This is a workaround: unless in the minibuffer,
-              ;; embark cannot pass TARGET to action when the action
-              ;; is an interactive command
-              (cl-letf (((symbol-function 'embark--act)
-                         (lambda (action target _)
-                           (if (where-is-internal action (list bibtex-actions-buffer-map))
-                               (funcall action target)
-                             (command-execute action)))))
-                (if bibtex-actions-embark-dwim
-                    (embark-dwim)
-                  (embark-act))))
+              (if bibtex-actions-embark-dwim
+                  (embark-dwim)
+                (embark-act)))
           (user-error
            (when (and (string-equal (error-message-string err) "No target found")
                       bibtex-actions-at-point-fallback)

```

minad:
@bdarcus @ilupin My recommendation is to remove the `bibtex-actions-at-point` entirely. This function is dubious since it overwrites `embark-target-finders/embark-general-map` locally. See also my comment https://github.com/oantolin/embark/issues/256#issuecomment-876987345. Furthermore the switching between `embark-dwim` and `embark-act` does not really make sense via `bibtex-actions-embark-dwim`. For example: `embark-act` is bound to `C-.` in my setup and `embark-dwim` is `M-.`. The only point of `embark-dwim` is to have a quick access key. Otherwise I would have to press `C-. RET` instead of `M-.`.

bdarcus:
Moving the next step discussion over here.

So I'm thinking maybe wes should try this, rather than what I have now?

> Since you are using CRM, your target finder could also return the candidates as a single string separated by crm-separator

This would be a stopgap until Embark is enhanced, but I think should work, and I think is a better stopgap than status quo.

bdarcus:
@minad

> My recommendation is to remove the `bibtex-actions-at-point` entirely.

To maintain compatibility in non-Embark setups, wouldn't we want to retain it as a wrapper of sorts, which uses `embark-dwim` if available, but has a fallback?

bibtex-actions.el
```diff
@@ -489,18 +489,19 @@ TEMPLATE."
   "Return citation keys at point as a list for `embark'."
   (when-let ((key (or (bibtex-actions-get-key-org-cite)
                       (bibtex-completion-key-at-point))))
-    (cons 'citation-key (if (listp key) key (list key)))))
+    ;; Currently, Embark only accepts a string. So we use a crm-like string to
+    ;; represent the list.
+    (cons 'citation-key (string-join (if (listp key) key (list key)) ","))))
```
bdarcus, line 494: @minad - is the above what you mean by adapting this to use a crm string?

I'm getting this error, which means the list isn't passed to the function.

``` console
bibtex-completion-get-entry1: Wrong type argument: stringp, nil
```

minad:
@bdarcus 

> To maintain compatibility in non-Embark setups, wouldn't we want to retain it as a wrapper of sorts, which uses embark-dwim if available, but has a fallback?

You could offer a command `bibtex-actions-at-point` which is completely independent of Embark. It looks at the citation at point and then opens the crm interface with these citations preselected. But what is this command supposed to do? Open the pdfs? Note that there exist no actions if Embark is not present! If you add such a command, the user can then decide if they want to use Embark or your command. But since such an independent command seems to be pretty useless, I would not offer it.

bibtex-actions.el
minad: This looks correct. I cannot say more without a stack trace.

bdarcus: > This looks correct. I cannot tell more without a stack trace.

Here's with point on a citation, with two references, and so two keys:

``` elisp
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  regexp-quote(nil)
  bibtex-completion-get-entry1(nil)
  bibtex-completion-get-entry(nil)
  bibtex-completion-open-url-or-doi((nil))
  bibtex-completion-open-pdf((nil) bibtex-completion-open-url-or-doi)
  bibtex-completion-open-any((nil))
  bibtex-actions-open((nil))
  funcall-interactively(bibtex-actions-open (nil))
  command-execute(bibtex-actions-open)
  #f(compiled-function () #<bytecode 0x11e8052d3efee9f9>)()
  embark--act(bibtex-actions-open "low2001,mcneill2011" t)
  embark-act(nil)
  funcall-interactively(embark-act nil)
  command-execute(embark-act)
```

minad: It seems the candidates are not injected correctly by Embark or lost somewhere on the way. But note that your target finder should separate the candidates with the `crm-separator`, in your case this seems to be `&`.

bdarcus: > It seems the candidates are not injected correctly by Embark or lost somewhere on the way. 

Right. I will add this piece to the Embark issue then.

> But note that your target finder should separate the candidates with the crm-separator, in your case this seems to be &.

Yes, I thought about that, but wasn't sure whether that would be active outside of bibtex-read.

But in any case, I tried using & and got the same result.

minad: Did you try this with emacs -Q? Now that I think about it, there could be an issue with how `consult-completing-read-multiple` handles such `crm-separated` input.

bdarcus: > Did you try this with emacs -Q? Now that I think about it, there could be an issue with how consult-completing-read-multiple handles such crm-separated input.

Yes, I used my test script, but commented out the `consult-completing-read-multiple` line.

bdarcus:
> @bdarcus
> 
> > To maintain compatibility in non-Embark setups, wouldn't we want to retain it as a wrapper of sorts, which uses embark-dwim if available, but has a fallback?
> 
> You could offer a command `bibtex-actions-at-point` which is completely independent of Embark. It looks at the citation at point and then opens the crm interface with these citations preselected. But what is this command supposed to do? Open the pdfs? Note that there exist no actions if Embark is not present! If you add such a command, the user can then decide if they want to use Embark or your command. But since such an independent command seems to be pretty useless, I would not offer it.

Open pdfs, links, notes, etc.

I need to provide a function for the org-cite follow processor.

``` elisp
  :follow (lambda (_datum _arg) (call-interactively 'bibtex-actions-at-point)))
```

Without Embark, it would just invoke the default bibtex-completion-at-point function.

With Embark, `embark-dwim`, which would be the same command.

For the user, however, they would be using `org-open-at-point` in both cases.

minad:
Why do you distinguish between the actions-map and the actions-buffer-map?

bdarcus:
> Why do you distinguish between the actions-map and the actions-buffer-map?

Because not all of the commands make sense in the buffer.

For example, it probably doesn't make sense to insert a key on a key.

minad:
Okay, makes sense. Still I feel that that this is overdoing it a bit. Embark also reuses the same keymap, so that not every action makes sense. @oantolin also likes to execute `embark-act` and then presses `M-x` to execute arbitrary actions - he likes to argue that the actions are entirely up to the user :)

ilupin:
> ```
> bibtex-completion-get-entry1: Wrong type argument: stringp, nil
> ```

This patch fixes this error and makes `embark-act` work with multiple keys at point.

```diff
diff --git a/bibtex-actions.el b/bibtex-actions.el
index cc384ab..fd5c668 100644
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -265,7 +265,8 @@ offering the selection candidates"
            'bibtex-actions-history bibtex-actions-presets nil)))
     (cl-loop for choice in chosen
              ;; Collect citation keys of selected candidate(s).
-             collect (cdr (assoc choice candidates)))))
+             collect (cdr (or (assoc choice candidates)
+                              (rassoc choice candidates))))))
 
 (defun bibtex-actions--format-candidates ()
   "Transform candidates from 'bibtex-completion-candidates'.
@@ -489,7 +490,7 @@ TEMPLATE."
   "Return citation keys at point as a list for `embark'."
   (when-let ((key (or (bibtex-actions-get-key-org-cite)
                       (bibtex-completion-key-at-point))))
-    (cons 'citation-key (if (listp key) key (list key)))))
+    (cons 'citation-key (if (listp key) (string-join key " & ") key))))
 
 (defvar bibtex-actions-buffer-map
   (let ((map (make-sparse-keymap)))
@@ -502,12 +503,7 @@ TEMPLATE."
     (define-key map (kbd "p") 'bibtex-actions-open-pdf)
     (define-key map (kbd "RET") 'bibtex-actions-run-default-action)
     map)
-  "Keymap for Embark citation-key actions.
-
-This keymap is used by `bibtex-actions-at-point' internally.  It
-should not be added to `embark-keymap-alist' because interactive
-commands (i.e. bibtex-actions-*) currently do not work well with
-`embark-act' outside the minibuffer.")
+  "Keymap for Embark citation-key actions.")
 
 ;;; Command wrappers for bibtex-completion functions
 
@@ -611,25 +607,13 @@ interactively when `bibtex-actions-at-point-fallback' is non-nil.
 With prefix ARG, rebuild the cache before offering candidates."
   (interactive "P")
   (if (fboundp 'embark-dwim)
-      (let ((embark-keymap-alist '((bibtex . bibtex-actions-map))))
+      (let ((embark-keymap-alist '((bibtex . bibtex-actions-map)
+                                   (citation-key . bibtex-actions-buffer-map))))
         (condition-case err
-            (let ((embark-target-finders '(bibtex-actions-citation-key-at-point))
-                  (embark-keymap-alist '((citation-key . bibtex-actions-buffer-map)))
-                  ;; `embark-general-map' is useless in this case
-                  ;; because we are not in the minibuffer and the
-                  ;; TARGET is not string
-                  (embark-general-map embark-meta-map))
-              ;; This is a workaround: unless in the minibuffer,
-              ;; embark cannot pass TARGET to action when the action
-              ;; is an interactive command
-              (cl-letf (((symbol-function 'embark--act)
-                         (lambda (action target _)
-                           (if (where-is-internal action (list bibtex-actions-buffer-map))
-                               (funcall action target)
-                             (command-execute action)))))
-                (if bibtex-actions-embark-dwim
+            (let ((embark-target-finders '(bibtex-actions-citation-key-at-point)))
+              (if bibtex-actions-embark-dwim
                     (embark-dwim)
-                  (embark-act))))
+                  (embark-act)))
           (user-error
            (when (and (string-equal (error-message-string err) "No target found")
                       bibtex-actions-at-point-fallback)

```

bdarcus:
Thank you @ilupin!

I pushed the commit here: https://github.com/bdarcus/bibtex-actions/pull/171/commits/7d255f4111c165fd28fd3839ee4a98ca112c6fdb

Still need to test it more.

bibtex-actions.el
```diff
@@ -611,25 +607,13 @@ interactively when `bibtex-actions-at-point-fallback' is non-nil.
 With prefix ARG, rebuild the cache before offering candidates."
   (interactive "P")
   (if (fboundp 'embark-dwim)
```
bdarcus, line 609: @ilupin why do we need this now?

bdarcus: I guess it's a way to test for presence of Embark.

bibtex-actions.el
```diff
@@ -611,25 +607,13 @@ interactively when `bibtex-actions-at-point-fallback' is non-nil.
...
-              ;; embark cannot pass TARGET to action when the action
-              ;; is an interactive command
-              (cl-letf (((symbol-function 'embark--act)
-                         (lambda (action target _)
-                           (if (where-is-internal action (list bibtex-actions-buffer-map))
-                               (funcall action target)
-                             (command-execute action)))))
-                (if bibtex-actions-embark-dwim
+            (let ((embark-target-finders '(bibtex-actions-citation-key-at-point)))
+              (if bibtex-actions-embark-dwim
```
bdarcus, line 614: OTOH, we shouldn't need this; right?

bdarcus:
The code looks good to me, but I'm running into some weird issues trying to test this that I can't figure out.

For example, I was going to try to modify `test/bibtex-actions.el` for this, but then got this odd type error on the `(require 'bibtex-actions)` line.

But WDYT of this small change @ilupin?

https://github.com/bdarcus/bibtex-actions/pull/171/commits/75ddbf898306cbda1713c41a64641f11a96a1fd8

It just moves the Embark config out of the at-point function.

bdarcus:
Final update for tonight, @ilupin:

As I've said else here, I have some org/org-cite issues ATM that make this difficult for me to test.

But if I run the test script, even `embark-act` from the minibuffer is no longer working correctly for me.

bibtex-actions.el
```diff
@@ -611,35 +607,25 @@ interactively when `bibtex-actions-at-point-fallback' is non-nil.
...
     (if-let ((keys (bibtex-actions-citation-key-at-point)))
         (funcall bibtex-actions-default-action keys)
       (when bibtex-actions-at-point-fallback
         (bibtex-actions-run-default-action
          (bibtex-actions-read :rebuild-cache arg))))))
 
+(with-eval-after-load "embark"
+  (add-to-list 'embark-target-finders 'bibtex-actions-citation-key-at-point)
+  (add-to-list 'embark-keymap-alist '((bibtex . bibtex-actions-map)
+                                      (citation-key . bibtex-actions-buffer-map))))
```
ilupin, line 628:
```suggestion
(with-eval-after-load "embark"
  (add-to-list 'embark-target-finders 'bibtex-actions-citation-key-at-point)
  (add-to-list 'embark-keymap-alist '(bibtex . bibtex-actions-map))
  (add-to-list 'embark-keymap-alist '(citation-key . bibtex-actions-buffer-map)))
```

ilupin: I think this should solve the problem that `embark-act` doesn't work.

test/bibtex-actions.el
```diff
@@ -14,18 +14,23 @@
...
-(global-set-key (kbd "M-o") #'embark-act)
+(global-set-key (kbd "M-;") #'embark-act)
+(global-set-key (kbd "M-.") #'embark-dwim)
 
 ;; replace CRM with consult alernative
 ;(advice-add #'completing-read-multiple :override #'consult-completing-read-multiple)
 
 ;; ensure that embark knows which map to use with bibtex entries
-(add-to-list 'embark-keymap-alist '(bibtex . bibtex-actions-map))
+;(add-to-list 'embark-keymap-alist '(bibtex . bibtex-actions-map))
```
ilupin, line 24:
```suggestion
(add-to-list 'embark-target-finders 'bibtex-actions-citation-key-at-point)
(add-to-list 'embark-keymap-alist '(bibtex . bibtex-actions-map))
(add-to-list 'embark-keymap-alist '(citation-key . bibtex-actions-buffer-map))
```

bibtex-actions.el
minad: Indeed, the add-to-list statement was wrong.

bibtex-actions.el
```diff
@@ -611,35 +607,25 @@ interactively when `bibtex-actions-at-point-fallback' is non-nil.
...
-                  (embark-act))))
-          (user-error
-           (when (and (string-equal (error-message-string err) "No target found")
-                      bibtex-actions-at-point-fallback)
-             (bibtex-actions-run-default-action
-              (bibtex-actions-read :rebuild-cache arg))))))
+      (condition-case err
+          (if bibtex-actions-embark-dwim
+              (embark-dwim)
+            (embark-act))
```
minad, line 613: Small suggestion here - why not set a `bibtex-actions-at-point-function` to `embark-dwim/embark-act` by default. If this function is not fbound, then call the bibtex-actions-run-default-action?

ilupin:
> WDYT of this small change @ilupin?
> 
> [0c794aa](https://github.com/bdarcus/bibtex-actions/commit/0c794aac95b600d36400f8917c1b48dfa76e91c9)
> 
> It just moves the Embark config out of the at-point function.

I think this change should be reverted because it breaks `bibtex-actions-at-point`: if there is no citation key at point, `bibtex-actions-at-point` will not call the fallback function when other embark targets exist at point.

> But if I run the test script, even `embark-act` from the minibuffer is no longer working correctly for me.

This is fixed in https://github.com/bdarcus/bibtex-actions/pull/171#discussion_r667415153

> I have some org/org-cite issues ATM that make this difficult for me to test.

https://github.com/bdarcus/bibtex-actions/blob/26e83fef37de89e7a5143fd7b897b2c7df183dcd/bibtex-actions.el#L481

There is a problem in the `:insert` processor. As a result, the `:follow` processor is not evaluated. If you remove the `:insert` processor, which can be fixed later, I think there is no problem with this PR.

bdarcus:
@ilupin 

> I think this change should be reverted because it breaks bibtex-actions-at-point: if there is no citation key at point, bibtex-actions-at-point will not call the fallback function when other embark targets exist at point.

Ok, but I don't understand why the embark maps need to be let-bound there, and why if they are, it works outside of that function.

Can you clarify, so I understand, and can put a comment in the code?

bibtex-actions.el
bdarcus: @ilupin - can you answer @minad's question here?

bdarcus: @minad - I may be misunderstanding your suggestion here, but my understanding is this is what this function does. 

The variable `bibtex-actions-embark-dwim` is set to `t` by default, which means if one does `M-x bibtex-actions-at-point`, or in org `M-x org-open-at-point` with Embark installed, it will run `embark-dwim`.

bdarcus:
Actually, one problem I see @ilupin:

When doing `embark-act` and choosing the default action (`RET`), it returns a type error.

But the command it runs (by default `bibtex-actions-open`) does not return that error if I choose it directly.

bibtex-actions.el
```diff
@@ -489,7 +490,7 @@ TEMPLATE."
   "Return citation keys at point as a list for `embark'."
   (when-let ((key (or (bibtex-actions-get-key-org-cite)
                       (bibtex-completion-key-at-point))))
-    (cons 'citation-key (if (listp key) key (list key)))))
+    (cons 'citation-key (if (listp key) (string-join key " & ") key))))
```
bdarcus, line 493: Actually, one problem I see @ilupin:

When doing `embark-act` on a single key and choosing the default action (`RET`), it returns a type error. Same with doing `embark-dwim`.

But the command it runs (by default `bibtex-actions-open`) does not return that error if I choose it directly.

If I change the above to this, I still get an error.

``` elisp
    (cons 'citation-key (if (listp key) (string-join key " & ") (list key)))))
```

bdarcus: Fixed in https://github.com/bdarcus/bibtex-actions/pull/171/commits/8296dbaa7f0ab2a315705c8fabf431c4ad704fa7

On `bibtex-actions-run-default-action` just needed to split if a string.

bdarcus:
I've merged this. We can address remaining small issues in a separate PR.

bdarcus:
> I've merged this. We can address remaining small issues in a separate PR.

Started at #174, but no success so far.

bibtex-actions.el
ilupin: `bibtex-actions-at-point` was actually added as an entrypoint command equivalent to `ivy/helm-bibtex`, rather than `embark-act/dwim`/`org-open-at-point`, at first. Therefore it is more than an `at-point` function, though the name may be misleading. In this sense, the code can be simplified as suggested.

minad: My proposal is to simplify the function as follows:

~~~ elisp
(defvar bibtex-actions-at-point-function 'embark-dwim)
(defvar bibtex-actions-at-point-fallback t)

(defun bibtex-actions-at-point (&optional arg)
  (interactive "P")
  (if (fboundp bibtex-actions-at-point-function)
      (condition-case err
          (funcall bibtex-actions-at-point-function)
        (user-error
         (when (and (string-equal (error-message-string err) "No target found")
                    bibtex-actions-at-point-fallback) ;; fallback when no target is found
           (let ((bibtex-actions-at-point-function)) ;; disable bibtex-actions-at-point-function and retry
             (bibtex-actions-at-point arg)))))
    (bibtex-actions-run-default-action ;; prompt directly when bibtex-actions-at-point-function is nil or not bound
     (bibtex-actions-read :rebuild-cache arg))))
~~~

PR #174, opened 2021-07-11
Add separate bibtex-actions-org-cite

bdarcus:
### Summary

I've decided to move org-cite specific stuff into a separate `bibtex-actions-org-cite`. We can start with this, and expand it as needed, including possibly later considering a minor mode and such.

It reflects where I ended up, where I was finding it difficult to fix the problem with the insert processor that @ilupin [discovered](https://github.com/bdarcus/bibtex-actions/pull/171#issuecomment-877738027), and it seemed easier to split out the org-cite stuff into a separate `bibtex-actions-org-cite` file, and remove the `with-eval-after-load` embark config from the main file.

I think this is the right approach, but could use some more review and testing.

You should just be able to load the file and it will work with `org-cite` (assuming you have that properly setup).

### Menu experiments

Note also I am experimenting here with the menu descriptions. 

So here I'm using a mix of `bibtex-actions` and native `org` commands (though where we are plugging in `bibtex-actions` commands anyway), but hidden behind descriptions with a common "cite" context prefix.

Screenshot from 2021-07-12 15-10-12

I'd actually like to also remove the embark map here,  and the other menu items, so it's just a very clean menu, focused only on the contextual commands.

### Other issues

I still need to:

- get the requires statements on the file correct; not sure on that
- I'm also looking for suggestions on how to modify the `use-package` example on the README to use this

### DIFF
--- a/bibtex-actions-org-cite.el
+++ b/bibtex-actions-org-cite.el
@@ -82,9 +82,10 @@
 (defvar bibtex-actions-org-cite-map
   (let ((map (make-sparse-keymap)))
     (define-key map "c" '("cite | insert/edit" . org-cite-insert))
-    (define-key map "o" '("cite | open-at-point" . org-open-at-point))
+    (define-key map "o" '("cite | open-at-point" . bibtex-actions-open))
     (define-key map "e" '("cite | open/edit entry" . bibtex-actions-open-entry))
     (define-key map "n" '("cite | open notes" . bibtex-actions-open-notes))
+    (define-key map "RET" '("cite | default action" . bibtex-actions-run-default-action))
     map)
   "Keymap for 'bibtex-actions-org-cite'.")
 
--- a/bibtex-actions-org-cite.el
+++ b/bibtex-actions-org-cite.el
@@ -45,14 +45,6 @@
 (require 'oc-basic)
 (require 'embark)
 
-(declare-function bibtex-actions-at-point "bibtex-actions")
-(declare-function bibtex-actions-open "bibtex-actions")
-(declare-function bibtex-actions-open-pdf "bibtex-actions")
-(declare-function bibtex-actions-open-notes "bibtex-actions")
-(declare-function bibtex-actions-open-entry "bibtex-actions")
-(declare-function org-open-at-point "org")
-(declare-function org-cite-insert "org-cite")
-
 ;TODO
 ;(defvar bibtex-actions-org-cite-open-default
 
--- a/bibtex-actions-org-cite.el
+++ b/bibtex-actions-org-cite.el
@@ -62,7 +62,7 @@
 
 (defun bibtex-actions-org-cite-follow (_datum _arg)
   "Follow processor for org-cite."
-  (call-interactively 'bibtex-actions-at-point))
+  (call-interactively 'embark-dwim))
 
 (org-cite-register-processor 'bibtex-actions-org-cite
   :insert (org-cite-make-insert-processor
--- a/bibtex-actions-org-cite.el
+++ b/bibtex-actions-org-cite.el
@@ -79,10 +79,10 @@
 
 (defvar bibtex-actions-org-cite-map
   (let ((map (make-sparse-keymap)))
-    (define-key map "c" '("(cite) insert/edit" . org-cite-insert))
-    (define-key map "o" '("(cite) open-at-point" . org-open-at-point))
-    (define-key map "e" '("(cite) open/edit entry" . bibtex-actions-open-entry))
-    (define-key map "n" '("(cite) open notes" . bibtex-actions-open-notes))
+    (define-key map "c" '("cite | insert/edit" . org-cite-insert))
+    (define-key map "o" '("cite | open-at-point" . org-open-at-point))
+    (define-key map "e" '("cite | open/edit entry" . bibtex-actions-open-entry))
+    (define-key map "n" '("cite | open notes" . bibtex-actions-open-notes))
     map)
   "Keymap for 'bibtex-actions-org-cite'.")
 
--- a/bibtex-actions-org-cite.el
+++ b/bibtex-actions-org-cite.el
@@ -45,6 +45,9 @@
 (require 'oc-basic)
 (require 'embark)
 
+(declare-function bibtex-actions-at-point "bibtex-actions")
+(declare-function org-open-at-point "org")
+
 ;TODO
 ;(defvar bibtex-actions-org-cite-open-default
 
--- a/bibtex-actions-org-cite.el
+++ b/bibtex-actions-org-cite.el
@@ -85,9 +85,9 @@
     (define-key map "o" '("cite | open-at-point" . bibtex-actions-open))
     (define-key map "e" '("cite | open/edit entry" . bibtex-actions-open-entry))
     (define-key map "n" '("cite | open notes" . bibtex-actions-open-notes))
-    (define-key map "RET" '("cite | default action" . bibtex-actions-run-default-action))
+    (define-key map (kbd "RET") '("cite | default action" . bibtex-actions-run-default-action))
     map)
-  "Keymap for 'bibtex-actions-org-cite'.")
+  "Keymap for 'bibtex-actions-org-cite' `embark' at-point functionality.")
 
 ;; Bibtex-actions-org-cite configuration
 
--- a/bibtex-actions-org-cite.el
+++ b/bibtex-actions-org-cite.el
@@ -8,7 +8,7 @@
 ;; License: GPL-3.0-or-later
 ;; Version: 0.1
 ;; Homepage: https://github.com/bdarcus/bibtex-actions
-;; Package-Requires: ((emacs "26.3") (bibtex-actions "0.5"))
+;; Package-Requires: ((emacs "26.3"))
 ;;
 ;; This file is not part of GNU Emacs.
 ;;
--- a/bibtex-actions-org-cite.el
+++ b/bibtex-actions-org-cite.el
@@ -45,6 +45,11 @@
 (require 'oc-basic)
 (require 'embark)
 
+(declare-function bibtex-actions-at-point "bibtex-actions")
+(declare-function bibtex-actions-open "bibtex-actions")
+(declare-function bibtex-actions-open-pdf "bibtex-actions")
+(declare-function bibtex-actions-open-notes "bibtex-actions")
+(declare-function bibtex-actions-open-entry "bibtex-actions")
 (declare-function org-open-at-point "org")
 (declare-function org-cite-insert "org-cite")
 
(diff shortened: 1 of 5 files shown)


bibtex-actions.el
```diff
@@ -472,17 +472,22 @@ TEMPLATE."
 
 (defun bibtex-actions-org-cite-insert (&optional multiple)
   "Return a list of keys when MULTIPLE, or else a key string."
+  ;; FIX
```
bdarcus, line 475: Somewhere in here is the problem, I think.

In the wip-cite-new branch, just before merger, I added this code so the function would return a list, or an individual string.

But since mainlining, this has broken, and nothing I had tried has fixed it.

minad:
I think a minor mode is not needed for setting two variables. I tend to define modes only for more intrusive things and more complicated setups. Furthermore modes are useful if it is planned that the feature is regularly turned on and off, which isn't the case here, I assume. 

But if you are still convinced to use a mode, make sure that you can turn it off and on, such that it resets the state cleanly when turning if off. See for example https://github.com/minad/vertico/blob/54f8a4af0c632155ec6d38c0496da33df6211ab2/extensions/vertico-mouse.el#L81-L91.

bdarcus:
But do you agree it's better as a separate package?

The minor mode was just a small idea with that. But I wasn't seen a clean way to do this in the main package.

bdarcus:
> I would not make it a fully separate package (on the package level), just an extra file like you have it. For example in consult, I have the consult-org.el, consult-flymake.el, consult.el files, which are all part of the consult package. Only consult-flycheck is a really separate package since it does depend on flycheck, which is not an Emacs core component.

Yes, that's what I mean. It would live here, and there would be no new MELPA recipe, or change in the existing one. It would just allow someone to do:

``` elisp
(require 'bibtex-actions-org-cite)
```

bdarcus:
> But if you are still convinced to use a mode, make sure that you can turn it off and on, such that it resets the state cleanly when turning if off. See for example

@minad - so in this case, it would mean resetting the embark and org-cite variables to the state they were in before turning it on.

How do you do that?

Do I have to first store the state of those variables, and then reset them to those?

minad:
> Do I have to first store the state of those variables, and then reset them to those?

Yes. But you could also just not provide the mode ;)

bdarcus:
> Yes. But you could also just not provide the mode ;)

I'll remove it.

I want to get this working ASAP, and can worry about details like that later.

minad:
> Little question on naming, @minad: is not the separate file still technically a "package" from an emacs standpoint?
> Or what do we call it?

I don't really know to be honest. I call it file, package file, .. It would probably be more correct to say package only for the whole thing. And the file is called feature, see the `features` variable and the `require` documentation. But who calls a file a feature?

bdarcus:
> I'd actually like to also remove the embark map here

@minad - how do I do this?

I basically only want the embark general commands available in the minibuffer menu since I think they're only useful there; not at point, with the `citation-key` targets.

PS - I know my annoyance with these menus is mostly because of `which-key`, but since Doom uses `which-key` ...

PS 2 - I'm trying to get these details write because I want to update screenshots, and maybe do a short screencast, so people can see the possibilities of org-cite and embark et al!

bdarcus:
> I call it file, package file, .. It would probably be more correct to say package only for the whole thing. And the file is called feature, see the `features` variable and the `require` documentation. But who calls a file a feature?

I just leave that out in the README :-)

minad:
I agree about fixing oantolin/embark#139, but nothing should be done about oantolin/embark#207. As we discussed in 207, the general map can be easily rebound to another key. Furthermore if https://github.com/oantolin/embark/issues/258 is fixed, some keymaps could opt out of the general map.

bdarcus:
> I agree about fixing [oantolin/embark#139](https://github.com/oantolin/embark/issues/139), but nothing should be done about [oantolin/embark#207](https://github.com/oantolin/embark/issues/207). As we discussed in 207, the general map can be easily rebound to another key in the user config. Furthermore if [oantolin/embark#258](https://github.com/oantolin/embark/issues/258) is fixed, some keymaps could opt out of the general map.

You're right; I keep forgetting the state of different details.

bdarcus:
@minad - how is loading for something like `consult-org.el` handled?

I noticed in the vertico doom PR, for example, that file is never explicitly loaded, and there's no `use-package` for it, but the commands within it are there.

So does that mean the Emacs, or whatever is handling this loading, is smart enough to know that `consult-org` is a file associated with the `consult` package?

I want to edit the `use-package` example here, and of course am also working a Doom PR that will include this.

bibtex-actions-org-cite.el
```diff
@@ -0,0 +1,103 @@
...
+(defun bibtex-actions-org-cite-insert (&optional multiple)
+  "Return a list of keys when MULTIPLE, or else a key string."
+  (let ((references (bibtex-actions-read)))
+    (if multiple
+        references
+      (car references))))
+
+(defun bibtex-actions-org-cite-follow (_datum _arg)
+  "Follow processor for org-cite."
+  (call-interactively 'embark-dwim))
```
bdarcus, line 65: @minad @ilupin - how does `embark-dwim` know what default function to call?

I've been driving myself crazy this morning trying to figure this out, and have fixed one piece. But still getting a void symbol error with `embark-dwim`.

bdarcus: Of course the problem in the end was some stupid syntax issue with the keymap. Now fixed.

bdarcus:
OK, finally got this working.

Just need some more testing, especially with the test script.

bdarcus:
I merged this, since it also fixes a bug.

publicimageltd:
Just an aside: I am currently too busy IRL to test stuff. I follow the development, however, and it's nice to see how the package grows.

iyefrat:
> I noticed in the vertico doom PR, for example, that file is never explicitly loaded, and there's no use-package for it, but the commands within it are there.
>
>So does that mean the Emacs, or whatever is handling this loading, is smart enough to know that consult-org is a file associated with the consult package?

This is done with [autoloading](https://www.gnu.org/software/emacs/manual/html_node/eintr/Autoload.html). As I understand it, there are roughly two steps to adding a package to emacs. First, you add the directory that contains the elisp files you want (or just the elisp files) to load to your `load-path`, so Emacs knows about them. Second, you `require` the files to load the functions. Autoloading allows you to mark specific functions with the files in your `load-path` that they're defined in, so you can load the relevant files whenever you first use the autoloaded function (which is very useful for shortening startup times).

 Emacs has some built in ways to make this whole process easier, see [here](https://www.gnu.org/software/emacs/manual/html_node/elisp/Autoload.html#Autoload). Generally you mark all the functions you want to autoload with a `;;;###autoload` above the function, and then something (that I haven't looked into that much yet because Doom does it automatically) scoops all of those up to a list of autoloaded functions that Emacs knows about.

bdarcus:
> Emacs has some built in ways to make this whole process easier, see [here](https://www.gnu.org/software/emacs/manual/html_node/elisp/Autoload.html#Autoload). Generally you mark all the functions you want to autoload with a `;;;###autoload` above the function, and then something (that I haven't looked into that much yet because Doom does it automatically) scoops all of those up to a list of autoloaded functions that Emacs knows about.

So if the file is in the same root directory as the main package file, AND I have the autoload cookie attached to some function(s) I need in that separate file, I should be good?

So practically speaking, I should add an autoload function to [bibtex-actions-org-cite.el](https://github.com/bdarcus/bibtex-actions/blob/main/bibtex-actions-org-cite.el)?

What I merged here was a first cut to get it working. I know it could some refinement!

iyefrat:
> So if the file is in the same root directory as the main package file, AND I have the autoload cookie attached to some function(s) I need in that separate file, I should be good?

Yep, that's how `consult-org` gets autoloaded.

> So practically speaking, I should add an autoload function to bibtex-actions-org-cite.el?
>
>Like, maybe I wrap much of that in a bibtex-actions-org-cite-setup autoload function?

Generally you should only autoload user-facing functions in the package that are likely to be the first functions used in a given session. For example in my package `evil-tex`, the only function we autoloaded is the one that defines `evil-tex-mode`, because you can't use the package if that isn't on. `consult` doesn't have a minor mode like that, so they instead autoloaded all the `consult-*` commands that users are expected to use, so consult gets loaded whenever you use one of those.

In `bibtex-actions-org-cite` you'd probably want to autoload `bibtex-actions-org-cite-insert`, and maybe `bibtex-actions-org-cite-insert-follow`, not super sure what the follow function does. 

> Edit: I will say that what I have here does appear to work. Maybe as others test the doom PR they might find otherwise.

What do you mean?

bdarcus:
> In `bibtex-actions-org-cite` you'd probably want to autoload `bibtex-actions-org-cite-insert`, and maybe `bibtex-actions-org-cite-insert-follow`, not super sure what the follow function does.

The follow function just plugs in to `org-open--at-point`, which is the command the user is interacting with; so:

- "follow" -> `M-x org-open-at-point`
- "insert" -> `M-x org-cite-insert`

This bit is the key part that exposes the org-cite "processor":

``` elisp
(org-cite-register-processor 'bibtex-actions-org-cite
           #'org-cite-basic--complete-style)
  :follow #'bibtex-actions-org-cite-follow)
```

But there's no defun there; I guess that's macro that's creating the processor (?), so that a user can then do:

``` elisp
(setq org-cite-follow-processor 'bibtex-actions-org-cite)
```

> What do you mean?

I meant my Doom PR. As in, I don't think there's a problem ATM, but people may find otherwise.

iyefrat:
> I meant my Doom PR. As in, I don't think there's a problem ATM, but people may find otherwise.

Oh, there it works because your `use-package` statement just loads `bibtex-actions-org-cite` as soon as `org`, `oc`, and `bibtex-actions` are all loaded. That's probably the best way to go about it, since it doesn't look like the user runs functions from here directly. I'd add that `use-package` statement to the readme as recommend configuration.

You also probably want to add `(org "9.5")` to the requirements there, and add similar workaround for the embark portion depending on what you end up doing with the embark part of this package.

bdarcus:
> You also probably want to add `(org "9.5")` to the requirements there ...

Do I have to wait to do this, given there's no official 9.5 release yet?

Or is it enough that `org-version` now reports "9.5"?

When I run package-lint it complains about org not being installable.

iyefrat:
> When I run package-lint it complains about org not being installable.
I think this is a false positive that stems from it trying to install it from gnu elpa. The version in the org package is `9.5-dev`, and the version from  gnu-devel elpa is `9.5snapshot0.20210702.140319`. Both of these should satisfy a `9.5` version requirement I think. Do you still get the error after you add gnu-devel elpa to your archives?

bdarcus:
@minad 

> ... if [oantolin/embark#258](https://github.com/oantolin/embark/issues/258) is fixed, some keymaps could opt out of the general map.

Now that Omar fixed it, how do I opt out?

PR #175, opened 2021-07-11
docs: Clarify search fields

### LINKED ISSUE #173: Adding `doi` and `url` to `bibtex-completion-additional-search-fields`
Currently, `bibtex-actions--format-candidates` defines the value of `has:link` by searching for `doi` or `url` fields in each entry returned by `(bibtex-completion-candidates)`. However, by default, this function only returns `authors`, `editors`, `title`, `year`, `type`, and `key` for each entry.

So I added the relevant fields to `bibtex-completion-additional-search-field` and links are properly filtered.

```elisp
(dolist (field '(url doi))
    (add-to-list 'bibtex-completion-additional-search-fields field))
```
I think this modification should be included in `bibtex-actions.el`, or we should at least mention this in the README?

bdarcus:
Update the README and the test script.

Fix #173.

### DIFF
--- a/README.org
+++ b/README.org
@@ -51,9 +51,9 @@ To setup bibtex-actions using =use-package=, you can do:
          ("M-b" . bibtex-actions-insert-preset))
   :after embark
   :config
-  ;; Make the 'bibtex-actions' bindings available from `embark-act'.
-  (add-to-list 'embark-keymap-alist '(bibtex . bibtex-actions-map))
   :custom
+  ;; make sure to set this to ensure open commands work correctly
+  (bibtex-completion-additional-search-fields '(doi url))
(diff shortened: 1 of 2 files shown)


hieutkt:
Using `add-to-list` should be safer than `setq`, no? In case users already set the default value of `bibtex-completion-additional-search-fields` elsewhere in their configs? That's why in my original issue I had:

```emacs-lisp
(dolist (field '(url doi))
    (add-to-list 'bibtex-completion-additional-search-fields field))
```

bdarcus:
> On second thought I think it's trivial to users to manage themselves.

That was my thought.

It's just an example, and normally a user should set it in one place. But some want to use multiple packages that depend on that backend, so ...

PR #176, opened 2021-07-12
oc: Use bibtex-actions-at-point-function for follow processor

bdarcus:
This is @minad's suggestion: https://github.com/bdarcus/bibtex-actions/pull/171#discussion_r667755127

Edit: actually, I integrated a part of that suggestion in #174. This just refines it.

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -161,12 +161,10 @@ means no action."
   :type '(choice (const :tag "Prompt" 'prompt)
                  (const :tag "Ignore" nil)))
 
-(defcustom bibtex-actions-embark-dwim t
-  "Whether to run the default action on citation keys at point.
-If non-nil, call the default embark action through `embark-dwim'.
-If nil, prompt the user for an action through `embark-act'."
+(defcustom bibtex-actions-at-point-function 'embark-dwim
+  "The function to run for 'bibtex-actions-at-point'."
   :group 'bibtex-actions
-  :type 'boolean)
+  :type 'function)
 
 ;;; History, including future history list.
 
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -569,8 +569,8 @@ With prefix, rebuild the cache before offering candidates."
 (defun bibtex-actions-dwim ()
   "Run the default action on citation keys found at point."
   (interactive)
-    (if-let ((keys (bibtex-actions-citation-key-at-point)))
-        (funcall bibtex-actions-default-action keys)))
+  (if-let ((keys (bibtex-actions-citation-key-at-point)))
+      (funcall bibtex-actions-default-action keys)))
 
 (provide 'bibtex-actions)
 ;;; bibtex-actions.el ends here
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -161,7 +161,7 @@ means no action."
   :type '(choice (const :tag "Prompt" 'prompt)
                  (const :tag "Ignore" nil)))
 
-(defcustom bibtex-actions-at-point-function 'embark-dwim
+(defcustom bibtex-actions-at-point-function 'bibtex-actions-dwim
   "The function to run for 'bibtex-actions-at-point'."
   :group 'bibtex-actions
   :type 'function)
--- a/bibtex-actions-org-cite.el
+++ b/bibtex-actions-org-cite.el
@@ -36,7 +36,7 @@
 
 (require 'bibtex-actions)
 (require 'org)
-(require 'org-cite)
+(require 'oc)
 (require 'oc-basic)
 (require 'embark)
 
--- a/bibtex-actions-org-cite.el
+++ b/bibtex-actions-org-cite.el
@@ -57,7 +57,7 @@
 
 (defun bibtex-actions-org-cite-follow (_datum _arg)
   "Follow processor for org-cite."
-  (call-interactively 'embark-dwim))
+  (call-interactively 'bibtex-actions-at-point-function))
 
 (org-cite-register-processor 'bibtex-actions-org-cite
   :insert (org-cite-make-insert-processor
@@ -88,6 +88,7 @@
 
 (setq org-cite-follow-processor 'bibtex-actions-org-cite)
 (setq org-cite-insert-processor 'bibtex-actions-org-cite)
+(setq bibtex-actions-at-point-function 'embark-dwim)
 
 ;; Embark configuration for org-cite
 
--- a/bibtex-actions-org-cite.el
+++ b/bibtex-actions-org-cite.el
@@ -57,7 +57,7 @@
 
 (defun bibtex-actions-org-cite-follow (_datum _arg)
   "Follow processor for org-cite."
-  (call-interactively 'bibtex-actions-at-point-function))
+  (call-interactively bibtex-actions-at-point-function))
 
 (org-cite-register-processor 'bibtex-actions-org-cite
   :insert (org-cite-make-insert-processor


bibtex-actions.el
```diff
@@ -161,7 +161,9 @@ means no action."
   :type '(choice (const :tag "Prompt" 'prompt)
                  (const :tag "Ignore" nil)))
 
-(defcustom bibtex-actions-embark-dwim t
+(defvar bibtex-actions-at-point-fallback t)
```
minad, line 164: Make the fallback variable a defcustom too?

bdarcus: Funny; I was just writing that question.

bdarcus: Do you want me to credit you on the commit, BTW?

I know it's a small suggestion, but figure I may as well ask.

minad: No need, it is just a small suggestion.

bdarcus: There's already a defcustom for this. Do you see any issue with this?

``` elisp
(defcustom bibtex-actions-at-point-fallback 'prompt
  "Fallback action for `bibtex-actions-at-point'.
The action is used when no citation key is found at point.
`prompt' means choosing entries via `bibtex-actions-read' and nil
means no action."
  :group 'bibtex-actions
  :type '(choice (const :tag "Prompt" 'prompt)
                 (const :tag "Ignore" nil)))
```

minad: Ah okay right. Okay to keep it like this I guess.

bdarcus: Hmm ... I'm not understanding why it's not doing the fallback prompting, even in your version with the boolean.

minad: Oh no idea, I am sorry - I didn't test this. I just wrote this quickly to illustrate the simplification I thought is useful. Something along these lines should work of course.

bdarcus:
@ilupin - this works, except for the fallback prompting.

Let me know WYT.

bibtex-actions.el
```diff
@@ -168,6 +168,18 @@ If nil, prompt the user for an action through `embark-act'."
   :group 'bibtex-actions
   :type 'boolean)
 
+(defcustom bibtex-actions-at-point-fallback t
```
bdarcus, line 171: Not sure about this.

bibtex-actions.el
```diff
@@ -607,21 +584,16 @@ If no citation key is found, target entries can be chosen
...
-    (if-let ((keys (bibtex-actions-citation-key-at-point)))
-        (funcall bibtex-actions-default-action keys)
-      (when bibtex-actions-at-point-fallback
-        (bibtex-actions-run-default-action
-         (bibtex-actions-read :rebuild-cache arg))))))
+                    bibtex-actions-at-point-fallback) ;; fallback when no target is found
+           (let ((bibtex-actions-at-point-function)) ;; disable bibtex-actions-at-point-function and retry
+             (bibtex-actions-at-point arg)))))
+    (bibtex-actions-run-default-action ;; prompt directly when bibtex-actions-at-point-function is nil or not bound
+     (bibtex-actions-read :rebuild-cache arg))))
```
ilupin, line 596:
```suggestion
    (if-let ((keys (bibtex-actions-citation-key-at-point)))
        (funcall bibtex-actions-default-action keys))))
```

ilupin: We should just run the default action on the citation keys if `bibtex-actions-at-point-function` is not available.

I think `bibtex-actions-at-point-fallback` should also be removed because (1) it depends on the specific error raised by `embark` and (2) it is unnecessary in an `at-point` function. I suggest the following 

```lisp
(defun bibtex-actions-at-point ()
  "Run the default action on citation keys found at point."
  (interactive)
  (if (fboundp bibtex-actions-at-point-function)
      (funcall bibtex-actions-at-point-function)
    (if-let ((keys (bibtex-actions-citation-key-at-point)))
        (funcall bibtex-actions-default-action keys))))
```
and set the default value of `bibtex-actions-at-point-function` to nil.

minad: Oh I am happy with that suggestion of yours @ilupin. This is much better than what I had proposed (I had also thought about this).

The good thing is you can take this suggestion to the natural conclusion and remove `bibtex-actions-at-point` entirely without loss of functionality :grimacing:

~~~ elisp
(defun bibtex-dwim ()
  "Run the default action on citation keys found at point."
  (interactive)
    (if-let ((keys (bibtex-actions-citation-key-at-point)))
        (funcall bibtex-actions-default-action keys)))
~~~

Then there is no need for a `bibtex-actions-at-point-function` or a `bibtex-actions-at-point`. The user either calls `embark-dwim`, `embark-act` or `bibtex-dwim`. This is actually the best solution and what I had originally in mind when asking @bdarcus why there is even a `bibtex-actions-at-point` command wrapping `embark-act/dwim`.

bibtex-actions.el
```diff
@@ -161,12 +161,10 @@ means no action."
   :type '(choice (const :tag "Prompt" 'prompt)
                  (const :tag "Ignore" nil)))
 
-(defcustom bibtex-actions-embark-dwim t
-  "Whether to run the default action on citation keys at point.
-If non-nil, call the default embark action through `embark-dwim'.
-If nil, prompt the user for an action through `embark-act'."
+(defcustom bibtex-actions-at-point-function 'embark-dwim
```
ilupin, line 164:
```suggestion
(defcustom bibtex-actions-at-point-function nil
```

bibtex-actions.el
bdarcus: WDYT @ilupin ?

And then on #174, how would that impact this key (now broken) function?

``` elisp
(defun bibtex-actions-org-cite-follow (_datum _arg)
  "Follow processor for org-cite."
  (call-interactively 'bibtex-actions-at-point))
```
What would we be calling here?

Maybe we need the defcustom for this case?

To remind: this is the function that hooks up to `org-open-at-point`.

ilupin: > But maybe we need the defcustom here, in case a user wants `embark-act`?

There is `bibtex-actions-at-point-function`, which can be called directly in the `:follow` processor. Users can set it to `bibtex-actions-dwim`(the default) or `embark-act/dwim`, then we don't need to wrap `embark-act/dwim` in any `bibtex-actions` functions.

```lisp
(defun bibtex-actions-org-cite-follow (_datum _arg)
  "Follow processor for org-cite."
  (call-interactively bibtex-actions-at-point-function))

;; or keep the original name `bibtex-actions-at-point'
(defun bibtex-actions-dwim ()
  "Run the default action on citation keys found at point."
  (interactive)
    (if-let ((keys (bibtex-actions-citation-key-at-point)))
        (funcall bibtex-actions-default-action keys)))
```

bdarcus: I'm realizing there a couple of things specific to `org-cite` here we should address.

1. this
2. also, I mention in the main post that I think we should have a target finder specific to org-cite citations, so we can bind a different map to it

I'm hoping to sort all this out today, but I will probably merge #174 first, with hard-coded `embark-dwim`, and then finish this, and probably make the follow processor configurable there.

That's because this is more complicated.

bdarcus:
OK, as I planned, I got #174 working first and merged, and then rebased main on here.

I think I'm going to turn off CI.

bibtex-actions.el
bdarcus: I'm unsure the best approach.

We could do this change, and reset on `bibtex-actions-org-cite`?

bdarcus: > We could do this change, and reset on `bibtex-actions-org-cite`?

I made this change. I think that's my last commit for the day.

ilupin:
> 1. `embark-act` -> `RET`

It works well for me, and I don't find any other problem.

bdarcus:
> It works well for me, and I don't find any other problem.

OK, thanks.

One thing I forgot to test is fallback behavior, which doesn't seem to work.

I'll look into it.

PR #177, opened 2021-07-12
ensure `style` is set in bibtex-actions--format-citation-org

### LINKED ISSUE #162: `bibtex-actions--format-citation-org` inserts malformed link when not prompting for optional args
**Describe the bug**
If `bibtex-completion-cite-prompt-for-optional-arguments` is set to `nil`,  the inserted citation ends up with an extraneous `/`, e.g.

`[cite/:ReevesHowchilipepper2008]`

FIx is easy, just change line 199 to

`(style  (if (or (not style) (string-equal style "default")) "" (concat "/" style)))` 

or alternatively, add "default" to the final branch of the if statement in the let-binding of 'style in that same function.

titaniumbones:
if `bibtex-completion-cite-prompt-for-optional-arguments` is nil, set `style` to "default" before inserting citation string.

fixes #162


bdarcus:
Thank you!

bdarcus:
BTW, the `org-cite-insert` based "insert processor" alternative: I'm working on that on #174.

PR #180, opened 2021-07-14
Update use-package examples

bdarcus:
This updates the README to offer three config options to best match recent org-cite/embark-related enhancements:

1. basic
2. embark (embark, but works the same across markdown, latex, org)
3. org-cite (same as 2, but has custom menus and commands for org-cite)

I can then work backward from there; tweak code to match what we want in the `use-package` statements.

In particular, a key question: what's the best way to deal with the fact that org-cite-related functionality depends on Org 9.5?

Aside: `(featurep 'oc)` will test for presence of `org-cite`.

Right now, I put all of that in [`bibtex-actions-org-cite.el`](https://github.com/bdarcus/bibtex-actions/blob/main/bibtex-actions-org-cite.el).

But should I move the functions on which org-cite functionality depends back into the main package file, and configure the variables in `use-package` (see https://github.com/bdarcus/bibtex-actions/pull/180#discussion_r670394355)?

Or is status quo on this basically the right approach?

### DIFF
--- a/README.org
+++ b/README.org
@@ -32,6 +32,14 @@ Here's a screenshot with [[https://github.com/raxod502/selectrum][vertico]], [[h
 
 Bibtex-actions is available for installation from [[https://melpa.org][MELPA]].
 
+In addition, the following packages are strongly recommended for the best experience.
+
+1. [[https://github.com/minad/vertico][Vertico]] or [[https://github.ccm/raxod502/selectrum][Selectrum]] (completion interface)
+2. [[https://github.com/oantolin/orderless][Orderless]] (completion style)
+3. [[https://github.com/oantolin/embark][Embark]] (contextual actions)
+4. [[https://github.com/minad/marginalia][Marginalia]] (annotations, and also candidate classification for Embark)
+4. [[https://github.com/minad/consult][Consult]] (enhanced interface for multiple candidate selection)
+
 ** Configuration
    :PROPERTIES:
    :CUSTOM_ID: configuration
@@ -42,7 +50,24 @@ Bibtex-actions is available for installation from [[https://melpa.org][MELPA]].
     :CUSTOM_ID: basic
     :END:
 
-To setup bibtex-actions using =use-package=, you can do:
+This is the minimal configuration, and will work with any completing-read compliant vertical completion UI, like Vertico, Selectrum, or the built-in icomplete-vertical, with actions available via =M-x= commands.
+
+#+BEGIN_SRC emacs-lisp
+(use-package bibtex-actions
+  :bind (("C-c b" . bibtex-actions-insert-citation)
+         :map minibuffer-local-map
+         ("M-b" . bibtex-actions-insert-preset))
+  :config
+  ;; make sure to set this to ensure open commands work correctly
+  (bibtex-completion-additional-search-fields '(doi url))
+  (bibtex-completion-bibliography '("~/bib/references.bib")))
+#+END_SRC
+
+Since most of the command logic resides in bibtex-completion, that is where to look for different [[https://github.com/tmalsburg/helm-bibtex#basic-configuration-recommended][configuration options]].
+
+*** Embark
+
+Highly recommended, this option adds embark, for contextual access to actions in the minibuffer and at-point.
 
 #+BEGIN_SRC emacs-lisp
 (use-package bibtex-actions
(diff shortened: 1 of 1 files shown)


README.org
```diff
@@ -59,28 +57,53 @@ To setup bibtex-actions using =use-package=, you can do:
...
 
 #+BEGIN_SRC emacs-lisp
-  (setq bibtex-completion-bibliography "~/bib/references.bib")
+(use-package bibtex-actions
+  :bind (("C-c b" . bibtex-actions-insert-citation)
+         :map minibuffer-local-map
+         ("M-b" . bibtex-actions-insert-preset))
+  :after embark
+  :config
+  ;; Make the 'bibtex-actions' bindings and targets available to `embark'.
```
bdarcus, line 71: Right now, I have these variables set in `bibtex-actions-org-cite`.

So I should align these; either move these into a separate `bibtex-actions-embark.el` file, or remove vars from `bibtex-actions-org-cite`, and move them to the below config.

iyefrat: The main issue from my perspective is that you essentially need two extra files for consistency in the first approach, both `bibtex-actions-embark` and `bibtex-actions-org-cite-embark` (correct me if I'm wrong, but it seems that you want to have `bibtex-actions-org-cite` be optional for users). That feels a bit ridiculous. But having all the bindings be in the `use-package` means more work on the user side, which would be nice to avoid if possible.

 After some more thought I'm not sure what are the best practices here for dealing with optional dependencies. I'd ask about this in #emacs-lisp in the Doom discord, I feel like you'd get a better answer there for the best way to go about having an optional dependency, I might be missing something.

> Also, I don't need :custom here; right?

I'm not sure what `:custom` actually does. It relates to Emacs' `M-x customize` feature, which is disabled in Doom due to some weird stuff it does involving themes, so I never looked into how it works.

bdarcus: > But having all the bindings be in the `use-package` means more work on the user side, which would be nice to avoid if possible.

Right.

One options to remove `bibtex-actions-org-cite` again, put the functions in the main file, and then have the `use-package` look more like what I have here for the "embark" option.

That seems on reflection a reasonable balance?

bdarcus: I will update the README with what I'm thinking, and then can make the code change if it seems sensible.

bdarcus: > I will update the README with what I'm thinking, and then can make the code change if it seems sensible.

Done.

Let me know WYT. If seems OK, I'll update the code.

iyefrat: Are you moving the `bibtex-actions-org-cite` stuff back to the main file?

bdarcus: > Are you moving the `bibtex-actions-org-cite` stuff back to the main file?

That's what I'm proposing. I thought I'd do that here before making the code change.

WDYT?

iyefrat: The primary issue is that it depends on Org 9.5. You should wrap up the `org-cite` stuff in something that makes sure org is new enough so that users of the package on older Org versions don't get left out. Other than that I think it's a good idea.

iyefrat: Another option would be to declare variables for the byte compiler. Not too sure what happens if you do it that way and use an older version of Org though.

bdarcus: > The primary issue is that it depends on Org 9.5. You should wrap up the `org-cite` stuff in something that makes sure org is new enough so that users of the package on older Org versions don't get left out. Other than that I think it's a good idea.

"Primary issue" you mean just if, say, one were to drop this in a config now, as is?

But, for example, in the doom PR, I can just keep the org-cite variables associated with the "oc" feature? If the feature isn't available, it won't be loaded?

Or I could modify this here to do that as well? See below for alternate.

> Another option would be to declare variables for the byte compiler. 

I already do this for the embark variables.

``` elisp
;; Set bibliography paths so they are the same.
(my/bibs '("~/bib/references.bib"))

(use-package bibtex-actions
  :after (embark oc) ; or :require?
  :config
  ;; Make the 'bibtex-actions' org-cite target and bindings available to `embark'.
  (add-to-list 'embark-target-finders 'bibtex-actions-org-cite-citation-finder)
  (add-to-list 'embark-keymap-alist '(oc-citation . bibtex-actions-org-cite-map))
  ;; Make sure to set this to ensure 'bibtex-actions-open-link' command works correctly.
  (bibtex-completion-additional-search-fields '(doi url))
  (bibtex-completion-bibliography my/bibs))

(use-package oc
  :config
  ;; activate processor for fontification, preview, etc
  ;; currently using basic, but would prefer org-cite-csl-activate
  (org-cite-activate-processor 'basic)
  (org-cite-follow-processor 'bibtex-actions-org-cite
  (org-cite-insert-processor 'bibtex-actions-org-cite)
  ;; setup export processor; default csl/citeproc-el, with biblatex for latex
  (org-cite-export-processors
       '((latex biblatex nil nil)
        (t csl nil nil)))
  (org-cite-global-bibliography '("~/bib/references.bib"))))

;; Use consult-completing-read for enhanced interface.
(advice-add #'completing-read-multiple :override #'consult-completing-read-multiple)

Comment (iyefrat) [reply README.org:79]: The issue is the following portion:
```lisp
(use-package bibtex-actions
  :after (embark oc) ; or :require?
  ;...
  )
```
Using `:after` is what you want here, but this means that if you try to use `bibtex-actions` with an old version of Org the configuration won't load because it'll be waiting for `oc` forever. I missed that bit.

Due to the fact the `org-cite` stuff has to be loaded after `oc` it's probably best to keep it in it's own file, otherwise the load order stuff gets weird. The embark configuration looks simple enough to be relegated to the `use-package` statements though.

bdarcus: So the upshot is you think I should keep the separate file?

iyefrat: Yeah I think so. But you need to add the following line to the `bibtex-actions-org-cite` use-package:

```lisp
(add-to-list 'embark-keymap-alist '(oc-citation . bibtex-actions-org-cite-map))
```

bdarcus: > But you need to add the following line to the `bibtex-actions-org-cite` use-package:

Actually, I don't currently, as I include that in `bibtex-actions-org-cite`.

I am a little uncertain if all the variables I set there should be set there though.

BTW, I added this question as an "issue" on the doom PR, with a link here.

iyefrat: > BTW, I added this question as an "issue" on the doom PR, with a link here.

Yeah that's probably best, I'm not super sharp on the best practices here.

bdarcus: Org itself has code like this:

``` elisp
(when (featurep 'org-inlinetask)
```

``` elisp
(unless (featurep 'org-element)
  (signal 'missing-test-dependency "org-element"))
```

If I were to post something on the emacs-list channel, what's the most concise way to put the question?

bdarcus: @ilupin - can you take a look at this? I have not yet tested it, but this commit experiments with moving back the org-cite and embark target loading back to the main file, but using a featurep:

https://github.com/bdarcus/bibtex-actions/pull/180/commits/43a78637d2f689ef8905215435fb4cff066791d0

If it's a bad idea, I can just drop the commit.

bdarcus: gagbo concludes:

> you can just make bibtex-actions-embark  file that manually requires embark
> so users can be told to manually require it (it = bibtex-actions-embark)

So I think we end with more-or-less status quo: no real change in code.

I'll update the README accordingly, and drop the code commits.

Issue #168, opened 2021-07-08
oc: Enhance style UI


bdarcus:
With the new "insert" processor I added, I am just using the same simple UI as oc-basic.

But there's room to do much more; say something like this mockup:

![Screenshot from 2021-07-18 12-22-26](https://user-images.githubusercontent.com/1134/126074767-bf4e0825-67fd-42c5-8600-ca287fb411e0.png)

Here's what `org-cite-supported-styles` returns:

``` elisp
((("author" "a")
  ("caps" "c"))
 (("noauthor" "na")
  ("bare" "b"))
 (("nocite" "n"))
 (("note" "ft")
  ("caps" "c")
  ("bare-caps" "bc"))
 (("numeric" "nb"))
 (("text" "t")
  ("caps" "c")
  ("bare-caps" "bc"))
 (("nil")
  ("caps" "c")
  ("bare-caps" "bc")
  ("bare" "b")))
```

Possibilities:

1. allow users to configure whether to use full style name or shortcuts
2. support sub-style variants somehow (but would need to be elegant and efficient)
3. add annotations/affixations to make it easier for users to know what a given choice will result in; but this would have to be adjustable by style category (author-date, note, etc.), or be a kind of live preview (which would need to rely on citeproc/CSL, so not sure how general that could be)

bdarcus:
@andras-simonyi - moving that sub-thread here, for consolidation.

https://github.com/andras-simonyi/citeproc-el/issues/15#issuecomment-882037958
https://github.com/andras-simonyi/citeproc-el/issues/15#issuecomment-882045167

For context, my question was:

>> I'm wanting to write a function that presents the list of (correct) styles, but also includes a preview of the result.... Do you know of a clever way to do this that would work across processors?

And @andras-simonyi's response:

> Depends on what you mean by "a preview of the result". Would the ensuing LaTeX citation command be an acceptable preview for the LaTeX processors? 

Yes; in fact, that's what I want.

I do see there's `org-cite-natbib--style-to-command`, but not an equivalent for `oc-biblatex`, and not a public function (though that's a minor issue, and I imagine if this is generally valuable, it could be added).

> For the CSL processor, on the other hand, it seems absolutely doable to provide a rendered preview as it is done by the CSL activation processor.

What's the best/easiest way to do that?

It doesn't have to be even close to perfect; it just has to allow the user to see what they are choosing in the completion UI.

andras-simonyi:
@bdarcus wrote
[...]

PR #182, opened 2021-07-18
oc: Add style select function

bdarcus:
Draft branch to Implement #168. This is RTM, but see my question below on whether this should be moved to a separate project.

I will make a decision today (July 22) or tomorrow, which may include merging this, with the current comment that I may remove it later in favor of the separate project. 

I have not tested the code over at `oc-blocks` yet, but if it doesn't work, it's very close, and simply adds the "insert processor" support from here, but makes it more configurable (bibtex-actions there is just currently the default option).

I think regardless, it may be better to keep the follow processor here.

## Status

Current status of code, with Emacs 28 `completing-read` grouping and annotation (here with natbib):

oc-styles

I ran into trouble when I tried to implement support for multiple previews. For now, I've decided to remove that, and just allow users to choose one option.

Also, the previews are not generated; I just have an alist with them, including the start for CSL, which I am manually creating.

## Ultimate Goals

I think ideally we want:

1. Possibility for **multiple preview targets** (user-configurable, with maybe default being automatically derived from the user org-cite export processor config); I, for example, have the biblatex and csl export processors configured, so probably want to see both of them.
2.  **Automatic style lists**; the actual style candidates should be extracted from `org-cite-supported-styles` (it's not currently, in part because I couldn't figure out elegant elisp to do it!).
3. **Previews on actual citation data**; related to 2, rather than using an alist, just use the export processors to generate the previews, ideally with the keys at point. 

That would make the code simpler and more general, and also more accurate, because it would account for details like fallback behavior.

OTOH, it may require adding back code since this may not work with `org-cite-make-insert-processor`.

## Separate project?

I am also wondering again if the insert processor(s) should be a separate, possibly collaborative, project.

I mean, I want to use completing-read with this, and to be able to plugin bibtex-actions for the "select-keys" functionality, but otherwise, it's conceptually independent of this project, which is really just a completing-read front-end to bibtex-completion.

Here's a test of what that could look like, where I have a `oc-blocks-insert-completion-function` defcustom for keys selection.

https://github.com/bdarcus/oc-blocks

Note: notwithstanding load-order problems with `org-cite-make-insert-processor`, I do think there are advantages to using it, and hope to see that enhanced if possible, as needed. But I think it's already pretty good, and can't see any obvious limitation.

## Other issues

Eye candy: what face should the preview face inherit from for best coherence with themes? For now, I use `minibuffer-prompt`.


bibtex-actions-org-cite.el
```diff
@@ -65,6 +65,23 @@
...
+"Complete a citation style for org-cite with KEYS preview.
+
+The previews include all configured export processors, with the
+default processor highlighted with the 'bibtex-actions-org-cite-styles-default' face."
+;; TODO
+)
+
+(defun bibtex-actions-csl-render-citation (citation)
+  "Render CITATION."
+  (let ((proc (your-cached-processor)))
```
bdarcus, line 78: @andras-simonyi - I added your start here. 

What is the cached-processor? I assume it's some `oc-csl` variable?

bdarcus:
@minad - do you have a minimal grouping example I could take a look at?

Google is not helpful here.

minad:
Emacs 28 master has a few group-functions if you grep.

bibtex-actions-org-cite.el
andras-simonyi: I meant the `citeproc-el` citation processor used for rendering the previews. It makes sense to use a cached one to avoid  reloading the used CSL style unnecessarily. It should be indeed the value of a variable (as in `org-cite-csl-activate`), and the safest option would be to use a buffer-local one which is dedicated to this task, i.e., rendering previews. Thinking about it a bit more, actually it could be shared with `org-cite-csl-activate`, but we probably don't want to engage in this kind of premature optimization at this stage :-).

bdarcus: > but we probably don't want to indulge in this kind of premature optimization at this stage

So bottomline, it's not shared with `oc-csl`?

Therefore I would create one here?

bdarcus:
> Emacs 28 master has a few group-functions if you grep.

I only found two. The xref one was the closest to what I need.

Where do I find the documentation for `group-function`; the meaning of the arguments?

bibtex-actions-org-cite.el
andras-simonyi: I think so. An even more lighweight option would be not using a full-fledged processor but just the style object and `citeproc-render-item`  (see the "Rendering isolated references" section in the README. Tomorrow I can post some code along these lines.

minad:
Transform could remove the prefix t/. This makes sense when the t is already part of the title such that the information is not duplicated.

See for example consult-grep once with vertico-group-format=nil and once non-nil.

bdarcus:
> Transform could remove the prefix t/. This makes sense when the t is already part of the title such that the information is not duplicated.
> 
> See for example consult-grep once with vertico-group-format=nil and once non-nil.

Can you please explain what's going on with this code from `consult-grep` `group-function`?

```elisp
(defun consult--grep-group (cand transform)
  "Return title for CAND or TRANSFORM the candidate."
  (if transform
      (substring cand (1+ (length (get-text-property 0 'consult--grep-file cand))))
    (get-text-property 0 'consult--grep-file cand)))
```

I think there's some small thing I'm not currently understanding; I guess specifically what "transform" means, and that argument.

What's going on, for example, with the `if transform` bit? How does that argument get set, so that `if` is `t` there?

This was my attempt here, but it's not working as I expect:

```elisp
(defun bibtex-actions--org-cite-styles-completing-read-group (cand transform)
  "Return group title of style CAND or TRANSFORM the candidate."
  (if transform
      ;; If there's a variant, remove it from the group title.
      ; so with input of "t/c", group is "t"
      (car (split-string cand "/")))
    cand)
```

minad:
The group function either transforms the candidate or returns a title if transform=nil. Your function above is wrong, since it returns the candy for transform=nil. It should return a title.

bdarcus:
> The group function either transforms the _candidate_ or returns a title if transform=nil. Your function above is wrong, since it returns the cand for transform=nil. It should return a title string.

OK, that makes sense. But how do I get them grouped?

I'm defining the data structure here, so I have flexibility, but currently I have an alist with examples like this:

```elisp
("text/f" . "\citet*")
("text/c" . "\Citet")
```

In org-cite, "text" (earlier I used the shortcut "t") is a "style", and "f" (shortcut for "full") is a variant.

I want to group by "style".

I will use the cdr for the preview annotation/affixation.

So this is what I currently get:

Screenshot from 2021-07-19 09-47-55

bdarcus:
I got it; nevermind.

bibtex-actions-org-cite.el
bdarcus: > I'd copy the solution in `org-cite-csl-activate`, and later think about whether more sharing/reuse can be achieved.

I haven't tested it (need to hook it to annotations, etc.), but should be done. Thanks!

minad:
What exactly do you want to achieve? I can take another look at this in a few days. I am away right now, so I cannot give you proper examples.

bdarcus:
Oh sorry; no rush.

But would look this, without the annotations. Not sure on the "default" part, but this is the general goal.

```
---- default -----------------------------------------------------------------------
/
/b
/c
/f
/bc
/bf
---- text ---------------------------------------------------------------------------
text
text/b
text/c
text/f
text/bc
text/bf
---- author ---------------------------------------------------------------------------
author
author/b
author/c
author/f
author/bc
author/bf
```

Ideally the group titles could be configurable, but that's not as important.

minad:
Something like this should work:

~~~ elisp
(defun my-group-fun (cand transform)
  (if transform
      cand
   (car (split-string cand "/"))))

bdarcus:
> Something like this should work:

It did; thank you!

Here's the final-ish function logic, which also handles the "default" grouping.

```elisp
      (if transform
          style
        (if (string-match "^/[bcf]*" style) "default" (car (split-string style "/")))))
```

bibtex-actions-org-cite.el
bdarcus: So @andras-simonyi, I have basic grouping annotation working, but in looking at the CSL angle I'm a little unsure.

I need to use the org-cite-make-insert function, where the insert-style argument to hook my function up. But that function is defined as taking 0 arguments. So I don't see how to get the citation to preview.

Maybe I can just select one at random?

andras-simonyi: I'm not sure about the details but my hunch is that the ideal solution would be work with the citation object at point, which you can get, I think, using `org-element-context`. (You might find it useful to look at  `org-cite-csl-activate--sensor-fun`, `element` is the citation object to be rendered in the code.) If you manage to get hold of the citation object, then you can probably generate from it all the required stylistic variants and render them.

bdarcus: > ... my hunch is that the ideal solution would be working with the citation object at point, which you can get, I think, using `org-element-context`. (You might find it useful to look at `org-cite-csl-activate--sensor-fun`, `element` is the citation object to be rendered in the code.) If you manage to get hold of the citation object then you can probably generate from it all the required stylistic variants and render them.

Right, but I can't see how to do that ATM; from _within_ the `SELECT-STYLE` function.

So let's review, in part so I can think this through while writing it.

The annotation function, where I am putting these previews, takes a candidate as input, and returns some (propertized) string.

The candidate is a style+variant string.

The `bibtex-actions-org-cite-select-style` function handles the UI, but cannot (at least ATM) take any arguments; so unlike some org-cite functions, it cannot take the citation as argument.

Here's the code that constructs the insert processor. 

```elisp
(org-cite-register-processor 'bibtex-actions-org-cite
           #'bibtex-actions-org-cite-select-style) ; this what I'm changing here, but this can't take any args
  :follow #'bibtex-actions-org-cite-follow)
```
To be clear, I don't have to use `org-cite-make-insert-processor`; it's just easier, and I'm trying to keep this simple as possible :-)

`Org-cite-insert`, which this is using, is context aware, but I'm not seeing how the `SELECT-STYLE` function can know about that.

So the issue is how to get the citation within that `bibtex-actions-org-cite-select-style` function.

If I can do that, not only will it be a better UX, but it will drastically reduce the amount of code here, because I can use the export processors for all the previews, rather than the alists I'm currently using, and treat the csl and tex export processors the same.

I was thinking about maybe having some cache of previews that is constantly updated, but not sure.

Or maybe I'm missing something obvious, which is why I asked.

cc  @jkitchin, in case he has any ideas, since i know in his experiments, he is not using `org-cite-make-insert-processor`, but also had to write a lot more code for this.

https://github.com/jkitchin/scimax/blob/3ce9942ddadb09176d0fae8db161cf3f5324766a/org-ref3/oc-bibtex.el#L336

bdarcus: PS - I posted a simplified version of this question to the list. I think that's the best solution, if I'm understanding correctly.

jkitchin: Here are a few reasons my current code doesn't use `org-cite-make-insert-processor`:

0. It feels complicated to me to use a function that takes two functions and then returns a function; I found it hard to debug, so I wrote out a full function I could debug.
 
1. Mostly it is because (I think) it is not compatible with the way ivy does sequential actions. This needs to call a function with the selected candidate as the arg. To solve that, I separated the insert into two functions, one that uses ivy to select, with an action that calls a second function that looks more like the  one created by `org-cite-make-insert-processor`. This is probably specific to ivy, and would not apply the same to helm, or probably other completion tools.  

2. Second, as @bdarcus discusses above, there are places where I need a new context, because the citation has changed since `org-insert-cite` was initially called, e.g. due to an insertion. I get the context inside my insert function on every call.

bdarcus: Thanks @jkitchin. Yes, I can see your 0 point.

I may end up going that path as well, though will see if Nicolas pops his head up and has a response.

On this:

> Mostly it is because (I think) it is not compatible with the way ivy does sequential actions. 

Was this because you wanted to enable moving and sorting of multi-cites?

I was actually wondering how to do that in just `completing-read`/`completing-read-multiple`, but hadn't thought about it deeply yet. It may be possible, but am not sure.

jkitchin: With ivy, the following options exist for sequential inserts (without closing the selection buffer), is you type C-M-m. That will call the default action with the current selection, but it does not close the selection buffer.  If you want an alternate action, you type C-M-o, select the action, press Ret, and then you go back to the selection buffer. There is an alternative way to mark candidates, and then the action is applied to each candidate in the marked list. Neither of these  approaches works well with the way `org-cite-insert` was working.

The moving/sorting is all done by a keymap that goes on in the activate function, or by functions that work at point in the follow hydra.

bdarcus: > With ivy, the following options exist for sequential inserts (without closing the selection buffer), is you type C-M-m. That will call the default action with the current selection, but it does not close the selection buffer.

Because ivy doesn't have support for multi-selection a la helm?

In bibtex-actions, I use `completing-read-multiple` for that.

> The moving/sorting is all done by a keymap that goes on in the activate function ...

I'm curious now; will take a look.

jkitchin: ivy does do something kind of like helm where you can mark a set of candidates, and then apply a function to each candidate. It isn't quite like helm though. In the beginning of ivy though, it was not like helm, and the ability to mark things came later.

bdarcus: > 1. I found it hard to debug, so I wrote out a full function I could debug.

This bit me hard yesterday. I think at minimum that function should have a more helpful error message, to make clear where the problem is.

I ran into trouble when I tried to implement support for multiple previews. For now, I've decided to remove that, and just allow users to choose one option.

Also, the previews are not generated; I just have an alist with them, including the start for CSL, which I am manually creating.

I will probably merge this as is, but I think ideally we want:

1. possibility for multiple preview targets (user-configurable, with maybe default being automatically derived from the user org-cite export processor config)
2.  the actual style candidates should be extracted from `org-cite-supported-styles` (it's not currently)
2. related, rather than using an alist, just use the export processors to generate the previews, ideally with the keys at point. That's less code and more general (works the same for tex and for csl).

I was also wondering again if the insert processor(s) should be a separate, possibly collaborative, project.

I mean, I want to use completing-read with this, and to be able to plugin bibtex-actions for the "select-keys" functionality, but otherwise, it's conceptually independent of this project, which is really just a completing-read front-end to bibtex-completion.

bdarcus:
Pending fixing the annoying `org-cite-make-insert-processor` bug (which I don't think is my fault), this is RTM.

I've updated the main post to reflect my current thinking on next steps, etc.

bibtex-actions-org-cite.el
bdarcus: > > 1. I found it hard to debug, so I wrote out a full function I could debug.
> 
> This bit me hard yesterday. I think at minimum that function should have a more helpful error message, to make clear where the problem is.

I fixed it by placing `org-cite-make-insert-processor` at end of the file.

bdarcus:
@minad an aside on `group-function`; shouldn't this be added to `completion-extra-properties` in minibuffer.el, to accompany the current affixation and annotation options?

PR #186, opened 2021-07-23
Add transient menu as default at-point

bdarcus:
Screenshot from 2021-07-23 07-43-10

I've been wanting to get embark which-key menus to do something like this, but it's not really designed to.

Edit: I have decided to remove mention of `which-key` in the repo, though I still need to update screenshots.

So this proposes to add a `transient.el` map, and have that the default at-point command, across the different configuration types, including with org-cite.

It has the advantage that it will work with a plain installation, without Embark, but is also probably better even with it.

Doesn't currently work, though; getting this error, which is probably because current code is designed around Embark:

```console
concat: Wrong type argument: stringp, citation-key
```

If I add a check for embark before stringifying the keys, I get farther, but this error:

```console
bibtex-actions-dwim: Wrong type argument: number-or-marker-p, bibtex-actions-open
```

If I look in debugger, seems close, but odd about the nil here:

```console
 funcall-interactively(bibtex-actions-open-link (nil "berlet2000"))
```

WDYT @ilupin - is there an easy fix to allow this?

Is yes, can build on it; the final menu will be more complete, and we can expand the available actions over time.

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -483,7 +483,9 @@ TEMPLATE."
   "Return citation keys at point as a list for `embark'."
   (when-let ((keys (or (bibtex-actions-get-key-org-cite)
                       (bibtex-completion-key-at-point))))
-    (cons 'citation-key (bibtex-actions--stringify-keys keys))))
+    (if (= bibtex-actions-default-action 'embark-act)
+        (cons 'citation-key (bibtex-actions--stringify-keys keys))
+      keys)))
 
 
 ;;; Command wrappers for bibtex-completion functions
@@ -579,7 +581,9 @@ With prefix, rebuild the cache before offering candidates."
 (defun bibtex-actions-run-default-action (keys)
   "Run the default action `bibtex-actions-default-action' on KEYS."
   (funcall bibtex-actions-default-action
-           (if (stringp keys) (split-string keys " & ") keys)))
+           (if (or (= bibtex-actions-default-action 'embark-act)
+                   (stringp keys))
+               (split-string keys " & ") keys)))
 
 ;;;###autoload
 (defun bibtex-actions-dwim ()
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -38,6 +38,7 @@
 ;;; Code:
 
 (require 'bibtex-completion)
+(require 'transient)
 
 (declare-function org-element-context "org-element")
 (declare-function org-element-property "org-element")
@@ -161,7 +162,7 @@ means no action."
   :type '(choice (const :tag "Prompt" 'prompt)
                  (const :tag "Ignore" nil)))
 
-(defcustom bibtex-actions-at-point-function 'bibtex-actions-dwim
+(defcustom bibtex-actions-at-point-function 'bibtex-actions-transient-menu
   "The function to run for 'bibtex-actions-at-point'."
   :group 'bibtex-actions
   :type 'function)
@@ -209,6 +210,18 @@ means no action."
     (define-key map (kbd "RET") '("cite| default action" . bibtex-actions-run-default-action))
     map)
   "Keymap for Embark citation-key actions.")
+
+(transient-define-prefix bibtex-actions-transient-menu ()
+  "Transient menu for bibtex-actions."
+  ["Reference(s)\n"
+   ["Open"
+    ("p" "PDF" bibtex-actions-open-pdf)
+    ("b" "BibTEX" bibtex-actions-open-entry)
+    ("l" "URL" bibtex-actions-open-link)
+    ("n" "Notes" bibtex-actions-open-notes)]
+
+   ["Library"
+    ("r" "Refresh" bibtex-actions-refresh)]])
 
 ;;; Org-cite citation function
 


ilupin:
I think `transient` is very suitable for bibtex-actions and I have also used transients to replace hydras in my configuration since it will be included in Emacs 28. However, it is not included by default in Emacs 27 and earlier. Do you really want `bibtex-actions` to depend on `transient` (i.e. `(require 'transient)`)?

> is there an easy fix to allow this?

The error doesn't seem a `transient` issue. `bibtex-actions-transient-menu` needs to be changed and should be able to pass the keys to bibtex-actions-* commands.

If `transient` becomes a dependency of `bibtex-actions`, I think it is possible to use only `transient` to provide contextual actions.

bdarcus:
> I think `transient` is very suitable for bibtex-actions and I have also used transients to replace hydras in my configuration since it will be included in Emacs 28. However, it is not included by default in Emacs 27 and earlier. Do you really want `bibtex-actions` to depend on `transient` (i.e. `(require 'transient)`)?

That's a good point I hadn't thought about.

I looked into transient after discussing this and hydra with John Kitchin over the past few days, who is working on an org-cite version of org-ref. Here was my experiment for his package.

Screenshot from 2021-07-23 04-45-43

Transient is available on ELPA for pre-28; right?

I guess this goes back to previous discussions about embark. There's no good way to include that transient map somehow without formally requiring it, using `fboudp` or `bound-and-true-p`??
 
> > is there an easy fix to allow this?
> 
> The error doesn't seem a `transient` issue. `bibtex-actions-transient-menu` needs to be changed and should be able to pass the keys to bibtex-actions-* commands.

Oh, you don't think the issue is with the surrounding functions?

> If `transient` becomes a dependency of `bibtex-actions`, I think it is possible to use only `transient` to provide contextual actions.

WDYT?

On one hand, maybe it's not a bad dependency to have. I certainly think the functionality could be nice to provide.

OTOH, might be nice to keep things minimal. There will be a version of org-cite with a transient and/or hydra menu, which will probably more functionality than here, so could always just leave that to him.

ilupin:
> Transient is available on ELPA for pre-28; right?

Yes.

> 
> I guess this goes back to previous discussions about embark. There's no good way to include that transient map somehow without formally requiring it (say using `fboundp` or `bound-and-true-p`?)??

You can wrap all transient related code with `with-eval-after-load`:
```
(with-eval-after-load 'transient
  (transient-define-prefix bibtex-actions-transient-menu ()
 ...
```

> Oh, you don't think the issue is with the surrounding functions?

The main issue is `bibtex-actions-` commands don't find the keys at point by themselves. Then the transient needs to find the keys and passes them to `bibtex-actions-`,  which will make the transient a bit complex.

I have created a working transient, but there are still other unclear transient issues.

> > If `transient` becomes a dependency of `bibtex-actions`, I think it is possible to use only `transient` to provide contextual actions.
> 
> WDYT?
> 
> On one hand, maybe it's not a bad dependency to have. I certainly think the functionality could be nice to provide.

I agree, but it is not in urgent need since `embark` already provides similar functionality. Specifically, I think `transient` is more flexible than embark because of infix arguments. However, `transient` cannot be used in the minibuffer ATM. `embark` is still needed if you want to access various actions in the minibuffer.

bdarcus:
> > Transient is available on ELPA for pre-28; right?
> 
> Yes.
> 
> > I guess this goes back to previous discussions about embark. There's no good way to include that transient map somehow without formally requiring it (say using `fboundp` or `bound-and-true-p`?)??
> 
> You can wrap all transient related code with `with-eval-after-load`:
> 
> ```
> (with-eval-after-load 'transient
>   (transient-define-prefix bibtex-actions-transient-menu ()
>  ...
> ```
> 
> > Oh, you don't think the issue is with the surrounding functions?
> 
> The main issue is `bibtex-actions-` commands don't find the keys at point by themselves. Then the transient needs to find the keys and passes them to `bibtex-actions-`, which will make the transient a bit complex.
> 
> I have created a working transient, but there are still other unclear transient issues.
> 
> > > If `transient` becomes a dependency of `bibtex-actions`, I think it is possible to use only `transient` to provide contextual actions.
> > 
> > 
> > WDYT?
> > On one hand, maybe it's not a bad dependency to have. I certainly think the functionality could be nice to provide.
> 
> I agree, but it is not in urgent need since `embark` already provides similar functionality. Specifically, I think `transient` is more flexible than embark because of infix arguments. However, `transient` cannot be used in the minibuffer ATM. `embark` is still needed if you want to access various actions in the minibuffer.

Alright, I'll close this for now then.

So then maybe I'll keep the embark requirement in `oc-bibtex-actions` (I was thinking about removing it), and we can basically say this provides the actions via embark, and for other options, look elsewhere?

Thanks!

bdarcus:
BTW, on the Doom discord channel, a few people have had positive things to say about his package:

https://gitlab.com/jjzmajic/hercules.el

It is basically like hydra for which-key, allowed group and such.

bdarcus:
If someone DOES want to do this at some point, it could go well in a `oc-bibtex-actions-follow-transient.el` file, so a user can do:

```elisp
(setq org-cite-follow-processor 'oc-bibtex-actions-follow-transient)
```

PR #193, opened 2021-07-26
Add face, modify existing face

bdarcus:
Add a primary candidate face, and a "highlight" one; the former would replace the previous "suffix" one.

Key question is just what parent faces to reference.

A few theme examples:

Screenshot from 2021-07-26 13-04-39

Screenshot from 2021-07-26 13-32-08

Screenshot from 2021-07-26 13-31-50

Thoughts?

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -64,7 +64,7 @@
   :group 'bibtex-actions)
 
 (defface bibtex-actions-highlight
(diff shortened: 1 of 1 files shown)


bdarcus:
I am adding this, but removing the highlight face parent for now, since I can't find an ideal one.

PR #197, opened 2021-07-27
oc: Refactor into separate files

bdarcus:
FYI, @andras-simonyi, I'm working on splitting the "insert processor", including the "select-style" function, into a separate file.


bdarcus:
I started to do this in #222. Will close this.

PR #204, opened 2021-08-01
Use correct oc-bibtex-actions package name in documentation

meliache:
The readme and use-package example in it mention `bibtex-actions-org-cite`, but it seems that the currently provided name is `oc-bibtex-actions`, so I fixed it.


bdarcus:
Thanks!

PR #208, opened 2021-08-03
support local bibs

### LINKED ISSUE #150: Add support for local bibs
> Sometimes, specially when working collaboratively in a paper, it is useful to have the .bib file of the current document (that sometimes is a local file and not your bib database) added to the bibtex-completion-bibliography.

_Originally posted by @ramos in https://github.com/bdarcus/bibtex-actions/discussions/148_

Not sure what's involved, or when I'll get to it.

Should work both for latex-mode, but also for other relevant options; certainly (the soon to be merged) org-cite. Pandoc?

A PR is welcome.

bdarcus:
This adds support for local bib files, by adding a buffer-local cache.

When visiting a buffer, you should get the correct mix of local and global bib files loaded automatically.

Also adds a hidden `is:local` string to local candidates, so that you can filter local items like so:

image

Open questions:

1. Should local candidates already included in the global cache be removed from the local cache?
2. ~Best options for auto-freshing of cache so users don't have to think about this? Maybe we include an [example auto-timer hook](https://github.com/bdarcus/bibtex-actions/pull/208#issuecomment-892589239) in the README?~ settled; working on this
3. Should there be some kind of UI hint for local candidates, building on the above filtering? We discuss below grouping as one option, but there are more subtle alternatives. Think, for example, how presence of related PDFs or notes is currently indicated in the affixation prefix (though if we go this way, we need a good default symbol). For example:

image

... or, different symbol:

image

Fix #150.

### DIFF
--- a/README.org
+++ b/README.org
@@ -222,61 +222,26 @@ You then have two ways to access these strings from the completion prompt:
 =Bibtex-actions= also preserves the history of your selections (see caveat below about multiple candidate selection though), which are also accessible in your completion UI, but by using =M-p=.
 You can save this history across sessions by adding =bibtex-actions-history= to =savehist-additional-variables=.
 
-** Pre-filtering entries
-    :PROPERTIES:
-    :CUSTOM_ID: prefiltering-entries
-    :END:
-
-By default, =bibtex-actions= will, assuming you are using =orderless= or =prescient= to filter candidates, pre-filter entries for the following commands.
-
-1. =bibtex-actions-open=: pre-narrows the list to those which have associated pdf or links
-2. =bibtex-actions-open-link=: pre-narrows the list to those which have associated links
-3. =bibtex-actions-open-pdf=: -pre-narrows the list to those which have associated pdf(s)
-
-That is, upon running the command, an =initial-input= value will be inserted to narrow the results. 
-You can also delete that if you prefer to see the full list of candidates.
-
-By default, pre-filtering of =bibtex-actions-open-notes= is off, because the command by default will create a new note if none is available, and therefore it makes sense to have access to your full library. 
-But you can customize this to pre-filter if you prefer.
-
-If you want to modify those values, or remove them entirely, you can set =bibtex-actions-initial-inputs= like so; in this case turning off pre-filtering for =bibtex-actions-open-pdf=:
-
-#+begin_src elisp
-(setq bibtex-actions-initial-inputs
-  '((pdf    . nil)
-    (note   . nil)
-    (link   . "has:link")
-    (source . "has:link\\|has:pdf"))
-#+end_src
-
 ** Refreshing the library display
     :PROPERTIES:
     :CUSTOM_ID: refreshing-the-library-display
     :END:
 
-Bibtex-actions uses a cache to speed up library display.
+=Bibtex-actions= uses two caches to speed up library display; one for the global bibliography, and another for local files specific to a buffer.
 This is great for performance, but means the data can become stale if you modify it.
 
-The =bibtex-actions-refresh= command will reload the cache, and you can call this manually. 
+The =bibtex-actions-refresh= command will reload the caches, and you can call this manually.
 You can also call any of the =bibtex-actions= commands with a prefix argument: =C-u M-x bibtex-actions-insert-key=.
 
-Finally, another option is to add =bibtex-completion=-style proactive loading externally by using =filenotify= something like this:
+Another option is to add a hook, something like this:
 
 #+BEGIN_SRC emacs-lisp
-  ;; Of course, you could also use `bibtex-completion-bibliography` here, but would need 
-  ;; to adapt this if you specify multiple files.
-  (file-notify-add-watch 
-    "/path/to/file.bib" '(change) 'bibtex-actions-refresh)
-#+END_SRC
+(defun gen-bib-cache-idle ()
+  "Generate bib item caches with idle timer"
+  (run-with-idle-timer 0.5 nil #'bibtex-actions-refresh))
 
-You can also extend this to do the same thing for your PDF files, or notes:
-
-#+BEGIN_SRC emacs-lisp
-  (file-notify-add-watch 
-    bibtex-completion-library-path '(change) 'bibtex-actions-refresh)
-
-  (file-notify-add-watch 
-    bibtex-completion-note-path '(change) 'bibtex-actions-refresh)
+(add-hook 'LaTeX-mode-hook #'gen-bib-cache-idle)
+(add-hook 'org-mode-hook #'gen-bib-cache-idle)
 #+END_SRC
 
 For additional configuration options on this, see [[https://github.com/bdarcus/bibtex-actions/wiki/Configuration#automating-path-watches][the wiki]].
--- a/README.org
+++ b/README.org
@@ -233,7 +233,7 @@ This is great for performance, but means the data can become stale if you modify
 The =bibtex-actions-refresh= command will reload the caches, and you can call this manually.
 You can also call any of the =bibtex-actions= commands with a prefix argument: =C-u M-x bibtex-actions-insert-key=.
 
-Another option is to add a hook, something like this:
+Another option to make the completion interface more seamless is to add a hook which generates the cache after a buffer is opened. This can be done when emacs has been idle (half a second in the example below) with something like this:
 
 #+BEGIN_SRC emacs-lisp
 (defun gen-bib-cache-idle ()
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -293,10 +293,10 @@ key associated with each one."
                     (s-join bibtex-actions-symbol-separator
                             (list pdf note link))"	") suffix))))
 
-(defvar bibtex-actions--candidates-cache nil
+(defvar bibtex-actions--candidates-cache 'uninitialized
   "Store the global candidates list.")
 
-(defvar-local bibtex-actions--local-candidates-cache nil
+(defvar-local bibtex-actions--local-candidates-cache 'uninitialized
   ;; We use defvar-local so can maintain per-buffer candidate caches.
   "Store the local (per-buffer) candidates list.")
 
@@ -305,12 +305,12 @@ key associated with each one."
 If the cache is nil, this will load the cache.
 If FORCE-REBUILD-CACHE is t, force reloading the cache."
   (when (or force-rebuild-cache
-            (not bibtex-actions--candidates-cache)
-            (not bibtex-actions--local-candidates-cache))
+            (eq 'uninitialized bibtex-actions--candidates-cache)
+            (eq 'uninitialized bibtex-actions--local-candidates-cache))
     (bibtex-actions-refresh force-rebuild-cache))
   (seq-concatenate 'list
-                   ((lambda (x) (if (eq x t) nil x)) bibtex-actions--local-candidates-cache)
-                   ((lambda (x) (if (eq x t) nil x)) bibtex-actions--candidates-cache)))
+                   bibtex-actions--local-candidates-cache
+                   bibtex-actions--candidates-cache))
 
 ;;;###autoload
 (defun bibtex-actions-refresh (&optional force-rebuild-cache)
@@ -324,10 +324,10 @@ is non-nil, also run the `bibtex-actions-before-refresh-hook' hook."
   ;; itself is used to indicate a cache that hasn't been computed this is to
   ;; work around the fact that we can't differentiate between empty list and nil
   (setq bibtex-actions--candidates-cache
-        ((lambda (x) (if x x t)) (bibtex-actions--format-candidates)))
+        (bibtex-actions--format-candidates))
   (let ((bibtex-completion-bibliography (bibtex-actions--local-files-to-cache)))
     (setq bibtex-actions--local-candidates-cache
-        ((lambda (x) (if x x t)) (bibtex-actions--format-candidates "is:local")))))
+        (bibtex-actions--format-candidates "is:local"))))
 
 ;;;###autoload
 (defun bibtex-actions-insert-preset ()
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -207,8 +207,26 @@ offering the selection candidates"
              collect (cdr (or (assoc choice candidates)
                               (rassoc choice candidates))))))
 
-(defun bibtex-actions--format-candidates ()
-  "Transform candidates from 'bibtex-completion-candidates'.
+(defun bibtex-actions--normalize-paths (file-paths)
+  "Return a list of FILE-PATHS normalized with truename."
+  (if (stringp file-paths)
+      ;; If path is a string, return as a list.
+      (list (file-truename file-paths))
+    (delete-dups (mapcar (lambda (p) (file-truename p)) file-paths))))
+
+(defun bibtex-actions--local-files-to-cache ()
+  "The local bibliographic files not included in the global bibliography."
+  ;; We cache these locally to the buffer.
+  (let* ((local-bib-files
+          (bibtex-actions--normalize-paths
+           (bibtex-completion-find-local-bibliography)))
+         (global-bib-files
+          (bibtex-actions--normalize-paths
+           bibtex-completion-bibliography)))
+    (seq-difference local-bib-files global-bib-files)))
+
+(defun bibtex-actions--format-candidates (&optional context)
+  "Format candidates, with optional hidden CONTEXT metadata.
 This both propertizes the candidates for display, and grabs the
 key associated with each one."
   (let* ((main-template
@@ -240,7 +258,7 @@ key associated with each one."
             ;; We display this content already using symbols; here we add back
             ;; text to allow it to be searched, and citekey to ensure uniqueness
             ;; of the candidate.
-            (candidate-hidden (s-join " " (list pdf note link citekey))))
+            (candidate-hidden (s-join " " (list pdf note link context citekey))))
        (cons
         ;; If we don't trim the trailing whitespace, 'completing-read-multiple' will
         ;; get confused when there are multiple selected candidates.
@@ -276,16 +294,23 @@ key associated with each one."
                             (list pdf note link))"	") suffix))))
 
 (defvar bibtex-actions--candidates-cache nil
-  "Store the candidates list.")
+  "Store the global candidates list.")
+
+(defvar-local bibtex-actions--local-candidates-cache nil
+  ;; We use defvar-local so can maintain per-buffer candidate caches.
+  "Store the local (per-buffer) candidates list.")
 
 (defun bibtex-actions--get-candidates (&optional force-rebuild-cache)
   "Get the cached candidates.
 If the cache is nil, this will load the cache.
 If FORCE-REBUILD-CACHE is t, force reloading the cache."
   (when (or force-rebuild-cache
-            (not bibtex-actions--candidates-cache))
+            (not bibtex-actions--candidates-cache)
+            (not bibtex-actions--local-candidates-cache))
     (bibtex-actions-refresh force-rebuild-cache))
-  bibtex-actions--candidates-cache)
+  (seq-concatenate 'list
+                   ((lambda (x) (if (eq x t) nil x)) bibtex-actions--local-candidates-cache)
+                   ((lambda (x) (if (eq x t) nil x)) bibtex-actions--candidates-cache)))
 
 ;;;###autoload
 (defun bibtex-actions-refresh (&optional force-rebuild-cache)
@@ -295,8 +320,14 @@ is non-nil, also run the `bibtex-actions-before-refresh-hook' hook."
   (interactive "P")
   (when force-rebuild-cache
     (run-hooks 'bibtex-actions-force-refresh-hook))
+  ;; t in the result indicates that cache was computed but came out nil nil
+  ;; itself is used to indicate a cache that hasn't been computed this is to
+  ;; work around the fact that we can't differentiate between empty list and nil
   (setq bibtex-actions--candidates-cache
-        (bibtex-actions--format-candidates)))
+        ((lambda (x) (if x x t)) (bibtex-actions--format-candidates)))
+  (let ((bibtex-completion-bibliography (bibtex-actions--local-files-to-cache)))
+    (setq bibtex-actions--local-candidates-cache
+        ((lambda (x) (if x x t)) (bibtex-actions--format-candidates "is:local")))))
 
 ;;;###autoload
 (defun bibtex-actions-insert-preset ()


bdarcus:
Do we agree at this point the solution to getting the right files is solved, and the remaining piece is just how to make sure the candidates are appropriately in sync with whatever document we're working on?

For you, that means the candidate list will change, and so necessarily the cache refresh, depending on which document you have open; for me, usually not.

> But no with a buffer local cache, which seems ideal to me. 

If you still think this is an idea you want to champion, how would the buffer local cache work? Does that mean a separate candidates list? What makes it "ideal", with what cost, if any?

> I can also see it being annoying for some one with global bib file but it should be about the same as refreshing it one opening a new buffer in terms of time spent doing it.

Ideally, none of us find anything about the ultimate solution "annoying"; it should be elegant, and we basically shouldn't have to think about it. Oh, and easy to maintain.

That's the goal anyway.

bdarcus:
In any case, can I delegate this piece to you?

Could be input from the doom discord might be valuable here?

bdarcus:
Idea: rather than cache candidates, cache bib file names.

We know we need to reset the cache if the file list of the current buffer differs from the file list cache.

Just a question of how to initiate the check automatically.

Or is that what you have in mind?

aikrahguzar:
> Do we agree at this point the solution to getting the right files is solved, and the remaining piece is just how to make sure the candidates are appropriately in sync with whatever document we're working on?

Yes

> If you still think this is an idea you want to champion, how would the buffer local cache work? Does that mean a separate candidates list? What makes it "ideal", with what cost, if any?

I have been using a buffer local cache (i.e with cache variable defined by `defvar-local` instead of `defvar`) in my config. The reason this is ideal to me that I can work with latex documents with different bib files have the correct cache by construction. The only downside I can think of is memory usage because the candidate list is unique to the buffer. For me the effect has been negligible with 2-3 latex-mode buffers with bib files containing about 1500 entries. On the positive it does save time spent on refreshing the cache and we don't have to worry about refreshing it when the focused document changes. 

I have a hook that run the refresh command in my `latex-mode-hook` and it works pretty well 
 
> Ideally, none of us find anything about the ultimate solution "annoying"; it should be elegant, and we basically shouldn't have to think about it. Oh, and easy to understand and maintain.
> 
> That's the goal anyway.

To me the `defvar-local` seems the simplest in terms of maintainability but then I have never really maintained software used by other than me.

aikrahguzar:
> Alternately, we can just not worry about this part. As you note, one can already refresh cache manually (with a prefix arg, etc).

This would be my second preferred solution.

aikrahguzar:
And asking on doom discord seems like a good thing. While trying to familiarize myself with how bibtex-completion works under the hook, I saw that `reftex` which is used under the hood has a function for generating a new bib file consisting of all the entries cited in the document from a global bib file. Some people might have a workflow built around that.

bdarcus:
> I have been using a buffer local cache (i.e with cache variable defined by defvar-local instead of defvar) in my config. The reason this is ideal to me that I can work with latex documents with different bib files have the correct cache by construction. 

But what's the practical difference between the global cache that is refreshed vs a local one? Isn't it the same UX in the end?

If we do add this, how would we integrate it with this PR? I assume in `bibtex-actions--get-candidates`?

aikrahguzar:
> But what's the practical difference between the global cache that is refreshed vs a local one? Isn't it the same UX in the end?
> 
> If we do add this, how would we integrate it with this PR? I assume in `bibtex-actions--get-candidates`?

The only change needed would be to change
```lisp
(defvar bibtex-actions--candidates-cache nil)
``` 

to 

```lisp
(defvar-local bibtex-actions--candidates-cache nil)
```

I don't think there would be any practical difference in ux but it would be much less code.

bdarcus:
> I don't think there would be any practical difference in ux but it would be much less code.

I can make the change here and test it.

How would it impact the case of the user who doesn't use local bibs; only global?

aikrahguzar:
> How would it impact the case of the user who doesn't use local bibs; only global?

From how the cache is computed in this pr, I think the only effect would be that would get the same cache duplicated in different buffers. So higher memory usage is all I can see.

But even for them compared to global but refreshed cache it would smoother time wise if they switch buffers frequently since the cache wouldn't be refreshed every time.

bdarcus:
> > How would it impact the case of the user who doesn't use local bibs; only global?
> 
> From how the cache is computed in this pr, I think the only effect would be that would get the same cache duplicated in different buffers. So higher memory usage is all I can see.

Higher memory usage, and a brief pause when changing buffers. Could be an issue if one has a very large global bibliography?

> But even for them compared to global but refreshed cache it would smoother time wise if they switch buffers frequently since the cache wouldn't be refreshed every time.

OIC. You're saying the local caches would be maintained across the buffers?

If yes, I get it now.

But the above issues seems like something worth thinking about.

aikrahguzar:
> Higher memory usage, and a brief pause when changing buffers. Could be an issue if one has a very large global bibliography?

There shouldn't any pause when changing a buffer. 
 
> OIC. You're saying the local caches would be maintained across the buffers?

Yes

> If yes, I get it now.

I am using this and it seems to be working great for me.

bdarcus:
> There shouldn't any pause when changing a buffer.

Say I'm working on a book, with a single large global bibliography.

I have two buffers open, representing two chapters.

Would not they each get their own buffer-local cache, created upon opening those buffers (though not when changing between them)?

Maybe I can just merge this change and document it, in case we need to address this later?

bdarcus:
> Maybe I can just merge this change and document it, in case we need to address this later?

No; I think we need to figure this out.

Consider someone using org-roam, where they have many buffers opening.

aikrahguzar:
> No; I think we need to figure this out.
> 
> Consider someone using org-roam, where they have many buffers opening.

One solution would be to use a global cache containing the entries from global file only and one buffer local cache containing the local bib only and concat them before offering candidates. This will work if concatenation is fast. But the complication would be that we can no longer simply use`bibtex-completion-find-local-bibliography` since it return both local and global files for `org`.

bdarcus:
> I ran (memory-report-object-size bibtex-actions--candidates-cache) on a buffer with almost 1500 entries, I see 723275 bytes ~ 0.7 MiB. On another buffer with about 300 entries I see 155518 bytes ~ 0.15 MiB which doesn't seem too bad.

But if, as with the org-roam case, there are 20 buffers open, with a 2000 item global bib?

If this is worth addressing, it seems that would suggest your original two caches idea, which I didn't understand earlier:

```elisp
(defvar bibtex-actions--global-candidates-cache nil)
(defvar-local bibtex-actions--local-candidates-cache nil)
```

We then only use the later when it's needed; e.g. when the local bib list differs from the global one.

Right?

> This will happen (assuming a hook runs then). I have one on in my latex-mode-hook that runs with an idle timer so actual bibtex-actions ui always seems fast. This is the hook ...

We should probably add this to the README regardless.

aikrahguzar:
> But if, as with the org-roam case, there are 20 buffers open, with a 2000 item global bib?

Even 20MiB doesn't seem too bad to me. 
 
> If this is worth addressing, it seems that would suggest your original two caches idea, which I didn't understand earlier:
> 
> ```emacs-lisp
> (defvar bibtex-actions--global-candidates-cache nil)
> (defvar-local bibtex-actions--local-candidates-cache nil)
> ```
> We then only use the later when it's needed; e.g. when the local bib list differs from the global one.
> 
> Right?

This does seem worth doing and I will try to get something like this working over the weekend.

And my idea earlier was wound up in too many details and without a good ux but I think we have settled on a good one here.


> We should probably add this to the README regardless.

Please go ahead.

bdarcus:
> This does seem worth doing and I will try to get something like this working over the weekend.

:+1: 

> We then only use the later when it's needed; e.g. when the local bib list differs from the global one.

Something like this on `bibtex-actions--get-candidates`:

```elisp
  ;; code to determine which cache to use; if nil, use local
  (let ((global-cache (bibtex-actions--use-global-cache))) ; return t if local and global file lists are the same
```

aikrahguzar:
Question: What is our working definition of global bib files? The simplest answer seem to be whatever is in `bibtex-completion-bibliography` and I am inclined to work with it for the time being at least. But I don't know what if any is the  `org`'s idea of global bib files.

bdarcus:
> Question: What is our working definition of global bib files? The simplest answer seem to be whatever is in `bibtex-completion-bibliography` and I am inclined to work with it for the time being at least. But I don't know what if any is the `org`'s idea of global bib files.

Exactly; here's what I was playing with:

```elisp
(defun bibtex-actions--normalize-paths (file-paths)
  "Return a list of FILE-PATHS sorted and normalized with truename."
  (cl-sort
   (delete-dups
    (mapcar (lambda (p) (file-truename p)) file-paths)) 'string-lessp))

(defun bibtex-actions--use-global-cache ()
  "When non-nil, use the global cache."
  (let* ((local-bib-files
          (bibtex-actions--normalize-paths (bibtex-completion-find-local-bibliography)))
         (global-bib-files
          (bibtex-actions--normalize-paths bibtex-completion-bibliography)))
    (equal local-bib-files global-bib-files)))
```

aikrahguzar:
What I was thinking of was actually slightly different, we should always use both caches but concatenate them when offering candidates to the user. The global caches is populated with candidates from global files, while the local cache contains the differences of local files and global files, i.e morally 
```lisp
(cl-set-difference local-files global-files)
```
This seems ideal in terms of memory usage. If there are 20 buffers each with the same big global file but small local files the big global file is present in a single cache. 

This is assuming that while constructing the cache most of the time is spent formatting and the concatenation is quick.

I think we should take the same approach with the latex too. if the user has a declared a global file they probably want to use it and `reftex` can be used to extract a locally used entries from the global file and we should put in the README explaining how to do that.

bdarcus:
I'm agnostic on the details.

Whatever is cleanest (including easiest to understand), and works the best.

bdarcus:
Do note, though, that I'm pretty sure bibtex-completion normalizes across
bib files, removing duplicates and such.

bdarcus:
> the local cache contains the differences of local files and global files,

So more like this (cl-set-difference doesn't work as I expect here):

```elisp
(defun bibtex-actions--local-files-to-cache ()
  "The local bibliographic files not included in the global bibliography."
  (let* ((local-bib-files
          (bibtex-actions--normalize-paths (bibtex-completion-find-local-bibliography)))
         (global-bib-files
          (bibtex-actions--normalize-paths bibtex-completion-bibliography)))
    (seq-difference local-bib-files global-bib-files)))
```

aikrahguzar:
> (cl-set-difference local-files global-files)
> Seems we want seq-difference though? I'm agnostic on the details. Whatever is cleanest (including easiest to understand), and works the best.

I don't know emacs api that well and don't understand the differences between data structures manipulation modules, I picked up `cl-set-difference` because it seemed to do what we want and there is a lot of `cl-` stuff in the file.

aikrahguzar:
> Do note, though, that I think bibtex-completion normalizes across bib files, removing duplicates and such. Not sure it matters, but ...

Duplication is definitely a concern, from going through the function, it seems like it does deduplications by comparing hashes derived from bib data. We should be still able deduplicate by first constructing the global cache and checking against it in addition while constructing the local cache. Though I think this should be a second order concern

bdarcus:
> > (cl-set-difference local-files global-files)
> > Seems we want seq-difference though? I'm agnostic on the details. Whatever is cleanest (including easiest to understand), and works the best.
> 
> I don't know emacs api that well and don't understand the differences between data structures manipulation modules, I picked up `cl-set-difference` because it seemed to do what we want and there is a lot of `cl-` stuff in the file.

I don't really understand these well either, but we want the first result; right?

```elisp
ELISP> (seq-difference '("local-1.bib" "global-1.bib") '("global-1.bib" "global-2.bib"))
("local-1.bib")

ELISP> (cl-set-difference '("local-1.bib" "global-1.bib") '("global-1.bib" "global-2.bib"))
("local-1.bib" "global-1.bib")
```

aikrahguzar:
> You'd want to confirm bibtex-completion doesn't already have some feature for this. I actually don't know.

I did look through the symbols in bibtex-completion and didn't find anything.

BTW, the groups in vertico are very nice, I just had to scroll down to the function heading. I was actually thinking that if we want to emphasize what is local and global we can show them in different groups and then we don't need to worry about deduplication

aikrahguzar:
> > > (cl-set-difference local-files global-files)
> > > Seems we want seq-difference though? I'm agnostic on the details. Whatever is cleanest (including easiest to understand), and works the best.
> > 
> > 
> > I don't know emacs api that well and don't understand the differences between data structures manipulation modules, I picked up `cl-set-difference` because it seemed to do what we want and there is a lot of `cl-` stuff in the file.
> 
> I don't really understand these well either, but we want the first result; right?
> 
> ```emacs-lisp
> ELISP> (seq-difference '("local-1.bib" "global-1.bib") '("global-1.bib" "global-2.bib"))
> ("local-1.bib")
> 
> ELISP> (cl-set-difference '("local-1.bib" "global-1.bib") '("global-1.bib" "global-2.bib"))
> ("local-1.bib" "global-1.bib")
> ```

I should have tried evaluating them but this is a very strange result from just reading the documentation

bdarcus:
> BTW, the groups in vertico are very nice, I just had to scroll down to the function heading. I was actually thinking that if we want to emphasize what is local and global we can show them in different groups and then we don't need to worry about deduplication

You're talking about the Emacs 28 `group-function`?

Yes, it is, and I hadn't thought about that. But it makes sense?

And I've recently been playing with the feature for org-cite; here's the style selection UI.

Screenshot from 2021-07-20 07-23-17

aikrahguzar:
> You're talking about the Emacs 28 `group-function`?
> 
> Yes, it is, and I hadn't thought about that. But it makes sense?
> 
> And I've recently been playing with the feature for org-cite; here's the style selection UI.
> 
> Screenshot from 2021-07-20 07-23-17

Exactly this. I got sidetracked in my though but I was going to say that it can work but I am not sure if the global vs local distinction is worth emphasizing.

bdarcus:
> I am not sure if the global vs local distinction is worth emphasizing. But I can also see emphasizing it make the behavior less surprising.

Neither am I. I lean towards thinking it's a good idea though.

I'd need to think about whether and how it would impact the implementation though. I _think_ it would require including some kind of flag (string; for sake of argument `:local:`) in the candidate data, that could be hidden, which designated which group it was a part of, and then using that in the group-function.

But it's a minor thing, I think; we can worry about that later.

bdarcus:
So trying to figure out how to hook this up ...

Right now, the main format function iterates through `bibtex-completion-candidates` to create the candidate list.

By definition, that I think will use the value of `bibtex-completion-bibliography`.

So therefore I _think_ the appropriate approach would be to run it twice, let-binding that variable to specify which files to use.

The first is just as is now; with the variable value.

The second would be the subset for the local cache.

bibtex-actions.el
```diff
@@ -367,7 +367,12 @@ is non-nil, also run the `bibtex-actions-before-refresh-hook' hook."
   (when force-rebuild-cache
     (run-hooks 'bibtex-actions-force-refresh-hook))
   (setq bibtex-actions--candidates-cache
-        (bibtex-actions--format-candidates)))
+        (bibtex-actions--format-candidates
+         (bibtex-actions--normalize-paths
```
bdarcus, line 371: This isn't necessary here.

aikrahguzar:
> So trying to figure out how to hook this up ...
> 
> Right now, the main format function iterates through `bibtex-completion-candidates` to create the candidate list.
> 
> By definition, that I think will use the value of `bibtex-completion-bibliography`.
> 
> So therefore I _think_ the appropriate approach would be to run it twice, let-binding that variable to specify which files to use; if there's any local bib files.

I think one way to do this if we set aside deduplication across local and global caches would be to remove the let binding of  `bibtex-completion-bibliography` from `bibtex-actions--format-candidates` and instead use 
```lisp
(defun bibtex-actions-refresh (&optional force-rebuild-cache)
  "Reload the candidates cache.
If called interactively with a prefix or if FORCE-REBUILD-CACHE
is non-nil, also run the `bibtex-actions-before-refresh-hook' hook."
  (interactive "P")
  (when force-rebuild-cache
        (bibtex-actions--format-candidates))
  (let (bibtex-completion-bibliography bibtex-actions--local-files-to-cache)
    (setq bibtex-actions--local-candidates-cache
        (bibtex-actions--format-candidates))))
```
Deduplication can then be done if we write the function by applying it before the second `setq`.

aikrahguzar:
> ```emacs-lisp
> (cl-remove-duplicates '(("one" . 1)("two" . 2) ("ONE" . 1)) :test (lambda (x y) (equal (cdr x)(cdr y))))
> ```

Is it fine to assume that keys are unique across all the files? It seems like a sensible thing to me but bibtex-completion instead seems to compare the hashes derived from the entry (I am not completely sure what is happening in `bibtex-completion-candidates`).

aikrahguzar:
I think this should work up to deduplication across the caches,

```lisp
(defun bibtex-actions--get-candidates (&optional force-rebuild-cache)
  "Get the cached candidates.
If the cache is nil, this will load the cache.
If FORCE-REBUILD-CACHE is t, force reloading the cache."
  (when (or force-rebuild-cache
            (not bibtex-actions--candidates-cache)
            (not bibtex-actions--local-candidates-cache))
    (bibtex-actions-refresh force-rebuild-cache))
  (seq-concatenate 'list bibtex-actions--local-candidates-cache bibtex-actions--candidates-cache))

;; t in the result indicates that cache was computed but came out nil
;; nil itself is used to indicate a cache that hasn't been computed
;; this is to work around the fact that we can't differentiate between empty list and nil
(defun bibtex-actions-refresh (&optional force-rebuild-cache)
  "Reload the candidates cache.
If called interactively with a prefix or if FORCE-REBUILD-CACHE
is non-nil, also run the `bibtex-actions-before-refresh-hook' hook."
  (interactive "P")
  (when force-rebuild-cache
        ((lambda (x) (if x x t)) (bibtex-actions--format-candidates)))
  (let (bibtex-completion-bibliography (bibtex-actions--local-files-to-cache))
    (setq bibtex-actions--local-candidates-cache
        ((lambda (x) (if x x t)) (bibtex-actions--format-candidates)))))
```

bdarcus:
> I think one way to do this if we set aside deduplication across local and global caches would be to remove the let binding of `bibtex-completion-bibliography` from `bibtex-actions--format-candidates` and instead use

Right; that's easier.

I reverted that earlier change and updated this.

Also simplified how attaching the hidden metadata works (just pass the string).

> I think this should work up to deduplication across the caches,

Added.

I haven't tested it yet, so that's the next step for us.

Oh, and will have to document on the README, including cache updating.

BTW, see [this doc](https://docs.github.com/en/github/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#adding-line-comments-to-a-pull-request) for info on line comments, including "suggested changes," which are super convenient.

bdarcus:
> I haven't tested it yet, so that's the next step for us.

Dedup isn't working correctly yet :-)

image

BTW, WDYT about the `group-function` now?

bibtex-actions.el
```diff
@@ -235,11 +235,33 @@ offering the selection candidates"
...
+(defun bibtex-actions--local-files-to-cache ()
+  "The local bibliographic files not included in the global bibliography."
+  ;; We cache these locally to the buffer.
+  (let* ((local-bib-files
+          (bibtex-actions--normalize-paths
+           (bibtex-completion-find-local-bibliography)))
+         (global-bib-files
+          (bibtex-actions--normalize-paths
+           bibtex-completion-bibliography)))
+    (seq-difference local-bib-files global-bib-files)))
```
bdarcus, line 254: I think this will need adjustment for org-cite.

bibtex-actions.el
```diff
@@ -341,8 +370,14 @@ is non-nil, also run the `bibtex-actions-before-refresh-hook' hook."
...
   (when force-rebuild-cache
     (run-hooks 'bibtex-actions-force-refresh-hook))
+  ;; t in the result indicates that cache was computed but came out nil nil
+  ;; itself is used to indicate a cache that hasn't been computed this is to
+  ;; work around the fact that we can't differentiate between empty list and nil
   (setq bibtex-actions--candidates-cache
-        (bibtex-actions--format-candidates)))
+        ((lambda (x) (if x x t)) (bibtex-actions--format-candidates)))
+  (let (bibtex-completion-bibliography (bibtex-actions--local-files-to-cache))
+    (setq bibtex-actions--local-candidates-cache
```
bdarcus, line 379: This is wrong, as that's not a variable.

aikrahguzar: Yes, I missed a pair of parenthesis here

aikrahguzar:
> 
> Dedup, and/or the logic for the local cache for org-cite at least, isn't working correctly yet (the TEST record is local, BTW).
> 
> image

There are two problems
1) `bibtex-completion-bibliography` is still let bound in `bibtex-actions--format-candidates` so the two variables are duplicate copies of each other.

2) I don't know how to let bound :) . Let me check the link you sent and add the missing parenthesis. You go there before me.

Also I think I wasn't clear  but right now there is no dedup if the local and global files contain the same entries. This works up to that.

> BTW, WDYT about the `group-function` now?

bdarcus:
Latest push fixes a syntax error on the let-binding.

image

bibtex-actions.el
```diff
@@ -235,11 +235,33 @@ offering the selection candidates"
...
+          (bibtex-actions--normalize-paths
+           bibtex-completion-bibliography)))
+    (seq-difference local-bib-files global-bib-files)))
+
+(defun bibtex-actions--format-candidates (&optional context)
+  "Format candidates, with optional hidden CONTEXT metadata.
 This both propertizes the candidates for display, and grabs the
 key associated with each one."
-  (let* ((main-template
+  (let* ((bibtex-completion-bibliography
```
aikrahguzar, line 260: We also need to remove the let-binding on `bibtex-completion-bibliography` here. SIce it will override the binding in refresh function.

bdarcus: Done.

bdarcus:
OK, it looks like it works as expected now, including in org-cite.

Let's test a bit more and think about what we want to add for documentation?

aikrahguzar:
> OK, it looks like it works as expected now, including in org-cite.

Working well in latex too. At least for me. The remaining question is if we can remove entries from local cache which are already present in global one. This might be involved and I need to understand exactly how `bibtex-completions` does it if we want to do that. 

> Let's test a bit more and think about what we want to add for documentation?
:+1:

bdarcus:
> The remaining question is if we can remove entries from local cache which are already present in global one. This might be involved and I need to understand exactly how bibtex-completions does it if we want to do that.

Could be bibtex-completion had a convenience function for this?

aikrahguzar:
> Could be bibtex-completion had a convenience function for this?

I don't see an obvious candidate but it is worth asking

bdarcus:
One wrinkle is the data I add to the "local" candidates means they will be unique across the caches at that point.

So if removed, would probably want to remove before running the formatting function.

bdarcus:
Last change for the day, I think:

I separately culled some outdated org-cite code no longer needed, and rebased everything here on top of that, and merged it all together into one commit.

bdarcus:
BTW, on your hook, can you just include the code you include run the hook?

Maybe we can include an example that works for all three primary modes: latex, org, markdown?

aikrahguzar:
> BTW, on your hook, can you just include the code you use to run the hook?

It is this
```lisp
(add-hook! 'LaTeX-mode-hook #'outline-minor-mode #'outline-hide-body #'auto-fill-mode #'gen-bib-cache-idle)
```

Though `add-hook!` is a doom convenience.

bdarcus:
OK, updated the README with the basics, and the hook example. 

Does it need anything more?

README.org
```diff
@@ -222,61 +222,26 @@ You then have two ways to access these strings from the completion prompt:
...
-    "/path/to/file.bib" '(change) 'bibtex-actions-refresh)
-#+END_SRC
-
-You can also extend this to do the same thing for your PDF files, or notes:
+Another option is to add a hook, something like this:
 
 #+BEGIN_SRC emacs-lisp
-  (file-notify-add-watch 
-    bibtex-completion-library-path '(change) 'bibtex-actions-refresh)
+(defun gen-bib-cache-idle ()
```
aikrahguzar, line 239: This hook won't help with stale cache, I include so that a cache is generated when a new buffer is opened making the ui seem more responsive.

For stale cache `filenotify`is the solution though it won't work with local cache as is.

bdarcus: > For stale cache `filenotify`is the solution though it won't work with local cache as is.

As in, they conflict, or (as I assume) they are complementary? If the latter, I'll add it back and adjust the language.

aikrahguzar: They are complemtary. I think another hook will be needed that adds a watch on the files returned by `bibtex-completion-find-local-bibliography` to makes it work with local cache.

bdarcus: Can you just use "suggested changes" to suggest the specific line changes then? Just click the left most icon on the toolbar.
 
 Allows me to just click to accept, and then you get credit in the git history.

README.org
```diff
@@ -222,61 +222,26 @@ You then have two ways to access these strings from the completion prompt:
...
-
-#+BEGIN_SRC emacs-lisp
-  ;; Of course, you could also use `bibtex-completion-bibliography` here, but would need 
-  ;; to adapt this if you specify multiple files.
-  (file-notify-add-watch 
-    "/path/to/file.bib" '(change) 'bibtex-actions-refresh)
-#+END_SRC
-
-You can also extend this to do the same thing for your PDF files, or notes:
+Another option is to add a hook, something like this:
```
aikrahguzar, line 236:
```suggestion
Another option to make the completion interface more seamless is to add a hook which generates the cache after a buffer is opened. This can be done when emacs is idle with something like this:
```

README.org
aikrahguzar: I adjusted the language, I will have to look into filenotify a bit to write and test the hook for cache update.

localauthor:
I just gave the local-bibs branch a shot and figured I'd report that all worked for me as expected, straightaway. All it took was adding a local bib file to the `#+bibliography:` header argument and refreshing the setup (`C-c C-c` with point on header arguments). And I made sure the local bib had some entries overlapping with the global bib and some unique entries. Results:

- After calling `org-cite-insert`, all local and global candidates were listed together, without any duplicates.

- After removing the local `#+bibliography:` header argument and refreshing the setup, the cites from the local bib are marked as no longer valid (face changes to `error`), while those from the global bib, including the cites listed in both bibs, remain valid.

- Added entry to the local bib, refreshed setup (`C-c C-c` on the header args), and found the new entry upon calling `org-cite-insert`. (Surprisingly, it wasn't necessary to call `bibtex-actions-refresh`. Is that expected?) Same in reverse, for removing entries.

- Exports generate as expected, no missing item data.

Any other testing you're interested in while I'm on this branch? Seems to be in order, from here. In any case, thanks for your work on this package!

bdarcus:
> I just gave the local-bibs branch a shot and figured I'd report that all worked for me as expected, straightaway. 

Great; thanks for confirming!

> * Added entry to the local bib, refreshed setup (`C-c C-c` on the header args), and found the new entry upon calling `org-cite-insert`. (Surprisingly, it wasn't necessary to call `bibtex-actions-refresh`. Is that expected?) 

I'm not sure actually. I think it will create the local cached data the first time you run a bibtex-actions command on a "new" buffer with local bib. Does that explain it?

> Any other testing you're interested in while I'm on this branch? Seems to be in order, from here. In any case, thanks for your work on this package!

Do you have any input on the "open questions" in the main post; particularly 1 and 3 (we have already started to add 2 to the README)?

localauthor:
> Does that explain it?

 Yes, that makes sense.

>  input on the "open questions"

On question 1, not really; I don't really know how caching works, or what's at stake in removing entries from local cache.

On question 3, some way of distinguishing local from global would be nice:

- Grouping seems the more actionable option, in terms of quickly narrowing to either local or global entries. I can see if I were to collaborate with someone, and wanted to use their bib entries in a document without 'polluting' my precious global bib, it might be helpful to be able to select from a grouping their bib entries. But a complication might be: which group would any common entries be in? Could they be in both?

- Affixation prefixes are nice, but mostly visual, rather than actionable, right? Or, would there be a way to narrow to just local or just global entries? As for indicators, rather than try to find properly communicative symbols, why not simply letters: `l`, `g,` or `l/g`?

bdarcus:
Per the screenshot, you can already filter the local candidates from the
prompt.

The ideas are just designed to visually indicate that.

bibtex-actions.el
```diff
@@ -276,16 +294,23 @@ key associated with each one."
                             (list pdf note link))"	") suffix))))
 
 (defvar bibtex-actions--candidates-cache nil
```
aikrahguzar, line 296:
```suggestion
(defvar bibtex-actions--candidates-cache 'uninitialized
```

bibtex-actions.el
```diff
@@ -276,16 +294,23 @@ key associated with each one."
...
-  "Store the candidates list.")
+  "Store the global candidates list.")
+
+(defvar-local bibtex-actions--local-candidates-cache nil
```
aikrahguzar, line 299:
```suggestion
(defvar-local bibtex-actions--local-candidates-cache 'uninitialized
```

bibtex-actions.el
```diff
@@ -276,16 +294,23 @@ key associated with each one."
...
+  "Store the local (per-buffer) candidates list.")
 
 (defun bibtex-actions--get-candidates (&optional force-rebuild-cache)
   "Get the cached candidates.
 If the cache is nil, this will load the cache.
 If FORCE-REBUILD-CACHE is t, force reloading the cache."
   (when (or force-rebuild-cache
-            (not bibtex-actions--candidates-cache))
+            (not bibtex-actions--candidates-cache)
+            (not bibtex-actions--local-candidates-cache))
```
aikrahguzar, line 309:
```suggestion
            (eq 'uninitialized bibtex-actions--candidates-cache)
            (eq 'uninitialized bibtex-actions--local-candidates-cache))
```

bibtex-actions.el
```diff
@@ -276,16 +294,23 @@ key associated with each one."
...
     (bibtex-actions-refresh force-rebuild-cache))
-  bibtex-actions--candidates-cache)
+  (seq-concatenate 'list
+                   ((lambda (x) (if (eq x t) nil x)) bibtex-actions--local-candidates-cache)
+                   ((lambda (x) (if (eq x t) nil x)) bibtex-actions--candidates-cache)))
```
aikrahguzar, line 313:
```suggestion
                   bibtex-actions--local-candidates-cache
                   bibtex-actions--candidates-cache))
```

bibtex-actions.el
```diff
@@ -295,8 +320,14 @@ is non-nil, also run the `bibtex-actions-before-refresh-hook' hook."
   (interactive "P")
   (when force-rebuild-cache
     (run-hooks 'bibtex-actions-force-refresh-hook))
+  ;; t in the result indicates that cache was computed but came out nil nil
+  ;; itself is used to indicate a cache that hasn't been computed this is to
+  ;; work around the fact that we can't differentiate between empty list and nil
   (setq bibtex-actions--candidates-cache
-        (bibtex-actions--format-candidates)))
+        ((lambda (x) (if x x t)) (bibtex-actions--format-candidates)))
```
aikrahguzar, line 327:
```suggestion
        (bibtex-actions--format-candidates))
```

bibtex-actions.el
```diff
@@ -295,8 +320,14 @@ is non-nil, also run the `bibtex-actions-before-refresh-hook' hook."
...
+  (let ((bibtex-completion-bibliography (bibtex-actions--local-files-to-cache)))
+    (setq bibtex-actions--local-candidates-cache
+        ((lambda (x) (if x x t)) (bibtex-actions--format-candidates "is:local")))))
```
aikrahguzar, line 330:
```suggestion
        (bibtex-actions--format-candidates "is:local"))))
```

aikrahguzar (changes requested):
I changed the handling to setting caches but mostly superficially. But I think they made the code more readable and simpler. I have tested these changes and they work (apart from any mistakes I might have introduced going through them on github piecewise).

bibtex-actions.el
bdarcus: Why this?

aikrahguzar:
>     2. ~Best options for auto-freshing of cache so users don't have to think about this? Maybe we include an [example auto-timer hook](https://github.com/bdarcus/bibtex-actions/pull/208#issuecomment-892589239) in the README?~ settled; working on this

I got it work but it is somewhat non-trivial 
```lisp
(defun bib-referesh-add-watches (func) "Add watches for the files that contribute to the local cache"
       (let ((buffer (buffer-name)))
         (seq-map (lambda (bibfile)
                    (file-notify-add-watch bibfile '(change)
                                           (lambda (x) (unless (eq 'stopped (cadr x))
                                                    (with-current-buffer buffer (funcall func))))))
                  (bibtex-actions--local-files-to-cache))))

(defun bib-refresh-filenotify (func)
  "Hook to add watches on buffer local bib files and remove them when the buffer is killed"
  (let ((descs (bib-referesh-add-watches func)))
    (add-hook 'kill-buffer-hook
              (lambda () (seq-map #'file-notify-rm-watch descs)) nil t)))
```

Adding 
```lisp
(lambda () (bib-refresh-filenotify #'bibtex-actions-refresh))
```
to the appropriate mode hook will auto-refresh the cache when a local bib file changes.

While something
```lisp
(lambda () (bib-refresh-filenotify (lambda () (setq bibtex-actions--candidates-cache-local 'uninitialized))))
```
will invalidate the cache so that it gets refreshed the next time a bibtex-actions command is called.

Right now one problem with this is that `bibtex-actions--local-files-to-cache` is not autloaded so the user has to do the autolading.

bdarcus:
> > ```
> > 2. ~Best options for auto-freshing of cache so users don't have to think about this? Maybe we include an [example auto-timer hook](https://github.com/bdarcus/bibtex-actions/pull/208#issuecomment-892589239) in the README?~ settled; working on this
> > ```
> 
> I got it work but it is somewhat non-trivial
> 
> ```lisp
> (defun bib-referesh-add-watches (func) "Add watches for the files that contribute to the local cache"
>        (let ((buffer (buffer-name)))
>          (seq-map (lambda (bibfile)
>                     (file-notify-add-watch bibfile '(change)
>                                            (lambda (x) (unless (eq 'stopped (cadr x))
>                                                     (with-current-buffer buffer (funcall func))))))
>                   (bibtex-actions--local-files-to-cache))))
> 
> (defun bib-refresh-filenotify (func)
>   "Hook to add watches on buffer local bib files and remove them when the buffer is killed"
>   (let ((descs (bib-referesh-add-watches func)))
>     (add-hook 'kill-buffer-hook
>               (lambda () (seq-map #'file-notify-rm-watch descs)) nil t)))
> ```
> 
> Adding
> 
> ```lisp
> (lambda () (bib-refresh-filenotify #'bibtex-actions-refresh))
> ```
> 
> to the appropriate mode hook will auto-refresh the cache when a local bib file changes.
> 
> While something
> 
> ```lisp
> (lambda () (bib-refresh-filenotify (lambda () (setq bibtex-actions--candidates-cache-local 'uninitialized))))
> ```
> 
> will invalidate the cache so that it gets refreshed the next time a bibtex-actions command is called.
> 
> Right now one problem with this is that `bibtex-actions--local-files-to-cache` is not autloaded so the user has to do the autolading.

1. Should we include this, or should it go in the README?
2. Either way, maybe we can save this for a separate commit, and you can submit that as a PR?
3. Should it be autoloaded?

bibtex-actions.el
aikrahguzar: For clarity and also because it makes the refresh `bibtex-actions--get-candidates` and `bibtex-actions--refresh` somewhat simpler since we lambda's

bdarcus: Gotcha; I just batch-committed these.

aikrahguzar:
>     1. Should we include this, or should it go in the README?

filenotify stuff is in core eamcs so it seems ok to  include this and I think it is complicated enough that I would rather not have it in my personal configuration if possible.

>     2. Either way, maybe we can save this for a separate commit, and you can submit that as a PR?

I need a git tutorial so I can probably do it this weekend unless I have to do some work stuff. 

>     3. Should it be autoloaded?

`bibtex-actions--local-files-to-cache` probably not but if we include something like `bib-refresh-filenotify` then that should be. Or we can document how to autoload it in the readme. I am agnostic.

bibtex-actions.el
```diff
@@ -324,10 +324,10 @@ is non-nil, also run the `bibtex-actions-before-refresh-hook' hook."
   ;; itself is used to indicate a cache that hasn't been computed this is to
   ;; work around the fact that we can't differentiate between empty list and nil
```
aikrahguzar, line 325: This comment is outdated now. I forgot to remove it earlier.

bdarcus: Oops; I merged it.

You can remove it when you do the related PR :-)

aikrahguzar: That was quick. I hope these late changes don't introduce any errors. They are pretty trivial but still fingers crossed.

bdarcus:
> I need a git tutorial so I can probably do it this weekend unless I have to do some work stuff.

FWIW, for the PRs, you don't need to worry about advanced stuff like rebase, as I usually just "squash rebase" anyway (e.g. bundle whatever commits into one single commit). Just make your changes and commit, then push, them, to your fork.

bdarcus:
> And asking on doom discord seems like a good thing. While trying to familiarize myself with how bibtex-completion works under the hook, I saw that `reftex` which is used under the hood has a function for generating a new bib file consisting of all the entries cited in the document from a global bib file. Some people might have a workflow built around that.

FYI, parsebib just merged major enhancements, that could _maybe_ allow bypassing bibtex-completion for parsing. It handles deduping. Here's it parsing two files with same data; one biblatex, and the other CSL JSON.

```elisp
ELISP> (defvar my/bibtest3 (parsebib-parse '("~/org/bib/newer.json" "~/org/bib/newer.bib")))
my/bibtest3
ELISP> my/bibtest3
#<hash-table equal 234/325 0x1e587cd>
ELISP> (gethash "low2001" my/bibtest3)
(("subtitle" . "gated communities and the discourse of urban fear")
 ("title" . "The Edge and the Center")
 ("number" . "1")
 ("langid" . "english")
 ("urldate" . "2016-02-18")
 ("doi" . "10.1525/aa.2001.103.1.45")
 ("issn" . "1548-1433")
 ("pages" . "45--58")
 ("volume" . "103")
 ("journaltitle" . "American Anthropologist")
 ("date" . "2001-03-01")
 ("author" . "Low, Setha M.")
 ("shorttitle" . "The Edge and the Center")
 ("=type=" . "article")
 ("=key=" . "low2001"))
```

PR #199, opened 2021-07-29
Fix bibtex-actions-run-default-action

bdarcus:
Fix #195

PR #213, opened 2021-08-06
Fix embark-act > RET

### LINKED ISSUE #195: default actions not working in oc-bibtex-actions
Hi. Thanks for your work. Following closely, but find it challenging. 

I'm trying to setup for org-cite with citeproc-org (my understanding is that it'd be possible to remove pandoc and org-ref all together) in doom Emacs, where I also have the "biblio" module enabled (which I think you contributed) but I'm not sure how to do it correctly, or how to get the default embark action on a citation to open for example the associated pdf.

bdarcus:
This is a minimal alternative to #199.

The following works in org-cite:

- `M-x bibtex-actions-dwim`
- `M-x embark-dwim`
- `M-x org-open-at-point`

The only thing that doesn't work (for me at least):

- `embark-act` -> `RET` 

I do now get this message when clicking on a key, though, which might be a hint:

```console
Embark: The action indicator takes three arguments, KEYMAP, TARGET and TYPE.
```

If we can't get this working, I'll just remove that binding, which isn't that useful anyway.

Fix #195.

cc @ilupin

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -530,8 +530,8 @@ With prefix, rebuild the cache before offering candidates."
 (defun bibtex-actions-dwim ()
   "Run the default action on citation keys found at point."
   (interactive)
-  (if-let ((keys (bibtex-actions-citation-key-at-point)))
-      (funcall bibtex-actions-default-action keys)))
+  (if-let ((keys (cdr (bibtex-actions-citation-key-at-point))))
+      (bibtex-actions-run-default-action keys)))
 
 (provide 'bibtex-actions)
 ;;; bibtex-actions.el ends here
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -522,7 +522,9 @@ With prefix, rebuild the cache before offering candidates."
 (defun bibtex-actions-run-default-action (keys)
   "Run the default action `bibtex-actions-default-action' on KEYS."
   (funcall bibtex-actions-default-action
-           (if (stringp keys) (split-string keys " & ") keys)))
+           (if (stringp keys)
+               (split-string keys " & ")
+             (split-string (cdr keys) " & "))))
 
 ;;;###autoload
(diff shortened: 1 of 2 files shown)


oc-bibtex-actions.el
```diff
@@ -245,7 +245,7 @@ strings by style."
     (define-key map (kbd "n") '("open notes" . bibtex-actions-open-notes))
     (define-key map (kbd "p") '("open PDF" . bibtex-actions-open-pdf))
     (define-key map (kbd "r") '("refresh library" . bibtex-actions-refresh))
-    (define-key map (kbd "RET") '("default action" . bibtex-actions-run-default-action))
+    (define-key map (kbd "RET") '("default action" . bibtex-actions-dwim))
```
bdarcus, line 248: Actually, I don't know if this makes sense, because what if we want `embark-act` the default?

bdarcus, line 248: Hypothesis: embark doesn't like the description here.

bdarcus:
The latest commit I pushed simply removes this binding. I'd like to figure it out, but on reflection, I don't think this is that useful anyway.

WDYT?

bdarcus:
I'm not really happy with this, but this what the menu looks like now in which-key:

image

bibtex-actions.el
```diff
@@ -522,14 +522,16 @@ With prefix, rebuild the cache before offering candidates."
 (defun bibtex-actions-run-default-action (keys)
   "Run the default action `bibtex-actions-default-action' on KEYS."
   (funcall bibtex-actions-default-action
-           (if (stringp keys) (split-string keys " & ") keys)))
+           (if (stringp keys)
+               (split-string keys " & ")
+             (split-string (cdr keys) " & "))))
```
bdarcus, line 527: Why do we need both dwim and run-default? Couldn't we just add an optional keys arg on the former?

PR #214, opened 2021-08-06
Filenotify convenience functions

aikrahguzar:
Added filenotify based convenience functions and updated the README with example usage. They have been mildly tested in my personal config.

Also touched up comments a bit.

### DIFF
--- a/README.org
+++ b/README.org
@@ -233,7 +233,9 @@ This is great for performance, but means the data can become stale if you modify
 The =bibtex-actions-refresh= command will reload the caches, and you can call this manually.
 You can also call any of the =bibtex-actions= commands with a prefix argument: =C-u M-x bibtex-actions-insert-key=.
 
-Although not default =Bibtex-actions= also provides convenience functions for auto-refreshing cache when bib files change using filenotify. There are two such functions. =bibtex-actions-with-filenotify-local= and =bibtex-actions-filenotify-global=. The following snippet will refresh caches when one of the bib files change
+Although not default, =bibtex-actions= also provides convenience functions for auto-refreshing cache when bib files change using filenotify. 
+There are two such functions: =bibtex-actions-with-filenotify-local= and =bibtex-actions-filenotify-global=. 
+The following snippet will refresh caches when one of the bib files change:
 
 #+BEGIN_SRC emacs-lisp
 (bibtex-actions-with-filenotify-global #'bibtex-actions-refresh)
--- a/README.org
+++ b/README.org
@@ -249,7 +249,8 @@ A different option which might be useful is to merely invalidate the cache so th
 This can be achieved by the following code:
 
 #+BEGIN_SRC emacs-lisp
-(bibtex-actions-with-filenotify-global (lambda () (setq bibtex-actions--candidates-cache 'uninitialized)))
+(bibtex-actions-with-filenotify-global
+ (lambda () (setq bibtex-actions--candidates-cache 'uninitialized)))
 
 (add-hook 'LaTeX-mode-hook
           (lambda () (bibtex-actions-with-filenotify-local
--- a/README.org
+++ b/README.org
@@ -242,8 +242,8 @@ Although not default =Bibtex-actions= also provides convenience functions for au
 (add-hook 'org-mode-hook (lambda () (bibtex-actions-with-filenotify-local #'bibtex-actions-refresh)))
 #+END_SRC
 
-This will reload the cache whenever a bib file changes. A different option which might be useful to merely invalidate the cache so that it s regenerated whenever a
-bibtex-actions is called next. This can be achieved by the following code,
+This will reload the cache whenever a bib file changes. A different option which might be useful to merely invalidate the cache so that it is regenerated whenever a
+bibtex-actions command is called next. This can be achieved by the following code,
 
 #+BEGIN_SRC emacs-lisp
 (bibtex-actions-with-filenotify-global (lambda () (setq bibtex-actions--candidates-cache 'uninitiated)))
--- a/README.org
+++ b/README.org
@@ -242,14 +242,14 @@ Although not default =Bibtex-actions= also provides convenience functions for au
 (add-hook 'org-mode-hook (lambda () (bibtex-actions-with-filenotify-local #'bibtex-actions-refresh)))
 #+END_SRC
 
-This will reload the cache whenever a bib file changes. A different option which might be useful to merely invalidate the cache so that it is regenerated whenever a
+This will reload the cache whenever a bib file changes. A different option which might be useful is to merely invalidate the cache so that it is regenerated whenever a
 bibtex-actions command is called next. This can be achieved by the following code,
 
 #+BEGIN_SRC emacs-lisp
-(bibtex-actions-with-filenotify-global (lambda () (setq bibtex-actions--candidates-cache 'uninitiated)))
+(bibtex-actions-with-filenotify-global (lambda () (setq bibtex-actions--candidates-cache 'uninitialized)))
 
-(add-hook 'LaTeX-mode-hook (lambda () (bibtex-actions-with-filenotify-local (lambda () (setq bibtex-actions--local-candidates-cache 'uninitiated)))))
-(add-hook 'org-mode-hook (lambda () (bibtex-actions-with-filenotify-local (lambda () (setq bibtex-actions--candidates-cache 'uninitiated)))))
+(add-hook 'LaTeX-mode-hook (lambda () (bibtex-actions-with-filenotify-local (lambda () (setq bibtex-actions--local-candidates-cache 'uninitialized)))))
+(add-hook 'org-mode-hook (lambda () (bibtex-actions-with-filenotify-local (lambda () (setq bibtex-actions--local-candidates-cache 'uninitialized)))))
 #+END_SRC
 
 Another option to make the completion interface more seamless is to add a hook which generates the cache after a buffer is opened. This can be done when emacs has been idle (half a second in the example below) with something like this:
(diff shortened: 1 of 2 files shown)


README.org
```diff
@@ -233,6 +233,25 @@ This is great for performance, but means the data can become stale if you modify
 The =bibtex-actions-refresh= command will reload the caches, and you can call this manually.
 You can also call any of the =bibtex-actions= commands with a prefix argument: =C-u M-x bibtex-actions-insert-key=.
 
+Although not default =Bibtex-actions= also provides convenience functions for auto-refreshing cache when bib files change using filenotify. There are two such functions. =bibtex-actions-with-filenotify-local= and =bibtex-actions-filenotify-global=. The following snippet will refresh caches when one of the bib files change
```
bdarcus, line 236:
```suggestion
Although not default, =bibtex-actions= also provides convenience functions for auto-refreshing cache when bib files change using filenotify. 
There are two such functions: =bibtex-actions-with-filenotify-local= and =bibtex-actions-filenotify-global=. 
The following snippet will refresh caches when one of the bib files change:
```

I use one-sentence-per-line here, for cleaner git diffs.

README.org
```diff
@@ -233,6 +233,25 @@ This is great for performance, but means the data can become stale if you modify
...
+
+#+BEGIN_SRC emacs-lisp
+(bibtex-actions-with-filenotify-global #'bibtex-actions-refresh)
+
+(add-hook 'LaTeX-mode-hook (lambda () (bibtex-actions-with-filenotify-local #'bibtex-actions-refresh)))
+(add-hook 'org-mode-hook (lambda () (bibtex-actions-with-filenotify-local #'bibtex-actions-refresh)))
+#+END_SRC
+
+This will reload the cache whenever a bib file changes. A different option which might be useful is to merely invalidate the cache so that it is regenerated whenever a
+bibtex-actions command is called next. This can be achieved by the following code,
```
bdarcus, line 246:
```suggestion
This will reload the cache whenever a bib file changes. 
A different option which might be useful is to merely invalidate the cache so that it is regenerated whenever a bibtex-actions command is called next. 
This can be achieved by the following code:
```

Same, except why include this? I know we can, but it adds perhaps unnecessary complexity?

bdarcus (changes requested):
Looks good. 

Just a few formatting issues, and also a question on one chunk.

You can click to commit my suggestions, and "resolve" when you're done with each.

README.org
```diff
@@ -233,6 +233,25 @@ This is great for performance, but means the data can become stale if you modify
 The =bibtex-actions-refresh= command will reload the caches, and you can call this manually.
 You can also call any of the =bibtex-actions= commands with a prefix argument: =C-u M-x bibtex-actions-insert-key=.
 
+Although not default =Bibtex-actions= also provides convenience functions for auto-refreshing cache when bib files change using filenotify. There are two such functions. =bibtex-actions-with-filenotify-local= and =bibtex-actions-filenotify-global=. The following snippet will refresh caches when one of the bib files change
+
+#+BEGIN_SRC emacs-lisp
+(bibtex-actions-with-filenotify-global #'bibtex-actions-refresh)
+
+(add-hook 'LaTeX-mode-hook (lambda () (bibtex-actions-with-filenotify-local #'bibtex-actions-refresh)))
+(add-hook 'org-mode-hook (lambda () (bibtex-actions-with-filenotify-local #'bibtex-actions-refresh)))
```
bdarcus, line 242: Question: do we need to do anything to remove the inotify watches, or does the fact that this is a hook take care of that?

README.org
aikrahguzar: I am fine with leaving it out. It is basically if someone doesn't like the cpu hit if they add a few entries in a row. But I am personally using the first one. My bib files don't change that often.

README.org
aikrahguzar: The function `bibtex-actions-with-filenotify` takes care of removing the watches. It adds a buffer local hook to `kill-buffer-hook` to remove these watches when the buffer is killed.

README.org
bdarcus: WDYT about moving it to the wiki?

aikrahguzar: Seems like a good idea

bdarcus: I'm actually fine either way; just asking.

bdarcus:
PS - when you're done you can click the "re-request" review icon at the top right, next to my name.

README.org
aikrahguzar: Me too, I usually prefer it someone else makes organizational decisions:)

README.org
aikrahguzar: I should probably clarify `bibtex-actions-with-filenotify-local` takes care of it. The global version just returns the descriptors. The global watches need to be killed manually.

README.org
```diff
@@ -233,6 +233,28 @@ This is great for performance, but means the data can become stale if you modify
...
+
+This will reload the cache whenever a bib file changes. 
+A different option which might be useful is to merely invalidate the cache so that it is regenerated whenever a bibtex-actions command is called next. 
+This can be achieved by the following code:
+
+#+BEGIN_SRC emacs-lisp
+(bibtex-actions-with-filenotify-global (lambda () (setq bibtex-actions--candidates-cache 'uninitialized)))
+
+(add-hook 'LaTeX-mode-hook (lambda () (bibtex-actions-with-filenotify-local (lambda () (setq bibtex-actions--local-candidates-cache 'uninitialized)))))
+(add-hook 'org-mode-hook (lambda () (bibtex-actions-with-filenotify-local (lambda () (setq bibtex-actions--local-candidates-cache 'uninitialized)))))
```
bdarcus, line 255:
```suggestion
(add-hook 'LaTeX-mode-hook
          (lambda () (bibtex-actions-with-filenotify-local
                      (lambda () (setq bibtex-actions--local-candidates-cache 'uninitialized)))))

(add-hook 'org-mode-hook
          (lambda () (bibtex-actions-with-filenotify-local
                      (lambda () (setq bibtex-actions--local-candidates-cache 'uninitialized)))))
```

README.org
```diff
@@ -233,6 +233,33 @@ This is great for performance, but means the data can become stale if you modify
...
+(add-hook 'LaTeX-mode-hook (lambda () (bibtex-actions-with-filenotify-local #'bibtex-actions-refresh)))
+(add-hook 'org-mode-hook (lambda () (bibtex-actions-with-filenotify-local #'bibtex-actions-refresh)))
+#+END_SRC
+
+This will reload the cache whenever a bib file changes. 
+A different option which might be useful is to merely invalidate the cache so that it is regenerated whenever a bibtex-actions command is called next. 
+This can be achieved by the following code:
+
+#+BEGIN_SRC emacs-lisp
+(bibtex-actions-with-filenotify-global (lambda () (setq bibtex-actions--candidates-cache 'uninitialized)))
```
bdarcus, line 252:
```suggestion
(bibtex-actions-with-filenotify-global
 (lambda () (setq bibtex-actions--candidates-cache 'uninitialized)))
```

bdarcus (approved):
Thanks!

README.org
bdarcus: > I should probably clarify `bibtex-actions-with-filenotify-local` takes care of it. The global version just returns the descriptors. The global watches need to be killed manually.

Is that a necessary limitation?

aikrahguzar: No, I just don't know when the global watches should be removed.

bdarcus: I think bibtex-completion creates and destroys them per session?

https://github.com/tmalsburg/helm-bibtex/blob/12079bb09f203dda5cc2dd003bd60a6ad490f762/bibtex-completion.el#L396

bdarcus: BTW, one of the reasons I bypassed the bibtex-completion support for this is I wanted more flexibility, and wasn't sure how best to do that.

For example, I want to have a watch on my notes directory.

aikrahguzar: > I think bibtex-completion creates and destroys them per session?
> 
> https://github.com/tmalsburg/helm-bibtex/blob/12079bb09f203dda5cc2dd003bd60a6ad490f762/bibtex-completion.el#L396

I think this is because the design of bibtex-completions is a bit adhoc. In particular since local bib is used through a let binding the collection of bib files can change from call to call. This is not the case for us. 

Let us assume for a moment that the collection doesn't changes. In that case what we have is correct. The watches on the global bin should exist unless manually killed and those on local files should be killed with the buffer. Though there is a problem with the code as is. If the major mode hook is run multiple times which can happen there will be multiple watches on the file. I have fixed this and will do another pr.

Now let us consider the case when a bib file is added and removed which I think should be rare. For the local bib files this isn't handled by bibtex-completions. To detect this a reftex scan is needed. I think the best solution here is to have a manual refresh function that runs the scan and adjusts the watches. A user can run then call it after adding/removing bib files.

bdarcus: > I think the best solution here is to have a manual refresh function that runs the scan and adjusts the watches. A user can run then call it after adding/removing bib files

I agree.

But per my earlier comment, I would allow it to be configured to include note and library paths too.

aikrahguzar: > But per my earlier comment, I would allow it to be configured to include note and library paths too.

Ok, so basically add an optional list parameter that contains any extra files to be watched?

bdarcus: > so basically add an optional list parameter that contains any extra files to be watched?

Yes, but where the paths can be specific via variable.

E.g. if I have the notes-path variable set, I should be able to tell it to use that.

aikrahguzar: > E.g. if I have the notes-path variable set, I should be able to tell it to use that.

You can pass the variable as the optional argument to the function. Or are you thinking of something else?

bdarcus: Only that it's easy and flexible for the user.

aikrahguzar: I created the pr so we can discuss there if something needs changing.

bdarcus:
Just wanted to put this somewhere relevant for possible future reference.

A clever solution solution from the author of oc-basic:

> oc-basic relies on a cache. The cache key contains a hash of the
contents of the bib file. So whenever the bib file is modified, the
cache is invalidated, and oc-basic parses again the file.

PR #218, opened 2021-08-10
Generate style list dynamically

bdarcus:
Generate style candidates via `org-cite-supported-styles`, and for the
time being, comments out the annotation-function.

Also:

- add section on styles to the README
- removing sorting on group-function

### DIFF
--- a/README.org
+++ b/README.org
@@ -309,6 +309,18 @@ The "insert processor" will use =bibtex-actions-read= to browse your library to 
 The "follow processor" provides at-point functionality accessible via the =org-open-at-point= command.
 By default, in org-mode with org-cite support, when point is on a citation or citation-reference, and you invoke =org-open-at-point=, it will run the default command, which is =bibtex-actions-open=.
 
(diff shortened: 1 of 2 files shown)


bdarcus:
@minad - is there an easy way for me to sort the groups in a group-function?

I want my "Default" group (where candidates look like `/` "/b` etc), for example, to go first, and ideally to specify order that follows.

minad:
The group order follows the candidate order. The group of the first candidate comes first. If you want the groups to be sorted, you should sort the candidates first and specifiy display-sort-function=identity. The reason for this is, that completion styles like flex actively reorder candidates in order to get the best match first. Furthermore the completion UI may sort by history position. Grouping should not interfere with these abilities. If you set display/cycle-sort-function to identity you opt out of this, which is a good idea in some special cases.

bdarcus:
> But I doubt you have such a case here and I generally discourage interfering with sorting.

Yes, I think you're right; will just omit `display-sort`.

Thank you!

Issue #191, opened 2021-07-25
Make it quick and easy to open a "file" referenced from a BibTeX entry


salutis:
From https://github.com/tmalsburg/helm-bibtex/issues/383:

>> I added `file = {monet-1877a.jpg}` to my BibTeX entry. Then, I ran `bibtex-actions-open`, and it showed `[multiple] References: has:link\|has:pdf`. I wondered why the prompt is not`has:url\|has:file` instead, as those are the relevant BibTeX fields. (Right?)
> 
> So this piece is my problem :-)

I am not even sure what to ask for, but here we go.

Brainstorming off the top of my head:

I would expect `bibtex-actions-open-link` to be called `bibtex-actions-open-url`, if that is what the function does, and I would expect `bibtex-actions-open-file` to exist.

Then, the big question is, what shoud `bibtex-actions-open` do?

First and foremost, I would expect it to be intelligent. If the BibTeX entry has one `file` and nothing else (no notes, no URL, no nothing), then it should call `bibtex-actions-open-file`, as there is nothing else to open. If the BibTeX entry has more going on, I would expect Emacs to ask me.

Here is a basic 101 use-case:

1. `M-x bibtex-actions-open`
2. select some item and press `RET`
3. see the item (URL, file, ...)

I think the basic 101 use-case should "just work".

bdarcus:
> I would expect `bibtex-actions-open-link` to be called `bibtex-actions-open-url`, if that is what the function does, and I would expect `bibtex-actions-open-file` to exist.

If you search the issue tracker and PRs, you'll find this isn't a new question. But it's not super straightforward.

Here's the current situation:

- Open-pdf = open local (PDF) file
- Open-url = open remote file/resource
- Open-doi = same
- Ooen-link = url OR doi (bibtex-completion uses the latter language, but I saw no need for that distinction)
- Open = open any of the above, in order (if PDF, do that, else ...)

These commands are just wrappers around bibtex-completion functions. I tried to name them to simplify, but in an earlier issue we concluded ideally "open" is really "open source."

But the specific command and function name that I think is at issue is really the first one; "pdf" is in the very function and command names. You really want that to be "open-local-file" (keep in mind there's always a tension between brevity and clarity though).

> Then, the big question is, what shoud `bibtex-actions-open` do?
> 
[...]

Issue #188, opened 2021-07-24
Support non-PDF files in "bibtex-completion-library-path"


salutis:
I have, for instance, the following `.bib` entry:

```
@Misc{monet-1877a,
  author       = {Claude Monet},
  title        = {La Gare Saint-{Lazare}},
  year         = 1877,
  type         = {painting},
  publisher    = {Musée d'Orsay, Paris, France}
}
```

The `.bib` entry has a corresponding `monet-1877a.jpg` file.

When I use `bibtex-actions-open`, instead of seeing the painting, I am informed that:

```
Wrong type argument: stringp, nil
```

bdarcus:
That's a bug in bibtex-completion, which that command uses.

It should either open it, or give a useful error report.

Good luck!

salutis:
@bdarcus As a user, where can I ask about this?

bdarcus:
Sorry; here:

https://github.com/tmalsburg/helm-bibtex

bdarcus:
PS - have you had a chance to try org-cite yet?

salutis:
@bdarcus Not yet. I am waiting for the release (into `master`, a.k.a. Emacs 28). For now, I am using Pandoc.

salutis:
@bdarcus Once `org-cite` is out, will `helm-bibtex` still be used? I have no idea how this tremendous patchwork works under the hood. (I thought I do not use Helm.) Basically, I want to be able to open JPG, PDF, HTML, ePub, or even a directory based on the key.

salutis:
All right, I asked over at https://github.com/tmalsburg/helm-bibtex/issues/383.

bdarcus:
Just to explain, helm-bibtex, ivy-bibtex, and his package are all front-ends for the same backend bibtex-completion code.

Usually when something breaks here it's a general problem; like completion doesn't work or something.

salutis:
> Just to explain, `helm-bibtex`, `ivy-bibtex`, and this package are all front-ends to the same backend `bibtex-completion` code.
> 
> Usually when something breaks here it's a general problem; like completion doesn't work or something.

But then, why is the problem in `helm-bibtex` (which is just another front-end) and not some `bibtex-completion` package? Is that because that `bibtex-completion` is coupled with `helm-bibtex` for historical reasons?

bdarcus:
> Is that because that `bibtex-completion` is coupled with `helm-bibtex` for historical reasons?

This.

[...]

PR #222, opened 2021-08-11
refactor: switch to parsebib 3.0 for candidates

bdarcus:
Not yet sure of the implications of this broadly, but this switches to using parsebib 3.0 for parsing the candidates, and so:

1. bypasses bibtex-completion for this task
2. adds CSL JSON support

image

TODO:

1. for some reason, I get a bug on `bibtex-actions-read` when trying to add history to the default `completing-read` part, even though it's the exact same code as the CRM part


bibtex-actions.el
```diff
@@ -115,6 +117,23 @@ manager like Zotero or JabRef."
...
+(defvar bibtex-actions-field-map
+  '(("date" "year" "issued")
+    ("=key=" "id")
+    ("=type=" "type")
+    ("tags" "keywords" "keyword")
+    ("booktitle" "container-title")
+    ("journaltitle" "journal" "container-title")
+    ("number" "issue")))
+
+(defvar bibtex-actions-field-mapping
```
bdarcus, line 129: Need to make a decision on which approach to take on mapping.

bdarcus:
So @denismaier - how valuable do you think CSL JSON support is?

It basically works, and my (not extensive) tests find only one issue: `insert-reference` gives empty dates.

I'm not seeing any other problems, so one option is just to merge and see what happens?

One thing I have not tested is if it handles bibtex similar to bibtex-completion. What happens if you have a bibtex file with a bunch of latex markup in a title for example?

WDYT?

bibtex-actions.el
```diff
@@ -225,53 +244,75 @@ offering the selection candidates"
...
+       (car (split-string name ", "))))
+   (split-string names " and ") ", "))
+
 (defun bibtex-actions--format-candidates (&optional context)
   "Format candidates, with optional hidden CONTEXT metadata.
 This both propertizes the candidates for display, and grabs the
 key associated with each one."
-  (let* ((main-template
+  (let* ((candidates
+          (ht-values (parsebib-parse bibtex-completion-bibliography)))
```
aikrahguzar, line 272: I think it would be better if this function takes a list of files as a mandatory arguments instead of relying on  `bibtex-completion-bibliography`. I think the only place this will need to change is `bibtex-actions-refresh` where we can just `bibtex-completion-bibliography` as the argument for local files and similarly deal with the local files without having to let-bind `bibtex-completion-bibliography` which I find somewhat clumsy.

bibtex-actions.el
```diff
@@ -225,53 +244,75 @@ offering the selection candidates"
...
-         (concat
-          ;; We need all of these searchable:
-          ;;   1. the 'candidate-main' variable to be displayed
-          ;;   2. the 'candidate-suffix' variable to be displayed with a different face
-          ;;   3. the 'candidate-hidden' variable to be hidden
-          (propertize candidate-main 'face 'bibtex-actions-highlight) " "
-          (propertize candidate-suffix 'face 'bibtex-actions) " "
-          (propertize candidate-hidden 'invisible t)))
-        citekey)))))
+    (cl-loop for candidate in candidates
```
aikrahguzar, line 281: `cl-loop` can loop over the values of a hash table directly, I think this should work
```suggestion
    (cl-loop for candidate being the hash-values of (parsebib-parse bibtex-completion-bibliography)
```
and will allow you to get rid of dependency on `ht`.

bibtex-actions.el
bdarcus: Yeah, I was thinking about this.

bibtex-actions.el
bdarcus: So your suggestion is two-fold:

1. add a required arg on the function
2. do the suggestion here, but instead reference that arg?

BTW, have you had a chance to test this branch?

bdarcus: I was also wondering if I need to add a `bibtex-actions-bibliography` defcustom to hold that list?

bdarcus: I just pushed a change with these three things.

Still have to test, but WDYT?

aikrahguzar: > So your suggestion is two-fold:
> 
>     1. add a required arg on the function
> 
>     2. do the suggestion here, but instead reference that arg?

Yes.

> BTW, have you had a chance to test this branch?

No, I will wait till filenotify stuff is merged since I will need to change my config without it.

bdarcus: > Still have to test, but WDYT?

Actually, don't think it works as is, as the `ht-values` function returns the values for all the keys.

But should be easy to fix.

bibtex-actions.el
```diff
@@ -322,7 +366,7 @@ is non-nil, also run the `bibtex-actions-before-refresh-hook' hook."
   (when force-rebuild-cache
     (run-hooks 'bibtex-actions-force-refresh-hook))
   (setq bibtex-actions--candidates-cache
-        (bibtex-actions--format-candidates))
+        (bibtex-actions--format-candidates bibtex-actions-bibliography))
   (let ((bibtex-completion-bibliography (bibtex-actions--local-files-to-cache)))
     (setq bibtex-actions--local-candidates-cache
         (bibtex-actions--format-candidates "is:local"))))
```
aikrahguzar, line 372:
```suggestion
    (setq bibtex-actions--local-candidates-cache
        (bibtex-actions--format-candidates (bibtex-actions--local-files-to-cache) "is:local")))
```

bibtex-actions.el
aikrahguzar: > Nevermind; missed something :-)
> 
> Works now.

Local cache still needs to be adjusted. I added a suggestion (hopefully with balanced parens).

aikrahguzar: > I was also wondering if I need to add a `bibtex-actions-bibliography` defcustom to hold that list?

What is the eventual goal? Removal of dependency on `bibtex-completion`? It seems worthwhile though the commands will need to be re-implemented.

Maybe for the time being it should be defaulted `bibtex-completions-bibliography` otherwise it will break some people's setup

bdarcus: > > I was also wondering if I need to add a `bibtex-actions-bibliography` defcustom to hold that list?
> 
> What is the eventual goal? Removal of dependency on `bibtex-completion`? 

Yes.

> It seems worthwhile though the commands will need to be re-implemented.
> 
> Maybe for the time being it should be defaulted `bibtex-completions-bibliography` otherwise it will break some people's setup

Right.

I guess I should really remove the defcustom for now then?

bdarcus: ... or just keep it but set to default to bibtex-completion-bibliography?

aikrahguzar: > ... or just keep it but set to default to bibtex-completion-bibliography?

I think you can use `bibtex-completion` as an additional choice and the default value in the defcustom, set it to the default and maybe say it will be removed later.

bibtex-actions.el
```diff
@@ -322,10 +368,9 @@ is non-nil, also run the `bibtex-actions-before-refresh-hook' hook."
   (when force-rebuild-cache
     (run-hooks 'bibtex-actions-force-refresh-hook))
   (setq bibtex-actions--candidates-cache
-        (bibtex-actions--format-candidates))
-  (let ((bibtex-completion-bibliography (bibtex-actions--local-files-to-cache)))
-    (setq bibtex-actions--local-candidates-cache
-        (bibtex-actions--format-candidates "is:local"))))
+        (bibtex-actions--format-candidates bibtex-actions-bibliography))
+  (setq bibtex-actions--local-candidates-cache
+        (bibtex-actions--format-candidates bibtex-actions--local-files-to-cache "is:local")))
```
aikrahguzar, line 373: Missing parens around `bibtex-actions--local-files-to-cache`
```suggestion
        (bibtex-actions--format-candidates (bibtex-actions--local-files-to-cache) "is:local")))
```

bibtex-actions.el
bdarcus: I'm actually thinking seriously about standardizing on the csl names, as they should work better for this use case.

bdarcus:
OK @aikrahguzar - could you please give this a close look, including at the revised docstrings, and suggest any revisions?

Conflicts are resolved, but I am hoping I didn't introduce any regressions.

bibtex-actions.el
```diff
@@ -340,20 +385,24 @@ If FORCE-REBUILD-CACHE is t, force reload the cache."
 ;;;###autoload
 (defun bibtex-actions-refresh (&optional force-rebuild-cache scope)
   "Reload the candidates cache.
+
 If called interactively with a prefix or if FORCE-REBUILD-CACHE
 is non-nil, also run the `bibtex-actions-before-refresh-hook' hook.
-If SCOPE is `global' only glboal cache is refreshed, if it is `local' only local cache is
-refrehged. With any other value both are refreshed."
+
+If SCOPE is `global' only glboal cache is refreshed, if it is
```
bdarcus, line 392:
```suggestion
If SCOPE is `global' only global cache is refreshed, if it is
```

bibtex-actions.el
```diff
@@ -340,20 +385,24 @@ If FORCE-REBUILD-CACHE is t, force reload the cache."
...
 (defun bibtex-actions-refresh (&optional force-rebuild-cache scope)
   "Reload the candidates cache.
+
 If called interactively with a prefix or if FORCE-REBUILD-CACHE
 is non-nil, also run the `bibtex-actions-before-refresh-hook' hook.
-If SCOPE is `global' only glboal cache is refreshed, if it is `local' only local cache is
-refrehged. With any other value both are refreshed."
+
+If SCOPE is `global' only global cache is refreshed, if it is
+`local' only local cache is refrehged. With any other value both
```
bdarcus, line 393:
```suggestion
`local' only local cache is refreshed. With any other value both
```

aikrahguzar:
> OK @aikrahguzar - could you please give this a close look, including at the revised docstrings, and suggest any revisions?
> 
> Conflicts are resolved, but I am hoping I didn't introduce any regressions.



> OK @aikrahguzar - could you please give this a close look, including at the revised docstrings, and suggest any revisions?
> 
> Conflicts are resolved, but I am hoping I didn't introduce any regressions.

Everything looks good to me! I want to try this branch but I am still figuring out how to get git to show it to me.

bdarcus:
> I want to try this branch but I am still figuring out how to get git to show it to me.

`git checkout csl-json`

You may also have to do something like this first:

`git fetch` or `git fetch origin`

bdarcus:
One thing we lose here: the `has:pdf` and `has:note` metadata.

Would need to add functions to assess that, which if we do that, might as well just rewrite those commands here.

bibtex-actions.el
```diff
@@ -69,8 +70,15 @@
   "Face used to highlight content in `bibtex-actions' candidates."
   :group 'bibtex-actions)
 
+(defcustom bibtex-actions-bibliography (bibtex-actions--normalize-paths
+                                        bibtex-completion-bibliography)
```
aikrahguzar, line 74: This caused a really hard to figure out problem for me where I would get `void function bibtex-actions--normalize-paths` type error. I fixed it by changing it to nil. I think it is an issue of order where this is evaluated before the definition of `bibtex-actions--normalize-paths` is.

aikrahguzar: With this changed, this branch works for me

bdarcus: So you think it's fine as is then, or is there something I should change?

BTW, I just pushed a fix for the above pdf/note issue. Was easy.

bdarcus: > With this changed, this branch works for me

What do you mean by "this"?

Is this resolve, in other words?

aikrahguzar: > > With this changed, this branch works for me
> 
> What do you mean by "this"?
> 
I am changing the default to `nil` fixed it for me. I think I have a proper fix. I will add a suggestion for it.

aikrahguzar: Also I think I figured out the file-notify notes problem too. It was my mixing up local and global in the`unless` clauses in `bibtex-actions-filenotify-refresh`. I think exchanging those should fix it.

aikrahguzar:
After the latest changes I am seeing a lot of warnings like `Warning (:warning): Bibtex-completion couldn't find entry with key "Eynard:2008we".  Disable showing Disable logging` on cache refreshes and refreshing itself is much slower. The entries themselves are present in the cache. My guess is that it that the slowdown is probabaly due a combination to double parsing by both `bibtex-actions` and `bibtex-completion` and the flood of warnings. But I don't have any idea what is causing the refreshes themselves.

bdarcus:
You're probably right, which means of I want to do this I need to bite the
bullet and reimplements those functions.

bdarcus:
> After the latest changes I am seeing a lot of warnings like `Warning (:warning): Bibtex-completion couldn't find entry with key "Eynard:2008we". Disable showing Disable logging` on cache refreshes and refreshing itself is much slower. The entries themselves are present in the cache. My guess is that it that the slowdown is probabaly due a combination to double parsing by both `bibtex-actions` and `bibtex-completion` and the flood of warnings. But I don't have any idea what is causing the warnings themselves.

I reverted and pushed the commit that is the most likely culprit.

That means the pdf/note status won't work again, but I'm curious if that fixes your errors and warnings.

aikrahguzar:
> That means the pdf/note status won't work again, but I'm curious if that fixes your errors and warnings.

Yes the warnings are gone now.

Also I somehow edited one of my replies yesterday instead of replying to it :). This one https://github.com/bdarcus/bibtex-actions/pull/222#discussion_r688985172

Let me know what I should do there.

aikrahguzar:
> That means the pdf/note status won't work again, but I'm curious if that fixes your errors and warnings.

Some further investigation, the warning is caused by `bibtex-completion-get-entry1` which is called by `bibtex-completion-find-pdf-in-field` which is in turn called by `bibtex-completion-find-pdf`. The warning itself is due to bibtex-completions no longer having parsed entries and I think it is flood of warnings which slows down the refresh and not double parsing.

A copy paste of `bibtex-completion-find-pdf-in-field` can be fixed pretty easily to accommodate this. That apptoach seems problematic since the function itself is huge and I think should be broken apart. In particular it involves the interfacing with bibliography mangers like zotero and I think it would be better to have this interface refactored out and even made pluggable.

bdarcus:
> That appoach seems problematic since the function itself is huge and I think should be broken apart. In particular it involves the interface with bibliography mangers like zotero and I think it would be better to have this interface refactored out and even made pluggable.

Exactly; that's why I suggested better to just rewrite those functions, ideally much more simply.

But as you suggest; it's kind of hairy, so not sure how best to do that.

Do you see a good path?

bdarcus:
> > That means the pdf/note status won't work again, but I'm curious if that fixes your errors and warnings.
> 
> Yes the warnings are gone now.
> 
> Also I somehow edited one of my replies yesterday instead of replying to it :). This one [#222 (comment)](https://github.com/bdarcus/bibtex-actions/pull/222#discussion_r688985172)
> 
> Let me know what I should do there.

I don't really understand, but I at least can delete a comment revision. If you see a disclosure triangle near the "edited" bit, you should be able to as well.

aikrahguzar:
> I don't really understand, but I at least can delete a comment revision. If you see a disclosure triangle near the "edited" bit, you should be able to as well.

Yeah, I see that my wording got pretty awkward.  What I meant was should I do about the changes I suggested in 
https://github.com/bdarcus/bibtex-actions/pull/222#discussion_r688985172 

Those should have been a separate reply but I somehow ended up editing my original ones.

bibtex-actions.el
bdarcus: > I don't think I can do this as a suggestion but I have all this on my branch and I think I can make pr to `csl-json` branch?

Yes, try a PR to this branch.

aikrahguzar:
> But as you suggest; it's kind of hairy, so not sure how best to do that.
> 
> Do you see a good path?
> 
> Maybe just pluggable find note and pdf (might as well change it to a more general "file" though) functions?

The function is hairy but from my skimming the bulk of it is trying to work with other software. It is unclear to me exactly what we need from that external software. Is it used for anything other than finding the pdf files corresponding to an entry? Ideally I think there should be a well defined plist of functions we need to interface with an external software and a variable containing these plists so that they can be added without without changing the rest of the package.

If we forget about external software, I think the logic is:
1) see if there is a pdf in the bib entry
2) if not try to find one in the library directory.
This seems easier and can be done as a first step

bdarcus:
>  Is it used for anything other than finding the pdf files corresponding to an entry? 

Yes, and the same with the notes functions.

It's just, that project supports two approaches for each:

1. a) PDF etc flies in directory with key names (doe2001.pdf smith2020.jpg) OR b) with file paths defined in the records
2. a) Notes files same as above in directory, OR b) stored in a single (for example org) file

I think the latter of both, and particularly one, is where some of that complexity comes from.

I don't think 1b is technically "interfacing" with the software; I think it's just that different software may use different conventions for that field.

> Ideally I think there should be a well defined plist of functions we need to interface with an external software and a variable containing these plists so that they can be added without without changing the rest of the package.
> 
> If we forget about external software, I think the logic is:
> 
> 1. see if there is a pdf in the bib entry
> 2. if not try to find one in the library directory.
> 
> This seems easier and can be done as a first step

Yes, except it's possible I might have the entry field, but prefer to ignore it.

BTW, on my point about generalizing "PDF", it came from these issues, which raises a good point.

https://github.com/bdarcus/bibtex-actions/issues/188
https://github.com/bdarcus/bibtex-actions/issues/191

bibtex-actions.el
aikrahguzar: Done.

aikrahguzar: And see if it fixes the notes issue with filenotify watches too.

bdarcus: Merged.

aikrahguzar:
> Yes, except it's possible I might have the entry field, but prefer to ignore it.
> 
> BTW, on my point about generalizing "PDF", it came from these issues, which raises a good point.

Here it seems like the best ui would be to gather a list of all related and prompt the user to select one if we find more than one. This is done in the case of citation commands and finding the list of files containing the key in their name is not hard and I think will actually be simpler than how it is done. It might be slow though

bdarcus:
> > Yes, except it's possible I might have the entry field, but prefer to ignore it.
> > BTW, on my point about generalizing "PDF", it came from these issues, which raises a good point.
> 
> Here it seems like the best ui would be to gather a list of all related and prompt the user to select one if we find more than one. Like the second prompt in the case of citation commands and finding the list of files containing the key in their name is not hard and I think will actually be simpler than how `bibtex-completions` does it. It might be slow though.

Maybe also, as I think BC does, have a list of extensions that are valid, with PDF an obvious default.

aikrahguzar:
> ```emacs-lisp
> (defun bibtex-actions-find-files-in-library (key) ; or key-or-entry, as in BC
>   "Search `bibtex-actions-library-path' for file(s).
> 
> Return file paths where file names are composed of KEY plus
> 'bibtex-actions-file-extensions". 
> ```

I think this should more or less work for this
```elisp
(defcustom bibtex-actions-file-extensions '("pdf")
  "List of possible extensions for files related to an entry"
  :group 'bibtex-actions
  :type '(repeat string))

(defun bibtex-actions--files-for-key (key)
  (cl-flet ((possible-file-names-with-extension (extension)
                                                 (seq-map
                                                  (lambda (directory)
                                                    (file-name-concat directory
                                                                      (concat key "." extension)))
                                                  (bibtex-actions--normalize-paths bibtex-completion-library-path))))
    (seq-filter #'file-exists-p (seq-mapcat #'possible-file-names-with-extension 
                                            bibtex-actions-file-extensions))))

```
The function can be written more succinctly if we use `-zip` but I don't know what you think of extra dependencies. You can change the names though, I just didn't change them to agree with message I am replying to.

aikrahguzar:
> Probably the `completing-read` (or CRM) prompt would come in the open function?

I think so

bdarcus:
> > ```emacs-lisp
> > (defun bibtex-actions-find-files-in-library (key) ; or key-or-entry, as in BC
> >   "Search `bibtex-actions-library-path' for file(s).
> > 
> > Return file paths where file names are composed of KEY plus
> > 'bibtex-actions-file-extensions". 
> > ```
> 
> I think this should more or less work for this
> 
> ```emacs-lisp
> (defcustom bibtex-actions-file-extensions '("pdf")
>   "List of possible extensions for files related to an entry"
>   :group 'bibtex-actions
>   :type '(repeat string))
> 
> (defun bibtex-actions--files-for-key (key)
>   (cl-flet ((possible-file-names-with-extension (extension)
>                                                  (seq-map
>                                                   (lambda (directory)
>                                                     (file-name-concat directory
>                                                                       (concat key "." extension)))
>                                                   (bibtex-actions--normalize-paths bibtex-completion-library-path))))
>     (seq-filter #'file-exists-p (seq-mapcat #'possible-file-names-with-extension 
>                                             bibtex-actions-file-extensions))))
> ```
> 
> The function can be written more succinctly if we use `-zip` but I don't know what you think of extra dependencies. You can change the names though, I just didn't change them to agree with message I am replying to.
> 
> And I haven't tested it beyond that it correctly returns `nil` since I don't have `bibtex-completion-library-path` set

I"m out of my element on this, but would one of the other newer seq functions work?

Like `seq-mapn`?

bdarcus:
Also, just occurred to me (not having yet looked at your code) that we can
use the same function for pdf, etc files and notes; just add a PATHS arg?

bdarcus:
Or even KEY, PATHS, EXT?

aikrahguzar:
> Or even KEY, PATHS, EXT?

This would be it,
```elisp
(defun bibtex-actions--files-for-key (key dirs extensions)
  "Find files related to KEY in DIRS with extension in EXTENSIONS"
  (cl-flet ((possible-file-names-with-extension (extension)
                                                (seq-map
                                                 (lambda (directory)
                                                   (file-name-concat directory
                                                                     (concat key "." extension)))
                                                 dirs)))
    (seq-filter #'file-exists-p
                (seq-mapcat #'possible-file-names-with-extension 
                            extensions))))
```
BTW `-zip` or `seq-mapn` wouldn't do it since those are element-wise.

bdarcus:
When I'm back at a computer, I'm going to create a separate
bibtex-actions-utils.el file to hold all of these, and then require it in
the main file.

That will also address load order issues.

bdarcus:
What package am I requiring for the file stuff?

bdarcus:
> files is one, but I don't see anything for file-name-concat.

I just pushed the start of the separate utils.el file etc, but this line is wrong; wasn't sure initially how to fix it.

bdarcus:
We might want to use `expand-file-name` or some such here?

https://stackoverflow.com/questions/3964715/what-is-the-correct-way-to-join-multiple-path-components-into-a-single-complete

bdarcus:
OK, think I fixed it.

Let me know WYT?

Also, what functions from the notify work might be moved into utils? All of them, for example?

aikrahguzar:
> OK, think I fixed it.
> 
> Let me know WYT?

Looks good!

> Also, what functions from the notify work might be moved into utils? All of them, for example?

`bibtex-actions-filenotify-setup` is an autload so that should probably be left in the main file. I actually don't understand the emacs loading that well. But those functions refer to variables defined in the `bibtex-extensions.el`. I guess you can try moving the all except `bibtex-actions-filenotify-setup` to see if it works.

bdarcus:
> `bibtex-actions-filenotify-setup` is an autload so that should probably be left in the main file. I actually don't understand the emacs loading that well. But those functions refer to variables defined in the `bibtex-extensions.el`. I guess you can try moving the all except `bibtex-actions-filenotify-setup` to see if it works.

I think I can get it to work, but I first want to solve one annoying detail with the current code. I get this error:

```console
 bibtex-actions.el    40   1 error           Cannot open load file: No such file or directory, bibtex-actions-utils (emacs-lisp)
```

Any ideas?

aikrahguzar:
> > `bibtex-actions-filenotify-setup` is an autload so that should probably be left in the main file. I actually don't understand the emacs loading that well. But those functions refer to variables defined in the `bibtex-extensions.el`. I guess you can try moving the all except `bibtex-actions-filenotify-setup` to see if it works.
> 
> I think I can get it to work, but I first want to solve one annoying detail with the current code. I get this error:
> 
> ```
>  bibtex-actions.el    40   1 error           Cannot open load file: No such file or directory, bibtex-actions-utils (emacs-lisp)
> ```
> 
> Any ideas?
> 
> Edit:could be a false positive though, as it compiles fine.

Maybe try `doom build -r`? This seems like straight not being aware of the file.

aikrahguzar:
BTW without `bibtex-completion` we will need
```elisp
(require 'dash)
(require 's)
(require 'filenotify)
```
Out of these `'dash` can be removed easily, there is single use of `(-cons* a b c)` which can be replaced with `(cons a (cons b c))`. But `s-format` seems non-trivial enough that it seems better to keep it.

bdarcus:
OK.

With the new file, I can move back to defaulting to bibtex-completion variables on the new path defcustoms.

Any reason not to?

And then ` bibtex-actions--global-files-to-cache ` is no longer necessary?

bdarcus:
> Out of these `'dash` can be removed easily, there is single use of `(-cons* a b c)` which can be replaced with `(cons a (cons b c))`. But `s-format` seems non-trivial enough that it seems better to keep it.

Latest commit implements this change.

I'm not sure if this PR will completely remove the bibtex-completion requirement, but it's a big step.

bdarcus:
OK, fix a few issues, and with latest commit, everything appears to work correctly now.

WDYT: is this ready to merge?

A second step could be to replace the bibtex-completion open functions; e.g. expand the bibtex-actions commands to use this new stuff.

aikrahguzar:
> OK.
> 
> With the new file, I can move back to defaulting to bibtex-completion variables on the new path defcustoms.
> 
> Any reason not to?

I don't see any.
 
> And then `bibtex-actions--global-files-to-cache` is no longer necessary?

The only only this it does with that change is to pass `bibtex-actions-bibliograohy` through `bibtex-actions--normalize-paths`, which for it is essentially `file-truename`. I don't think it matters but I am not sure.

aikrahguzar:
> OK, fixed a few issues, and with latest commit, everything appears to work correctly now.
> 
> WDYT: is this ready to merge?

This is going to break pdf lookup for people using something like Zotero I think. I don't use that functionality and I am not sure how important it is to people.

> A second step could be to replace the bibtex-completion open functions; e.g. expand the bibtex-actions commands to use this new stuff.

bdarcus:
> This is going to break pdf lookup for people using something like Zotero I think. I don't use that functionality and I am not sure how important it is to people.

I'll sleep on it.

BTW, I just added a single "extensions" defcustom, that I'm thinking should apply to notes and files.

Is that what you had in mind too?

aikrahguzar:
> BTW, I just added a single "extensions" defcustom, that I'm thinking should apply to notes and files.
> 
> Is that what you had in mind too?

It seems fine at the moment and I think should be easy to change if people need it different for some reason.

Also at least the definition of `bibtex-actions--global-files-to-cache` should be changed to reflect the new default even if it isn't removed.

bdarcus:
> Also at least the definition of `bibtex-actions--global-files-to-cache` should be changed to reflect the new default even if it isn't removed.

In the recent changes, I readded that step to the defcustom, since there should no longer be any load-order issues, which is why I was thinking to remove this function.

WDYT?

bibtex-actions.el
```diff
@@ -259,50 +321,58 @@ key associated with each one."
...
-          (propertize candidate-suffix 'face 'bibtex-actions) " "
-          (propertize candidate-hidden 'invisible t)))
-        citekey)))))
+    (cl-loop for candidate being the hash-values of (parsebib-parse files)
+             collect
+             (let* ((files
+                     (when (bibtex-actions--files-for-key
+                            (bibtex-actions-get-value "=key=" candidate)
+                            bibtex-actions-library-paths bibtex-actions-file-extensions)
+                           " has:files"))
```
aikrahguzar, line 330:
```suggestion
                     (when (or (assoc "=has-pdf=" (cdr candidate))
                          (bibtex-actions--files-for-key
                            (bibtex-actions-get-value "=key=" candidate)
                            bibtex-actions-library-paths bibtex-actions-file-extensions)
                           " has:files"))
```

bibtex-actions.el
```diff
@@ -259,50 +321,58 @@ key associated with each one."
...
+                    (notes
+                     (when (bibtex-actions--files-for-key
+                            (bibtex-actions-get-value "=key=" candidate)
+                            bibtex-actions-notes-paths bibtex-actions-file-extensions)
+                           " has:notes"))
```
aikrahguzar, line 335:
```suggestion
                     (when (or (assoc "=has-note=" (cdr candidate))
                     (bibtex-actions--files-for-key
                            (bibtex-actions-get-value "=key=" candidate)
                            bibtex-actions-notes-paths bibtex-actions-file-extensions)
                           " has:notes"))
```

aikrahguzar (commented):
I think we should still be checking for in-entry pdf and notes.

bibtex-actions.el
bdarcus: Probably, but then should use `bibtex-actions-get-value` there, since that will properly deal with the tex formats and json.

aikrahguzar:
> > The only only this it does with that change is to pass bibtex-actions-bibliograohy through bibtex-actions--normalize-paths,
> 
> > Also at least the definition of `bibtex-actions--global-files-to-cache` should be changed to reflect the new default even if it isn't removed.
> 
> In the recent changes, I readded that step to the defcustoms, since there should no longer be any load-order issues, which is why I was thinking to remove this function.
> 
> The normalize function also ensures the paths are a list.
> 
> WDYT?

We no longer need to ensure it is a list of path so I am on the side of removing it.

bibtex-actions.el
bdarcus: Actually, I don't think this will work, as the files don't have this info.

bibtex-actions.el
bdarcus: Actually, this isn't right. Need to check on this.

bibtex-actions.el
aikrahguzar: Which info? I am a bit confused. The changes just check for a pdf in the bib entry in addition to finding the files in the directories we are searching

bdarcus: I am referring to notes, which are only stored external to the files.

aikrahguzar: > I am referring to notes, which are only stored external to the files.

Same with pdfs. I though this was because the entry can contain the locations of related pdfs or notes.

bdarcus: > > I am referring to notes, which are only stored external to the files.
> 
> Same with pdfs. I though this was because the entry can contain the locations of related pdfs or notes.

Sorry; AFAIK, there are no equivalent fields for notes.

But I guess I should confirm that.

aikrahguzar: > Sorry; AFAIK, there are no equivalent fields for notes.

Oh, I was just going by what was there before.

bdarcus:
> We no longer need to ensure it is a list of path so I am on the side of removing it.

Done https://github.com/bdarcus/bibtex-actions/pull/222/commits/e189bd49fe32af89b5d44342d81abd0dfe355003

bibtex-actions-utils.el
```diff
@@ -0,0 +1,171 @@
...
+              (mapc #'file-notify-rm-watch bibtex-actions--local-watches)
+              (setq bibtex-actions--local-watches 'uninitialized))
+            nil t))
+
+(defun bibtex-actions--filenotify-files ()
+  "Get the list of files to watch from `bibtex-actions-filenotify-files'."
+  (seq-mapcat (lambda (x)
+                (bibtex-actions--normalize-paths
+                 (cl-case x
+                   (bibliography (bibtex-actions--global-files-to-cache))
```
aikrahguzar, line 128:
```suggestion
                   (bibliography bibtex-actions-bibliography)
```

bibtex-actions.el
```diff
@@ -229,26 +280,36 @@ offering the selection candidates"
...
+           (bibtex-completion-find-local-bibliography))))
+    (seq-difference local-bib-files bibtex-actions-bibliography)))
+
+(defun bibtex-actions-get-value (field item &optional _default)
+  "Return biblatex FIELD value for ITEM."
+  (or (cdr (assoc-string field item 'case-fold))
+      (cl-loop for fname in (cdr (assoc field bibtex-actions-field-map))
+               when (cdr (assoc-string fname item 'case-fold))
+                         return (cdr (assoc-string fname item 'case-fold)))
+      ""))
```
aikrahguzar, line 297: Why are we returning empty string instead of `nil` of the value is not found? It seems strange and as it is is causing the candidates to display as if all of them have a pdf.

And the optional `_default` argument is unused. Maybe both are related?

bdarcus: > Why are we returning empty string instead of nil of the value is not found?

I can try again, but I think I ran into errors when I didn't do this.

> It seems strange and as it is is causing the candidates to display as if all of them have related files.

That's weird; I don't see that.

image

> And the optional _default argument is unused.

I think that's a vestige of what started out as a BC function.

bdarcus: > I can try again, but I think I ran into errors when I didn't do this.

I did, but it seems a bug in `bibtex-actions-shorten-names`:

```console
bibtex-actions-shorten-names: Wrong type argument: characterp, ""
```

aikrahguzar: > That's weird; I don't see that.

Maybe evaluate `bibtex-actions--format-candidates` again,  refresh the cache and then check. It wouldn't show until a cache refresh

bdarcus: What's really weird is now I see this, after re-adding the `""`.

> It seems strange and as it is is causing the candidates to display as if all of them have related files.

I am guessing the problem is in the format-entry function.

aikrahguzar: > I am guessing the problem is in the format-entry function.

Yes, it seems like there a string is always expected. One way of dealing with that would be to replace `""` with "_default" and then pass the empty string in the format-entry function

bdarcus: > Yes, it seems like there a string is always expected. One way of dealing with that would be to replace `""` with `_default` and then pass the empty string in the format-entry function.

I'm not seeing where to do the latter ATM, as I'm looking at BC.

aikrahguzar: > I'll look at it again tomorrow. It's probably something obvious.

Sounds good. 

I am not seeing any errors but I have an example of a bib entry causing them, send them my way, though I will mostly disappear from tomorrow.

bdarcus: I may have just fixed it, and it was indeed a simple thing.

bdarcus:
Hey @salutis - can you look at this and let me know WYT?

In the process of implementing this, we also added some functionality that points the way to addressing some of your earlier points about PDFs etc.

bdarcus:
This commit rewrites open-pdf as open-library-files: https://github.com/bdarcus/bibtex-actions/pull/222/commits/1b87cf580b0b4ef41a9081fbce7e6551a54a94bc.

A lot of open questions about how best to do this, but this basically runs `find-fille` (by default, but this function is configurable) on any relevant files, without user intervention.

bibtex-actions-utils.el
```diff
@@ -0,0 +1,182 @@
...
+(defun bibtex-actions-open-files (keys dirs)
+  "Open files related to KEY in DIRS."
+  (cl-loop for key in keys do
+           (let ((files
+                  (bibtex-actions--files-for-key
+                   key
+                   dirs
+                   bibtex-actions-file-extensions)))
+             (cl-loop for file in files do
+                      (funcall bibtex-actions-open-file-function file)))))
```
bdarcus, line 70: Need to add code here when a file isn't found; e.g:

```elisp
       (message "No file(s) found for this entry: %s"
                 key))
```

aikrahguzar:
@bdarcus  what is the last `cdr` in `bibtex-actions--process-display-formats` supposed to be? 
a) sum of all widths in the format string?
b) sum of all widths in the format string + number of sections + 2

It seems to be the seond one experimentally but I don't know why and the implementation is very convoluted and should be simplified, I went with the first one and don't see any adverse effects and changing to second one is easy too. 

I will pr that along with a couple of improvements to the filenotify stuff (I think I am done with it after this) after I know the answer.

bdarcus:
> @bdarcus what is the last `cdr` in `bibtex-actions--process-display-formats` supposed to be?

TBH, am not sure, as it's been awhile that I was working on that, and I struggled with it; namely the multiple templates were initially causing rendering problems.

It's likely if the rendering is fine, even when you have content in the suffix, then it's fine.

aikrahguzar:
> TBH, am not sure, as it's been awhile that I was working on that, and I struggled with it; namely the multiple templates were initially causing rendering problems.
> 
> It's likely if the rendering is fine, even when you have content in the suffix, then it's fine.

Looking around that data is entirely unused in the package. It is ultimately passed to the `bibtex-actions--foramat-entry` which ignores it. Compared to `bibtex-completion` the difference is we don't support the `*` so all our fields have to be fixed length. 

The maximum lengths of main body and suffix are determined by
```elisp
         (main-width (truncate (* (frame-width) 0.65)))
         (suffix-width (truncate (* (frame-width) 0.34)))
```
in `bibtex-completions-format-entry`.

I removed `bibtex-actions--process-display-formats` and calls to it and nothing changes even. Both main and suffix are populated in these entries.

bdarcus:
> > TBH, am not sure, as it's been awhile that I was working on that, and I struggled with it; namely the multiple templates were initially causing rendering problems.
> > It's likely if the rendering is fine, even when you have content in the suffix, then it's fine.
> 
> Looking around that data is entirely unused in the package. It is ultimately passed to the `bibtex-actions--foramat-entry` which ignores it. Compared to `bibtex-completion` the difference is we don't support the `*` so all our fields have to be fixed length.

I do think the `*` is currently supported; it's just there can only be one.

> The maximum lengths of main body and suffix are determined by
> 
> ```emacs-lisp
>          (main-width (truncate (* (frame-width) 0.65)))
>          (suffix-width (truncate (* (frame-width) 0.34)))
> ```
> 
> in `bibtex-completions-format-entry`.
> 
> I removed `bibtex-actions--process-display-formats` and calls to it and nothing changes even. Both main and suffix are populated in these entries.

OK. And there's no problem with the width exceeding the frame or anything?

bdarcus:
BTW, it's possible in the future, all of that will get rewritten and greatly simplified, IF minad gets around to upstreaming the idea for a new `decoration-function` to replace affixation.

aikrahguzar:
> I do think the `*` is currently supported; it's just there can only be one.

In BC the '*' support comes down to this
```elisp
(if (> field-width 0)
                field-width
              (- width (cddr format)))
```
We don't use `(cddr format)` at all `ba--format-entry`
 
> OK. And there's no problem with the width exceeding the frame or anything?

Width exceeding is the frame is possible even as is. The only restrictions on the widths are that 
1) width of a single field of main template  exceed 2/3 of frame width
2) width of a single field of suffix template can't exceed 1/3 of frame width

But same thing will happen with BC if the frame width is small enough, some fields will get cut off. There is no mechanism in either for keeping all the fields visible but truncated if there isn't enough space. This is just a consequence of width in format being absolute instead of being relative to frame width.

aikrahguzar:
>     1. width of a single field of main template  exceed 2/3 of frame width
> 
>     2. width of a single field of suffix template can't exceed 1/3 of frame width

Actually this isn't correct. We always use the given fields widths unless it is the '*' field. If it is a '*' field it will take up 2/3 of frame width if it is in the main section and 1/3 if it is in the suffix section.

bdarcus:
BTW, I wasn't really explicitly thinking about this with the recent work, but it's consistent with this idea I have the other day.

https://github.com/bdarcus/bibtex-actions/discussions/225

Aside: nobody uses the discussion forum, so I'm thinking to shut it down :-)

bdarcus:
I just updates `files-for-key` to `files-for-keys`: https://github.com/bdarcus/bibtex-actions/pull/222/commits/56f62491ce4ec46bc43aa358e092a436274ba690

Any objections?

I still have a bug I need to fix.

bdarcus:
Ugh ... just realizing you use this function in the revised `format-candidates`.

aikrahguzar:
> Ugh ... just realizing you use this function in the revised `format-candidates`.
> 
> But I don't think that should have any impact, since it's just checking if it returns non-nil.
> 
> EDIT: should now all be fixed.

You can also define `files-for-keys` in terms `files-for-key` and one advantage of that would that if you later want to add additional information to show in open commands like grouping that depends on the key, it would be easier. Overall I am agnostic about the two options.

> On another note, I just ran `package-lint`, and it gives me a lot of warnings like this:
> 
> ```
> 208:0: error: "bibtex-actions-filenotify-refresh" doesn't start with package's prefix "bibtex-actions-utils".
> ```
> 
> ... which I may as well fix.
> 

I see those warnings too. And this change makes sense. Quite a few function already have `bibtex-actions-filenotify` prefix.

bdarcus:
> You can also define `files-for-keys` in terms `files-for-key` and one advantage of that would that if you later want to add additional information to show in open commands like grouping that depends on the key, it would be easier. Overall I am agnostic about the two options.

You mean have two functions there: `files-for-keys` and `files-for-key`?

If yes, that would be more flexible; I can't see a downside.

aikrahguzar:
> You mean have two functions there: `files-for-keys` and `files-for-key`?

Yes. Though I feel names might need to be differentiated a bit more than a single `s`.

> If yes, that would be more flexible; I can't see a downside.

I quite like the style of writing a lot of small functions!

bdarcus:
> Yes. Though I feel names might need to be differentiated a bit more than a single `s`.

Could be:

1. `bibtex-actions-utils--files-for-key`
2. `bibtex-actions-utils--multi-files-for-keys`

aikrahguzar:
Also questions:

1) Why do we have two variables `bibtex-actions-template` and `bibtex-actions-template-suffix` instead of a single `bibtex-actions-template` which is list with elements `(type . (main-template suffix-template))`. With the single variable option it is easier to calculate the widths since the two template are together.

2) What is the reason for allowing multiple templates. I assume to allow a different set of fields to be displayed for different types? But it also allows the possibility of jagged alignment between the entries unless all the templates are compatible. 

When I am free later today, I will do another pr that removes the hardcoded widths given  in
```
(main-width (truncate (* (frame-width) 0.65)))
(suffix-width (truncate (* (frame-width) 0.34)))
```
in `bibtex-actions--format-candidate` and computes these based on template. Though it will still only work flawlessly if like now
a) total widths specified is less than frame width
b) all templates are compatible

bdarcus:
> I would go for `bibtex-actions-utils--files-for-multi-keys` for `bibtex-actions-utils--files-for-multiple-keys`

Yes.

But the args would be the same across the two functions; wouldn't they?

So then maybe better one function with a parameter?

> Why do we have two variables bibtex-actions-template and bibtex-actions-template-suffix instead of a single bibtex-actions-template which is list with elements (type . (main-template suffix-template)). With the single variable option it is easier to calculate the widths since the two template are together.

No good reason.

> What is the reason for allowing multiple templates. I assume to allow a different set of fields to be displayed for different types? But it also allows the possibility of jagged alignment between the entries unless all the templates are compatible.

That's just borrowed from `bibtex-completions` directly. I'm happy to remove it.

BTW, I had a DM conversation with minad yesterday. It looks likely he'll put forward a proposal to replace affixation in Emacs-next with a new `decoration-function`.

The key feature of that if accepted: it will handle display of random data as columns in the UI.

So all this code could be deprecated.

bdarcus:
> But the args would be the same across the two functions; wouldn't they?
> 
> So then maybe better one function with a parameter?

Or may should save this for a separate PR :-)

aikrahguzar:
> > I would go for `bibtex-actions-utils--files-for-multi-keys` for `bibtex-actions-utils--files-for-multiple-keys`
> 
> Yes.
> 
> But the args would be the same across the two functions; wouldn't they?

Why? I was thinking the first argument would be a key for `bibtex-actions-utils--files-for-key`
and a list for kyes for `bibtex-actions-utils--files-for-multiple-keys`?

> So then maybe better one function with a parameter?

I generally dislike functions that accept arguments of multiple types but it can be convenient so I don't mind too much.

> BTW, I had a DM conversation with minad yesterday. It looks likely he'll put forward a proposal to replace affixation in Emacs-next with a new `decoration-function`.
> 
> The key feature of that if accepted: it will handle display of random data as columns in the UI.
> 
> So all this code could be deprecated.

Is this emacs 28 only? In that case I wouldn't say soon.

But it would require a restructuring of how candidates are cached , but I would say that would be better.

bdarcus:
This commit splits the function: https://github.com/bdarcus/bibtex-actions/pull/222/commits/8a5b1d23896a7037f42f096ebbeddb31f097648f.

I have not yet modified format-entry. Coming next.

bibtex-actions-filenotify.el
```diff
@@ -0,0 +1,159 @@
...
+(defun bibtex-actions--invalidate-cache (&optional scope)
+  "Invalidate local or global caches according to SCOPE.
+If it is other than 'global or 'local invalidate both"
+  (unless (eq 'local scope)
+    (setq bibtex-actions--candidates-cache 'uninitialized))
+  (unless (eq 'glocal scope)
+    (setq bibtex-actions--local-candidates-cache 'uninitialized)))
+
+(defun bibtex-actions--make-default-callback (func scope &optional change)
+  "The callback FUNC by SCOPE used to update cache for default options.
```
bdarcus, line 58: @aikrahguzar - in which file should these go?

aikrahguzar: `bibtex-actions--make-default-callback` should be in `bibtex-actions-filenotify`. It is pretty much an  internal detail. 

I don't have a strong opinion on `bibtex-actions--invalidate-cache`. I don't  see a reason to use it over `bibtex-actions-refresh` for anything other the filenotify callback so I think it can be moved to `bibtex-actions-filenotify` too.

bdarcus:
I think I'm going to also create a separate `bibtex-actions-files.el` file, and split the `open-files` function.

bibtex-actions-filenotify.el
aikrahguzar: I think even moving two `defcustoms` and `bibtex-actions-filenotify-setup` as the only autload of the file is an option. The only minor downside I see is that it will require a separate use-package call to configure it in that case. On the positive side it will make the `filenotify` stuff self-contained and we can remove the duplicate `defvar`s from `bibtex-actions` in `bibtex-actions-filenotify`, remove `(require filenotify)` in `bibtex-actions` and instead do `(require bibtex-actions)` in `bibtex-actions-filenotify` which seems a more logical dependency.

bdarcus:
> I think I'm going to also create a separate `bibtex-actions-files.el` file, and split the `open-files` function.

I'm starting to work on this, but if you have any feedback, here's the skeleton: https://github.com/bdarcus/bibtex-actions/pull/222/commits/0a5f5c5f70b40f34ac0bd11cf6ed01fffc46be08.

My plan is for `bibtex-actions-open` to become a function that includes groups of files, notes and links.

bdarcus:
OK, good progress, but stuck on a little thing:

I need to `open-notes` function to create a new file if a file doesn't exist.

Best way to adapt the code to do that?

bibtex-actions-filenotify.el
bdarcus: > I think even moving two `defcustoms` and `bibtex-actions-filenotify-setup` as the only autload of the file is an option. The only minor downside I see is that it will require a separate use-package call to configure it in that case. On the positive side it will make the `filenotify` stuff self-contained and we can remove the duplicate `defvar`s from `bibtex-actions` in `bibtex-actions-filenotify`, remove `(require filenotify)` in `bibtex-actions` and instead do `(require bibtex-actions)` in `bibtex-actions-filenotify` which seems a more logical dependency.

I just did this in the latest commit.

Aside from the question I have below, I think this is ready for a look, if and when you have time.

I think this is much cleaner.

aikrahguzar: > Aside from the question I have below, I think this is ready for a look, if and when you have time.

I don't see any question?

bdarcus:
`bibtex-actions-open` with grouping!

image

Now just need to figure out how to add links.

aikrahguzar:
> OK, good progress, but stuck on a little thing:
> 
> I need to `open-notes` function to create a new file if a file doesn't exist.
> 
> Best way to adapt the code to do that?
> 
> Edit: with this commit, I have a simple default, and users can substitute `orb-edit-notes` for the function (still needs work though).
> 
> [28ba351](https://github.com/bdarcus/bibtex-actions/commit/28ba3510e50d92325f3c6788030a70e36a587a53)



> OK, good progress, but stuck on a little thing:
> 
> I need to `open-notes` function to create a new file if a file doesn't exist.
> 
> Best way to adapt the code to do that?

Is there anything need beyond calling `find-file` on the path? If you call `find-file` on a non-existent buffer it will create it. Since the name of the file is just supposed to be the key we only need to determine the extension and emacs will take care of the rest. We can even offer the possible extension in the `file-extensions` variable with something to indicate that they will be created.

aikrahguzar:
> `bibtex-actions-open` with grouping!
> 
> image
> 
> Now just need to figure out how to add links.

This looks great!

bibtex-actions-filenotify.el
```diff
@@ -41,20 +42,45 @@
 (defvar bibtex-actions-open-file-function)
 (defvar bibtex-actions-file-extensions)
```
aikrahguzar: I don't think these should be here

bdarcus: Oh yeah; an oversight.

bdarcus:
> Is there anything need beyond calling `find-file` on the path? If you call `find-file` on a non-existent buffer it will create it.

Right, but the path is removed from the files-for-key if it doesn't exist.

bibtex-actions-filenotify.el
bdarcus: actually, I fixed that it seems.

aikrahguzar:
> Right, but the path is removed from the files-for-key if it doesn't exist.

I guess then we can take `possible-file-names-with-extension` out of `cl-flet` into its own `defun`. I think this is a start
```lisp 
(defun bibtex-actions-file-possible-names (key dirs exntensions)
  "Possible names for files correponding to KEY with EXTENSIONS in DIRS"
  (seq-map
   (lambda (directory)
     (expand-file-name
      (concat key "." extension) directory))
   dirs))

(defun bibtex-actions-file--files-for-key (key dirs extensions)
  "Find files related to KEY in DIRS with extension in EXTENSIONS."
  (cl-flet ((possible-file-names-with-extension (extension)
              (bibtex-actions-file-possible-names key extension dirs)))
    (seq-filter #'file-exists-p
                (seq-mapcat #'possible-file-names-with-extension
                            extensions))))

(defun bibtex-actions-file--files-for-multiple-keys (keys dirs extensions)
  "Find files related to a list of KEYS in DIRS with extension in EXTENSIONS."
  (cl-flet ((files-for-key (key)
                           (let ((existing-files (bibtex-actions-file--files-for-key key dirs extensions)))
                             (if existing-files
                                 (seq-map (lambda (file) (cons file 'exists)) existing-files)
                               (seq-map (lambda (file) (cons file 'new))
                                        (bibtex-actions-file-possible-names key dirs extensions))))))
    (seq-mapcat #'files-for-key keys)))

```
`bibtex-actions-file--files-for-key` is exactly as before just written using new function. `bibtex-actions-file--files-for-multiple-keys` is changed to return a list of conses. If it finds existing files for a key it returns them consed with `'exists` and if no existing file is found it return possible paths for them consed with `new`.

aikrahguzar:
Also I am getting warnings to change the `emacs` versions to `25.1` in both `ba-filenotify` and `ba-files`

bibtex-actions-filenotify.el
```diff
@@ -0,0 +1,200 @@
...
+;; Modified: August 17, 2021
+;; Version: 0.0.1
+;; Keywords: abbrev bib c calendar comm convenience data docs emulations extensions faces files frames games hardware help hypermedia i18n internal languages lisp local maint mail matching mouse multimedia news outlines processes terminals tex tools unix vc wp
+;; Homepage: https://github.com/bruce/bibtex-actions-filenotify
+;; Package-Requires: ((emacs "24.3"))
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; Commentary:
+;;
```
aikrahguzar, line 17:
```suggestion
;; A companion to the bibtex-actions for auto-invalidation and auto-refreshing
;; of cache when a bib file or a related file such notes directory or library
;; changes. Uses filenotify api to acheive this.
```

bibtex-actions-filenotify.el
```diff
@@ -0,0 +1,200 @@
...
+;; Copyright (C) 2021 Bruce D'Arcus
+;;
+;; Author: Bruce D'Arcus <https://github.com/bdarcus>
+;; Maintainer: Bruce D'Arcus <bdarcus@gmail.com>
+;; Created: August 17, 2021
+;; Modified: August 17, 2021
+;; Version: 0.0.1
+;; Keywords: abbrev bib c calendar comm convenience data docs emulations extensions faces files frames games hardware help hypermedia i18n internal languages lisp local maint mail matching mouse multimedia news outlines processes terminals tex tools unix vc wp
+;; Homepage: https://github.com/bruce/bibtex-actions-filenotify
+;; Package-Requires: ((emacs "24.3"))
```
aikrahguzar, line 12:
```suggestion
;; Package-Requires: ((emacs "25.1"))
```
I am getting a warning with "24.3".

bibtex-actions-filenotify.el
```diff
@@ -0,0 +1,200 @@
...
+(defvar bibtex-actions-bibliography)
+(defvar bibtex-actions-library-paths)
+(defvar bibtex-actions-notes-paths)
+(defvar bibtex-actions-create-note-function)
+(defvar bibtex-actions-filenotify-files)
+(defvar bibtex-actions--candidates-cache)
+(defvar bibtex-actions--local-candidates-cache)
+(defvar bibtex-actions-filenotify-callback)
+(defvar bibtex-actions-open-file-function)
+(defvar bibtex-actions-file-extensions)
```
aikrahguzar, line 43:
```suggestion
```
`(require 'bibtex-actions)` should take care of these being defined and most aren't needed.

bibtex-actions-filenotify.el
```diff
@@ -0,0 +1,200 @@
...
+function can run several times without adding duplicate watches."
+  (when (eq 'uninitialized bibtex-actions-filenotify--local-watches)
+    (bibtex-actions-filenotify--add-local-watches))
+  (add-hook 'kill-buffer-hook
+            (lambda ()
+              (mapc #'file-notify-rm-watch bibtex-actions-filenotify--local-watches)
+              (setq bibtex-actions-filenotify--local-watches 'uninitialized))
+            nil t))
+
+(defun bibtex-actions--filenotify-files ()
```
aikrahguzar, line 137:
```suggestion
(defun bibtex-actions-filenotify--files ()
```
Taking care of package prefix warning.

bibtex-actions-filenotify.el
```diff
@@ -0,0 +1,200 @@
...
+watches have to be removed manually. To remove them call
+`bibtex-actions-rm-global-watches'"
+  (setq bibtex-actions-filenotify--global-watches
+        (seq-map
+         (lambda (bibfile)
+           (file-notify-add-watch
+            bibfile '(change)
+            (lambda (x)
+              (bibtex-actions-filenotify--callback 'global x))))
+         (bibtex-actions--filenotify-files))))
```
aikrahguzar, line 161:
```suggestion
         (bibtex-actions-filenotify--files))))
```
Propagating the change

bdarcus:
> Also I am getting warnings to change the emacs versions to 25.1 in both ba-filenotify and ba-files.

I accidentally editing your comment. Sorry.

The version numbers were an oversight also; I'll change to 26.3, and perhaps even 27.1, since Ubuntu now uses that, and Doom and Vertico require it.

bdarcus:
> I guess then we can take `possible-file-names-with-extension` out of `cl-flet` into its own `defun`. I think this is a start

There's a little problem with that first function; still trying to figure it out.

aikrahguzar:
> > I guess then we can take `possible-file-names-with-extension` out of `cl-flet` into its own `defun`. I think this is a start
> 
> There's a little problem with that first function; still trying to figure it out.

Sorry, that was me trying to write code too late and not testing it. That snippets is completely messed up. This works

```elisp
(defun bibtex-actions-file--possible-names (key dirs extensions)
  "Possible names for files correponding to KEY with EXTENSIONS in DIRS"
  (cl-flet ((possible-file-names-with-extension
             (extension)
             (seq-map
              (lambda (directory)
                (expand-file-name
                 (concat key "." extension) directory))
              dirs)))
    (seq-mapcat #'possible-file-names-with-extension
                extensions)))

(defun bibtex-actions-file--files-for-key (key dirs extensions)
    "Find files related to KEY in DIRS with extension in EXTENSIONS."
    (seq-filter #'file-exists-p
                (bibtex-actions-file--possible-names key dirs extensions)))

(defun bibtex-actions-file--files-to-open-or-create (keys dirs extensions)
  "Find files related to a list of KEYS in DIRS with extension in EXTENSIONS."
  (cl-flet ((files-for-key (key)
                           (let ((existing-files (bibtex-actions-file--files-for-key key dirs extensions)))
                             (if existing-files
                                 (seq-map (lambda (file) (cons file 'exists)) existing-files)
                               (seq-map (lambda (file) (cons file 'new))
                                        (bibtex-actions-file--possible-names key dirs extensions))))))
    (seq-mapcat #'files-for-key keys)))

```

bdarcus:
Thank you!

I've merged your PR, and applied this code as well.

Also added open-link support here: https://github.com/bdarcus/bibtex-actions/pull/222/commits/51203f1c154d181954071e900a420ffab83ade6c

.. which also required a change in the candidate list to include the full data, and then a `bibtex-actions-get-entry` function to access the cached data (for example, for link lookup).

There are a few other tweaks I'd like to make, but this should hopefully be ready to merge sometime today.

bdarcus:
Hmm ... I've uncovered a bug that has cropped up on this branch: embark-act at point returns nil. Not sure what change could have caused that.

aikrahguzar:
> Thank you!
> 
> I've merged your PR, and applied this code as well.
> 
> Also added open-link support here: [51203f1](https://github.com/bdarcus/bibtex-actions/commit/51203f1c154d181954071e900a420ffab83ade6c)
> 
> ... which also required a change in the candidate list to include the full data, and then a `bibtex-actions-get-entry` function to access the cached data (for example, for link lookup).

I am somewhat concerned about the memory implications of this change. Since the bibtex entries can be huge and as far as I know can contain whole abstracts. Does parsebib only parse a limited set of predefind keys or everything? It would be better to keep only those key value pairs which we need, though this can wait for the actual complaints. On the positive side this move us towards the kind of cache we would need if cr adds a columnar view.

> There are a few other tweaks I'd like to make, but this should hopefully be ready to merge sometime today.

Sounds great!

bdarcus:
> I am somewhat concerned about the memory implications of this change. Since the bibtex entries can be huge and as far as I know can contain whole abstracts. Does parsebib only parse a limited set of predefind keys or everything? It would be better to keep only those key value pairs which we need, though this can wait for the actual complaints. 

Here's what a candidate looks like now, and all I do is preserve these data in the cached candidate list.

```elisp
(#("Low                              2001      The Edge and the Center                                    low2001            article                                                                                                       has:notes has:link  low2001" 0 91
   (face bibtex-actions-highlight)
   92 228
   (face bibtex-actions)
   229 258
   (invisible t))
  ("subtitle" . "gated communities and the discourse of urban fear")
  ("title" . "The Edge and the Center")
  ("number" . "1")
  ("langid" . "english")
  ("abstract" . "Across America, middle-class and upper-middle-class gated communities are creating new forms of exclusion and residential segregation, exacerbating social cleavages that already exist (Blakely and Snyder 1997; Higley 1995; Lang and Danielson 1997; Marcuse 1997). While historically secured and gated communities were built in the United States to protect estates and to contain the leisure world of retirees, these urban and suburban developments now target a much broader market, including families with children (Guterson 1992; Lofland 1998). This retreat to secured enclaves with walls, gates, and guards materially and symbolically contradicts American ethos and values, threatens public access to open space, and creates yet another barrier to social interaction, building of social networks, as well as increased tolerance of diverse cultural/ racial/social groups (Davis 1992;Devine 1996;Etzoni 1995; Judd 1995; McKenzie 1994). In this paper, I explore how the discourse of fear of violence and crime and the search for a secure community by those who live in gated communities in the United States legitimates and rationalizes class-based exclusion strategies and residential segregation. I examine whether residents of cities experiencing increasing cultural diversity are fleeing neighborhoods because they have experienced a \"loss of place\" and therefore feel unsafe and insecure (Altaian and Low 1992). Some people are responding to this loss by choosing to buy into a defensive space, a walled and guarded community that they can call home, [gated communities, United States, urban fear]")
  ("urldate" . "2016-02-18")
  ("doi" . "10.1525/aa.2001.103.1.45")
  ("issn" . "1548-1433")
  ("pages" . "45--58")
  ("volume" . "103")
  ("journaltitle" . "American Anthropologist")
  ("date" . "2001-03-01")
  ("author" . "Low, Setha M.")
  ("shorttitle" . "The Edge and the Center")
  ("=type=" . "article")
  ("=key=" . "low2001"))
```

I can check if we can exclude the abstracts upfront, as I don't think we need them.

> On the positive side this move us towards the kind of cache we would need if cr adds a columnar view.

Yes.

bdarcus:
> I can check if we can exclude the abstracts upfront, as I don't think we need them.

You can explicitly tell it what fields you want, but I don't see a way to tell it what fields to leave out.

```elisp
 (parsebib-parse '("~/org/bib/newer.bib") :fields '("title" "doi" "author"))
```

WDYT; should I ask for that?

aikrahguzar:
> You can explicitly tell it what fields you want, but I don't see a way to tell it what fields to leave out.
> 
> ```emacs-lisp
>  (parsebib-parse '("~/org/bib/newer.bib") :fields '("title" "doi" "author"))
> ```
> 
> WDYT; should I ask for that?

No, we should be able to figure out the fields need from the templates in `bibtex-actions-template` and `bibtex-actions-template-suffix`. I will do try that later but I think it shouldn't be a pressing concern. It can be separate pr later.

bdarcus:
> I will do try that later but I think it shouldn't be a pressing concern. It can be separate pr later.

OK, I opened a new linked issue for this, with a key tip.

aikrahguzar:
> I know you haven't been working with that code, but can you see anything?
> 
> If not, I may just merge this anyway and fix it in a subsequent commit.

I looked briefly but I am not too familiar with how embark works and don't have embark configuration in my config. But is the problem with `bibtex-actions-citation-key-at-point` ? It seems to be working in `latex`. I made it interactive and changed it to message the key and it does that. Though it seems it splits the key at `:`

bdarcus:
> I looked briefly but I am not too familiar with how embark works and don't have embark configuration in my config. But is the problem with `bibtex-actions-citation-key-at-point` ? It seems to be working in `latex`. I made it interactive and changed it to message the key and it does that.

I did the same!

> Though it seems it splits the key at `:` .

That uses whatever bibtex-completion function for latex.

> That leaves `bibtex-actions-get-key-org-cite` which is entirely only an org thing independent of the rest of package.

And that seems to work correctly too; nothing changed there.

> Maybe a problem with embark config?

I don't think so, because it works if I check out the main branch.

aikrahguzar:
> I don't think so, because it works if I check out the main branch.

Maybe a name that embark config uses changed between the two branches?

bdarcus:
> > I don't think so, because it works if I check out the main branch.
> 
> Maybe a name that embark config uses changed between the two branches?

I checked that, but haven't found anything yet.

aikrahguzar:
> I checked that, but haven't found anything yet.

I evaluated `(add-to-list 'embark-target-finders 'bibtex-actions-citation-key-at-point)` and `emabrk-act` finds the key to act on in `latex` with the same split at `:` due to bibtex-completion.

bdarcus:
> > I checked that, but haven't found anything yet.
> 
> I evaluated `(add-to-list 'embark-target-finders 'bibtex-actions-citation-key-at-point)` and `emabrk-act` finds the key to act on in `latex` with the same split at `:` due to bibtex-completion.

As in, it fails; right? And that suggests something wrong with `bibtex-completion-get-key-latex`?

But it doesn't get `nil`, which is what I see.

aikrahguzar:
> But it doesn't get `nil`, which is what I see.
Yup, not `nil`.

This is very weird. This is about the only thing that hasn't been touched at all in this branch!

I don't have an org file with citations or I would try a bit more debugging.

bdarcus:
> This is very weird. This is about the only thing that hasn't been touched at all in this branch!

Exactly!

In the meantime, I merged the branch, since this was getting out of hand :-)

> I don't have an org file with citations or I would try a bit more debugging.

I opened a new linked issue (#233) on this, and will post a minimal file if you get time.

You should use embark anyway; it's a core part of this ecosystem ;-)

aikrahguzar:
> I opened a new linked issue (#233) on this, and will post a minimal file if you get time.

I just tried it but that example doesn't see to work for me probably because I don't have org unpinned. It is whatever doom has.

> You should use embark anyway; it's a core part of this ecosystem ;-)

I am aware of embark and I like the idea but I haven't found a use case yet though I use similar features in  doom.

bdarcus:
> > I opened a new linked issue (#233) on this, and will post a minimal file if you get time.
> 
> I just tried it but that example doesn't see to work for me probably because I don't have org unpinned. It is whatever doom has.

Oh right; you're on Doom!

This should work; Doom is already pinned to a version with org-cite:

https://github.com/hlissner/doom-emacs/pull/5290

But we should also get it working again with latex.

aikrahguzar:
> [hlissner/doom-emacs#5290](https://github.com/hlissner/doom-emacs/pull/5290)

I replied on the issue but org is giving me trouble

> But we should also get it working again with latex.

That shouldn't be hard. There the problem is just `bibtex-completion` handing embark only part of the key.

PR #227, opened 2021-08-16
Refactor bibtex-actiona--format-entry

aikrahguzar:
I think I have managed to refactor `bibtex-actiona--format-entry` into a form where it is easier to see what is happening.


bdarcus:
Nice!

PR #228, opened 2021-08-16
bibtex-actions--process-display-formats refactor + filenotify improvements

aikrahguzar:
bibtex-actions--process-display-formats refactor is exactly as before but refactored. 

Misc filenotify improvements, mostly that only the cache that is invalidated is recomputed when offering candidates.

### DIFF
--- a/bibtex-actions-utils.el
+++ b/bibtex-actions-utils.el
@@ -114,8 +114,13 @@ If it is other than 'global or 'local invalidate both"
 
 CHANGE refers to the notify argument."
   (cl-case (cadr change)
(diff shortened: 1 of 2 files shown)


bdarcus:
Thank you!

bdarcus:
Question: should we change `bibtex-actions-filenotify.el` to `bibtex-actions-file-notify.el`, for consistency?

Thinking about this in the context of trying to figure out where I would put the completion functions, if I wanted to put them in a separate file.

I had originally thought `bibtex-actions-select.el`, but `bibtex-actions-select-file` should likely also maybe a little confusing.

bdarcus:
I created #248 to experiment.

PR #232, opened 2021-08-18
Limit fields being parsed.

### LINKED ISSUE #231: limit parsebib fields to those in templates
> You can explicitly tell it what fields you want, but I don't see a way to tell it what fields to leave out.
> 
> ```emacs-lisp
>  (parsebib-parse '("~/org/bib/newer.bib") :fields '("title" "doi" "author"))
> ```
> 
> WDYT; should I ask for that?

No, we should be able to figure out the fields needed from the templates in `bibtex-actions-template` and `bibtex-actions-template-suffix`. I will do try that later but I think it shouldn't be a pressing concern. It can be separate pr later.

_Originally posted by @aikrahguzar in https://github.com/bdarcus/bibtex-actions/pull/222#issuecomment-901104333_

There's a nuance here, however, when you do get to it:

Note how `bibtex-actions-get-value` works in terms of mapping field names across bibtex, biblatex, csl-json.

I am actually thinking of standardizing on the csl names for the templates, but in any case, it would need to include all relevant fields in that mapping table.

This simple approach works (maybe there's an easy flatten function that doesn't use dash?):

```elisp
(parsebib-parse 
  bibtex-actions-bibliography :fields (-flatten bibtex-actions-field-map)))
```

But the mapping variable currently only includes fields where there are differences across the formats.

```elisp
ELISP> (gethash "zukin2009" bibtex-actions-bibliography :fields (-flatten bibtex-actions-field-map)))
(("number" . "1")
 ("journaltitle" . "City \\& Community")
 ("date" . "2009-03-01")
 ("=type=" . "article")
 ("=key=" . "zukin2009"))

ELISP> (gethash "zukin2009" bibtex-actions-bibliography :fields (-flatten bibtex-actions-field-map)))
((id . "zukin2009")
 (type . "article-journal")
 (container-title . "City & Community")
 (issue . "1")
 (issued . "2009"))
```

aikrahguzar:
This turned out to be straightforward.
This is what `(bibtex-actions--fields-to-parse)` produces with default templates,
```elisp
("author" "date" "title" "=key=" "=type=" "tags" "doi" "url" "file" "year" "issued" "id" "type" "keywords" "keyword")
```

Fix #231

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -345,11 +345,12 @@ personal names of the form 'family, given'."
 (defun bibtex-actions--fields-to-parse ()
   "Determine the fields to parse from the template and field map"
   (let* ((fields-in-format (bibtex-actions--fields-in-formats))
-        (fields-for-symbols (list "doi" "url" bibtex-actions-file-variable))
-        (canonical-fields (seq-concatenate 'list fields-in-format fields-for-symbols)))
+         (fields-for-symbols (list "doi" "url" bibtex-actions-file-variable))
+         (canonical-fields (seq-concatenate 'list fields-in-format fields-for-symbols))
+         (equivalent-fields (cons '("author" "editor") bibtex-actions-field-map)))
     (seq-concatenate 'list
                      canonical-fields
-                     (seq-mapcat (lambda (field) (cdr (assoc field bibtex-actions-field-map)))
+                     (seq-mapcat (lambda (field) (cdr (assoc field equivalent-fields)))
                                  canonical-fields))))
 
 (defun bibtex-actions--format-candidates (files &optional context)
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -329,6 +329,19 @@ personal names of the form 'family, given'."
          (car (split-string name ", "))))
      (split-string names " and ") ", ")))
 
+(defun bibtex-actions--fields-to-parse ()
+  "Determine the fields to parse from the template"
+  (cl-flet ((fields-for-format
+             (format-string)
+             (split-string
+              (s-format format-string
+                        (lambda (fields-string) (car (split-string fields-string ":")))))))
+    (cl-remove-duplicates
+     (seq-mapcat (lambda (format) (fields-for-format (cdr format)))
+              (seq-concatenate 'list
+                               bibtex-actions-template
+                               bibtex-actions-template-suffix)))))
+
 (defun bibtex-actions--format-candidates (files &optional context)
   "Format candidates from FILES, with optional hidden CONTEXT metadata.
 This both propertizes the candidates for display, and grabs the
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -329,18 +329,28 @@ personal names of the form 'family, given'."
          (car (split-string name ", "))))
      (split-string names " and ") ", ")))
 
-(defun bibtex-actions--fields-to-parse ()
-  "Determine the fields to parse from the template"
+(defun bibtex-actions--fields-in-formats ()
+  "Find the fields to mentioned in the templates."
   (cl-flet ((fields-for-format
              (format-string)
              (split-string
               (s-format format-string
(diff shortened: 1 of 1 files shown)


bdarcus:
We need somewhere to add "editor", which is an alternate for "author".

aikrahguzar:
> We need somewhere to add "editor", which is an alternate for "author".

Maybe add it to `bibtex-actions-field-map` ? This will make sure they go together as long as we use the map. This will also simplify `bibtex-actions--field-value-for-formatting`

aikrahguzar:
Actually if we go with this the only point of `bibtex-actions--field-value-for-formatting` will be to truncate the date to 4 chracters. Why should we do this if the format specifies a wider date? If format specifies 4 characters this will happen by itself.

So we can entirely eliminate `bibtex-actions--field-value-for-formatting`

bdarcus:
> > We need somewhere to add "editor", which is an alternate for "author".
> 
> Maybe add it to `bibtex-actions-field-map` ? This will make sure they go together as long as we use the map. This will also simplify `bibtex-actions--field-value-for-formatting`

I originally included it, but it somehow felt the wrong place, since it's not strictly a mapping.

I'm not sure though.

bdarcus:
> Actually if we go with this the only point of `bibtex-actions--field-value-for-formatting` will be to truncate the date to 4 chracters. Why should we do this if the format specifies a wider date? If format specifies 4 characters this will happen by itself.

But the width is really the column width.

How would you address that if you removed that function?

aikrahguzar:
> But the width is really the column width.
> 
> How would you address that if you removed that function?

We already truncate to column width specified by the format. I am just saying that we should defer to that. Right now date gets truncated to 4 characters even if more width is available according to the format.

aikrahguzar:
> I originally included it, but it somehow felt the wrong place, since it's not strictly a mapping.
> 
> I'm not sure though.

I think the question is if we need to differentiate between the two in the context of this package, which is how I am interpreting the map. It is a list of equivalence classes for this package.

bdarcus:
> We already truncate to column width specified by the format. I am just saying that we should defer to that. Right now date gets truncated to 4 characters even if more width is available according to the format.

I'm not following something.

The reason to truncate there is we want a year representation of a date, and in CSL JSON and BibLaTeX, dates are typically longer than 4 characters.

That may not even be the best way to handle it, but having the full raw date string can look weird; eg:

``` console
2016
2019-03-05
2011
2017-03-05/2017-03-07
```

aikrahguzar:
> I originally included it, but it somehow felt the wrong place, since it's not strictly a mapping.
> 
> I'm not sure though.

A template can specify both author and editor so for the moment at least I added a special cases just for finding fields to parse.

bdarcus:
Come to think of it, maybe we just need a `bibtex-actions-shorten-date` to accompany `bibtex-actions-shorten-names`?

Or could be generalized to "format"?

aikrahguzar:
> Come to think of it, maybe we just need a `bibtex-actions-shorten-date` to accompany `bibtex-actions-shorten-names`?
> 
> Or could be generalized to "format"?

My idea would be that if someone wants only 4 characters they will set the format width to 4. Numbers have width 1 so it shouldn't be a problem. If there is more width available they probably wanted to see more precise date.

But I don't have a strong opinion beyond wanting to remove code if it doesn't hurt.

bdarcus:
> My idea would be that if someone wants only 4 characters they will set the format width to 4. Numbers have width 1 so it shouldn't be a problem.

So then we go back to my earlier comment; how does the user set the spacing between the columns?

Just using whitespace in the template?

If yes, that's fine; I just wanted to make sure you understand the scope of full picture.

aikrahguzar:
> Just using whitespace in the template?

Yes, white space in template is preserved as is.

> If yes, that's fine; I just wanted to make sure you understand the full picture.

bdarcus:
RTM then, or you still working on it?

aikrahguzar:
> RTM then, or you still working on it?

No, it is ready

PR #235, opened 2021-08-18
Fix the overflow

aikrahguzar:
I had though that invisible strings were not displayed but still occupied space. But I was wrong. Reducing the star-width by 1 fixes the appearance of overflow indicator as least for me.

My best pr yet :)


bdarcus:
Nice when they're small; eh?

Thanks!

BTW, github has a shorthand where you can write `fix` and the issue number and it will auto close the linked issue.

I added that to the commit message.

PR #238, opened 2021-08-19
Fix passing literal keys

aikrahguzar:
Fixes #222 

It was caused by change of consing the whole candidate with the candidate string instead of just the key.

I don't see a downside of just assuming that the string is a literal key if it is not the display string of a candidate and just passing it on.


bdarcus:
Oh, thank you!!

Couple things:

1. This fixes 233; no? Can you edit the post accordingly?
2. So when @ilupin was originally working on the embark at-point, he made the changes to also do the `rassoc` check. But a lot has changed in the interim, and gotten simplified along the way.

PR #240, opened 2021-08-19
Add basic template for default note fn

### LINKED ISSUE #239: Enhance default note function
There is now a defcustom to substitute your own open-note function, but I think we should still expand the default one as follows:

1. Add a basic (maybe `org-capture`, though may be overkill?) template that will work well with `org-roam v2`, and make clear it's for `org-mode`.

I think that should be enough for now actually.

bdarcus:
I made a mistake and pushed the initial commit to the main branch (sigh), but it was intended to go here.

So I'll finish it on this branch.

I'm not actually sure how best to do this with the template, but it's intended to be a simple thing, which is never-the-less compatible with `org-roam v2`.

Current code results in duplicate titles properties at the beginning, which is a Doom thing, so I can ignore that:

https://github.com/hlissner/doom-emacs/blob/develop/modules/editor/file-templates/README.org

FWIW, here's what inserting a note using the default capture template in `org-roam v2` looks like:

image

So only real difference currently is it includes the property drawer for the file, and an org id.

But it should be easy enough to write, or borrow, an OR function for this.

Fix #239

Just FYI @aikrahguzar

### DIFF
--- a/bibtex-actions-file.el
+++ b/bibtex-actions-file.el
@@ -21,6 +21,9 @@
 
 (require 'cl-lib)
 
+(declare-function bibtex-actions-get-entry "bibtex-actions")
+(declare-function bibtex-actions-get-value "bibtex-actions")
+
 ;;;; File related variables
 
 (defcustom bibtex-actions-file-open-function 'find-file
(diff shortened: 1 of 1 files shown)


bibtex-actions-file.el
```diff
@@ -126,8 +126,11 @@ If you use 'org-roam' and 'org-roam-bibtex, you should use
           (caar (bibtex-actions-file--files-to-open-or-create
                  (list key)
                  bibtex-actions-notes-paths '("org"))))
-         (template "#+title: ${title}\n"))
+         ;; how do I get the entry title here?
```
bdarcus, line 129: Answer: look it up in the cached data?

Issue #237, opened 2021-08-19
Refactor templates, remove map variable


bdarcus:
Per suggestion [below](https://github.com/bdarcus/bibtex-actions/issues/237#issuecomment-901846642), let's make templates a little more flexible, so we can simplify elsewhere.

------------

I accidentally left in a second field `mapping` table, which was an experiment in a template approach that would be more CSL field focused.

The advantage of that is you have a generic field name like `container-title`, which then maps to type-specific fields like `booktitle`.

Would make it easier to remove type-specific templates.

Also, I guess I should make that a `defcustom`?

Thoughts?

cc @aikrahguzar

aikrahguzar:
> I accidentally left in a second field `mapping` table, which was an experiment in a template approach that would be more CSL field focused.
> 
> The advantage of that is you have a generic field name like `container-title`, which then maps to type-specific fields like `booktitle`.
> 
> Would make it easier to remove type-specific templates.
> 
> Also, I guess I should make that a `defcustom`?
> 
> Thoughts?

I would go in the other direction and instead use template to determine this kind of mapping. I think as it is we have too many knobs for basically doing the same stuff. I think instead we should use a template of the form
```elisp 
"${author editor:36}    ${date year issued:4}   ${title:60}"
```
i.e the mapping is part of the template and can be adjusted there. I am assuming the field names don't have space in them. If they do we can use another separator like comma. We try each of the alternatives in turn and use the first one that succeeds. This wouldn't be hard to implement and simplify a bunch of details  by putting everything in one place. I think this is true even for the user. If they need an alternate just add it to the proper place in the format string.

This will also get rid of annoying but unlikely in practice scenarios like what happens if the template has both author and editor. They can both be present and with the current default if this happens and author is not present but the editor is we will be displaying the editor names in both columns.

bdarcus:
That makes sense in general.

But how would it interact with `bibtex-actions-get-value`? I'm particularly thinking of crucial fields like `=key=`/`id`.

That function would just look in the template, rather than the variable, maybe with some specific logic for the special fields?

bdarcus:
> EDIT: and is there anywhere else, outside of formatting, that variables need to be accessed, beyond the ones I note above?

Outside of the formatting, it is used elsewhere to access key, doi, url.

aikrahguzar:
> That makes sense in general.
> 
[...]

PR #241, opened 2021-08-19
New layout for cached strings and overhaul of templates and formatting

aikrahguzar:
For #237 

Mostly removes a lot of code.

`bibtex-actions-template` has a new type, so that might lead to type errors in a long session. 
Preferably test in a new session if everything works. Everything seems to work fine for me.

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -114,7 +114,7 @@
 
 (defcustom bibtex-actions-template
   (cons
-   "${author editor:30}   ${date year issued:8}  ${title:48}"
+   "${author editor:30}   ${date year issued:4}  ${title:48}"
    "          ${=key= id:15}    ${=type=:12}    ${tags keywords keywords:*}")
   "Configures formatting for the BibTeX entry.
 car is for the main body of the candidate and cdr is for suffix.
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -116,7 +116,7 @@
   (cons
    "${author editor:30}   ${date year issued:4}  ${title:48}"
    "          ${=key= id:15}    ${=type=:12}    ${tags keywords keywords:*}")
-  "Configures formatting for the BibTeX entry.
+  "Configures formatting for the bibliographic entry.
 car is for the main body of the candidate and cdr is for suffix.
 The same string is used for display and for search."
     :group 'bibtex-actions
@@ -272,7 +272,7 @@ offering the selection candidates"
     (seq-difference local-bib-files bibtex-actions-bibliography)))
 
 (defun bibtex-actions-get-value (fields item)
-  "Return biblatex FIELD value for ITEM."
+  "Return the first non nil biblatex field value for ITEM among FIELDS."
   (cl-flet ((get-value (field) (cdr (assoc-string field item 'case-fold))))
       (get-value (seq-find #'get-value fields))))
 
@@ -482,8 +482,11 @@ TEMPLATE."
             ;; Make sure we always return a string, even if empty.
             (field-value (or (bibtex-completion-clean-string
                               (bibtex-actions-get-value field-names entry))
-                             " ")))
-       (truncate-string-to-width field-value display-width 0 ?\s)))))
+                             ""))
+            (formatted-value (if (seq-intersection '("author" "editor") field-names)
+                                 (bibtex-actions-shorten-names field-value)
+                               field-value)))
+       (truncate-string-to-width formatted-value display-width 0 ?\s)))))
 
 ;;; At-point functions
 
@@ -555,7 +558,7 @@ With prefix, rebuild the cache before offering candidates."
 
 ;;;###autoload
 (defun bibtex-actions-open-entry (keys)
-  "Open BibTeX entry associated with the KEYS.
+  "Open bibliographic entry associated with the KEYS.
 With prefix, rebuild the cache before offering candidates."
   (interactive (list (bibtex-actions-select-keys
                       :rebuild-cache current-prefix-arg)))
@@ -607,7 +610,7 @@ With prefix, rebuild the cache before offering candidates."
 
 ;;;###autoload
 (defun bibtex-actions-insert-bibtex (keys)
-  "Insert BibTeX entry associated with the KEYS.
+  "Insert bibliographic entry associated with the KEYS.
 With prefix, rebuild the cache before offering candidates."
   (interactive (list (bibtex-actions-select-keys
                       :rebuild-cache current-prefix-arg)))
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -313,11 +313,10 @@ When nil, return DEFAULT."
 
 (defun bibtex-actions-get-entry (key)
   "Return the cached entry for KEY."
-  (seq-find
-   (lambda (entry)
-     (string-equal key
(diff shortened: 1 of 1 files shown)


bibtex-actions.el
```diff
@@ -113,31 +113,19 @@
 
 
 (defcustom bibtex-actions-template
-  '((t . "${author:30}   ${date:8}  ${title:48}"))
+  (cons
+   "${author editor:30}   ${date year issued:8}  ${title:48}"
+   "          ${=key= id:15}    ${=type=:12}    ${tags keywords keywords:*}")
   "Configures formatting for the BibTeX entry.
-When combined with the suffix, the same string is used for
```
bdarcus, line 118: I guess we should change "BibTeX" to "bibliographic in the line above now.

bibtex-actions.el
```diff
@@ -302,22 +271,17 @@ offering the selection candidates"
...
-(defun bibtex-actions-get-value (field item &optional default)
-  "Return biblatex FIELD value for ITEM.
-When nil, return DEFAULT."
-  (or (cdr (assoc-string field item 'case-fold))
-      (cl-loop for fname in (cdr (assoc field bibtex-actions-field-map))
-               when (cdr (assoc-string fname item 'case-fold))
-                         return (cdr (assoc-string fname item 'case-fold)))
-      default))
+(defun bibtex-actions-get-value (fields item)
+  "Return biblatex FIELD value for ITEM."
```
bdarcus, line 275: Is the "biblatex" word here now relevant, given that you've generalized this?

bibtex-actions.el
```diff
@@ -113,31 +113,19 @@
 
 
 (defcustom bibtex-actions-template
-  '((t . "${author:30}   ${date:8}  ${title:48}"))
+  (cons
+   "${author editor:30}   ${date year issued:8}  ${title:48}"
```
bdarcus, line 117:
```suggestion
   "${author editor:30}   ${date year issued:4}  ${title:48}"
```

Let's make this default, to make sure it works as expected.

bdarcus (changes requested):
I added a few minor comments and suggestions below. 

There's also this, which is I no longer get the shortened authors by default.

image

Also, another minor thing: maybe add a bit more padding between "columns" on the main template?

bdarcus:
One other thing:

I'm not sure why, but I got a rebase conflict when I downloaded your code into a new branch on my end.

If you haven't already, now would be a good time to read the link I sent earlier on rebasing, including of "upstream" branches.

You'll basically want to do `git pull --rebase upstream main` before you want to put together a PR to update your fork, and create a local branch from that, where you do your changes.

bdarcus:
Oh, one other thing: the overflow frame thing is back.

aikrahguzar:
> Oh, one other thing: the overflow frame thing is back.

I am not seeing it. Overflow will happen if your frame width is less the sum of widths in the frame or if you made the frame smaller but didn't refresh the cache. Is it happening without any of these?

bibtex-actions.el
aikrahguzar: I change all references fof BbTeX to bibliographic, Though the others one might be aspirational as of now.

aikrahguzar:
> I added a few minor comments and suggestions below.
> 
> There's also this, which is I no longer get the shortened authors by default.
> 
> image
> 
> Also, another minor thing: maybe add a bit more padding between "columns" on the main template?

That was my too hasty deletion of a function. Should be fixed now.

bdarcus:
> Overflow will happen if your frame width is less the sum of widths in the frame or if you made the frame smaller but didn't refresh the cache. Is it happening without any of these?

I have emacs full-screen, and have refreshed, and I still get it.

aikrahguzar:
> I have emacs full-screen, and have refreshed, and I still get it.

Default template? I have moved around the star in it too and it seems to be behaving for me. This is strange.

Kind of annoying thing to do but if you can: move the star somewhere other than the last column and count how many characters in that field are displayed and if they are different from the width that is specified in the template.

bdarcus:
Will do.

In the meantime, run `checkdoc`! I get a few warnings.

bdarcus:
If I move start to title, I get 47 characters.

bdarcus:
Actually, moving the star back, I don't get the overflow. :shrug: 

So maybe just fix the docstrings and I can merge, and we can worry about that separately if it pops up again.

When you're done, can you click the "request re-review" icon next to my reviewer name?

aikrahguzar:
> If I move start to title, I get 47 characters.

I meant how many characters are there for the last field? Set the width to 2 for the last field and see if you get only 2 characters.

aikrahguzar:
> Will do.
> 
> In the meantime, run `checkdoc`! I get a few warnings.

2021-08-19T20:14:34,403561647+02:00

:shrug:

aikrahguzar:
Also why is there `(defvar crm-separator)` instead of `(require crm)`?

bdarcus:
> > If I move start to title, I get 47 characters.
> 
> I meant how many characters are there for the last field? Set the width to 2 for the last field and see if you get only 2 characters.

image


```elisp
(defcustom bibtex-actions-template
  (cons
   "${author editor:30}   ${date year issued:4}"
   "          ${=key= id:15}    ${title:*}  ${=type=:12}    ${tags keywords keywords:2}")
```

bdarcus:
> Also why is there `(defvar crm-separator)` instead of `(require 'crm)`?

I don't know, but it ended up being needed so that the override I use on the completion function works.

Here's the checkdoc warnings; not sure why you're not seeing them.

image

bdarcus:
BTW, this is my fault, but while you're at it, can you first the typo in the very first line of the file?

aikrahguzar:
> image
> 
> ```emacs-lisp
> (defcustom bibtex-actions-template
>   (cons
>    "${author editor:30}   ${date year issued:4}"
>    "          ${=key= id:15}    ${title:*}  ${=type=:12}    ${tags keywords keywords:2}")
> ```

I think this was due to some combination of affixation not being fixed width and it having a tab character in it. It should be fixed now. Every column is now truncated or padded to a  fixed width.

aikrahguzar:
> BTW, this is my fault, but while you're at it, can you fix the typo in the very first line of the file?

Done

aikrahguzar:
> I think this was due to some combination of affixation not being fixed width and it having a tab character in it. It should be fixed now. Every column is now truncated or padded to a fixed width.

To be more precise, this treating the whole affixation as a single column, the individual sybols in it might not line up.

bdarcus (approved):

aikrahguzar:
> Here's the checkdoc warnings; not sure why you're not seeing them, but I am. Maybe because I'm using `flycheck` and you're not?

I find this weird, I have had that capitalization warning before. I am just using doom's default elisp setup without any configuration at all. And `SPC c x` is bound to `flycheck-list-errors` and it use it frequently.

bdarcus:
With this commit, I just bumped the packages in 5290.

bdarcus:
> You'll basically want to do `git pull --rebase upstream main` to update your fork, before you want to put together a PR, and then create a local branch from that, where you do your changes.

BTW, on this git stuff, if using`magit` (recommended), this is what the menu for "pull" ("P") would look like to update your fork from this main repo:

image

PR #248, opened 2021-08-20
filenotify-> file-notify

bdarcus:
Scaled this back to just change the file and function names for filenotify.

----------------------

I don't know if this is a good idea, or is overkill, and so this is an experiment ...

1. `bibtex-actions-filenotify` -> `bibtex-actions-file-notify`
2. add `bibtex-actions-select`

Idea is just to get it all cleanly modular for easier maintenance and evolution going forward.

But 2 is not so straight-forward, actually, given the current assumption of `bibtex-actions-bibliography`. Really `bibtex-actions-select-key` needs to a (probably optional)`bib-files` arg, and that needs to propagate down the functions. So one could optionally do something like:

```elisp
(bibtex-actions-select-keys
  random-files
  :library-files '("~/Documents")))
```

For that reason, my impulse is to drop the second commit for now, but merge the first.

Or, if I can do it cleanly, just move the code, but don't change it.


bdarcus:
Any objections to this change @aikrahguzar?

aikrahguzar:
> Any objections to this change @aikrahguzar?

No, I am agnostic.

bdarcus:
Thanks.

I changed my mind.

bdarcus:
I just saw this on emacs-devel, from the creator of org-roam. Is there any value for us with this?

https://lists.gnu.org/archive/html/emacs-devel/2021-08/msg00876.html
https://github.com/jethrokuan/filenotify-recursive

aikrahguzar:
> I just saw this on emacs-devel, from the creator of org-roam. Is there any value for us with this?
> 
> https://lists.gnu.org/archive/html/emacs-devel/2021-08/msg00876.html
> https://github.com/jethrokuan/filenotify-recursive
> 
> Maybe not, at least if and until it is upstreamed?

I think this is a question for those who use the notes and library functionality? Is the structure of these directories flat or not? 

But I think even then they can use this package pretty easily themselves. They just need to do 

```elisp
(setq bibtex-actions-filenotify-files 
      (seq-concatenate 'list (list (bibliography "file1")) (fnr--subdirectories-recursively dir)))
```
and the end result will what they want. 

The package basically places watches on all the files and directories in a list and ensures that they are removed together too, and we do both these things too and in essentially the same manner.

Issue #247, opened 2021-08-20
Affixation alignment issue


bdarcus:
@aikrahguzar:

![image](https://user-images.githubusercontent.com/1134/130226558-e6ce7223-ac8b-4ba1-a207-136481a4b608.png)

The display of the "file" symbol is I assume a bug in my Emacs 28 build, but note the alignment here; it actually impacts the entire row, though I only show the beginning.

Can you check this when you get a chance please? I assume the issue is in this line, though am unclear why:

https://github.com/bdarcus/bibtex-actions/blob/5356d118fd646d5aa04254a31c523c3e642b412b/bibtex-actions.el#L374

aikrahguzar:
This caused by the file symbol we have by default. Its width is 1 but it clearly occupies more than one column which is what is causing the issue. The really correct way to solve that would be to truncate by pixels not but I don't see anyway of doing it. 

What we can do here is to choose symbols which don't occupy more space than their widths. This works for me, 
```elisp 
(setq bibtex-actions-symbols
  `((file  .  ("📄" . "  "))
    (note .   ("✎" . " "))
    (link .   ("🔗" . "  ")))))
```

Previously there was a `tab` character in the affixation which was somehow taking care of the alignment though I don't know how but that causes problems with determining how many columns are available for the rest of the string.

I will make a pr to adjust change the file symbol to adjust this, which also fixes a quoting issue.

The change to file  symbol is not a proper fix since it is likely to be font dependent but this is the usual problem of mixing variable width fonts with character grid based displays.

bdarcus:
Fixed-ish in #249

PR #249, opened 2021-08-20
Fix (not well) affixation alignment + a quoting issue

aikrahguzar:
Alleviate #247

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -123,9 +123,9 @@ The same string is used for display and for search."
     :type  '(cons string string))
 
 (defcustom bibtex-actions-symbols
-  `((file  .  ("⌘" . " "))
+  `((file  .  ("📄" . "  "))
     (note .   ("✎" . " "))
-    (link .   ("🔗" . " ")))
+    (link .   ("🔗" . "  ")))
   "Configuration alist specifying which symbol or icon to pick for a bib entry.
 This leaves room for configurations where the absense of an item
 may be indicated with the same icon but a different face."
@@ -330,7 +330,7 @@ key associated with each one."
              collect
              (let* ((files
                      (when (or (bibtex-actions-get-value
-                                '(bibtex-actions-file-variable) candidate)
+                                (list bibtex-actions-file-variable) candidate)
                                (bibtex-actions-file--files-for-key
                                 citekey bibtex-actions-library-paths bibtex-actions-file-extensions))
                        " has:files"))
@@ -395,7 +395,7 @@ key associated with each one."
              (list (thing-string has-files 'file)
                    (thing-string has-note 'note)
                    (thing-string has-link 'link)))
-     "   ")))
+     "  ")))
 
 (defvar bibtex-actions--candidates-cache 'uninitialized
   "Store the global candidates list.


bdarcus:
Would't it also be addressed by adding a couple of spaces to that width?

aikrahguzar:
> Would't it also be addressed by adding a couple of spaces to that width?
> 
> That should make the width of the symbols irrelevant to alignment.

The problem is there because `string-width` of a symbol is 1 but it actually occupies 1.5 columns. So when we truncate a string containing it to say 8 columns by calling `truncate-string-to-width` the resulting string will actually use 8.5 columns. 

Normal spaces don't help but we can use Unicode spaces like 6 per em space which has has width less than 1 to compensate but there is no way to do a calculation about how much compensation we need. 

From the documentation of `string-width`

> The effect of faces and fonts used for non-Latin and other unusual characters (such as emoji) is ignored as well, as are display properties and invisible text. For these reasons, the results are not generally reliable; for accurate dimensions of text as it will be displayed, use window-text-pixel-size instead. 

Basically `truncate-string-to-width` is an approximation to the space occupied on screen that uses `string-width`.

From the manual it seems the right way to control what is displayed on screen would be using `display` text property but that need more reading. 

The only short term fix I see is to only use characters that don't occupy more space than their `character-width`.

bdarcus:
> The only short term fix I see is to only use characters that don't occupy more space than their `character-width`.

Should use just use characters for the short-term default then?

```console
F N L
```

I dunno; will merge this and we can think about it.

aikrahguzar:
> Should use just use characters for the short-term default then?
> 
> ```
> F N L
> ```
> 
> I dunno; will merge this and we can think about it.

That makes sense, though I don't know how short term it will be. My first quick perusal of `display` text property has just resulted in head scratching.

I suggest adding something like the following to the documentation for `bibtex-actions-symbols`
```
To avoid alignment issues make sure that both the car and cdr of a symbol have
the same width. In addition, the `string-width' of the symbols should be
accurate for the font being used. See the documentation of `string-width'
for the caveats.
```

bdarcus:
The approach I took to dealing with icons, where they are always displayed, just with different faces to indicate status, is another option.

Perhaps that should be generalized?

aikrahguzar:
> The approach I took to dealing with icons, where they are always displayed, just with different faces to indicate status, is another option.
> 
> Perhaps that should be generalized?

I think this is almost certainly a font issue, this is what I get for the fonts being used for the various icons (found using `describe-char`)
2021-08-20T20:42:38,579239337+02:00

As you can see there are three different fonts in use. Among them google noto plays well with my primary font but the other two fonts don't. There is no way for me to align all the `ftchrb` occurrences above because of the same issue. (I am surpised that the notes symbol hasn't caused you trouble yet but your fonts might be different) but it depends on what fonts get used and how they interact with the primary. 

So the technical response to this question (which I wouldn't don't like at all) would be to either use a monospace font which covers these characters or configure the fallback fonts such that they are compatible with it.

If all the symbols to come from the same monospace font, it will causes it will fix alignment issue but if this font has width bigger than the normal one it will cause the overflow. But that can at least be sidestepped by just reducing the star width by a couple of columns, which should mostly fix the issue at the expense of possibly leaving empty space at the end.

The same thing will happen with icons, they might not be compatible with the main font.  For me they cause row alignment issues where the icons seem to be shifted vertically w.r.t. rest of the text. But they will fix the issue of column alignment issue here because they all come from the same font.

bdarcus:
> I think this is almost certainly a font issue ...

What is the "this" to which you refer? The general issue, or the solution you quoted?

aikrahguzar:
> What is the "this" to which you refer? The general issue, or the solution you quoted?

The general issue of alignment

bdarcus:
And do you understand what I was saying about that solution?

Basically, you always output this:

```console
F L N
```

... but you propertize those strings to include either "on" or "off" depending.

That should work regardless of the font/icon; no?

aikrahguzar:
> And do you understand what I was saying about that solution?
> 
> Basically, you always output this:
> 
> ```
> F L N
> ```
> 
> ... but you propertize those strings to include either "on" or "off" depending.
> 
> That should work regardless of the font/icon; no?
> 

I hadn't understood this well. This should work. We can just use `succes` to indicate the presence of a resource `error` to indicate the absence. That should work with almost every theme.

PR #250, opened 2021-08-22
Customizable field display + better fix for symbol alignment

aikrahguzar:
First one is an experiment I am not too sure about. I was a bit bothered about the fact that the shortening of author and editor names is a special case. So I created a `defcustom` that controls transforming the fields from the raw parsed string to one for display and now name shortening goes through it. I like the uniformity but not sure if it anyone would want to do something with it. 

As a related change `bibtex-actions-get-value` is back to accepting a single field. There are now two related functions, `bibtex-action-has-a-value` which accepts a list of fields and returns the first field for which the item has a value and `bibtex-actions-display-value` which accpets also accepts a list of fields and returns the display string. 

I also managed to find a more robust solution for alignment issues for the symbols using the display property. This should now work correctly if a symbol with string-width `n` doesn't occupy more than `n+1` columns on screen and there is some padding available. The padding will be reduced by whatever fraction the symbol overflows. It still does require that the symbols for the absence or presence of a thing both have the same width.

### DIFF
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -123,7 +123,7 @@ The same string is used for display and for search."
     :type  '(cons string string))
 
 (defcustom bibtex-actions-display-transform-functions
-  '((t  . bibtex-completion-clean-string)
+  '((t  . bibtex-actions-clean-string)
     (("author" "editor") . bibtex-actions-shorten-names))
   "This variable configures the transformation of field values from raw values
 in bib files to those displayed when using `bibtex-actions-select-keys'.
@@ -303,7 +303,14 @@ The value is transformed using `bibtex-actions-display-transform-functions'"
                       (funcall (cdr fun) string)
                     string))
                 bibtex-actions-display-transform-functions
-                (bibtex-actions-get-value field item))))
+            ;; Make sure we always return a string, even if empty.
+                (or (bibtex-actions-get-value field item) ""))))
+
+;; Lifted from bibtex-completion
+(defun bibtex-actions-clean-string (s)
+  "Remove quoting brackets and superfluous whitespace from string S."
+  (replace-regexp-in-string "[\n\t ]+" " "
+         (replace-regexp-in-string "[\"{}]+" "" s)))
 
 (defun bibtex-actions-get-entry (key)
   "Return the cached entry for KEY."
@@ -514,7 +521,7 @@ FORMAT-STRING."
                                field-width
                              width))
             ;; Make sure we always return a string, even if empty.
-            (display-value (or (bibtex-actions-display-value field-names entry) "")))
+            (display-value (bibtex-actions-display-value field-names entry)))
        (truncate-string-to-width display-value display-width 0 ?\s)))))
 
 ;;; At-point functions
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -134,16 +134,14 @@ All functions that match a particular field are run in order."
 
 (defcustom bibtex-actions-symbols
   `((file  .  ("⌘" . " "))
-    (note .   ("✎ " . " "))
+    (note .   ("✎" . " "))
     (link .   ("🔗" . "  ")))
   "Configuration alist specifying which symbol or icon to pick for a bib entry.
 This leaves room for configurations where the absense of an item
 may be indicated with the same icon but a different face.
 
 To avoid alignment issues make sure that both the car and cdr of a symbol have
-the same width. In addition, the `string-width' of the symbols should be
-accurate for the font being used. See the documentation of `string-wdith'
-for the caveats."
+the same width."
   :group 'bibtex-actions
   :type '(alist :key-type string
                 :value-type (choice (string :tag "Symbol"))))
--- a/bibtex-actions.el
+++ b/bibtex-actions.el
@@ -122,13 +122,28 @@ The same string is used for display and for search."
     :group 'bibtex-actions
     :type  '(cons string string))
 
+(defcustom bibtex-actions-display-transform-functions
+  '((t  . bibtex-completion-clean-string)
+    (("author" "editor") . bibtex-actions-shorten-names))
+  "This variable configures the transformation of field values from raw values
+in bib files to those displayed when using `bibtex-actions-select-keys'.
+All functions that match a particular field are run in order."
+  :group 'bibtex-actions
+  :type '(alist :key-type   (choice (const t) (repeat string))
+                :value-type function))
+
 (defcustom bibtex-actions-symbols
   `((file  .  ("📄" . "  "))
     (note .   ("✎" . " "))
     (link .   ("🔗" . "  ")))
   "Configuration alist specifying which symbol or icon to pick for a bib entry.
 This leaves room for configurations where the absense of an item
-may be indicated with the same icon but a different face."
+may be indicated with the same icon but a different face.
+
+To avoid alignment issues make sure that both the car and cdr of a symbol have
+the same width. In addition, the `string-width' of the symbols should be
+accurate for the font being used. See the documentation of `string-wdith'
+for the caveats."
   :group 'bibtex-actions
   :type '(alist :key-type string
                 :value-type (choice (string :tag "Symbol"))))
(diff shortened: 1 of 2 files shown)


bibtex-actions.el
```diff
@@ -122,13 +122,26 @@ The same string is used for display and for search."
     :group 'bibtex-actions
     :type  '(cons string string))
 
+(defcustom bibtex-actions-display-transform-functions
+  '((t  . bibtex-completion-clean-string)
```
bdarcus, line 126: Why this?

bibtex-actions.el
```diff
@@ -271,10 +284,26 @@ offering the selection candidates"
...
-(defun bibtex-actions-get-value (fields item)
-  "Return the first non nil biblatex field value for ITEM among FIELDS ."
-  (cl-flet ((get-value (field) (cdr (assoc-string field item 'case-fold))))
-      (get-value (seq-find #'get-value fields))))
+(defun bibtex-actions-get-value (field item)
+  "Return the FIELD value for ITEM."
+  (cdr (assoc-string field item 'case-fold)))
+
+(defun bibtex-actions-has-a-value (fields item)
+  "Retuns the first field that has a value in ITEM among FIELDS ."
```
bdarcus, line 292:
```suggestion
  "Return the first field that has a value in ITEM among FIELDS ."
```

bibtex-actions.el
```diff
@@ -271,10 +284,26 @@ offering the selection candidates"
...
+  (seq-find (lambda (field) (bibtex-actions-get-value field item)) fields))
+
+(defun bibtex-actions-display-value (fields item)
+  "Return the first non nil value for ITEM among FIELDS .
```
bdarcus, line 296:
```suggestion
  "Return the first non nil value for ITEM among FIELDS.
```

bdarcus (commented):
I added a few minor suggestions and a question, but it may take me a bit of time to take a closer look, and to test.

Semester starts tomorrow.

bdarcus:
> First one is an experiment I am not too sure about.

For things unfinished, you can do draft PRs, including converting to draft and back.

bibtex-actions.el
aikrahguzar: This is to remove things like braces and quoting from fields, it has been there from the start.

> I added a few minor suggestions and a question, but it may take me a bit of time to take a closer look, and to test.
> 
> Semester starts tomorrow.

Take your time, my pace will slow down too.

aikrahguzar:
> I added a few minor suggestions and a question, but it may take me a bit of time to take a closer look, and to test.
> 
> Semester starts tomorrow.

Take your, I am going to slow down too due to travel.

aikrahguzar:
> > First one is an experiment I am not too sure about.
> 
> For things unfinished, you can do draft PRs, including converting to draft and back.

I will keep that one in mind though in this case it is not so much unfinished, as that I am not sure it is worth the added loc though I like the end of result of more uniformity and customizibility.

bibtex-actions.el
bdarcus: Yes, I know, but doesn't parsebib do that?

I know a lot of BC code was done when parsebib was very new.

bdarcus: For example, see `parsebib--parse-bib-value`

aikrahguzar: > Yes, I know, but doesn't parsebib do that?
> 
> I know a lot of BC code was done when parsebib was very new.

Mostly but not always. It is probably not possible to always remove them without introducing ambiguity.

2021-08-23T00:33:38,490186192+02:00

bdarcus: Here's the BC code:

```elisp
(defun bibtex-completion-clean-string (s)
  "Remove quoting and superfluous white space from BibTeX field value in S."
  (if s (replace-regexp-in-string "[\n\t ]+" " "
         (replace-regexp-in-string "[\"{}]+" "" s))
    nil))
```

bdarcus:
I'm needing, I think, a get-link function. Would this work?

```elisp
(defun bibtex-actions-get-link (key)
  "Return a link for a KEY."
  (let* ((entry (bibtex-actions-get-entry key))
         (field (bibtex-actions-has-a-value '(doi pmid pmcid url) entry))
         (base-url (pcase field
                     ("doi" "https://doi.org")
                     ("pmid" "https://www.ncbi.nlm.nih.gov/pubmed/")
                     ("pmcid" "https://www.ncbi.nlm.nih.gov/pmc/articles/"))))
         (concat base-url (bibtex-actions-get-value field entry))))
```

aikrahguzar:
> I'm needing, I think, a get-link function. Would this work?
> 
> ```emacs-lisp
> (defun bibtex-actions-get-link (key)
>   "Return a link for a KEY."
>   (let* ((entry (bibtex-actions-get-entry key))
>          (field (bibtex-actions-has-a-value '(doi pmid pmcid url) entry))
>          (base-url (pcase field
>                      ("doi" "https://doi.org") ; symbols or strings here?
>                      ("pmid" "https://www.ncbi.nlm.nih.gov/pubmed/")
>                      ("pmcid" "https://www.ncbi.nlm.nih.gov/pmc/articles/"))))
>     (if field
>         (concat base-url (bibtex-actions-get-value field entry))
>       (message "No link found for %s" key)))
> ```

It should be string in the list for `bibtex-actions-has-a-value` i.e.
```elisp 
(field (bibtex-actions-has-a-value '("doi" "pmid" "pmcid" "url") entry))
```
but at first glance it should work for finding the first link which is available.

I will try to add support for pre-prints from arxiv at some point in the future. It should be pretty easy.

bibtex-actions.el
aikrahguzar: > Here's the BC code:
> 
> ```emacs-lisp
> (defun bibtex-completion-clean-string (s)
>   "Remove quoting and superfluous white space from BibTeX field value in S."
>   (if s (replace-regexp-in-string "[\n\t ]+" " "
>          (replace-regexp-in-string "[\"{}]+" "" s))
>     nil))
> ```

Do you want define the function ourselves in view of the planned removal of BC dependency? I can do that.

BTW, since I didn't mentioned it earlier, the screenshot is what I get if I remove  `(t  . bibtex-completion-clean-string)` from `bibtex-actions-display-transform-functions`.

bdarcus:
> I will try to add support for pre-prints from arxiv at some point in the future. It should be pretty easy.

The basic idea of that function is prefer URLs based on the identifiers. Does that work for arxiv, or do I need to rethink?

bibtex-actions.el
bdarcus: > Do you want define the function ourselves in view of the planned removal of BC dependency? I can do that.

I think so, so that we can just replace little pieces as we go?

Might be it becomes simpler too; IDK.

aikrahguzar:
> > I will try to add support for pre-prints from arxiv at some point in the future. It should be pretty easy.
> 
> The basic idea of that function is prefer URLs based on the identifiers. Does that work for arxiv, or do I need to rethink?

That will work, I think. I don't know if this is standard bibtex but I get almost all my bib entries from https://inspirehep.net/ and they have entries  

```
    eprint = "1510.04430",
    archivePrefix = "arXiv",
```
For my case we can just assume that the entry is from arxiv and focus on the eprint field, the url would be https://arxiv.org/abs/1510.04430 

So we add `eprint` to the fields to parse (actually that should be done for pmid and pmcid too otherwise those links wouldn't be found) and then add it to the field list in `bibtex-has-a-value` too with the additional pcase expression,
```
("eprint" "https://arxiv.org/abs/")
``` 

The only thing is I don't know how much I am relying on my bibtex generator.

bdarcus:
Maybe the function needs to be more flexible?

bdarcus:
I also need to check how we do this in CSL JSON.

bibtex-actions.el
aikrahguzar: > I think so, so that we can just replace little pieces as we go?
> 
> Might be it becomes simpler too; IDK.

Done!

aikrahguzar:
> Maybe the function needs to be more flexible?

I have though about having some general configurable notion of resources related to a bib so that how files, urls etc are located can be configured. But it is another of those  things where I don't know if the added flexibility is worthwhile. I think arxiv will probably cover 99% of cases for preprints and we can wait till someone complains. 

Though if someone does we should look for a solution configurable so we just don't end up adding cases after cases.

bdarcus:
:-)

image

bdarcus:
Actually, I do have a question in order to address the second warning:

Is that 'width' variable no longer needed, such that it should be removed?

Issue #229, opened 2021-08-17
Redo bibtex-actions-open with a grouped list of related resources


bdarcus:
Related to #226, this is a proposal to redo `bibtex-actions-open` to present a UI like this after selecting the keys, where it shows related documents, both local files and external URLs

![image](https://user-images.githubusercontent.com/1134/129789881-fd1b66b1-0933-409b-b9b9-ed905d5c81b1.png)

I think this is a good idea; just a question of how best to do it technically. Here's the first step:

![image](https://user-images.githubusercontent.com/1134/129796737-bfe0a58e-7e04-423d-b6f9-4c277c8c114e.png)

```elisp
(defun bibtex-actions-open (keys)
  (interactive (list (bibtex-actions-read :rebuild-cache current-prefix-arg)))
  (let* ((files
         (bibtex-actions-file--files-for-multiple-keys
          keys
          (append bibtex-actions-library-paths bibtex-actions-notes-paths)
          bibtex-actions-file-extensions))
        (chosen-files (completing-read-multiple "Files: " files)))
        (cl-loop for file in chosen-files do
                 (funcall bibtex-actions-file-open-function file))))
```

bdarcus:
@minad - I have the basics of this working with grouped files:

![image](https://user-images.githubusercontent.com/1134/129800661-774c767e-e62c-4b7a-b4ec-682a58b6e09f.png)

How do I add links to the candidates list, with a different category?

I guess this is consult multi-source territory, but if it's easy, I'd prefer to avoid the dependency.

bdarcus:
This is partially addressed in #222.

bdarcus:
Seems easier to just use `consult--multi`, as `consult-dir` does:

![image](https://user-images.githubusercontent.com/1134/130327548-a230d8e0-2c6d-4103-a2e4-781948796c31.png)

bdarcus:
.. or even just customize `consult-buffer`?

In any case, what I want is conceptually similar, but for the sources to also include related files and links.

bdarcus:
I know you may not use this @aikrahguzar, but since you've worked on the related functionality, you may have some thoughts.

Here's a demo of what's possible with `consult--multi`; in this case mixing three sources specific to bibtex-actions, and one that consult.

![image](https://user-images.githubusercontent.com/1134/130358190-d13f4c09-c8c2-47a1-97c8-52de4fd5d24f.png)

Because they are each different categories, embark can attach different actions to them.

The questions I'm thinking about:

[...]

PR #251, opened 2021-08-22
refactor bibtex-actions-open

bdarcus:
Initial commit implement #229 as a simple flat list of related resources.

Still need:

1. ~add links~
2. `bibtex-actions-select-file`? 
3. probably also add a `consult--multi` option if present, rather than requiring consult; alternately, perhaps a better first step is to just add an annotation function to the simple flat version

If the latter, the annotation could just be types:

- DOI
- URL
- note
- file

...?


bdarcus:
@aikrahguzar  - I fixed the docstring issue and (more importantly) the compile error with the unused 'width' variable here. OK?

Issue #254, opened 2021-08-24
`cl-lib` vs `seq`


bdarcus:
@aikrahguzar I'm wondering if we should standardize on one of these?

I noticed you've used both in recent PRs.

Reminded of this because of a discussion on the doom discord, when Henrik expressed preference for `cl-lib` (larger API and older), but also thought `seq` was fine.

Other [opinions](https://www.reddit.com/r/emacs/comments/lp7zkw/why_not_cllib/goaroc0?utm_source=share&utm_medium=web2x&context=3) I've found have a preference for `seq`, but it seems to come down to personal preference/clarity, rather than any technical reason.

I chose to use `cl-loop` because it's intuitive to me (probably based on my experience with python), and because it's performance is supposedly really good.

If we should keep both, could you explain why, so I understand? I don't have much experience using `seq` ATM.

aikrahguzar:
> @aikrahguzar I'm wondering if we should standardize on one of these?
> 
> I noticed you've used both in recent PRs.
> 
> Reminded of this because of a discussion on the doom discord today, when Henrik expressed preference for `cl-lib` (larger API and older), but also thought `seq` was fine. He also, BTW, suggests ditching `f` and `s` for built-in functions, though I'm not sure there's a real alternative to `s-format` ATM, and this would be a short-term issue if we end up with `decoration-function`. Interestingly, `marginalia` doesn't use any kind of templating for it's annotation functions, but instead [this macro](https://github.com/minad/marginalia/blob/207eb405706f5415dd0daf490925648fd058cc42/marginalia.el#L375).
> 
> Other [opinions](https://www.reddit.com/r/emacs/comments/lp7zkw/why_not_cllib/goaroc0?utm_source=share&utm_medium=web2x&context=3) I've found have a preference for `seq`, but it seems to come down to personal preference/clarity, rather than any technical reason.
> 
> I chose to use `cl-loop` because it's intuitive to me (probably based on my experience with python), and because it's performance is supposedly really good.
> 
> If we should keep both, could you explain why, so I understand? I don't have much experience using `seq` ATM, but am happy to learn.
> 
> These are equivalent, and if anything, the former is faster.
> 
> ```emacs-lisp
> (seq-map #'1+ '(2 4 6))
> (cl-loop for i in '(2 4 6) collect (+ 1 i))
> ```
> 
> While I'm not sure if `seq` will work on hash-tables, `subr-x` has `hash-table-values`.

[...]

PR #257, opened 2021-08-24
refactor: cl-loop -> seq-map

bdarcus:
Picking up on #254, this is what it would look like.

WDYT @aikrahguzar?


bibtex-actions.el
```diff
@@ -355,65 +361,68 @@ personal names of the form 'family, given'."
...
-   for candidate in cands
-   collect
-   (let ((candidate-symbols (bibtex-actions--symbols-string
-                             (string-match "has:files" candidate)
-                             (string-match "has:note" candidate)
-                             (string-match "has:link" candidate))))
-     (list candidate candidate-symbols ""))))
+    (seq-map
+     (lambda (entry)
+       (let* ((citekey (bibtex-actions-get-value "=key=" entry))
```
bdarcus, line 373: Will this work for csl json?

I wasn't clear how to get the key from the hash here.

PR #265, opened 2021-08-31
fix: expand file name on file-field

bdarcus:
This should fix the path problem @thisirs

Let me know if you can test it.

### DIFF
--- a/bibtex-actions-file.el
+++ b/bibtex-actions-file.el
@@ -81,8 +81,9 @@ If you use 'org-roam' and 'org-roam-bibtex, you should use
            (results (seq-mapcat
                      #'possible-file-names-with-extension
                      extensions))
-           (file-field (bibtex-actions-get-value
-                        bibtex-actions-file-variable entry)))
+           (file-field (expand-file-name
+                        (bibtex-actions-get-value
+                         bibtex-actions-file-variable entry))))
       (when file-field (push file-field results))
       results)))
 


thisirs:
Thanks for the patch! This is currently expanding on `default-directory`. I was thinking of expanding it against all paths in `dirs`. Maybe something like this:

```lisp
(defun bibtex-actions-file--possible-names (key dirs extensions)
  "Possible names for files correponding to KEY with EXTENSIONS in DIRS."
  (cl-flet ((possible-file-names
             (basename &optional extension)
             (when basename
                  (seq-map
                   (lambda (directory)
                     (expand-file-name
                      (if extension (concat basename "." extension) basename)
                      directory))
                   dirs))))
    (let* ((entry (bibtex-actions-get-entry key))
           (file-field (bibtex-actions-get-value
                        bibtex-actions-file-variable entry)))
      (append
       (seq-mapcat (apply-partially #'possible-file-names key) extensions)
       (possible-file-names file-field)))))
```

bdarcus:
@aikrahguzar - can you look at this when you get a chance?

@thisirs  - in the meantime, it would still help to understand your use case. Could you post an example of say a tree of the dir in question, and what the contents of that field looks like?

thisirs:
Well I have a `biblio` directory where all the pdfs are stored. That tree is maintained by Zotero to reflect the structure of its internal library. Zotero also exports a bib file where the file field is the path of the pdf relative to `biblio`.
For example one file field is

    file = {RNN/Cho_et_al_2014_On_the_properties_of_neural_machine_translation.pdf}

That file is stored in the `RNN` subdir of `biblio`. So for now I can't reach for that file.

Things can get more complicated when the file field looks like this

      file = {graphs_and_deep_learning/Coates_Ng_2011_Selecting_Receptive_Fields_in_Deep_Networks.pdf;../../../../Zotero/storage/G37E6X3M/4293-selecting-receptive-fields-in-deep-networks.html}

And this is only the way Zotero exports its "attachments" in a bib file. Jabref and Mendeley are most likely encoding things differently...

So my problem is easily solved by changing `bibtex-actions-file--possible-names` as shown above but the problem is more general.

bdarcus:
Thanks; that will be helpful.

Have you not tried Zotfile?

That's what I use. So I have it rename the files based on the key, and move them all to a single directory.

bdarcus:
Also, the second example you have is a list of files, one PDF, and the other HTML.

> And this is only the way Zotero exports its "attachments" in a bib file. Jabref and Mendeley are most likely encoding things differently...

I'd like to know how these work too.

thisirs:
> That's what I use. So I have it rename the files based on the key, and move them all to a single directory.

That's what I use too! That would indeed solve my issue because we wouldn't need the file field in that case. But I still want pdfs with more meaningful names than just a bibtex key (I am sharing it with colleagues) and subdirectories (I don't want to share the whole library)

> I'd like to know how these work too.

Looking at the code, Jabref seems to be using a 2D array encoding in its file field with ; and : as separators :exploding_head:

bdarcus:
> Looking at the code, Jabref seems to be using a 2D array encoding in its file field with ; and : as separators

Seems Mendeley also does some weird things.

Maybe there should be a configurable function for parsing the field content?

PR #266, opened 2021-09-01
Implement custom file field parser

thisirs:


bdarcus:
Thank you!!

I'll take a look at this a bit later today.

@aikrahguzar

bibtex-actions-file.el
```diff
@@ -50,6 +50,12 @@ If you use 'org-roam' and 'org-roam-bibtex, you should use
   :group 'bibtex-actions
   :type '(function))
 
+(defcustom bibtex-actions-file-parser-function
+  'bibtex-actions-file-parser-default
+  "Function to parse file field"
```
bdarcus, line 55:
```suggestion
  "Function to parse file field."
```

bibtex-actions-file.el
```diff
@@ -68,6 +74,19 @@ If you use 'org-roam' and 'org-roam-bibtex, you should use
...
+  "Return a list of files from base directories DIRS and FILE-FIELD."
+  (seq-map (lambda (dir) (expand-file-name file-field dir)) dirs))
+
+(defun bibtex-actions-file-parser-zotero (dirs file-field)
+  "Return a list of files from DIRS and a Zotero formatted FILE-FIELD."
+  (cl-loop
+   for dir in dirs
+   for field in (split-string file-field ";")
+   for filename = (expand-file-name field dir)
+   if (file-regular-p filename)
```
bdarcus, line 87: I had recently removed use of `cl-loop` in favor of the `seq` and `dolist` functions for consistency, but I can't figure out an elegant way to do that here.

Should this be "when" though, since there's no "then" clause?

```suggestion
   when (file-regular-p filename)
```

bibtex-actions-file.el
```diff
@@ -50,6 +50,12 @@ If you use 'org-roam' and 'org-roam-bibtex, you should use
   :group 'bibtex-actions
   :type '(function))
 
+(defcustom bibtex-actions-file-parser-function
+  'bibtex-actions-file-parser-default
+  "Function to parse file field"
+  :group 'bibtex-actions
+  :type '(function))
```
bdarcus, line 57: Two questions, maybe either of which we need to answer for sure now:

1. should there be an explicit choice here?
2. do we think there might be need to allow more than one function?

bdarcus (commented):
A few of minor suggestions and questions below.

Hopefully @aikrahguzar will have time to take a look.

I've been trying to get the CI to be more helpful again, as it flags false positives these days. Sorry about that. But it did flag a minor docstring issue I note below.

bdarcus:
Should this return two results?

```elisp
ELISP> (bibtex-actions-file-parser-zotero '("~/org/roam") "biblio/ahrens2017.org;biblio/perl2015.org")
("/home/bruce/org/roam/biblio/ahrens2017.org")
```

thisirs:
Not a `cl-loop` fan either. I even got it wrong with nested for loops... Your version is working and even if files are not filtered by `file-regular-p` or `file-exists-p` they will be in `bibtex-actions-file--files-for-key`.

Force pushed changes sorry...

bibtex-actions-file.el
thisirs: Basic support (just a path) seems to be a good default. If more than one parser is required we can support a list of functions to call or provide the user a way to create another custom function doing the job (`seq-mapcat (lambda (fun) (apply fun ...)))

bdarcus:
> Not a `cl-loop` fan either.

I don't actually mind it, since it reminds me of previous experience with python and such.

But it seems not everyone agrees, and I was aiming for consistency.

bibtex-actions-file.el
aikrahguzar: I think a single functions is a good thing here. Especially since the functions takes arguments and is not just for side effects. I think if we allow multiple functions, the intent would be try a bunch of parsers until one return non nil? That can be done pretty easily directly too. 

I think there should be some explanations of what the arguments for the function are in the docstring of the `defcustom`.

aikrahguzar:
> Thank you!!
> 
> I'll take a look at this a bit later today.
> 
> @aikrahguzar

I looked very briefly at the merged changes and they look good. I am running around a lot but I think I can do a bit this weekend if something is pending, otherwise I will take a stab at the tasks in removing `BC` dependency. It seems like everything is shaping up really nicely overall.

thisirs:
I didn't test for a missing file filed in function `bibtex-actions-file-parser-default` and `expand-file-name` breaks. We can either add a test in `bibtex-actions-file-parser-default` or directly in `bibtex-actions-file--possible-names` to prevent future bugs. What do you think?

aikrahguzar:
> I didn't test for a missing file field in function `bibtex-actions-file-parser-default` and `expand-file-name` breaks. We can either add a test in `bibtex-actions-file-parser-default` or directly in `bibtex-actions-file--possible-names` to prevent future bugs. What do you think?

I think changing 
```elisp
(results-file (funcall bibtex-actions-file-parser-function dirs file-field))
```
to 
```elisp
(results-file (when file-field (funcall bibtex-actions-file-parser-function dirs file-field)))
```
in `bibtex-actions-file--possible-names` should fix this.

Issue #269, opened 2021-09-03
Add a note template


bdarcus:
The current default note function is simple, with no ability for a user to change the output, other than to use a different function.

It would be nice to have at least the title field be configurable with a template.

I see two options:

1. just have a template for this exactly the same as for the library display, and use `format-entry` on it
2. use `org-capture` somehow

1 seems simpler/more general, and would just be something like:

```elisp
(setq bibtex-actions-note-title-template 
  "Notes on {author editor:15}, {title:40}")
```

bdarcus:
@aikrahguzar - shouldn't this work?

```elisp
(defvar my/temp "Notes on ${author editor:15}, ${title:40}")

(bibtex-actions--format-entry
        (bibtex-actions-get-entry "low2001")
        250
        my/temp)
```

And if I go this way, how should that relate to `bibtex-actions-template`? Maybe it shouldn't be a cons?

Finally, while I think about, should I change the function name to `bibtex-actions-format-entry`?

aikrahguzar:
I am not sure where the notes template is going to be used. From the example it seems like a completely different display from `bibtex-actions-select-keys`, is that right? If that is the case I think we have two disjoint situations,
1) We can have a template with explicit widths with a column view
2) No explicit widths in any field so the view is that of a single string

i.e we don't have to deal with a situation where some fields have a width but others don't. In this case it seems simpler to just define another formatting function for the second case. I think the following will work, (really bad name but it is just for illustration)
```elisp
(defun bibtex-actions--format-entry-no-widths (entry format-string)
  "Formats a BibTeX ENTRY for display in results list. and the display format is governed by
FORMAT-STRING."
  (s-format
   format-string
   (lambda (raw-field)
     (let ((field-names (split-string (car field) "[ ]+")))
       (bibtex-actions-display-value field-names entry)))))
``` 
And then we can add the formatting function as an additional (possibly optional) argument to the `bibtex-actions--format-candidates` and `bibtex-actions--format-entry` and pass different functions depending on the command called, and in that case it does make sense to make them public.

bdarcus:
> I am not sure where the notes template is going to be used. From the example it seems like a completely different display from `bibtex-actions-select-keys`, is that right?

Yes. It's for inserting content in a new file.

But I can imagine it being useful for other, non-select-keys, use cases.

bdarcus:
Mostly implemented on linked PR, with following default note template string:

```elisp
[...]

PR #274, opened 2021-09-06
feat: add new format-entry function, note template

bdarcus:
Per #269, this adds a template for formatting new note titles, and a function to handle it.

It has occurred to me maybe this function should be called `format-entry`, and the original one renamed `format-entry-columns`?

Note: there's a [little change](https://github.com/bdarcus/bibtex-actions/pull/274/commits/6ddbe996b6d6d3c6b75379ac0ad7e80caa5b0c92#diff-43b58707c1adce5c6f3e1ed8d97ea5665182d94767472f7c196d630ff3ca0c3cL452) I made here that exposed a much bigger infinite recursion bug that I can't figure out how to fix otherwise.

```elisp
ELISP> (bibtex-actions--format-candidates bibtex-actions-bibliography)
*** Eval error ***  Lisp nesting exceeds ‘max-lisp-eval-depth’
```

Backtrace:

```elisp
Debugger entered--Lisp error: (error "Lisp nesting exceeds ‘max-lisp-eval-depth’")
  #<subr F616e6f6e796d6f75732d6c616d626461_anonymous_lambda_17>("${author editor:30}")
  replace-regexp-in-string("\\$\\({\\([^}]+\\)}\\|[0-9]+\\)" #<subr F616e6f6e796d6f75732d6c616d626461_anonymous_lambda_17> "${author editor:30}     ${date year issued:4}     ..." t t)
  s-format("${author editor:30}     ${date year issued:4}     ..." #<subr F616e6f6e796d6f75732d6c616d626461_anonymous_lambda_20>)
  #<subr F616e6f6e796d6f75732d6c616d626461_anonymous_lambda_21>("${author editor:30}     ${date year issued:4}     ...")
  mapcar(#<subr F616e6f6e796d6f75732d6c616d626461_anonymous_lambda_21> ("${author editor:30}     ${date year issued:4}     ..." "          ${=key= id:15}    ${=type=:12}    ${tags..."))
  #f(compiled-function #'sequence #<bytecode 0x18574107428dcbb4>)(#<subr F616e6f6e796d6f75732d6c616d626461_anonymous_lambda_21> ("${author editor:30}     ${date year issued:4}     ..." "          ${=key= id:15}    ${=type=:12}    ${tags..."))
  apply(#f(compiled-function #'sequence #<bytecode 0x18574107428dcbb4>) #<subr F616e6f6e796d6f75732d6c616d626461_anonymous_lambda_21> ("${author editor:30}     ${date year issued:4}     ..." "          ${=key= id:15}    ${=type=:12}    ${tags...") nil)
  seq-map(#<subr F616e6f6e796d6f75732d6c616d626461_anonymous_lambda_21> ("${author editor:30}     ${date year issued:4}     ..." "          ${=key= id:15}    ${=type=:12}    ${tags..."))
  seq-mapcat(#<subr F616e6f6e796d6f75732d6c616d626461_anonymous_lambda_21> ("${author editor:30}     ${date year issued:4}     ..." "          ${=key= id:15}    ${=type=:12}    ${tags..."))
  bibtex-actions--fields-in-formats()
  bibtex-actions--fields-to-parse()
  bibtex-actions--format-candidates(("/home/bruce/org/bib/newer.bib"))
```

### DIFF
--- a/bibtex-actions-file.el
+++ b/bibtex-actions-file.el
@@ -24,6 +24,8 @@
 
 (declare-function bibtex-actions-get-entry "bibtex-actions")
 (declare-function bibtex-actions-get-value "bibtex-actions")
+(declare-function bibtex-actions-get-template "bibtex-actions")
+(declare-function bibtex-actions--format-entry-no-widths "bibtex-actions")
 
 ;;;; File related variables
 
@@ -165,9 +167,13 @@ use 'orb-edit-note' for this value."
             (file-exists (file-exists-p file)))
       (funcall bibtex-actions-file-open-function file)
     (let* ((uuid (org-id-new))
-           (title (bibtex-actions-get-value "title" (bibtex-actions-get-entry key)))
+           (entry (bibtex-actions-get-entry key))
+           (note-title
+            (bibtex-actions--format-entry-no-widths
+             entry
+             (bibtex-actions-get-template 'note)))
            (content
-            (concat ":PROPERTIES:\n:ID:  " uuid "\n:END:\n#+title: Notes on " title "\n")))
+            (concat ":PROPERTIES:\n:ID:  " uuid "\n:END:\n" note-title "\n")))
       (funcall bibtex-actions-file-open-function file)
       (insert content))))
 
--- a/bibtex-actions-file.el
+++ b/bibtex-actions-file.el
@@ -45,7 +45,7 @@
   :type '(boolean))
 
 (defcustom bibtex-actions-file-open-note-function
-  'bibtex-actions-file-open-notes-default
+  'bibtex-actions-file-open-notes-default-org
   "Function to open and existing or create a new note.
 
 If you use 'org-roam' and 'org-roam-bibtex', for example, you can
@@ -158,7 +158,7 @@ use 'orb-edit-note' for this value."
                   nil 0 nil
                   file)))
 
-(defun bibtex-actions-file-open-notes-default (key)
+(defun bibtex-actions-file-open-notes-default-org (key)
   "Open a note file from KEY."
   (if-let* ((file
              (caar (bibtex-actions-file--files-to-open-or-create
@@ -168,12 +168,12 @@ use 'orb-edit-note' for this value."
       (funcall bibtex-actions-file-open-function file)
     (let* ((uuid (org-id-new))
            (entry (bibtex-actions-get-entry key))
-           (note-title
+           (note-meta
             (bibtex-actions--format-entry-no-widths
              entry
              (bibtex-actions-get-template 'note)))
            (content
-            (concat ":PROPERTIES:\n:ID:  " uuid "\n:END:\n" note-title "\n")))
+            (concat ":PROPERTIES:\n:ID:  " uuid "\n:END:\n" note-meta "\n")))
       (funcall bibtex-actions-file-open-function file)
       (insert content))))
 
--- a/README.org
+++ b/README.org
@@ -141,16 +141,6 @@ If you prefer to use a hydra-based in-buffer interface, you can also use that pr
 (setq org-cite-activate-processor 'org-ref-cite-follow)
 #+END_SRC
 
-** Test Script
-    :PROPERTIES:
-    :CUSTOM_ID: test-script
-    :END:
-
-The repository =test= directory also includes a script you can use to run this and associated packages in the =emacs -Q= sandbox.
-To do that, simply run =./run.sh= from the =test= directory.
-By default, this will use selectrum as the completion system.
-If you would like to try vertico instead, just do =M-x vertico-mode=.
-
 ** Rich UI
     :PROPERTIES:
     :CUSTOM_ID: rich-ui
@@ -168,17 +158,26 @@ For the prefix, you can filter for associated PDFs or notes using =has:pdf= or =
 #+CAPTION: UI sections
 [[file:images/ui-segments.png]]
 
-You can configure both of the last two similar as you do with bibtex-completion.
+*** Templates
+
+The =bibtex-actions-templates= variable configures formatting for these sections, as well as the default note function.
+Here's the defaults:
 
 #+BEGIN_SRC emacs-lisp
-  (setq bibtex-actions-template '((t . " ${title:*}")))
-  (setq bibtex-actions-template-suffix '((t . "          ${=key=:15}")))
+(setq bibtex-actions-templates
+      '((main . "${author editor:30}     ${date year issued:4}     ${title:48}")
+        (suffix . "          ${=key= id:15}    ${=type=:12}    ${tags keywords:*}")
+        (note . "#+title: Notes on ${author editor}, ${title}")))
 #+END_SRC
 
 Note:
 
-1. the asterisk signals to the formatter to use available space for the column: you should only use this on one field total, across the two templates, for the formatting to work correctly.
-2. the inclusion of CSL-JSON support is currently experimental, and relies on =bibtex-actions-field-map= variable to map among the different supported formats, and is subject to change.
+1. You may include multiple variables in a field; the formatter will print the first one it finds.
+2. If you plan to use CSL JSON at all, you can and should include CSL JSON variables names where appropriate as such options. The default main template dates field demonstrates this.
+3. The asterisk signals to the formatter to use available space for the column.
+4. The note template does not take widths, as formatting is inline there rather than columnar.
+
+*** Icons
 
 By default, this UI is plain text, but you can configure it to use icons instead.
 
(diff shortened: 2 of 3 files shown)


bibtex-actions-file.el
```diff
@@ -165,9 +167,13 @@ use 'orb-edit-note' for this value."
...
     (let* ((uuid (org-id-new))
-           (title (bibtex-actions-get-value "title" (bibtex-actions-get-entry key)))
+           (entry (bibtex-actions-get-entry key))
+           (note-title
+            (bibtex-actions--format-entry-no-widths
+             entry
+             (bibtex-actions-get-template 'note)))
            (content
-            (concat ":PROPERTIES:\n:ID:  " uuid "\n:END:\n#+title: Notes on " title "\n")))
+            (concat ":PROPERTIES:\n:ID:  " uuid "\n:END:\n" note-title "\n")))
```
bdarcus, line 176: There's an issue here:

Right now, I have half of this which is handled via template, and the other via hard-coded function logic.

It might be better to allow the template a `uuid-new` or some such field, so the template could include of it:

```elisp
 ":PROPERTIES:\n:ID: ${uuid-new}\n:END:\n#+title: Notes on ${author editor}, ${title}"
```

bdarcus: Maybe better to use something like this, to remove the `org-id` dependency?

http://ergoemacs.org/emacs/elisp_generate_uuid.html

aikrahguzar:
> Note: there's a [little change](https://github.com/bdarcus/bibtex-actions/pull/274/commits/6ddbe996b6d6d3c6b75379ac0ad7e80caa5b0c92#diff-43b58707c1adce5c6f3e1ed8d97ea5665182d94767472f7c196d630ff3ca0c3cL452) I made here that exposed a much bigger infinite recursion bug that I can't figure out how to fix otherwise.

Is it adding the `if` statement causing the infinite loop? That is very weird, because there is no recursion there at all. I checked by evaluating 

```elisp
(defun my-ba-get-entry (key)
  "Return the cached entry for KEY."
  (if (or (eq 'uninitialized bibtex-actions--candidates-cache)
          (eq 'uninitialized bibtex-actions--local-candidates-cache))
      (message "Something is wrong; your library is not initialized.")
    (cddr (seq-find
           (lambda (entry)
             (string-equal key (cadr entry)))
           (bibtex-actions--get-candidates)))))
```
and the calling it with a key and it seems to return the entry as expected. 

I am setting up at a new place so won't do much anytime soon but I will try to track down infinite recursion bug if I can reproduce it.

bdarcus:
> Is it adding the `if` statement causing the infinite loop?

No; the reverse. Adding that statement originally "fixed" it.

> I am setting up at a new place so won't do much anytime soon but I will try to track down infinite recursion bug if I can reproduce it.

OK.

bdarcus:
> Adding that statement, along with changing the load order of some functions, originally "fixed" it.

I added this check back (which again "fixed" it), but also added a "FIX" comment, because I don't think this check should be necessary if everything else is correctly coded.

aikrahguzar:
@bdarcus, does changing `ba--get-candidates` to
```elisp
(defun bibtex-actions--get-candidates (&optional force-rebuild-cache)
  "Get the cached candidates.
If the cache is unintialized, this will load the cache.
If FORCE-REBUILD-CACHE is t, force reload the cache."
  (when force-rebuild-cache
    (bibtex-actions-refresh force-rebuild-cache))
  (when (eq 'uninitialized bibtex-actions--candidates-cache)
    (bibtex-actions-refresh nil 'global))
  (when (eq 'uninitialized bibtex-actions--local-candidates-cache)
    (bibtex-actions-refresh nil 'local))
  (seq-concatenate 'list
                   bibtex-actions--local-candidates-cache
                   bibtex-actions--candidates-cache))
```
helps with the recursion blowing up? There is certainly a bug as currently the `'local` and `'global` checks are inside an `if` statement that won't be executed unless `force-rebuild-cache` is non-`nil`. Which it won't be in the normal first usage and that behavior would be fixed by checking for uninitialized variables for in `ba-get-entry`.

I don't see why it would cause infinite recursion instead of a type error though, unless it is something that happens during `completing-read` part. That seems unlikely to me but I don't understand that part well.

I would test and do a pr but I need to setup my laptop since the last one went with my last job.

bdarcus:
> does changing `ba--get-candidates` to ... helps with the recursion blowing up?

Alas, no.

aikrahguzar:
@bdarcus  I figured out the problem, but I don't have a good solution yet. It is the use of `bibtex-actions-file--files-for-key` in the `bibtex-actions-file--format-candidates`. `bibtex-actions-file--files-for-key` calls `bibtex-actions-file--possible-names` which calls `ba-get-entry` which calls `ba-get-candidates` and `ba-get-candidates` will call `bibtex-actions-file--format-candidates` because the cache isn't built yet causing bad recursion. The check for initialization then allow us to proceed.  This is a pretty subtle result of recent changes to parse the file field. I think the solution should be to change `ba-file` functions to take the extra `entry` argument and pass the entry directly for `ba--format-candidates` and use the `ba-get-entry` in interactive functions that need the files to supply this new entry argument.

This has me thinking that `ba-get-entry` exists only to overcome a shortcoming in the `ba-select-keys`. We store the entry but then throw it away when we return only the keys in `ba-select-keys` and then we need to go through the whole candidate list to find the entry later when we need it. It is inefficient but that doesn't really matter since we only need to lookup a few selected keys but it can also cause this kind of issue. We can change `ba-select-keys` to return an alist of `(key . entry)` but that will probably need some adjustment for embark integration too.

bdarcus:
Maybe get-entry needs to grab from the candidate cddr then?

If that's possible ...

aikrahguzar:
> Maybe get-entry (or just however to get the necessary data) needs to grab from the candidate cddr then? If that's possible; not sure it is.

Something like this would be my preferred solution. Where `select-keys` returns alist but a lot of (I think all the interactive) functions would needed to be update to deal with this change. Though the change would just be to change the `key` to `key-entry` and then get the key by  (car key-entry). The other problem is embark's at point and we would need to change that case to fetch the entry like we do in `get-entry` now. That is the only place where I think the current solution is the optimal one.

bdarcus:
Looks like the only field files-for-key is currently accessing (indirectly) is `files`.

aikrahguzar:
> Looks like the only field files-for-key is currently accessing (indirectly) is `files`.

Yes, but I think generally the set of functions that depend on the caches being populated should be fairly limited, I guess ideally only to interactive functions.

aikrahguzar:
> Looks like the only field files-for-key is currently accessing (indirectly) is `files`.

Yes, but I think generally the set of functions that depend on the caches being populated should be fairly limited, I guess ideally only to interactive functions.

bdarcus:
One issue, the details of which I'm forgetting such that I'm unsure if it applies here, is that completing-read throws out non-candidate data at a certain point.

bdarcus:
Occurs to me also that it might be wise at some to add some tests, that we can also hook up to the CI.

aikrahguzar:
> One issue, the details of which I'm forgetting such that I'm unsure if it applies here, is that completing-read throws out non-candidate data at a certain point.

I think we already take care of it to find the key corresponding to the formatted string. I was thinking of changing this
```elisp
(cl-loop for choice in chosen
             ;; Collect citation keys of selected candidate(s).
             collect (or (cadr (assoc choice candidates))
             ;; Key is literal coming from embark, just pass it on
                          choice))
```
to something like this
```elisp
(seq-map (lambda (choice) (seq-find
                            (lambda (cand) (cdr (or (equal choice (car cand))
                                             ;; For embark, find the cand with correct key
                                                    (equal choice (cadr cand)))))
                          candidates))
             chosen)
```
Basically change `cadr` to `cdr` so that the function calling `select-keys` can use both the key and entry.

bdarcus:
> Basically change `cadr` to `cdr` so that the function calling `select-keys` can use both the key and entry.

I pushed #274 before seeing this, but will take a look sometime today or tomorrow.

Are you thinking of this as a simpler alternative to #274?

aikrahguzar:
> I pushed #274 before seeing this, but will take a look sometime today or tomorrow.
> 
> Are you thinking of this as a simpler alternative to #274?

I briefly looked at #274 and it looks good. This is almost unrelated and the purpose is not to go through candidate list too many times. I think it cleaner if we have a separation of the part of the library that does caching and the rest of the library interacts with cached candidates only through `select-keys`.

bdarcus:
> I think it cleaner if we have a separation of the part of the library that does caching and the rest of the library interacts with cached candidates only through `select-keys`.

Yes, ideally they really should be cleanly-separated.

bdarcus:
> I pushed #274 before seeing this ...

Sigh ... I of course meant #281!

PR #291, opened 2021-09-15
Avoid empty dirs arg on possible-names

### LINKED ISSUE #285: bibtex-actions-open fails when bibtex-actions-library-paths is nil
**Describe the bug**
All of  the parsers defined in `bibtex-actions-file.el` map over the value `(append bibtex-actions-library-paths bibtex-actions-notes-paths)`. If both of these are set to nil, this expression evaluates to `nil` and the map function (which varies by parser) will fail.   Thus, even if file paths are absolute (which is the default in Better bibtex output), no file will be found.

**To Reproduce**
1. set both vars to `nil`
2. attempt to execute `bibtex-actions-file-open` on a valid key

**Expected behavior**
Absolute paths should be opened

**Emacs version:**

**Additional context**
I believe the sensible fix is to 
- set the defcustom to '("/") by default
- document both vars in the Readme. 

does that sound right?

bdarcus:
I don't have a ton of time ATM, and it's not easy for me to test this, but this is intended to fix #285.

Is something like this what you were meaning @titaniumbones? Feel free to submit another PR against this branch, or a new one against main, if not.

cc @thisirs

### DIFF
--- a/bibtex-actions-file.el
+++ b/bibtex-actions-file.el
@@ -124,7 +124,11 @@ If you use 'org-roam' and 'org-roam-bibtex', you can use
             (when file-field
               (seq-mapcat
                (lambda (func)
-                 (funcall func dirs file-field))
+                 (funcall
+                  func
+                  ;; Avoid an empty dirs arg.
+                  (or dirs (list default-directory))
+                  file-field))
                bibtex-actions-file-parser-functions))))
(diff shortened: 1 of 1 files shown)


bdarcus:
I had a chance to test this finally, and it doesn't work.

PR #293, opened 2021-09-16
refactor files functions to take KEY and ENTRY

### LINKED ISSUE #292: bibtex-actions-open-notes returns error
**Describe the bug**
[From the doom emacs PR](https://github.com/hlissner/doom-emacs/pull/5290). After setting up the biblio module according to the readme, including specifying my library, bib, and notes locations, `org-cite-insert` works correctly, and does accessing PDFs through `embark-act`. However, I cannot access notes through `embark-act` or by using `bibtex-actions-open-notes`. Instead, I get `find-file-noselect: Wrong type argument: stringp, nil`.

**Emacs version:**
28.0.50

bdarcus:
fix #292


bdarcus:
Can you please review this @aikrahguzar?

bibtex-actions-file.el
```diff
@@ -128,17 +127,17 @@ If you use 'org-roam' and 'org-roam-bibtex', you can use
...
-                (bibtex-actions-file--possible-names entry dirs extensions)))
+                (bibtex-actions-file--possible-names key entry dirs extensions)))
 
-(defun bibtex-actions-file--files-to-open-or-create (keys dirs extensions)
-  "Find files related to a list of KEYS in DIRS with extension in EXTENSIONS."
+(defun bibtex-actions-file--files-to-open-or-create (key entry dirs extensions)
+  "Find files related to a KEY, ENTRY in DIRS with extension in EXTENSIONS."
   (cl-flet ((files-for-key
-             (key)
+             (_k)
```
bdarcus, line 138: this seems unnecessary ATM

bdarcus:
Since this was a bug, I'm merging this. We can refine in a subsequent PR.

bibtex-actions-file.el
```diff
@@ -124,21 +123,24 @@ If you use 'org-roam' and 'org-roam-bibtex', you can use
             (when file-field
               (seq-mapcat
                (lambda (func)
-                 (funcall func dirs file-field))
+                 (funcall
+                  func
+                  ;; Make sure this arg is non-nil.
+                  (or dirs default-directory)
```
aikrahguzar, line 129: I don't think this a good idea, default directory is buffer local. While I think what was needed was to deal with absolute paths. This will add path relative to the default directory which is probably going to be the parent of latex/org file. I guess that can make sense for local bib.

bibtex-actions-file.el
```diff
@@ -124,21 +123,24 @@ If you use 'org-roam' and 'org-roam-bibtex', you can use
...
-                (bibtex-actions-file--possible-names entry dirs extensions)))
+                (bibtex-actions-file--possible-names key entry dirs extensions)))
 
-(defun bibtex-actions-file--files-to-open-or-create (keys dirs extensions)
-  "Find files related to a list of KEYS in DIRS with extension in EXTENSIONS."
-  (cl-flet ((files-for-key
-             (key)
+(defun bibtex-actions-file--files-to-open-or-create (key entry dirs extensions)
+  "Find files related to a KEY, ENTRY in DIRS with extension in EXTENSIONS."
+  (cl-flet ((files-for-key ()
```
aikrahguzar, line 141: This strange now, since the function definition in `cl-flet` takes no argument and the final result is just running it. The `cl-flet` here can be removed and the function body will then be just the sexp with `let*`.

bibtex-actions-file.el
```diff
@@ -177,11 +180,10 @@ If you use 'org-roam' and 'org-roam-bibtex', you can use
...
   "Open a note file from KEY and ENTRY."
-  ;; modify when this addressed:
-  ;; https://github.com/org-roam/org-roam-bibtex/issues/211
   (if-let* ((file
              (caar (bibtex-actions-file--files-to-open-or-create
-                    (list key)
+                    key
+                    entry
                     bibtex-actions-notes-paths '("org"))))
             (file-exists (file-exists-p file)))
```
aikrahguzar, line 188: `files-to-open-or-create` already checks whether file exists or not and that is in the cons cell so this is redundant. But this is also the only place `files-to-open-or-create`is used as far as I can tell. So maybe that can be removed and instead inlined here.

aikrahguzar:
> Since this was a bug, I'm merging this. We can refine in a subsequent PR.

Yes, I think it should work fine but I have left my nitpicks in the comments.

bibtex-actions-file.el
bdarcus: That's what i was thinking, and plan to do.

bibtex-actions-file.el
bdarcus: In my (non-exhaustive) experiments, if the file path is absolute, it just ignores this, but returns the absolute path.

I still don't like it though; there should be a more clear design that makes clear this will work with absolute paths and nil path variables for the library.

aikrahguzar: > In my (non-exhaustive) experiments, if the file path is absolute, it just ignores the dir arg, but returns the absolute path.
> 

I think empty string is a better choice in that case but it seems like end result will be the same. `expand-file-name` treats the directory argument relative to the `default-directory` if it isn't absolute.

bibtex-actions-file.el
```diff
@@ -177,11 +180,10 @@ If you use 'org-roam' and 'org-roam-bibtex', you can use
 
 (defun bibtex-actions-file-open-notes-default-org (key entry)
   "Open a note file from KEY and ENTRY."
-  ;; modify when this addressed:
-  ;; https://github.com/org-roam/org-roam-bibtex/issues/211
   (if-let* ((file
              (caar (bibtex-actions-file--files-to-open-or-create
-                    (list key)
+                    key
+                    entry
```
aikrahguzar, line 186: I think replacing `entry` with `nil` here should fix the other bug found today, which is due the `file` field which doesn't contain a note getting parsed and then opened. That field is the only reason `entry` is needed by `possible-names`.

bdarcus: And the `entry` is only needed there to get the file field. Hmm ...

aikrahguzar: > And the `entry` is only needed there to get the file field. Hmm ...
> 
> Maybe if I remove and inline this I can figure something out.

Inlining will move the same question to `possible-names`. Entry should be passed to `possible-names` if we want to parse the file-field and not otherwise. Setting entry to `nil` is slightly iffy but I think not worth changing `possible-names`.

bdarcus: I haven't looked closely yet, but what is entry optional?

aikrahguzar: Optional entry should work but would require changing the order of the arguments.

Issue #273, opened 2021-09-04
Refactor `bibtex-actions-open-link` to use `bibtex-actions-get-link`


bdarcus:
Currently the command hard-codes DOIs as alternative to URL; should get use the `bibtex-actions-get-link` function instead so more general. That function should also, perhaps, return a URL if other options not present.

Issue #226, opened 2021-08-16
How should new "open" commands work?


bdarcus:
The WIP refactor in #222 includes removing the bibtex-completion dependency for the "open" commands.

This gives us opportunities, but also can be a bit overwhelming.

So how should these work?

My current thinking is:

- `bibtex-actions-open` optionally prompts with grouped files, (maybe) links, notes options; a kind of dashboard for related resources
- `bibtex-actions-open-pdf` rename to `bibtex-actions-open-library-files`, but opens directly, but function is a configurable choice of `find-file` (that will open files in Emacs), `xdg-open` (that will open externally), etc
- `bibtex-actions-open-notes` uses `find-file` to open within Emacs

Other options:

- Consult has [`consult-find`](https://github.com/minad/consult/blob/271b89af7f59acaa54123d36abb9efdcea0ddc38/consult.el#L4323) and [`consult-file-externally`](https://github.com/minad/consult/blob/271b89af7f59acaa54123d36abb9efdcea0ddc38/consult.el#L3146) (supports wider platform range), though can just borrow some of this code.

See #191.

bdarcus:
OK, @salutis @aikrahguzar - here's where #222 stands; I have:

1. changed `bibtex-actions-read` to use `completing-read` by default, with an optional `:multiple` arg
2. changed `bibtex-actions-open-pdf` to `bibtex-actions-open-library-files`, which uses `completing-read` (so you select one candidate; not multiple)
3. removed the `bibtex-completion` dependency on this, so handle it all internally
4. added a `bibtex-actions-open-library-file-external` defcustom on the file-open function, and when t, `bibtex-actions-open-files` uses `xdg-open` by default, for example, but works on other platforms.

I'm kind of stuck now, though, on the question of how this should work. Ideally I just want to merge #222 and then discuss the broader possibilities at a more leisurely pace.

All of this stuff is connected though; if we want the open commands to use multiple, then I need to change other code.

bdarcus:
Actually, I think maybe all commands should just take a list of keys. Otherwise it just gets complicated.

aikrahguzar:
>     1. changed `bibtex-actions-read` to use `completing-read` by default, with an optional `:multiple` arg

Why this? What we have now seems really good to me. If I want 1 entry I can use enter to select one. If I want more multiple, I use tab to add them to selected group. I don't see any need for a separate option.   
 
[...]

PR #295, opened 2021-09-18
open file refinements

### LINKED ISSUE #282: PDF file always open externally, even with pdf-tools installed.
**To Reproduce**
`M-x bibtex-actions-open` and select an entry, hit enter, select a PDF file from the given options,
the PDF file is opened with browser somehow ... even tho I've pdf-tools installed.

Same thing happens if I call `embark-act` after `bibtex-actions-open` and select open library files.

Meanwhile `M-x bibtex-actions-open-library-files` will always open PDF files with pdf-tools.

**Expected behavior**
 Maybe I missed some settings, but I think bibtex-actions should use pdf-tools when possible.

my config:
```emacs-lisp
(straight-use-package 'bibtex-actions)

(with-eval-after-load 'bibtex-actions
  (setq bibtex-actions-open-library-file-external nil
		bibtex-actions-bibliography '("/path/to/bib")
		bibtex-actions-library-paths '("/path/to/lib")
		bibtex-actions-notes-paths '("/path/to/notes")
		bibtex-actions-file-variable "File"
		bibtex-actions-file-open-note-function #'orb-edit-note
		bibtex-actions-default-action #'bibtex-actions-open-library-files)
  (with-eval-after-load 'embark
	(add-to-list 'embark-target-finders 'bibtex-actions-citation-key-at-point)
	(add-to-list 'embark-keymap-alist '(bib-reference . bibtex-actions-map))
	(add-to-list 'embark-keymap-alist '(citation-key . bibtex-actions-buffer-map))))
```

bdarcus:
Fix #282, #273; see also #226 and #229

cc @pRot0ta1p @salutis

I've implemented changes here to `bibtex-actions-open-library-files` and `bibtex-actions-open` interactive functions, and related defcustoms.

So to configure flie handling, you can do this:

```elisp
(setq bibtex-actions-file-extensions '("org" "md" "pdf") ; open with the default open command; find-file
      bibtex-actions-file-extensions-external '("mp3" "png" "tiff" "html"))
```

WDYT? 

TODO:

1. need to make sure this works correctly with file-fields

Aside: occurs to me HTML may be a little tricky, because we pretty much only want to view these files; not edit them. But some may want to view in emacs.

### DIFF
--- a/README.org
+++ b/README.org
@@ -305,6 +305,15 @@ If you have a mix of entries created with Zotero and Calibre, for example, you c
     bibtex-actions-file-parser-calibre))
 #+END_SRC
 
+These two variables govern what functions will open which extensions.
+
+#+BEGIN_SRC emacs-lisp
+(setq bibtex-actions-file-extensions '("org" "md" "pdf")
+      bibtex-actions-file-extensions-external '("mp3" "png" "tiff" "html"))
+#+END_SRC
+
+The first will open using =bibtex-actions-open=, which uses the function specified in =bibtex-actions-file-open-function= (the default here is =find-file=), while the second will use =bibtex-actions-file-open-external=.
+
 ** Usage
    :PROPERTIES:
    :CUSTOM_ID: usage
--- a/README.org
+++ b/README.org
@@ -305,14 +305,7 @@ If you have a mix of entries created with Zotero and Calibre, for example, you c
     bibtex-actions-file-parser-calibre))
 #+END_SRC
 
-These two variables govern what functions will open which extensions.
-
-#+BEGIN_SRC emacs-lisp
-(setq bibtex-actions-file-extensions '("org" "md" "pdf")
-      bibtex-actions-file-extensions-external '("mp3" "png" "tiff" "html"))
-#+END_SRC
-
-The first will open using =bibtex-actions-open=, which uses the function specified in =bibtex-actions-file-open-function= (the default here is =find-file=), while the second will use =bibtex-actions-file-open-external=.
+The =bibtex-actions-file-extension= variable governs which file extensions the open commands will recognize.
 
 ** Usage
    :PROPERTIES:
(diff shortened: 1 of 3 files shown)


pRot0ta1p:
Thank you ! I have tried this pull request and its perfect :smiley:  .

As for HTMLs, my work flow only uses PDFs and EPUBs, I think maybe make it force open internally when user invoke with 2 prefix-arguments is good enough?

bdarcus:
> TODO:
> 
> 1. need to make sure this works correctly with file-fields

@thisirs @aikrahguzar - to merge this, I need to modify (I think) possible names further so that it will consider the `extensions` arg of that function.

Any thoughts on how best to do that?

salutis:
> That `test.pdf` file should only be listed once, and when chosen, should open either internally or externally depending on how the `bibtex-actions-file-extensions*` vars are set. So I can't just do `delete-dups` on the list returned by `bibtex-actions-file--possible-names`.

If `pdf` figures in both `bibtex-actions-file-extensions` and `bibtex-actions-file-extensions-external`, I would expect the file show twice, e.g. as `/tmp/test.pdf` and once `/tmp/test.pdf (external)`.

bdarcus:
> If `pdf` figures in both `bibtex-actions-file-extensions` and `bibtex-actions-file-extensions-external`, I would expect the file show twice, e.g. as `/tmp/test.pdf` and `/tmp/test.pdf (external)` or something along those lines.

True, and possible.

If pdf is only in one, however, it should only show up once, which was the expected here.

But I actually can't reproduce this after restarting Emacs, so it might have just been spurious behavior. OTOH, now I don't see `test.pdf`.

:shrug:

bdarcus:
It may be better to remove the "external" support.

Consult has `consult-external`, which one can access via `embark-act` on a file candidate.

Issue #302, opened 2021-09-23
Enhance citation editing


bdarcus:
Currently, `oc-bibtex-actions-insert` uses `org-cite-make-insert-processor`, which generally works well, and avoids us having to maintain code like this.

But there are some limitations in it. See, for example, #300.

One obvious simple improvement would be to add a delete-reference command, which would just pass an argument to `org-cite-insert`, and be available at-point via `embark-act`.

It also may make sense to port some of the org-ref-cite functions for this.

A final possibility is to rewrite `org-cite-make-insert-processor` for our needs, but I'm not eager to do that personally, both because of a lack of time, and lack of knowledge of that code in org, etc.

PR #303, opened 2021-09-23
add org-cite citation editing commands

bdarcus:
This adapts some of the functions from org-ref-cite (whose future is now unclear) for modifying citations.

The delete-citation and kill are simple no-brainers.

The "shift" ones seem useful to me, and would ideally work the same as in org-ref-cite; they'd be attached via keymap face, so one could use arrow keys. 

Not sure what else, given we already cover the key ones.

I could add the pre/post (affix) one, but not sure how useful that is, given how can just directly add the content in the buffer.

See #302; cc @andras-simonyi (both because related to previous discussion of activate processors, and because you might have input on my note below)

### DIFF
--- a/oc-bibtex-actions.el
+++ b/oc-bibtex-actions.el
@@ -44,7 +44,10 @@
 
 (declare-function bibtex-actions-at-point "bibtex-actions")
 (declare-function org-open-at-point "org")
+(declare-function org-element-property "org")
+(declare-function org-element-type "org")
 (declare-function org-cite-make-insert-processor "oc")
+(declare-function org-cite-get-references "oc")
 
 (defface oc-bibtex-actions-style-preview
   ;; Not sure if this is the best parent face.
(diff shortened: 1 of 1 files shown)


oc-bibtex-actions.el
```diff
@@ -223,6 +226,72 @@ strings by style."
...
+  lst)
+
+(defun oc-bibtex-actions--shift-reference (direction datum)
+  "When point is on a citation-reference DATUM, shift it in DIRECTION."
+  ;; adapted from org-ref-cite
+  (let*  ((current-citation (if (eq 'citation (org-element-type datum)) datum
+                             (org-element-property :parent datum)))
+         (current-ref (when (eq 'citation-reference (org-element-type datum)) datum))
+         (refs (org-cite-get-references current-citation))
+         (index
```
bdarcus, line 250: Index isn't currently calculated correctly; just nil.

bdarcus: @aikrahguzar - if and when you have a moment, can you take a look at this and see if something jumps out that I'm missing?

bdarcus: I got this working, but I don't quite understand how. It could be the code in ORC was more complex because of its internal details?

bdarcus:
The delete-citation and shift-left and shift-right functions are now working @andras-simonyi.

Issue #278, opened 2021-09-07
bibtex-actions-insert-citation vs org-cite-insert


bdarcus:
Not sure how to handle this, but when using oc-bibtex-actions with org, one should not use `bibtex-actions-insert-citation` (or even `bibtex-actions-insert-key`), and instead use the native `org-cite-insert`.

Ideas I can think of:

1. don't autoload those commands at all
2. only autoload them subject to some condition; maybe a variable
3. don't touch autoload at all, but add a message when using the commands in org

liebkne:
I have a large bibtex file (1500 entries)

`org-cite-insert` is super slow at the first time call (34s) while `bibtex-actions-insert-citation` is much faster (3s). I'd like to use `bibtex-actions-insert-citation` even with `oc`.

bdarcus:
It's slow because you're using oc-basic for the insert processor; not bibtex-actions.

liebkne:
Yes, that's why I said I still prefer to use `bibtex-actions-*` commands in this situation. It would be nice if this awesome package can deal with the new style `oc` link syntax.

bdarcus:
> Yes, that's why I said I still prefer to use `bibtex-actions-*` commands in this situation. It would be nice if this awesome package can deal with the new style `oc` link syntax.

It can ;-)

https://github.com/bdarcus/bibtex-actions#configuration

Note: I may need to fix the example to include explicitly doing `setq org-cite-insert-processor 'oc-bibtex-actions`.

liebkne:
Thanks, that works.

It would be helpful to include that in README. And I've thought `(setq org-cite-insert-processor 'oc-bibtex-actions)` will make `bibtex-action-insert-citations` to work with `oc` directly and now I understand I should use `org-cite-insert` even it's set :)

bdarcus:
> It would be helpful to include that in README. 

Well, the link I posted was to the README. What more should I add; something in the first lines?

> And I've thought `(setq org-cite-insert-processor 'oc-bibtex-actions)` will make `bibtex-action-insert-citations` to work with `oc` directly and now I understand I should use `org-cite-insert` even it's set :)

Yes, we basically plugin to the org-cite infrastructure, and let it take over.

That's what this issue is about, because it confuses users. Basically, when using org-cite, there's no reason for the insert commands.

liebkne:
I mean the `(setq org-cite-insert-processor 'oc-bibtex-actions)` part would be helpful for newbies :)

Issue #245, opened 2021-08-20
Rewrite format-reference to use citeproc-el


bdarcus:
If replacing bibtex-completion function for this, might as well use citeproc, since it would provide much more flexibility (mode-specific output, thousands of CSL styles for formatting specification, etc. etc.)..

Also, citeproc is needed anyway if and when we implement things like style previews in `oc-bibtex-actions`.

Downside is an additional dependency.

localauthor:
Simple question: is the intent here to replace `bibtex-completion-insert-reference`?

If so, would the attendant flexibility you mention include allowing references to be formatted in different styles (not just APA, but, Chicago, MLA, Harvard, etc.)? Inquiring humanists want to know ;)

bdarcus:
Yes, and if integrating citeproc-el, yes.

But while I have you, what's your use case for this? Are you using org here primarily, for example?

localauthor:
Yes, I write manuscripts and take readings notes in org.

In manuscripts I use org-cite, so I don't need `insert-reference`, because it's enough to have everything formatted upon export. But I don't export my reading notes, so if I want to include a formatted list of sources in a note---in an annotated bibliography, say---`insert-reference` is a quick way to do that. I'm just not used to APA, so being able to use Chicago would be nice.

bdarcus:
I'm wondering if for annotated bibs and CVs if there's a better way using `oc-csl|?

I think @titaniumbones had asked about this, and found a CSL style designed for this, where you could just use citations, and they would be formatted as references.

titaniumbones:
oops, sorry for the delay. I did find a CSL style for this -- it's apa-cv.csl, and I can't quite remember where I found it but it wasn't hard. I will make a Chicago version of it too, when I have a sec. If it seems useful to other people, I can put both up in github somewhere once I dig out a bit.

bdarcus:
Maybe we could move this command out of the main file, and create a new `bibtex-actions-citeproc.el` file where a revised version could go.

`oc-bibtex-actions` could then require that.

WDYT @aikrahguzar?

bdarcus:
> oops, sorry for the delay. I did find a CSL style for this -- it's apa-cv.csl, and I can't quite remember where I found it but it wasn't hard. I will make a Chicago version of it too, when I have a sec. If it seems useful to other people, I can put both up in github somewhere once I dig out a bit.

It was occurred to me it makes at this point to encourage more distributed style development and hosting. That was the original intention.

[...]

PR #307, opened 2021-09-26
Towards removing bibtex-completion dependency + misc changes

aikrahguzar:


bibtex-actions.el
```diff
@@ -169,6 +169,32 @@ and nil means no action."
...
+insert-keys: the corresponding function should insert the list of keys given
+to as the argument at point in the buffer.
+
+keys-at-point: the corresponding function should return the list of keys at
+point."
+  :group 'bibtex-actions
+  :type '(alist :key-type (repeat string :tag "Major modes")
+                :value-type (set (cons (const local-bib-files) function)
+                                 (cons (const insert-keys) function)
+                                 (cons (const keys-at-pont function)))))
```
aikrahguzar, line 196: The idea here is to gather the functionality needed to support a major mode. Are there more places we need mode specific functionality? I think `insert-citation` is one.

bibtex-actions.el
```diff
@@ -169,6 +169,32 @@ and nil means no action."
   :group 'bibtex-actions
   :type 'function)
 
+(defcustom bibtex-actions-major-mode-functions
+  '(((latex-mode) .
```
aikrahguzar, line 173: Should this be a list of major-modes as here or a function so that something like `derived-mode-p` can be used? The later choice seems more flexible. So we can just say `(derived-mode-p 'latex-mode)` instead of having to enumerate the derived ones.

aikrahguzar:
Question: Does `bibtex-actions-insert-reference` work? It doesn't for me because I think it requires the cache from `bibtex-completion` which isn't generated anymore. I get `Warning (:warning): Bibtex-completion couldn't find entry with key "thekey".` but also `Wrong type argument: char-or-string-p, nil`

Also what is it supposed to do? I haven't used it and was trying to figure it out when I hit that warning and type error.

bibtex-actions.el
```diff
@@ -274,14 +300,29 @@ offering the selection candidates."
          ((string= extension (or "org" "md")) "Notes")
           (t "Library Files")))))
 
+(defun bibtex-actions-latex--local-bib-files ()
+  "Local bibliographic for latex retrieved using reftex"
+  (reftex-access-scan-info t)
```
aikrahguzar, line 305: This causes a rescan which is fast enough for me and has the advantage that if a bib file has been added or removed, that is detected.

bibtex-actions.el
```diff
@@ -585,8 +624,7 @@ FORMAT-STRING."
 
 (defun bibtex-actions-citation-key-at-point ()
```
aikrahguzar, line 625: Is this supposed to be the single key at point or else a list of contiguous keys in a citation command? I went for the latter in latex but both the org one and the previous `bibtex-completion` implementation would return a single key.

bdarcus:
> Question: Does `bibtex-actions-insert-reference` work? 

I haven't tested that command in a long time.

> Also what is it supposed to do?

Format a bibliographic entry.

We could just remove it, and also consider #275.

bibtex-actions.el
bdarcus: Check the docstring.

aikrahguzar: > Check the docstring.

Thanks, fixed

aikrahguzar:
> Format a bibliographic entry.

It is only for bibtex or can it also do the in file bibliography which is possible in latex.
 
> We could just remove it, and also consider #275.

I am fine with removing it. But what about #275? 

But I do need to deal with a package management questions. At present I need to add

```elisp
(require 'org-cite)
(require 'reftex)
```
Which seem sub-optimal. The other option is to leave `bibtex-actions-major-mode-functions` `nil` by default and then `(add-to-list bibtex-actions-major-mode-functions ...)` in `oc-bibtex-actions` and a new `bibtex-actions-latex`. This has the better lazy loading behavior but makes configuration more complicated.

bdarcus:
> It is only for bibtex or can it also do the in file bibliography which is possible in latex.

It just outputs the entry plain text, for any text mode.

It doesn't work very well in my experience, and ...

> I am fine with removing it. But what about #275?

.. sorry; #245.

The idea would be to have citeproc-el handle the formatting.

bdarcus:
> But I do need to deal with a package management questions. At present I need to add
> 
> ```emacs-lisp
> (require 'org-cite)
> (require 'reftex)
> ```
> 
> Which seem sub-optimal. The other option is to leave `bibtex-actions-major-mode-functions` `nil` by default and then `(add-to-list bibtex-actions-major-mode-functions ...)` in `oc-bibtex-actions` and a new `bibtex-actions-latex`. This has the better lazy loading behavior but makes configuration more complicated.

I am learning this too, and so not certain.

But two things to consider:

1. `with-eval-after-load` (see [how consult uses this](https://github.com/minad/consult/blob/6f07e1bdb2023871855b44f44153ad87af15a9ee/consult.el#L4458))
2. `fbounp` and `boundp`

bibtex-actions.el
```diff
@@ -172,6 +172,7 @@ and nil means no action."
 (defcustom bibtex-actions-major-mode-functions
   '(((latex-mode) .
      ((local-bib-files . bibtex-actions-latex--local-bib-files)
+      (insert-keys . bibtex-actions-latex--insert-keys)
```
aikrahguzar, line 178: What should `insert-keys` be for `org`? Bibtex-completions uses the comma separated list for everything, but does the org-cite reqiuires `@` in front of the key?

bdarcus: Both pandoc and org-cite use the `@key` syntax, with semi-colon delimiter. 

See [bibtex-completion-format-citation-pandoc-citeproc](https://github.com/tmalsburg/helm-bibtex/blob/b85662081de98077f13f1a9fac03764702325d28/bibtex-completion.el#L1024).

But that has had me wonder whether insert-keys is a useful command; particularly in org-cite (where `org-cite-insert` works differently depending on where point it).

Is it useful in latex, given insert-citation?

aikrahguzar:
> I am learning this too, and so not certain.

I am leaning towards `add-to-list` option at the moment. We already do `(require 'oc)` in `oc-bibtex-actions` and looking up local bibliography and keys at point both require `org-cite` anyway so a similar but much smaller file for latex makes sense too.

My question would be: to what extent is `bibtex-actions` usable in org right now without `org-cite`? Is it a goal to keep it usable without `org-cite`? Maybe `bibtex-actions-insert` works even without `org-cite`?  
 
> But (in general) two things to consider:
> 
>     1. `with-eval-after-load` (see [how consult uses this](https://github.com/minad/consult/blob/6f07e1bdb2023871855b44f44153ad87af15a9ee/consult.el#L4458))
> 
>     2. `fbounp` and `boundp`

aikrahguzar:
> My question would be: to what extent is `bibtex-actions` usable in org right now without `org-cite`? Is it a goal to keep it usable without `org-cite`? Maybe `bibtex-actions-insert` works even without `org-cite`?

Actually even if we want some commands to be usable without `org-cite` we use the versions without `org-cite` in `bibtex-actions` and do `add-to-list` in `oc-bibtex-actions` to shadow them.

bdarcus:
> to what extent is `bibtex-actions` usable in org right now without `org-cite`? 

I think it's pretty usable, as the user has a choice of different citation options: pandoc, org-ref, etc.

But you could arguably say that's a problem too.

> Is it a goal to keep it usable without `org-cite`? 

IDK ATM.

> Maybe `bibtex-actions-insert` works even without `org-cite`?

Maybe. Need to think on this more. 

Question beyond whether it "works", is how.

I _think_ ideally:

1. if you're using org-cite, the bibtex-actions-insert commands (at least for citations and keys) shouldn't load (see #278).
2. we allow people to use insert-citation with pandoc or org-ref, at least for now

WDYT?

bibtex-actions.el
aikrahguzar: > Both pandoc and org-cite use the `@key` syntax, with semi-colon delimiter.

I added that. 

> 
> But that has had me wonder whether insert-keys is a useful command; particularly in org-cite (where `org-cite-insert` works differently depending on where point it).
> 
> Is it useful in latex, given insert-citation?

It is useful for me because `insert-citation` gives a menu with cite commands and I only use `\cite` so I have defined a command that use `\cite` using `insert-keys`.

I want to change `insert-citation` so that it can be passed the citation argument as an optional argument so that it gives you the menu only if it is not passed, in that case it will become less useful.

aikrahguzar:
> I _think_ ideally:
> 
>     1. if you're using org-cite, the bibtex-actions-insert commands (at least for citations and keys) shouldn't load (see [bibtex-actions-insert-citation vs org-cite-insert #278](https://github.com/bdarcus/bibtex-actions/issues/278)).
> 
>     2. we allow people to use insert-citation with pandoc or org-ref, at least for now
> 
> 
> WDYT?

I am not familiar enough with `org` to comment on the rest but

2) makes me think that the best approach is to do as in https://github.com/bdarcus/bibtex-actions/pull/307#issuecomment-927306608 

For 1) I don't see an easy way except to move the definitions of commands themselves to `oc-bibtex-actions` and the new `bibtex-actions-latex` which might not be a bad idea. But what we can pretty easily is to print a warning to use `org-cite` functions directly when these commands are invoked with `org-cite` present.

aikrahguzar:
> > I _think_ ideally:
> > ```
> > 1. if you're using org-cite, the bibtex-actions-insert commands (at least for citations and keys) shouldn't load (see [bibtex-actions-insert-citation vs org-cite-insert #278](https://github.com/bdarcus/bibtex-actions/issues/278)).
> > 
> > 2. we allow people to use insert-citation with pandoc or org-ref, at least for now
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > WDYT?
> 
> I am not familiar enough with `org` to comment on the rest but
> 
>     1. makes me think that the best approach is to do as in [Towards removing bibtex-completion dependency + misc changes #307 (comment)](https://github.com/bdarcus/bibtex-actions/pull/307#issuecomment-927306608)
> 
> 
> For 1) I don't see an easy way except to move the definitions of commands themselves to `oc-bibtex-actions` and the new `bibtex-actions-latex` which might not be a bad idea. But what we can pretty easily is to print a warning to use `org-cite` functions directly when these commands are invoked with `org-cite` present.

Anyhow I think the picture will be clearer if I manage to deconstruct `bibtex-completion-insert-citation` which is pretty hairy at first glance.

bibtex-actions.el
bdarcus: > It is useful for me because `insert-citation` gives a menu with cite commands and I only use `\cite` so I have defined a command that use `\cite` using `insert-keys`.
> 
> I want to change `insert-citation` so that it can be passed the citation argument as an optional argument so that it gives you the menu only if it is not passed, in that case it will become less useful.

Yeah, this is what I meant about `org-cite-insert`.

I'm not sure it's the 100% best UI ATM, but it's an improvement. So:

1. when you run it in bare text, it just inserts the citation with keys.
2. when you run it when point is on "cite", it gives you the style UI
3. when you run it when point is on a citekey, it prompts to replace the key

Not sure how much of that is possible in LaTeX, given the more lightweight syntax, but might be worth thinking about.

bdarcus:
PS - seems you're packing a lot into one PR. Can be easier to manage if more focused.

For example, cl-case -> pcase could be on its own.

bibtex-actions-latex.el
```diff
@@ -0,0 +1,118 @@
...
+command is assumed to have a single argument into which keys are inserted."
+  (if (bibtex-actions-latex-is-a-cite-command (TeX-current-macro))
+      (progn (skip-chars-forward "^,}")
+             (unless (equal ?} (preceding-char)) (insert ", ")))
+    (let ((macro (or command
+                     (completing-read "Cite command: "
+                                    (seq-mapcat #'car bibtex-actions-latex-cite-commands)
+                                    nil nil nil
+                                    'bibtex-actions-latex-cite-command-history nil nil))))
+      (TeX-parse-macro macro
```
aikrahguzar, line 106: This works very well except I am getting a `cite: info not available, use ‘C-c &’ to get it.` message I think sue to `TeX-parse-macro`. I need to figure out why and how to suppress it.

bibtex-actions-latex.el
```diff
@@ -0,0 +1,118 @@
...
+;; commands provided by bibtex-actions.
+;;
+;;; Code:
+
+(require 'bibtex-actions)
+(require 'tex)
+(require 'reftex)
+
+(defcustom bibtex-actions-latex-cite-commands
+  '((("cite" "Cite" "parencite"
```
bdarcus, line 43: I don't really use LaTeX citations directly anymore, so take my questions/feedback with suitable grain-of-salt ...

I know this is how this is often implemented; as a flat list of commands.

But in practice, not all commands are the same; that their relevance depends on which system one is using; for example, natbib vs biblatex.

If you're using one, do you really want to have access to the long list of options compared to the other?

Also, I don't know if there's any value in cross-fertilizing ideas, because org-cite is more general than latex, but here's what the current org-cite style selection UI looks like by default, using Emacs 28 group-function.

image

aikrahguzar: > I don't really use LaTeX citations directly anymore, so take my questions/feedback with suitable grain-of-salt ...
> 
> I know this is how this is often implemented; as a flat list of commands.
> 
> But in practice, not all commands are the same; that their relevance depends on which system one is using; for example, natbib vs biblatex.
> 
> If you're using one, do you really want to have access to the long list of options compared to the other?
> 
> Also, I don't know if there's any value in cross-fertilizing ideas, because org-cite is more general than latex, but here's what the current org-cite style selection UI looks like by default, using Emacs 28 group-function.
> 
> image

I don't like the flat list of commands either but since I don't need an interface, I am not sure what it should be. Maybe it is worth asking people who use the multiple command what would they like the interface to be on doom discord (what would be good forum?). 

I am actually not familiar with biblatex or natbib either but if there is an easy way to cut down the number of options that have to presented depending on the file, that would be good. But as above some input has to be come from who use it.

bdarcus: Could just not worry about it now, and iterate once we have a starting point?

aikrahguzar: > Could just not worry about it now, and iterate once we have a starting point?

Sounds good though responses can take time so asking earlyish would be better.

This is going to take at least a couple of weekends though to get into some decent shape.

bibtex-actions-latex.el
aikrahguzar: Seems to have resolved itself.

aikrahguzar:
What is left now is `insert-entry`, `insert-reference`, `add-PDF-attachment` and `add-pdf-to-library`. For all of these I am unsure how to go about them.

bibtex-actions.el
```diff
@@ -693,8 +694,10 @@ With prefix, rebuild the cache before offering candidates."
 With prefix, rebuild the cache before offering candidates."
   (interactive (list (bibtex-actions-select-refs
                       :rebuild-cache current-prefix-arg)))
- (bibtex-completion-show-entry
-  (bibtex-actions--extract-keys keys-entries)))
+ (let ((bibtex-files (seq-concatenate 'list
+                                      bibtex-actions-bibliography
+                                      (bibtex-actions--local-files-to-cache))))
+  (mapc (lambda (key) (bibtex-find-entry key t nil t)) (bibtex-actions--extract-keys keys-entries))))
```
aikrahguzar, line 700: As an upside this works for local bib files while bibtex-completion based one it repalced didn't.

As a possible down side (I am not sure) that might have unintentionally worked for csl-json files since it is regex based, while I don't think this will.

bdarcus: I doubt it will either.

BTW: you may have noticed I renamed the project. So this will be one large merge conflict now :-)

PR #309, opened 2021-09-27
add brief documentation about function forms

titaniumbones:
It took me a while to figure this out on my own, so I thought it would
be helpful to have it documented somewhere.  README might be more
appropriate.

### DIFF
--- a/CONTRIBUTING.org
+++ b/CONTRIBUTING.org
@@ -9,10 +9,51 @@ I imagine it will stay pretty simple and stable, and that enhancements will focu
 Otherwise, I expect most innovations to focus on configuration details laid out on the wiki.
 You should feel free to add your ideas there.
 
-You can of course write your own functions that rely on =bibtex-actions-read=, but if you want to share it more widely, you should consider submitting it for inclusion in [[https://github.com/tmalsburg/helm-bibtex][bibtex-completion]]. itself.
-
 If you do submit a PR, details:
 
 - Follow the [[https://github.com/bbatsov/emacs-lisp-style-guide][elisp style guide]].
 - If the CI flags any errors or warnings, which it will report on the "files" tab, please fix them.
 - For working on lists and such, we primarily use the =seq= functions, and would like to remove the dependence on =s= and =bibtex-completion= in general.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.org
```diff
@@ -9,10 +9,51 @@ I imagine it will stay pretty simple and stable, and that enhancements will focu
...
 If you do submit a PR, details:
 
 - Follow the [[https://github.com/bbatsov/emacs-lisp-style-guide][elisp style guide]].
 - If the CI flags any errors or warnings, which it will report on the "files" tab, please fix them.
 - For working on lists and such, we primarily use the =seq= functions, and would like to remove the dependence on =s= and =bibtex-completion= in general.
+
+
+** Extending bibtex-actions
+
+Most user-accessible bibtex-actions  functions take an argument ~key-entry~ or ~keys-entries~. These expect, respectively, a cons cell of a citation key (a string like "SmithWritingHistory1987") and the corresponding bibliography entry for that citation, or a list of such cons cells.  If you wish to extend bibtex-actions at the user-action level, perhaps by adding a function to one of the embark keymaps, you will find it easiest to reproduce this pattern.  If you need to build the cons cells manually, this can be accomplished via ~bibtex-actions--get-entry~.  So, for example, to insert the annotations from a pdf into a buffer, the following pair of functions might be used:
```
bdarcus, line 21:
```suggestion
Most user-accessible bibtex-actions functions take an argument ~key-entry~ or ~keys-entries~.
These expect, respectively, a cons cell of a citation key (a string like "SmithWritingHistory1987") and the corresponding bibliography entry for that citation, or a list of such cons cells.
If you wish to extend bibtex-actions at the user-action level, perhaps by adding a function to one of the embark keymaps, you will find it easiest to reproduce this pattern.
If you need to build the cons cells manually, this can be accomplished via ~bibtex-actions--get-entry~.
So, for example, to insert the annotations from a pdf into a buffer, the following pair of functions might be used:
```

Just changing to one-sentence-per-line.

bdarcus:
Thanks.

PR #312, opened 2021-09-29
If note template nil, don't insert content

### LINKED ISSUE #311: Note template is not empty when using `oc-bibtex-actions`
**Describe the bug**
I want to have an empty note template. I am not able to achieve this using `oc-bibtex-actions`.

**To Reproduce**
- load `oc-bibtex-actions`
- set `(add-to-list 'bibtex-actions-templates '(note . ""))`
- create a new note, the file contents are not empty but
```
:PROPERTIES:
:ID:  9c59f72b-d376-4862-97ca-f348752ead25
:END:

```
(there is also a newline after the property drawer).

**Expected behavior**
The note is empty.

bdarcus:
Fix #311

@dschrempf - can you give this a try and see what you think? If you have the note template set to `nil`, it opens a blank document.

I'm not sure it's the best general solution, but it's simple enough.

### DIFF
--- a/bibtex-actions-file.el
+++ b/bibtex-actions-file.el
@@ -181,16 +181,18 @@ Example: ':/path/to/test.pdf:PDF'."
             (file-exists (file-exists-p file)))
       (funcall bibtex-actions-file-open-function file)
     (let* ((uuid (org-id-new))
+           (template (bibtex-actions-get-template 'note))
            (note-meta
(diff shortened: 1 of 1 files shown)


bdarcus:
Actually, I'm going to merge this. Can possibly tackle the other pieces separately.

dschrempf:
Thank you this works perfectly!

PR #314, opened 2021-10-01
Add activate processor

bdarcus:
This adds a simple configurable activate processor, that includes a 
small function to attach a keymap to org citations.

------------------------------

This uses the oc-basic activate processor, but adds a keymap function, adapted from org-ref-cite, to enable using keybindings to edit citations at point.

But it's also configurable, so if you want to use this for basic fontification and preview instead, you can do that by setting `oc-bibtex-actions-activation-functions` accordingly.

https://github.com/andras-simonyi/org-cite-csl-activate

Here's the current mapping:

```console
key             binding
---             -------

C-d             oc-bibtex-actions-delete-citation
C-k             oc-bibtex-actions-kill-citation
S-<left>        oc-bibtex-actions-shift-reference-left
S-<right>       oc-bibtex-actions-shift-reference-right
```

The rest would be handled with `embark-act`, and `org-cite-insert`.

TODO: make the keymap a defcustom.

### DIFF
--- a/README.org
+++ b/README.org
@@ -116,7 +116,7 @@ In other supported modes, it will work the same as the embark option above.
         org-cite-global-bibliography my/bibs
         org-cite-insert-processor 'oc-bibtex-actions
         org-cite-follow-processor 'oc-bibtex-actions
-        org-cite-activate-processor 'basic))
+        org-cite-activate-processor 'oc-basic))
 
 ;; Use consult-completing-read for enhanced interface.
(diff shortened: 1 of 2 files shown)


bdarcus:
FYI, I got this working @andras-simonyi; it basically just wraps the oc-basic activate processor, and then adds a little function to attach the keymap.

I'd like see if this work well with your csl based processor.

PR #353, opened 2021-11-01
Autoload the call to org-cite-register-processor

roshanshariff:
Since all the required functions are now autoloaded (#342), we can also autoload the call to `org-cite-register-processor` (protected by `with-eval-after-load 'oc`). This will register the `citar` processor whenever org-cite is loaded, but will not load `citar-org` until:
* the `org-cite-{activate,follow,insert}-processor` variables have been set to `'citar`, and
* an Org-mode file is loaded and org-cite is activated, a citation is followed, or `org-cite-insert` is called.

The usage instructions are then simply "install citar" and "set the `org-cite-*-processor` variables to `'citar`". There is no need to arrange for `citar-org` to be loaded explicitly, which can pull in all of Org Mode if done too early, or can result in the org-cite processor not being registered if done too late.

### DIFF
--- a/citar-org.el
+++ b/citar-org.el
@@ -27,10 +27,15 @@
 ;;
 ;;; Commentary:
 ;;
-;;  This is a small package that intergrates citar and org-cite.  It
-;;  provides a simple org-cite processor with "follow" and "insert" capabilties.
+;;  This is a small package that integrates citar and org-cite.  It
+;;  provides a simple org-cite processor with "follow", "insert", and
+;;  "activate" capabilities.
 ;;
-;;  Simply load this file and it will configure them for 'org-cite.'
+;;  Simply load this file (or its generated autoloads) and it will
+;;  make the processor available to 'org-cite'.  To instruct org-cite
+;;  to use citar, set one or more of the customization variables
+;;  'org-cite-activate-processor', 'org-cite-follow-processor', and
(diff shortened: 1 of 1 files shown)


bdarcus:
Excellent; thank you!

bdarcus:
One follow-up: do you have any suggested amendments to the use-package example in the README as a consequence of this?

Is it enough to remove the `:demand` line?

roshanshariff:
In my opinion, the configuration for org-mode should be what is given in the "basic" and "embark" sections of the readme, and in addition:
```emacs-lisp
(use-package citar-org
  :no-require
  :custom
```
There are already bindings for `org-cite-insert` (`C-c C-x C-@`) and `org-open-at-point` (`C-c C-o`) in `org-mode-map`, so I don't think they need to be bound again (though perhaps there could be a comment directing users to create a shorter binding if they want).

Using `:no-require` means that use-package will not load the package at all, relying solely on autoloads. It just becomes a convenient place to put configuration.

I think the other use-package stanzas (for the "basic" and "embark" case can also be simplified) and there is a bug that needs fixing (which is fixed in @aikrahguzar's branch and was masked with the old instructions relying on bibtex-actions, which is why I hadn't reported it). I will submit a new pull request proposing these changes (and fixing the bug).

Issue #310, opened 2021-09-27
deprecate and remove insert-reference and add-pdf-to-library


bdarcus:
I've included I'm almost definitely removing the pdf commands (already marked obsolete and will be removed soon); not sure about the other two.

Now's the time for objections, and/or PRs, if you disagree.

----------

Another issue to pull out of #307. Feedback?

> What is left now is `insert-entry`, `insert-reference`, `add-PDF-attachment` and `add-pdf-to-library`. For all of these I am unsure how to go about them.

_Originally posted by @aikrahguzar in https://github.com/bdarcus/bibtex-actions/issues/307#issuecomment-928041463_

My thought is maybe we remove the last two, and for the second -> #245.

Not sure about the first.

But all of this needs user feedback, since I have no idea whether and how people are using these commands.

localauthor:
I've pieced together the following rewrite of citar-insert-reference using citeproc.

Caveat: I do not understand more than the smallest fraction of citeproc-el, if that, and I'm learning elisp as I go, so, grains of salt.

Thoughts?

```
(defun citar-insert-reference (keys-entries)
  (interactive (list (citar-select-refs
                      :rebuild-cache current-prefix-arg)))
  (let* ((citelist (mapcar (lambda (x) (cons (cons 'id (cdr (assoc "=key=" x))) nil)) keys-entries))
         (proc (citeproc-create "~/.csl/chicago-fullnote-bibliography.csl"
			        (citeproc-itemgetter-from-bibtex citar-bibliography)
			        (citeproc-locale-getter-from-dir org-cite-csl-locales-dir)
			        "en-US"))
         (cites (list (citeproc-citation-create :cites citelist
					        :mode 'nil
					        :capitalize-first nil
					        :suppress-affixes nil)))
         (rendered-citations (progn 
                               (citeproc-append-citations cites proc)
                               (citeproc-render-bib proc 'plain))))
    (insert (car rendered-citations))))
```

bdarcus:
Oh, cool.

I think this is probably the way to go, but there's one thing I'm a little leery of:

Citeproc is a pretty big library, and it doesn't seems wise to have it a hard dependency.

Will need to think on how to do this.

In any case, thanks!

bdarcus:
Notes on `citar-insert-bibtex`:

[...]

PR #355, opened 2021-11-01
Autoload keymaps required for embark and simplify readme

### LINKED ISSUE #340: docs: defer loading oc-bibtex-actions in config?
It take at least 0.5s+ (I have already require org first) to require oc-bibtex-actions on my side, which make opening the first org mode file is slow.
```emacs-lisp
(require 'org)
(benchmark-run (require 'oc-bibtex-actions))
```

roshanshariff:
Set up the Embark integration automatically when Citar is loaded, and also add it to the autoload file. This means that Embark and Citar can be loaded in any order, and neither will pull in the other. Citar will be loaded the first time its target finder is invoked by `embark-act`.

There seems to be some redundancy between the embark action maps for `citar` and `citar-org`. Currently, loading `citar-org` overrides the action map for `bib-reference` targets, from `citar-map` to `citar-buffer-map`. This is of course, not ideal, and perhaps both of these action maps should be merged.

This PR also includes the simplified use-package stanza for `citar-org` made possible by #353, allowing its loading to be deferred to the last possible moment (fixes #340).

### DIFF
--- a/README.org
+++ b/README.org
@@ -58,73 +58,57 @@ In addition, the following packages are strongly recommended for the best experi
 This is the minimal configuration, and will work with any completing-read compliant vertical completion UI, like Vertico, Selectrum, or the built-in icomplete-vertical, with actions available via =M-x= commands.
 
 #+BEGIN_SRC emacs-lisp
-(use-package bibtex-completion)
-
 (use-package citar
   :bind (("C-c b" . citar-insert-citation)
          :map minibuffer-local-map
          ("M-b" . citar-insert-preset))
-  :after (bibtex-completion)
-  :config
-  (setq citar-bibliography '("~/bib/references.bib"))
+  :custom
+  (citar-bibliography '("~/bib/references.bib"))
 #+END_SRC
 
 Since most of the command logic resides in bibtex-completion, that is where to look for different [[https://github.com/tmalsburg/helm-bibtex#basic-configuration-recommended][configuration options]].
 
 *** Embark
 
-Highly recommended, this option adds embark, for contextual access to actions in the minibuffer and at-point.
+Citar will automatically integrate with Embark if it is installed, offering contextual access to actions in the
+minibuffer and at-point.  Some citation commands prompt for multiple citation keys, for which Consult offers an enhanced
+interface:
 
 #+BEGIN_SRC emacs-lisp
-(use-package citar
-  :bind (("C-c b" . citar-insert-citation)
-         :map minibuffer-local-map
-         ("M-b" . citar-insert-preset))
-  :after (embark bibtex-completion)
-  :config
-  ;; Make the 'citar' bindings and targets available to `embark'.
-  (add-to-list 'embark-target-finders 'citar-citation-key-at-point)
-  (add-to-list 'embark-keymap-alist '(bib-reference . citar-map))
-  (add-to-list 'embark-keymap-alist '(citation-key . citar-buffer-map))
-  (setq citar-bibliography '("~/bib/references.bib"))
-
 ;; use consult-completing-read for enhanced interface
 (advice-add #'completing-read-multiple :override #'consult-completing-read-multiple)
 #+END_SRC
 
-When using this option, these actions are generic, and work the same across org, markdown, and latex modes.
+When using Embark, the Citar actions are generic, and work the same across org, markdown, and latex modes.
 
 *** Org-Cite
 
 #+CAPTION: org-cite at-point integration with =embark-act=
 [[file:images/org-cite-embark-point.png]]
 
-If you use org-mode, this is the best option.
-It includes the embark functionality above, but customizes menus and user-experience for org-cite.
-In other supported modes, it will work the same as the embark option above.
+If you use Org-Mode, this is the best option.  Use the basic configuration above, and /in addition/, configure Org-Cite
+to use Citar:
 
 #+BEGIN_SRC emacs-lisp
 ;; Set bibliography paths so they are the same.
 (defvar my/bibs '("~/bib/references.bib"))
 
 (use-package citar-org
-  :bind (("C-c b" . org-cite-insert)
-         ("M-o" . org-open-at-point)
-         :map minibuffer-local-map
-         ("M-b" . citar-insert-preset))
-  :after (embark oc)
-  :config
-  (setq citar-bibliography my/bibs
-        org-cite-global-bibliography my/bibs
-        org-cite-insert-processor 'citar
-        org-cite-follow-processor 'citar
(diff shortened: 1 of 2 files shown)


roshanshariff:
One issue with this approach is that merely installing Citar will enable its integration with Embark, and cause it to be loaded through its target finder when `embark-act` is invoked. This is perhaps a little too magical. I will think more about this tomorrow.

bdarcus:
> I will think more about this tomorrow.

OK; I'll wait to hear back from you.

Earlier I required Embark, but @minad convinced me to use `with-eval-after-load` on that bit of code instead, which I just merged yesterday.

My main goal is for users to have flexibility here, but also that if they have Embark installed, it's as easy as possible for them to get it working correctly.

minad:
Adding the with-eval-after-load to the autoload is a good idea to allow loading the package in any order. However Roshan is right that installing the target finder autoload citar which may be undesired. There are these possibilities:

1. Do not add this with-eval-after-load block to citar, instead add a function `citar-setup-embark-target-finders` which must be called manually by the user. But this will not really improve the situation, it will only move the responsibility to the user.
2. Mark the entire target finder as autoloaded by wrapping it in a progn+autoload-pragma, such that the function definition itself will be copied to the autoload file. This makes only sense if the target finder is a small function since otherwise we bloat the autoload file. Furthermore it must be possible to separate the target finder from the rest of the code.
3. Move the target finder to its own file, continue to use the with-eval-after-load block. This way only the small target finder file will always be autoloaded without loading citar entirely, reducing the impact this way. Maybe that's feasible if the target finder can be separated from the rest of the code.

roshanshariff:
Excellent, thanks @minad for running down the possible options (as an aside, when we're happy with how things work for Citar, we should add some Embark documentation on best practices for third-party extensions based on this advice).

So, to summarize, here are the pros and cons of this PR (`with-eval-after-load 'embark` added to autoloads).
1. **Running `embark-act` pulls in `citar`.** With some changes, this may be acceptable. Currently Citar has two heavy dependencies: `org-capture` via `bibtex-completion` and `org-id` via `citar-file`. The first will go away soon, hopefully. The second can probably be refactored away: `citar-file` requires `org-id` only in one function which is only called when a customization option is set. With these changes, `citar` itself will be fairly lightweight, with the Org dependency being limited to `citar-org`.
2. **The user expectation that installing a package doesn't change any behaviour.** This is an unavoidable trade-off with automatically integrating with Embark. I'm starting to think that this is fine; there is precedent for package autoloads modifying `auto-mode-alist`, and one could argue that modifying `embark-target-finders` is not too different. It's unlikely that someone has Citar installed but really doesn't like its target finder being used by Embark, especially since target cycling means other targets are not blocked from use.

So I'm leaning towards going with this PR, but eventually removing Citar's dependency on Org. What do you think, @bdarcus, especially about the second point?

---

@bdarcus, I misspoke about the action keymaps. I meant to say that `citar` uses `citar-map` for `bib-reference` targets whereas `citar-org` uses `citar-org-map`. The former seems like a superset of the latter, but they're both meant to be used from the minibuffer and should be unified, since you can't have two action keymaps for the same target type. Removing the actions in #310 would mostly make them identical anyway.

Similarly, `citar-buffer-map` and `citar-org-buffer-map` are basically the same already, though they don't conflict since they're attached to different target types: `citation-key` and `oc-citation`, respectively.

bdarcus:
> So I'm leaning towards going with this PR, but eventually removing Citar's dependency on Org. What do you think, @bdarcus, especially about the second point?

Fine by me.

The only reason I use `org-id` is because I didn't see another way to generate a uuid, which is the purpose there. Surely there must be a way without the dependency?

minad:
Regarding uuids - a simple search yields https://nullprogram.com/blog/2010/05/11/. It is always questionable if one should rather add a dependency or add a hand made function, but for such small tasks I usually lean towards a hand made function to avoid unnecessary dependencies and leftpad problems.

roshanshariff:
> Shall I merge then?

Yes, please :)

I think it's okay to have the `org-id` dependency, but it can be require'd within the function when needed, rather than at the top-level of the file. I can create a PR for this when I have a moment.

minad:
@roshanshariff 

> (as an aside, when we're happy with how things work for Citar, we should add some Embark documentation on best practices for third-party extensions based on this advice).

I admit this was part of my motivation, when I started discussing on the bibtex-completion PR, which proposed adding completing-read/embark support. Of course I really like what @bdarcus is pulling off here - citar looks very polished and I am excited to see a citation solution which fits the more minimal paradigm.

bdarcus:
> I think it's okay to have the `org-id` dependency, but it can be require'd within the function when needed, rather than at the top-level of the file. I can create a PR for this when I have a moment.

Per the link @minad posted, might be simpler to just write a little internal function for this?

Either way is fine though.

minad:
> I think it's okay to have the org-id dependency, but it can be require'd within the function when needed, rather than at the top-level of the file. I can create a PR for this when I have a moment.

I'd recommend against this paradigm. I would only use it under very rare circumstances. It is better to define an `(autoload ...)`. However you may want to consider if requiring org is indeed appropriate. Maybe it is not if you only need a ten line function - see the blog post by skeeto.

roshanshariff:
> I'd recommend against this paradigm. I would only use it under very rare circumstances.

Yes, I agree, and perhaps a better option would be to find a way to move this functionality to `citar-org` which already depends on Org. This function is used for generating an Org ID to be used in the headers of org-mode files, so it seems okay to depend on `org-id` itself and not some other UUID generator.

minad:
@bdarcus I looked at skeeto's uuid generator. It does a few dubious things like calling the garbage collector. I propose this function instead.

~~~ elisp
(defun uuid-create ()
  "Return a newly generated UUID."
  (let ((s (md5 (format "%s%s%s%s%s%s%s%s%s%s"
                        (user-uid)
                        (emacs-pid)
                        (system-name)
                        (user-full-name)
                        user-mail-address
                        before-init-time
                        gc-elapsed
                        (float-time)
                        (random)
                        (recent-keys)))))
    (format "%s-%s-3%s-%s-%s"
            (substring s 0 8)
            (substring s 8 12)
            (substring s 13 16)
            (substring s 16 20)
            (substring s 20 32))))
~~~

minad:
@roshanshariff 

> This function is used for generating an Org ID to be used in the headers of org-mode files, so it seems okay to depend on org-id itself and not some other UUID generator.

I see, then it makes of course sense to move this to the org part and just use the org functionality :+1:

bdarcus:
I created #356 while we're looking at moving org-related function calls.

Issue #351, opened 2021-10-31
Duplicate entried while in org mode


jonathan-sahar-tau:
When I call ```bibtex-actions-open``` while in an org buffer I get all the entries duplicated. This doesn't happen in other buffers, so I suspect it has something to do with org-cite.
I couldn't quite understand from the documentation whether I'm supposed to have both blocks of configuration (the bibtex-actions one and the bibtex-actions-oc one) or just one of them.. below is my config:

```
(use-package bibtex-actions
  :bind (("C-c b" . org-cite-insert)
         ("M-o" . org-open-at-point)
         :map minibuffer-local-map
         ("M-b" . bibtex-actions-insert-preset))
  :after (embark oc)
  :config
  (setq bibtex-actions-bibliography my/bibs
        org-cite-global-bibliography my/bibs
        org-cite-insert-processor 'oc-bibtex-actions
        org-cite-follow-processor 'oc-bibtex-actions
        org-cite-activate-processor 'oc-bibtex-actions)
  (setq bibtex-actions-at-point-function 'embark-act)
  (add-to-list 'embark-target-finders 'bibtex-actions-citation-key-at-point)
  (add-to-list 'embark-keymap-alist '(bib-reference . bibtex-actions-map))
  (add-to-list 'embark-keymap-alist '(citation-key . bibtex-actions-buffer-map))
  (setq bibtex-actions-bibliography '("/home/jonathan/google_drive/.bibliography/motor-cognition.bib"))

  (setq bibtex-actions-file-note-org-include '(org-id org-roam-ref))
  ;; (setq bibtex-actions-file-open-note-function 'orb-bibtex-actions-edit-note)
;; use consult-completing-read for enhanced interface
  (setq bibtex-actions-templates '((main . "${author editor:30}     ${date year issued:4}     ${title:48}")
                                   (suffix . "${tags keywords keywords:*}   ${=key= id:15}    ${=type=:12}")
                                   (note . "#+title: Notes on ${author editor}, ${title}
* main points
* findings
* methods
* summary and short reference
* general notes
* see also (notes, tags/ other papers):
")))

  (setq bibtex-actions-symbols
        `((file . (,(all-the-icons-icon-for-file "foo.pdf" :face 'all-the-icons-dred) .
                ,(all-the-icons-icon-for-file "foo.pdf" :face 'bibtex-actions-icon-dim)))
        (note . (,(all-the-icons-icon-for-file "foo.txt") .
[...]

PR #357, opened 2021-11-02
add citar-latex, citar-markdown

bdarcus:
Another big step to removing the bibtex-completion dependency.

- add a new major-mode "adapter" design
- add citar-latex and citar-markdown based on above
- move around code and rename symbols accordingly
- rewrite citar-insert-keys, citar-insert-citation
- replace bibtex-completion-at-point with citar-keys-at-point

@aikrahguzar did most of this work.

------
TODO:

1. [X] docs
2. [X] deprecate "add"/"attach" pdf commands 
3. [X] move org-cite target finder to `citar-org`, adjust [code](https://github.com/bdarcus/citar/blob/62a6ca420fb39b6f64a3b54ab415c9dd67a404b0/citar.el#L672)
4. [ ] change `citar-major-mode-function` from using `major-mode` to `derived-mode-p` (but I'm not sure how to do that with this code ATM; may defer to this to a later PR)
5. [ ] test

-------
Given how much #307 has drifted from main, let's use this instead and GTM.

Aside from rebasing and resolving conflicts and fixing a few things, I also added a similar, simpler, `citar-markdown` for the pandoc syntax, though it doesn't work ATM. I did this in part to support markdown, but also to test the generality of the design. Since it doesn't work, there may be some issues to resolve there.

To me the remaining questions are:

1. how to handle the richer biblatex model? (probably defer worrying about it, though per below, the at-point stuff doesn't quite work here)
2. we should move some of the functions (like the org target finder), and align `citar-org` with these other two.
3. whether to tie to derived-mode rather than major-mode?
4. why the markdown one doesn't work, and how to fix it?
5. should we add a "notes-function" to the major-mode functions list, so we can put the org note function in `citar-org`?

Perhaps 4 is the only blocker? 

See also review comments below; 

cc @aikrahguzar 

@roshanshariff can you review this, as it has implications for the other details we were discussing in #355. See my comments below, for example.

### DIFF
--- a/citar-latex.el
+++ b/citar-latex.el
@@ -1,0 +1,128 @@
+;;; citar-latex.el --- Latex adapter for citar -*- lexical-binding: t; -*-
+;;
+;; Copyright (C) 2021 Bruce D'Arcus
+;;
+;; Author: Bruce D'Arcus <https://github.com/bdarcus>
+;; Maintainer: Bruce D'Arcus <https://github.com/bdarcus>
+;; Created: February 27, 2021
+;; License: GPL-3.0-or-later
+;; Version: 0.4
+;; Homepage: https://github.com/bdarcus/citar
+;; Package-Requires: ((emacs "26.3"))
+;;
+;; This file is not part of GNU Emacs.
+;;
+;; This program is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+;;
+;;; Commentary:
+;;
+;; A small package that provides the functions required to use citar
+;; with latex.
+;;
+;; Simply loading this file will enable manipulating the citations with
+;; commands provided by citar.
+;;
+;;; Code:
+
+(require 'citar)
+(require 'tex)
+(require 'reftex-parse)
+
+(defvar citar-major-mode-functions)
+
+(defcustom citar-latex-cite-commands
+  '((("cite" "Cite" "citet" "Citet" "parencite"
+      "Parencite" "footcite" "footcitetext" "textcite" "Textcite"
+      "smartcite" "Smartcite" "cite*" "parencite*" "autocite"
+      "Autocite" "autocite*" "Autocite*" "citeauthor" "Citeauthor"
+      "citeauthor*" "Citeauthor*" "citetitle" "citetitle*" "citeyear"
+      "citeyear*" "citedate" "citedate*" "citeurl" "fullcite"
+      "footfullcite" "notecite" "Notecite" "pnotecite" "Pnotecite"
+      "fnotecite") . (["Prenote"] ["Postnote"] t))
+    (("nocite" "supercite") . nil))
+  "Citation commands and their argument specs.
+
+The argument spec is the same as the args argument of
+`TeX-parse-macro'. When calling `citar-insert-citation' the keys
+will be inserted at the position where `TeX-parse-macro' leaves
+the point."
+  :group 'citar-latex
+  :type '(alist :key-type (repeat string)
+                :value-type sexp))
+
+(defcustom citar-latex-prompt-for-extra-arguments t
+  "Whether to prompt for additional arguments when inserting a citation."
+  :group 'citar-latex
+  :type 'boolean)
+
+(add-to-list 'citar-major-mode-functions
+             '((latex-mode) .
+               ((local-bib-files . citar-latex--local-bib-files)
+                (insert-keys . citar-latex--insert-keys)
+                (insert-citation . citar-latex--insert-citation)
+                (keys-at-point . citar-latex--keys-at-point))))
+
+(defun citar-latex--local-bib-files ()
+  "Local bibliographic for latex retrieved using reftex."
+  (reftex-access-scan-info t)
+  (ignore-errors (reftex-get-bibfile-list)))
+
+(defun citar-latex--keys-at-point ()
+  "Return a list of keys at point in a latex buffer."
+    (when (citar-latex-is-a-cite-command (TeX-current-macro))
+      (split-string (thing-at-point 'list t) "," t "[{} ]+")))
+
+(defun citar-latex--insert-keys (keys)
+  "Insert comma sperated KEYS in a latex buffer."
+  (insert (string-join keys ", ")))
+
+(defvar citar-latex-cite-command-history nil
+  "Variable for history of cite commands.")
+
+(defun citar-latex--insert-citation (keys &optional command)
+  "Insert a citation consisting of KEYS.
+
+If the command is inside a citation command keys are added to it. Otherwise
+a new command is started.
+
+If the optional COMMAND is provided use it, otherwise prompt for one.
+
+The availiable commands and how to provide them arguments are configured
+by `citar-latex-cite-commands'.
+
+If `citar-latex-prompt-for-extra-arguments' is `nil`, every
+command is assumed to have a single argument into which keys are
+inserted."
+  (when keys
+    (if (citar-latex-is-a-cite-command (TeX-current-macro))
+        (progn (skip-chars-forward "^,}")
+               (unless (equal ?} (preceding-char)) (insert ", ")))
+      (let ((macro (or command
+                       (completing-read "Cite command: "
+                                        (seq-mapcat #'car citar-latex-cite-commands)
+                                        nil nil nil
+                                        'citar-latex-cite-command-history nil nil))))
+        (TeX-parse-macro macro
+                         (when citar-latex-prompt-for-extra-arguments
+                           (cdr (citar-latex-is-a-cite-command macro))))))
+    (citar-latex--insert-keys keys)
+    (skip-chars-forward "^}") (forward-char 1)))
+
+(defun citar-latex-is-a-cite-command (command)
+  "Return element of `citar-latex-cite-commands` containing COMMAND."
+  (seq-find (lambda (x) (member command (car x)))
+                 citar-latex-cite-commands))
+
+(provide 'citar-latex)
+;;; citar-latex.el ends here
--- a/citar-latex.el
+++ b/citar-latex.el
@@ -66,13 +66,6 @@ the point."
   :group 'citar-latex
   :type 'boolean)
 
-(add-to-list 'citar-major-mode-functions
-             '((latex-mode) .
-               ((local-bib-files . citar-latex-local-bib-files)
-                (insert-keys . citar-latex-insert-keys)
-                (insert-citation . citar-latex-insert-citation)
-                (keys-at-point . citar-latex-keys-at-point))))
-
 ;;;###autoload
 (defun citar-latex-local-bib-files ()
   "Local bibliographic for latex retrieved using reftex."
@@ -120,7 +113,7 @@ inserted."
         (TeX-parse-macro macro
                          (when citar-latex-prompt-for-extra-arguments
                            (cdr (citar-latex-is-a-cite-command macro))))))
-    (citar-latex--insert-keys keys)
+    (citar-latex-insert-keys keys)
     (skip-chars-forward "^}") (forward-char 1)))
 
 (defun citar-latex-is-a-cite-command (command)
--- a/citar-latex.el
+++ b/citar-latex.el
@@ -68,29 +68,33 @@ the point."
 
 (add-to-list 'citar-major-mode-functions
              '((latex-mode) .
-               ((local-bib-files . citar-latex--local-bib-files)
-                (insert-keys . citar-latex--insert-keys)
-                (insert-citation . citar-latex--insert-citation)
-                (keys-at-point . citar-latex--keys-at-point))))
+               ((local-bib-files . citar-latex-local-bib-files)
+                (insert-keys . citar-latex-insert-keys)
+                (insert-citation . citar-latex-insert-citation)
+                (keys-at-point . citar-latex-keys-at-point))))
 
-(defun citar-latex--local-bib-files ()
+;;;###autoload
+(defun citar-latex-local-bib-files ()
   "Local bibliographic for latex retrieved using reftex."
   (reftex-access-scan-info t)
   (ignore-errors (reftex-get-bibfile-list)))
 
-(defun citar-latex--keys-at-point ()
+;;;###autoload
+(defun citar-latex-keys-at-point ()
   "Return a list of keys at point in a latex buffer."
     (when (citar-latex-is-a-cite-command (TeX-current-macro))
       (split-string (thing-at-point 'list t) "," t "[{} ]+")))
 
-(defun citar-latex--insert-keys (keys)
+;;;###autoload
+(defun citar-latex-insert-keys (keys)
   "Insert comma sperated KEYS in a latex buffer."
   (insert (string-join keys ", ")))
 
 (defvar citar-latex-cite-command-history nil
   "Variable for history of cite commands.")
 
-(defun citar-latex--insert-citation (keys &optional command)
+;;;###autoload
+(defun citar-latex-insert-citation (keys &optional command)
   "Insert a citation consisting of KEYS.
 
 If the command is inside a citation command keys are added to it. Otherwise
--- a/citar-markdown.el
+++ b/citar-markdown.el
@@ -1,0 +1,81 @@
+;;; citar-markdown.el --- Markdown adapter for citar -*- lexical-binding: t; -*-
+;;
+;; Copyright (C) 2021 Bruce D'Arcus
+;;
+;; Author: Bruce D'Arcus <https://github.com/bdarcus>
+;; Maintainer: Bruce D'Arcus <https://github.com/bdarcus>
+;; Created: February 27, 2021
+;; License: GPL-3.0-or-later
+;; Version: 0.4
+;; Homepage: https://github.com/bdarcus/citar
+;; Package-Requires: ((emacs "26.3"))
+;;
+;; This file is not part of GNU Emacs.
+;;
+;; This program is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+;;
+;;; Commentary:
+;;
+;; A small package that provides the functions required to use citar
+;; with markdown.
+;;
+;; Simply loading this file will enable manipulating the citations with
+;; commands provided by citar.
+;;
+;;; Code:
+
+(require 'citar)
+
+(defvar citar-major-mode-functions)
+
+(defcustom citar-markdown-prompt-for-extra-arguments t
+  "Whether to prompt for additional arguments when inserting a citation."
+  ;; REVIEW this maybe shouldn't be mode specific?
+  :group 'citar-markdown
+  :type 'boolean)
+
+(add-to-list 'citar-major-mode-functions
+             '((markdown-mode) .
+               ((local-bib-files . citar-markdown--local-bib-files)
+                (insert-keys . citar-markdown--insert-keys)
+                (insert-citation . citar-markdown--insert-citation)
+                (keys-at-point . citar-markdown--keys-at-point))))
+
+(defun citar-markdown--local-bib-files ()
+  "Return local bib files for markdown."
+  nil)
+
+(defun citar-markdown--keys-at-point ()
+  "Return a list of keys at point in a markdown buffer."
+  (error "TODO"))
+
+(defun citar-markdown--insert-keys (keys)
+  "Insert comma sperated KEYS in a markdown buffer."
+  (insert (mapconcat (lambda (k) (concat "@" k)) keys "; ")))
+
+(defun citar-markdown--insert-citation (keys)
+  "Insert a pandoc-style citation consisting of KEYS."
+  (let* ((prenote  (if citar-markdown-prompt-for-extra-arguments
+                       (read-from-minibuffer "Prenote: ")))
+         (postnote (if citar-markdown-prompt-for-extra-arguments
+                       (read-from-minibuffer "Postnote: ")))
+         (prenote  (if (string= "" prenote)  "" (concat prenote  " ")))
+         (postnote (if (string= "" postnote) "" (concat ", " postnote))))
+    (insert (format "[%s%s%s]"
+                    prenote
+                    (mapconcat (lambda (k) (concat "@" k)) keys "; ")
+                    postnote))))
+
+(provide 'citar-markdown)
+;;; citar-markdown.el ends here
--- a/citar-markdown.el
+++ b/citar-markdown.el
@@ -47,24 +47,16 @@
 
 (add-to-list 'citar-major-mode-functions
              '((markdown-mode) .
-               ((local-bib-files . citar-markdown--local-bib-files)
-                (insert-keys . citar-markdown--insert-keys)
-                (insert-citation . citar-markdown--insert-citation)
-                (keys-at-point . citar-markdown--keys-at-point))))
+               ((insert-keys . citar-markdown-insert-keys)
+                (insert-citation . citar-markdown-insert-citation))))
 
-(defun citar-markdown--local-bib-files ()
-  "Return local bib files for markdown."
-  nil)
-
-(defun citar-markdown--keys-at-point ()
-  "Return a list of keys at point in a markdown buffer."
-  (error "TODO"))
-
-(defun citar-markdown--insert-keys (keys)
+;;;###autoload
+(defun citar-markdown-insert-keys (keys)
   "Insert comma sperated KEYS in a markdown buffer."
   (insert (mapconcat (lambda (k) (concat "@" k)) keys "; ")))
 
-(defun citar-markdown--insert-citation (keys)
+;;;###autoload
+(defun citar-markdown-insert-citation (keys)
   "Insert a pandoc-style citation consisting of KEYS."
   (let* ((prenote  (if citar-markdown-prompt-for-extra-arguments
                        (read-from-minibuffer "Prenote: ")))
--- a/citar-markdown.el
+++ b/citar-markdown.el
@@ -45,11 +45,6 @@
   :group 'citar-markdown
   :type 'boolean)
 
-(add-to-list 'citar-major-mode-functions
-             '((markdown-mode) .
-               ((insert-keys . citar-markdown-insert-keys)
-                (insert-citation . citar-markdown-insert-citation))))
-
 ;;;###autoload
 (defun citar-markdown-insert-keys (keys)
   "Insert comma sperated KEYS in a markdown buffer."
--- a/citar.el
+++ b/citar.el
@@ -1,4 +1,4 @@
-;;; bibtex-actions.el --- Bibliographic commands based on completing-read -*- lexical-binding: t; -*-
+;;; citar.el --- Bibliographic commands based on completing-read -*- lexical-binding: t; -*-
 ;;
 ;; Copyright (C) 2021 Bruce D'Arcus
 ;;
@@ -7,7 +7,7 @@
 ;; Created: February 27, 2021
 ;; License: GPL-3.0-or-later
 ;; Version: 0.4
-;; Homepage: https://github.com/bdarcus/bibtex-actions
+;; Homepage: https://github.com/bdarcus/citar
 ;; Package-Requires: ((emacs "26.3") (bibtex-completion "1.0") (parsebib "3.0"))
 ;;
 ;; This file is not part of GNU Emacs.
@@ -44,6 +44,8 @@
 (require 'bibtex-completion)
 (require 'parsebib)
 (require 's)
+;; Not ideal, find a better FIX
+(require 'oc)
 
 (declare-function org-element-context "org-element")
 (declare-function org-element-property "org-element")
@@ -53,6 +55,10 @@
 (declare-function org-cite-make-insert-processor "org-cite")
 (declare-function org-cite-basic--complete-style "org-cite")
 (declare-function embark-act "ext:embark")
+
+(declare-function reftex-access-scan-info "reftex")
+(declare-function reftex-get-bibfile-list "reftex")
+(declare-function TeX-current-macro "tex")
 
 ;;; Declare variables for byte compiler
 
@@ -78,21 +84,18 @@
   "Face used to highlight content in `citar' candidates."
   :group 'citar)
 
-(defcustom citar-bibliography bibtex-completion-bibliography
+(defcustom citar-bibliography nil
   "A list of bibliography files."
-  ;; The bibtex-completion default is likely to be removed in the future.
   :group 'citar
   :type '(repeat file))
 
-(defcustom citar-library-paths bibtex-completion-library-path
+(defcustom citar-library-paths nil
   "A list of files paths for related PDFs, etc."
-  ;; The bibtex-completion default is likely to be removed in the future.
   :group 'citar
   :type '(repeat path))
 
-(defcustom citar-notes-paths bibtex-completion-notes-path
+(defcustom citar-notes-paths nil
   "A list of file paths for bibliographic notes."
-  ;; The bibtex-completion default is likely to be removed in the future.
   :group 'citar
   :type '(repeat path))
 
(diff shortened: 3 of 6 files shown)


citar.el
```diff
@@ -44,6 +44,8 @@
 (require 'bibtex-completion)
 (require 'parsebib)
 (require 's)
+;; Not ideal, find a better FIX
+(require 'oc)
```
bdarcus, line 48: Maybe we should move the oc stuff to `citar-org`?

citar.el
```diff
@@ -272,15 +306,31 @@ offering the selection candidates."
...
+(defun citar-latex--local-bib-files ()
+  "Local bibliographic for latex retrieved using reftex."
+  (reftex-access-scan-info t)
+  (ignore-errors (reftex-get-bibfile-list)))
+
+(defun citar-latex--keys-at-point ()
+  "Return a list of keys at point in latex buffers."
+  (let ((macro (TeX-current-macro)))
+    (when (string-match-p "cite" macro)
+      (split-string (thing-at-point 'list t) "," t "[{} ]+"))))
```
bdarcus, line 322: Should this go in `citar-latex`?

citar.el
```diff
@@ -586,6 +636,10 @@ FORMAT-STRING."
       ('citation
        (org-cite-get-references elt t)))))
 
+(defun bibtex-actions--insert-keys-org-cite (keys)
+  "Insert KEYS in org-cite format"
+  (string-join (seq-map (lambda (key) (concat "@" key)) keys) ":"))
```
bdarcus, line 641: Move to `citar-org`?

citar-latex.el
```diff
@@ -0,0 +1,127 @@
...
+  (reftex-locate-bibliography-files (TeX-master-directory)))
+
+(defun citar-latex--keys-at-point ()
+  "Return a list of keys at point in a latex buffer."
+    (when (citar-latex-is-a-cite-command (TeX-current-macro))
+      (split-string (thing-at-point 'list t) "," t "[{} ]+")))
+
+(defun citar-latex--insert-keys (keys)
+  "Insert comma sperated KEYS in a latex buffer."
+  (insert (string-join keys ", ")))
```
bdarcus, line 87: I don't believe this works for biblatex, which has this annoying (from our POV) distinction between two types of citations. Examples:

```latex
Test 1: \autocite{low2001,mcneill2011}
Test 2: \autocites{low2001}[see][p. 12]{mcneill2011}
```

Perhaps we just follow bibtex-completion for initial merge and only support for the first variant?

I still think we should have a separate command list for natbib and biblatex, and rely on `reftex-using-biblatex-p` to know which to use?

citar.el
```diff
@@ -586,6 +636,10 @@ FORMAT-STRING."
       ('citation
        (org-cite-get-references elt t)))))
 
+(defun citar--insert-keys-org-cite (keys)
+  "Insert KEYS in org-cite format"
+  (string-join (seq-map (lambda (key) (concat "@" key)) keys) ":"))
+
```
bdarcus, line 642: Move to `citar-org`?

citar.el
```diff
@@ -165,6 +171,34 @@ and nil means no action."
...
+     ((local-bib-files . org-cite-list-bibliography-files)
+      (insert-keys . citar--insert-keys-org-cite)
+      (keys-at-point . citar-get-key-org-cite))))
+  "The variable determining the major mode specifc functionality.
+It is alist with keys being a list of major modes. The value is an alist
+with values being functions to be used for these modes while the keys
+are symbols used to lookup them up. The keys are
+
+local-bib-files: the corresponding functions should return the list of
+local bibliography files.
```
bdarcus, line 188: This cannot now be nil. What if a mode (like markdown) doesn't support local bibs?

citar-latex.el
```diff
@@ -0,0 +1,127 @@
...
+(defcustom citar-latex-prompt-for-extra-arguments t
+  "Whether to prompt for additional arguments when inserting a citation."
+  :group 'citar-latex
+  :type 'boolean)
+
+(add-to-list 'citar-major-mode-functions
+             '((latex-mode) .
+               ((local-bib-files . citar-latex--local-bib-files)
+                (insert-keys . citar-latex--insert-keys)
+                (insert-citations . citar-latex--insert-citations)
```
bdarcus, line 73:
```suggestion
                (insert-citations . citar-latex--insert-citation)
```

citar-latex.el
```diff
@@ -0,0 +1,127 @@
...
+      (split-string (thing-at-point 'list t) "," t "[{} ]+")))
+
+(defun citar-latex--insert-keys (keys)
+  "Insert comma sperated KEYS in a latex buffer."
+  (insert (string-join keys ", ")))
+
+(defvar citar-latex-cite-command-history nil
+  "Variable for history of cite commands.")
+
+(defun citar-latex--insert-citations (keys &optional command)
```
bdarcus, line 92:
```suggestion
(defun citar-latex--insert-citation (keys &optional command)
```

Should be singular.

roshanshariff:
I would like to propose an alternative, more flexible, approach to having major-mode-specific functionality (instead of the `citar-major-mode-functions` alist). For each `OPERATION`, (currently `local-bib-files`, `insert-keys`, `insert-citation`, and `keys-at-point`, we define a hook variable in `citar.el`:
```emacs-lisp
(defvar citar-OPERATION-functions nil)

;; arg1 arg2 are as appropriate for the operation
(defun citar--OPERATION (arg1 arg2)
  (run-hook-with-args-until-success 'citar-OPERATION-function arg1 arg2))
```
Then, in the various citar backends, you would just do something like
```emacs-lisp
(defun citar-org--keys-at-point ()
  (when (derived-mode-p 'org-mode)
    ...))

(add-hook 'citar-keys-at-point-functions #'citar-org--keys-at-point)
```

The nice thing about this approach is that if someone wants to override one of these operations in a buffer, they can just add their own function:
```emacs-lisp
(add-hook `citar-local-bib-files-functions #'my-local-bib-finder depth 'local)
```
where `'local` makes their hook buffer-local (if they want) and the depth parameter controls its priority.

This also doesn't impose any choice of how the functions are selected (major mode, derived mode, etc.) and seems more "emacsy". What do you think?

---

> should we add a "notes-function" to the major-mode functions list, so we can put the org note function in citar-org?

I think the notes function is more of a high level citar function than a mode-specific operation, so it seems fine to have it be a defcustom as it is now. If somebody wants to override it in a particular buffer, they can always use `setq-local`. We can provide an implementation of that function in `'citar-org` and ask people to set the defcustom to it if they want.

bdarcus:
Yeah, as I was playing with the notes stuff, I realized that was a bad idea. I moved the defcustom to the main file.

bdarcus:
> This also doesn't impose any choice of how the functions are selected (major mode, derived mode, etc.) and seems more "emacsy". What do you think?

This project is largely me learning emacs/elisp as I go. :-)

Hooks are one of those details I'm not very knowledgeable about. But it seems to make sense to me.

roshanshariff:
> This project is largely me learning emacs/elisp as I go. :-)

I'm in exactly the same boat and have only really learned about these things in the past few months, so my suggestions should be taken with a grain of salt :)

> Would nil really be the default?

Ideally, yes, and all their contents should be added to them using add-hook (which can be run even before the hook is defined; it'll just define the hook variable if it's not already). The reason is that defvar only sets the value if it's not already defined. If somebody used `(add-hook ...)` in their init file, for instance, before citar was loaded, then the hook would already be defined and the defvar would do nothing.

> And does your citar--OPERATION function become the interactive function; for example citar-insert-citation?

I *think* the best practice would be to have a separate interactive function, and to think of the `citar--OPERATION` functions as purely internal. In some cases, the interactive function would do nothing more than call the internal function, but people could add advice, override it, etc. without breaking citar's internal API.

bdarcus:
So to start, like this?

https://github.com/bdarcus/citar/pull/357/commits/59c03d73a78ee67714368049a2e48689387966b9

roshanshariff:
> So to start, like this?

Yes, looks good to me! (I think the function `citar--at-point` should be called `citar--keys-at-point` though?)

bdarcus:
Right; done.

I'm a little unclear on next steps. Is the idea if a hook isn't configured, I add an error message to tell the user to configure the appropriate mode adapter?

roshanshariff:
> Is the idea if a hook isn't configured, I add an error message to tell the user to configure the appropriate mode adapter?

Yeah, I think that's a reasonable way to go. In my opinion the error message should be produced in the user-facing interactive function, with the internal `citar--OPERATION` functions being non-signalling (they will return nil if they failed, but no errors). That will make it easier to reuse the internal functions in other code: you might write a function that tries to insert a citation, but does something else if that fails.

bdarcus:
Before I go too far down this path, do you endorse this strategy of using abnormal hooks to configure mode-specific behavior @minad?

@roshanshariff suggested this as an alternative to the approach of using an alist, used in bibtex-completion and the initial version of the PR on which this is based.

It makes sense to m, but not sure if we're missing something.

roshanshariff:
My motivation was that functions like `local-bib-files` may not just depend on the major mode, but might need to be customized by the user in other ways. For example, I might want to set a particular bibliography for files belonging to a certain project.

This is less important for the other operations (keys-at-point and insert-keys/citations) but I can imagine cases where you'd want to customize them per-buffer too (for example, I might have global and local bibliographies, and I can add a function that copies a bibtex entry to the local bibliography if its key is cited). The ability to add buffer-local hooks seems like it would be useful here.

But I'm new to all this, so I would also love feedback from someone more experienced like @minad :)

minad:
I am not sure if hooks are the best strategy. Hooks are which you run until success are certainly very flexible, since you can dynamically decide (good example: embark target finders, we even find multiple targets now). But there are multiple alternatives to consider. Flexibility is not always good, sometimes a more rigid design is better.

1. Use alists of major mode/function pairs. The default implementation is stored with the key t by convention. (@bdarcus' original proposal)
2. cl-defgeneric, cl-defmethod with &context major-mode

I would probably not go with hooks and use a more structured approach. A fixed dispatching criterion makes debugging and understanding easier. You can still use hooks if it turns out that these approaches are not flexible enough.

minad:
The per buffer bibliographies mentioned by @roshanshariff rather point at another issue imo. While you have functions which are directly dependent on the syntax, for which I would use a structured approach, maybe the bibliography finding is a separate concern which needs more flexibility. You could also use the hook base approach for this functionality only.

bdarcus:
Ok, thank you; will revert that commit.

Is there good reason to go with your 2 option, or just keep the original status quo with the alist?

minad:
The option 2 makes most sense when you also want to dispatch on the arguments and on the context. It is an oop style approach and allows multiple dispatch. It can be quite powerful but you probably don't need it here. I just wanted to point it out since it is another dispatching approach.

To be honest I wouldn't worry too much about getting these things right. Refactoring can be done quickly if you didn't get it right initially. This is particularly likely if one is not experienced with the programming language at hand. I also wouldn't call myself a lisp/elips expert - the main problem is that lisps are so hyper flexible, you are basically dangling helplessly and don't know which language feature to use. (At least elisp has some restrictions, no call-cc, no restartable exceptions etc). Furthermore advices allow modifications which haven't been intended in the original design. And if that's not enough you can invent your own language elements with macros. Generally the best approach is to keep it simple - and make it only as complex as needed ;)

bdarcus:
OK, then: I reverted the commit that adds the hooks, and we'll go with @aikrahguzar's original alist, but perhaps using the hook for the local-bibs instead.

I still need to get the basics working though. The markdown one doesn't work correctly ATM.

citar-markdown.el
```diff
@@ -0,0 +1,78 @@
...
+
+(defcustom citar-markdown-prompt-for-extra-arguments t
+  "Whether to prompt for additional arguments when inserting a citation."
+  ;; REVIEW this maybe shouldn't be mode specific?
+  :group 'citar-markdown
+  :type 'boolean)
+
+(add-to-list 'citar-major-mode-functions
+             '((markdown-mode) .
+               ((local-bibs-files . citar-markdown--local-bib-files)
```
roshanshariff, line 50:
```suggestion
               ((local-bib-files . citar-markdown--local-bib-files)
```
There's a typo in the alist key.

citar-markdown.el
```diff
@@ -0,0 +1,78 @@
...
+  "Return local bib files for markdown."
+  (error "Markdown does not support local bib files"))
+
+(defun citar-markdown--keys-at-point ()
+  "Return a list of keys at point in a markdown buffer."
+  (error "TODO"))
+
+(defun citar-markdown--insert-keys (keys)
+  "Insert comma sperated KEYS in a markdown buffer."
+  (mapconcat (lambda (k) (concat "@" k)) keys "; "))
```
roshanshariff, line 65:
```suggestion
  (insert (mapconcat (lambda (k) (concat "@" k)) keys "; ")))
```
This change allows at least `citar-insert-keys` to work for me in markdown-mode.

citar-markdown.el
```diff
@@ -0,0 +1,78 @@
...
+(add-to-list 'citar-major-mode-functions
+             '((markdown-mode) .
+               ((local-bibs-files . citar-markdown--local-bib-files)
+                (insert-keys . citar-markdown--insert-keys)
+                (insert-citation . citar-markdown--insert-citation)
+                (keys-at-point . citar-markdown--keys-at-point))))
+
+(defun citar-markdown--local-bib-files ()
+  "Return local bib files for markdown."
+  (error "Markdown does not support local bib files"))
```
roshanshariff, line 57:
```suggestion
  nil)
```
Throwing an error blocks selecting keys from the global bibliography.

citar-markdown.el
bdarcus: I can't even get the basic functions, like `citar-select-refs` or `citar-get-candidates` to work.

I suspect it's an installation issue.

bdarcus:
OK, markdown now working correctly on my end after updating and re-compiling.

I'm wondering if we even need `insert-keys` anymore. If we adapt `citar-insert-citation`, it could serve the same purpose, depending on where point is in the buffer.

A more specific question:

If we autoload the mode-specific internal functions, and include references to them in the main file, will that allow something like the experience with embark; that the user doesn't have to do any other config, but the correct code will load when needed?

roshanshariff:
> If we autoload the mode-specific internal functions, and include references to them in the main file, will that allow something like the experience with embark;

You'll probably need `declare-function` to make the byte-compiler happy, but I don't see why that wouldn't work.

> I'm wondering if we even need `insert-keys` anymore. If we adapt `citar-insert-citation`, it could serve the same purpose, depending on where point is in the buffer; as with `org-cite-insert`.

I agree; I tend to use insert-keys ad hoc in random modes without proper citation support. But this usage can probably be replaced with `embark-insert`, which is normally available through `embark-general-map` (but in 

In the long term, we could develop some kind of insert-or-edit-citation functionality that detects a citation at point, extracts the keys from it, shows a completing-read-multiple prompt allowing you to remove keys or add new ones, etc. But that can be done after the basic functionality is merged.

The remaining problem I'm experiencing is that functions like `citar-open` and `citar-library-files` give an error when run in some mode that's not in `citar-major-modes-functions`. There should be support for fallback implementations of `local-bib-files`, etc.

bdarcus:
> There should be support for fallback implementations of `local-bib-files`, etc.

I just realized there are two `citar-latex--local-bib-files` functions; one in `citar` and the other in `citar-latex`, but not the same.

roshanshariff:
> I'm not really sure how `citar-org` should fit here, given `org-cite-insert` and `org-open-at-point`. Ideally, if one only uses org, `citar-insert-citation` shouldn't even be available. Thoughts?

Isn't it enough if `citar-insert-citation` is just not bound? Instead, `org-cite-insert` is bound in `org-mode-map` (and can be given a shorter binding) allowing citar-org functionality to be used without going through anything in citar. But `citar` will still be loaded by `citar-org` since it provides most of the functionality.

```emacs-lisp
;; Should be the only configuration needed for org-only use.
;; TODO: make sure citar handles having only local bibliographies, no global ones
(use-package citar-org
  :no-require
  :bind ; optional
  (:map org-mode-map
        ("C-c b" . #'org-cite-insert)) ; Also bound to C-c C-x C-@
  :custom
  (org-cite-global-bibliography (list +global-bibliography))

**Comment by bdarcus** [comment]
> Isn't it enough if `citar-insert-citation` is just not bound?

I think it's still confusing for people that use `M-x`.

**Comment by bdarcus** [comment]
Also, I was partly wondering about the alist. I just decided to remove that entry entirely, so we'll see what happens.

**Comment by roshanshariff** [comment]
> I think it's still confusing for people that use `M-x`.

I guess you're saying there should be a some way of using citar so it doesn't provide any interactive commands, just an org-cite processor? This is of course possible through refactoring, but I'm not sure I see the utility.

The number of interactive commands is quite small, and they all "do the right thing" in org mode (but also happen to work in other modes). And even in org mode you want to support using Embark, and its action will require some citar-specific interactive commands (open notes, files, URLs, etc.). The *only* citar command that is redundant in org-mode (afaict) is `citar-insert-citation`, which has the same meaning as `org-cite-insert` with `org-cite-insert-processor` set to `'citar`.

> I just decided to remove that entry for org entirely, so we'll see what happens.

I'm not sure why citar-org should be treated differently from citar-latex, here? They both allow citar to work in specialized major modes, right? And they're both included in the citar package, so it seems reasonable that both should work out-of-the-box.

**Comment by bdarcus** [comment]
> The remaining problem I'm experiencing is that functions like `citar-open` and `citar-library-files` give an error when run in some mode that's not in `citar-major-modes-functions`.

I can't see where would be generating this error with those commands. Can you?

**Comment by roshanshariff** [comment]
> I can't see where would be generating this error with those commands. Can you?

Evaluating `(citar-major-mode-functions 'local-bib-files)` gives an error when there isn't an alist entry for the current mode, which propagates through
* `(citar--local-files-to-cache)`
* `(citar-refresh nil 'local)`
* `(citar--get-candidates)`
* `(citar--select-refs)`
And then to anything that needs to select a reference.

**Comment by roshanshariff** [comment]
> Does that make sense now?

Yes, you're right, there's some redundancy with org-cite functions. Once this branch is merged, I'd be happy to help with streamlining the API surface between the mode-specific backends and citar.el to a smaller/minimal set of operations. Most of this redundancy isn't user-facing however, so we should be able to refactor the code without affecting any workflows.

**Comment by bdarcus** [comment]
> Evaluating `(citar-major-mode-functions 'local-bib-files)` gives an error

Do you mean `citar--major-mode-functions`?

So the fact it was triggered with `citar-open` was merely coincidence?

And it's this function that needs the nil handling?

**Comment by roshanshariff** [comment]
> I just pushed a commit that might solve this, though I have not yet tested it.

Excellent, thanks! Seems to work now.

**Comment by bdarcus** [comment]
So what else do you think this needs to merge?

The autoloads, etc details discussed above + documentation?

Anything else?

**Comment by roshanshariff** [comment]
Yes, that about covers it, I think.

**Comment by bdarcus** [comment]
The linter complains about autoloads on private functions.

**Comment by roshanshariff** [comment]
Maybe it's reasonable to make the relevant functions public? They're not interactive, so the downsides seem minimal.

**Comment by bdarcus** [comment]
OK, that's pushed.

I'll hopefully get to the docs tomorrow.

I also pushed the start of a solution to replacing the open-entry insert-bibtex functions.

**Comment by roshanshariff** [comment]
> I'm unsure what the UX should be for opening multiple entries from the same file

I think we need a better design in general for dealing with multiple entries simultaneously. Perhaps for operations like `citar-open-entry` where it really only makes sense to select one, the behaviour should be:
* When run interactively, use `completing-read` instead of `completing-read-multiple` to select an entry from the bibliography.
* If argument has one key, just open it.
* If argument has multiple keys, use `completing-read` to select among them.

I think similar behaviour would be useful in other places as well --- I use citar as a general reference manager UI, so suppose I invoke `citar-open` to find a PDF file. It's slightly annoying to use the `completing-read-multiple` interface when I really only want to act on one item. (With `consult-completing-read-multiple`, I have to press RET to select the entry and C-RET to dismiss the completion.)

In general, I think it makes sense to use `completing-read` where possible and `completing-read-multiple` only if necessary, for example when inserting or editing a citation with multiple keys. But perhaps others have different workflows?

**Comment by bdarcus** [comment]
I've thought about that, and even experimented with it, but I found it adds technical complexity when you consider the embark at-point functionality.

If you always assume a list, it's easier and consistent.

There's probably a solution to that though.

More generally, minad and oantolin and I have talked about adding multi-actions to embark, which could allow best of both; completing-read + ability to act on multiple targets.

But that's beyond the scope of citar of course.

**Comment by bdarcus** [comment]
The other advantage of using completing-read, BTW, is consult--read and multi, the latter which could be particularly useful for `citar-open`.

**Comment by bdarcus** [comment]
Wait: did I miss that embark got support for multi selection?

https://github.com/oantolin/embark/issues/365

**Comment by roshanshariff** [comment]
If I'm not mistaken, that's the feature of embark that makes some actions "repeatable", so that selecting the action runs it but doesn't dismiss embark. Then you can run the action again (or a different one).

**Comment by bdarcus** [comment]
I should mention ...

> With consult-completing-read-multiple, I have to press RET to select the entry and C-RET to dismiss the completion.

The ideal behavior here IMO, which @minad had initially, in the vertico and consult prototypes is TAB selects multiple, and RET completes.

With that, there's no downside to multiple.

But @minad removed it, IiRC because he wasn't comfortable maintaining that additional code in consult.

Except ... In the doom vertico module at least, this is how it works now. I assume that's custom code?

But even if it's not in consult, you could always add it yourself.

**Comment by minad** [comment]
> The other advantage of using completing-read, BTW, is consult--read and multi, the latter which could be particularly useful for citar-open.

I've probably said this a few times already and I think you also agree with the overall philosophy. I'd prefer if you keep a `completing-read` implementation. You could still use `consult--read` if Consult is available. But I consider it important to keep your package independent of Consult. Of course it is a good idea if users install the advanced packages of the ecosystem, Consult, Marginalia, Embark, Orderless. 

However the point is that each component stands for itself and offers advantages by itself without being just pulled in by another package. As examples I've already heard criticism for example regarding some consult-* packages which are not decoupled enough. Then there was the question if consult-completing-read-multiple or consult-completion-in-region couldn't be part of Vertico itself, since some users wanted Vertico but not Consult. Of course I will not move this functionality, since Consult is the package which keeps all the completion UI agnostic code. But it is clear that users start at different places in the ecosystem and like to make small steps. Maybe they start with Embark, Vertico, ... I could very well imagine that many people may start with Citar since your package offers genuinely new functionality for writing papers or documentation. If Citar then pulls in half a completion system, I would certainly dislike this.

Furthermore it would be nice if you test your package against Vertico and default completion, such that we make sure that the package works well in different setups.

> Wait: did I miss that embark got support for multi selection?

No, you did not miss this. However Embark with `embark-quit-after-action=nil` (should be the default) and the repeating functionality is often almost as good as multi actions. The only difference is that with multi actions you firsts select, select, select, then execute. With Embark you act, act, act.

> Except ... In the doom vertico module at least, this is how it works now. I assume that's custom code?

Yes, the TAB/RET behavior currently needs a little custom code. I bind `vertico-exit` to `TAB` and `+consult-crm-exit` to `RET` in the `consult-crm-map`. The reason is that this code is completion UI specific and therefore not part of Consult currently. I could consider adding this to the consult-vertico/consult-selectrum.el integration modules. However for now I kept a pretty strict stance here and only added a minimum of integration code there. If you look at consult-vertico.el it is very short.

~~~ elisp
(defun +consult-crm-exit ()
  (interactive)
  (let ((exit (or (command-remapping #'exit-minibuffer) #'exit-minibuffer)))
    (run-at-time 0 nil exit)
    (funcall exit)))
~~~

File: citar.el [hunk 12]
```diff
@@ -160,11 +170,59 @@ and nil means no action."
...
   :type 'function)
 
+(defcustom citar-major-mode-functions
+  '(((latex-mode) .
+     ((local-bib-files . citar-latex--local-bib-files)
+      (keys-at-point . citar-latex--keys-at-point)))
+    ((markdown-mode) .
+     ((local-bib-files . citar-markdown--local-bib-files)
+      (insert-keys . citar-markdown--insert-keys))))
+  "The variable determining the major mode specifc functionality.
```
andersjohansson, line 200: An entry for org-mode is needed here as well, isn’t it? Getting local bib-files previously depended on `bibtex-completion-find-local-bibliography` (which actually looked for file-extensions rather than major modes, which means it doesn’t work in org capture buffers), but now it should be enough with `org-cite-list-bibliography-files` here?
This seemed to work well (also in my org capture buffers):
```
((org-mode) .
         ((local-bib-files . org-cite-list-bibliography-files)))
```

bdarcus: I hadn't actually tested that, but was assuming org handled it.

Are you aware of "suggested change" comments? There's ba like plus icon on the toolbar.

bdarcus: PS - reminds me: should change the major mode check to use `derived-mode-p`.

bdarcus: @andersjohansson - I added a little function to handle this `citar-org-local-bibs` and have hooked it up.

If you have time, could you please test?

I still have to write the documentation, hopefully sometime today, but am hoping to merge soon.

bdarcus:
I've added a TODO at the top, and also just pushed a commit that replaces the important `bibtex-completion-at-point` call and replaced it with a new function `citar-keys-at-point` which uses `citar-major-mode-function`.

This is somewhat delicate code prone to break, so this will need testing. On my end, it seems to work correctly in latex and org.

bdarcus:
@minad -

> > The other advantage of using completing-read, BTW, is consult--read and multi, the latter which could be particularly useful for citar-open.
> 
> I've probably said this a few times already and I think you also agree with the overall philosophy. I'd prefer if you keep a `completing-read` implementation.

Yeah, me too.

Is there a relatively easy way for me to get consult-multi-like behavior in a UI that is using grouping?

Basically, `citar-open` uses grouping, but I don't know how to give each group a different category.

> Furthermore it would be nice if you test your package against Vertico and default completion, such that we make sure that the package works well in different setups.

For the longest time I did, so it should still work fine (I haven't changed anything that I can imagine would have any impact on compatibility). Alas, I just tested `icomplete-vertical`, and it doesn't work correctly!

> > Wait: did I miss that embark got support for multi selection?
> 
> No, you did not miss this. However Embark with `embark-quit-after-action=nil` (should be the default) and the repeating functionality is often almost as good as multi actions. The only difference is that with multi actions you firsts select, select, select, then execute. With Embark you act, act, act.

Sounds like I need to experiment with that.

minad:
> Is there a relatively easy way for me to get consult-multi-like behavior in a UI that is using grouping?

Okay, I see. You are interested in the consult-multi functionality. I'd argue that there is no point in this functionality if you control the entire generation of the candidates. It only matters if one wants to recombine multiple candidate sources ala consult-buffer/consult--source-*. consult-multi candidates have a certain structure, which is understood by marginalia and embark. The marginalia support should not be interesting for citar, since you define your own categories and define your own annotations (if you even do, since as far as I know you generate long invisible candidate strings to allow filtering). 

However the embark support for consult-multi is more interesting. Embark understands consult-multi via a target transformer, which unpacks the actual category and the actual candidate from the consult-multi candidate.

https://github.com/oantolin/embark/blob/0b524bda8d9c5f9e26898e383e98a4ec689fa144/embark-consult.el#L207-L216

Now you could do the following:

1. You could define your own citar-multi category and register your own target transformer with embark. Most flexible approach, you don't rely on the protocol defined by embark and consult.
2. You use the consult-multi protocol.
3. We rename the "consult-multi" category to "multi" in Embark/Consult/Marginalia. By doing that we make it clear that the protocol can be used by more libraries than only Consult. However we also overstep a little bit by defining unprefixed categories, since these may in principle interfere with other Emacs packages. 

Regarding custom category names - Marginalia already defines custom unprefixed category names, which oversteps the boundaries of the package a bit. However the categories chosen by Marginalia are names which are used in Emacs itself, like (read-file -> category file, read-color -> category color, bookmark-completing-read -> category bookmark, etc). Not all of these categories are 1-1 the names in Emacs though, I just saw. For example Marginalia has the category environment-variable, but Emacs uses read-envvar-name, so probably the category envvar-name would have been better. But so far there was no need to change this - it matters in particular if Emacs defines more of its own categories. Then Marginalia should be adapted.

https://github.com/minad/marginalia/blob/09d8ab38a5a4aa55a83968dc3e454d11fee05255/marginalia.el#L78-L106

> Alas, I just tested icomplete-vertical, and it doesn't work correctly!

What is the problem? Candidates too long or issues with the text properties?

minad:
For which command do you want to use consult-multi multi categories/grouping? Intuitively I think consult-multi should be used only for very few things (consult-buffer for files or buffers/buffer-like candidates or karthik's consult-dir for directory candidates). Maybe your use case can be solved in an alternative better way?

consult-multi works well for consult-buffer+vertico. But for example as Prot told me consult-buffer+mct/default-completion does not really work well since you don't see the candidates immediately and mixing many candidate sources leads to too many similar possible candidates. For most commands it works better if one does not mix the candidates. 

I implemented consult--multi as a slightly weaker and simpler model than Helm. Helm allows to combine multiple asynchronous sources, while consult--multi does not allow this. Helm designs its entire model around multi sources as basic building blocks. While one gets general building blocks this way, I'd argue that you almost never want to recombine multiple sources. I am sure Helm aficionados disagree with me about this. Helm also has a long history (anything or quicksilver on mac), where the idea was to throw everything together and narrow down. But this throwing together everything is of course not efficient, since with too many similar candidates the number of key presses you need to reach your goal goes up.

So what I want to say - Consult makes a compromise here. It uses multi sources sparsely for very few configurable commands. For most commands it stays with the standard Emacs approach of not merging candidate sources and uses specialized commands for each category.

bdarcus:
> For which command do you want to use consult-multi multi categories/grouping?

It would only maybe be relevant for `citar-open`.

There are currently five (!) "open" action commands:

1. `citar-open-library-files`; for PDFs, etc related to the reference
2. `citar-open-link`; for the same as 1, but online, as often derived from DOIs
3. `citar-open-notes`; open an existing note associated with the reference or creates a new one
4. `citar-open`: groups all of the above, or opens them directly (depending on configuration)
5. `citar-open-entry`: the bibtex entry that is

My interest in multi-like functionality was so that annotations and actions would work consistently, but also contemplating the possibility of removing 1 and 2 above (5 is trickier, but it can be multiple entries by in the same file, which is why @roshanshariff was suggesting `completing-read` here instead).

But I could also remove 4.

Basically, I'm trying to figure out how to simplify.

cc @salutis, who I knows loves `citar-open`, despite the limitation I note.

minad:
I see. `citar-open` sounds like a reasonable use case for multi functionality. You could also remove `citar-open` and introduce a small `consult-citar` package which adds consult sources and a `consult-citar-open` command which merges these sources. Alternatively implement `citar-open` on top of `consult--multi` if available (dynamic detection at runtime), if not, keep the current version.

In my comment above I forgot to mention that `consult--multi` in Consult also adds narrowing and this somehow mitigates the disadvantages of the merging of many similar candidates. It is important to mention this if one wants to judge the up- and downsides of `consult--multi`. If you want to implement grouping, narrowing, multi categories etc without Consult you are replicating a bit of work. This points at reusing `consult--multi`.

bdarcus:
> I see. `citar-open` sounds like a reasonable use case for multi functionality. You could also remove `citar-open` and introduce a small `consult-citar` package which adds consult sources and a `consult-citar-open` command which merges these sources.

Yes, I was wondering about that too.

> In my comment above I forgot to mention that `consult--multi` in Consult also adds narrowing and this somehow mitigates the disadvantages of the merging of many similar candidates. It is important to mention this if one wants to judge the up- and downsides of `consult--multi`. If you want to implement grouping, narrowing, multi categories etc without Consult you are replicating a bit of work. This points at reusing `consult--multi` if you indeed want to have such a multi command.

Yes, I was aware of those other features, and so also assuming it'd be too much work to bother, particularly if I can do a little package like the above (or keep it here but test for presence of consult, as you mention).

minad:
> Yes, I was aware of those other features, and so also assuming it'd be too much work to bother, particularly if I can do a little package like the above (or keep it here but test for presence of consult, as you mention).

Yes, the point is "how much" of consult you use. For example if you only want grouping or annotations consult is of course not necessary. But if you want to use all the features combined (multi sources, narrowing, preview etc), Consult is worth it. This has to be decided on a case by case basis.

I would also like to point out one small detail of consult multi sources: Consult sources can be defined without a dependency to Consult, since they are simple plists (https://github.com/minad/consult/blob/3b18b04524f0e963070ab6c61d4d3a3f10507d12/consult.el#L3937-L3948). Helm originally also used plists if I recall correctly. However Helm now uses objects instead and constructor functions (e.g. `helm-build-sync-source`). Therefore you cannot define Helm sources without a dependency to Helm, while you could do that with Consult. So if you define a bunch of citar sources, the user could use them via `(consult--multi '(citar-source1 citar-source2 other-custom-source))`.

minad:
> Edit: but the downside of consult--multi is it's completing-read; not multiple.

I know you like `completing-read-multiple` very much :) But I'd argue that `completing-read-multiple` is not necessarily better. It is used very sparsely throughout Emacs and generally has bad support. I think one could live well without it, use only `completing-read` in combination with Embark in order to support acting on multiple times. However as we discussed, Embark could profit from support for multi selections. But if that support is given `completing-read-multiple` seems almost obsolete to me. 

On the other if one looks at `describe-face` one would have to jump through hoops to implement the same behavior with Embark+`completing-read`, since all the selected faces are described in the same buffer in the end. But it is of course possible - if one Embark-acts multiple times in `describe-face`, each time it would be necessary to check if a face-help buffer is already open and if it is add the newly selected face to the buffer.

Anyways, to me `completing-read-multiple` and `completing-read` are a bit in conflict/overlapping in particular if we have Embark. Maybe we can figure out a good solution to harmonize this a bit better in the future, in particular in combination with Embark multi selections.

bdarcus:
> I think one could live well without it, use only `completing-read` in combination with Embark in order to support acting on multiple times.

Oh yes; totally agree.

Really there's only one command here that truly requires multiple: `citar-insert-citation`. 

Everything else would benefit from simply being able to quickly act sequentially on multiple targets. That includes with the open commands.

minad:
> Really there's only one command here that truly requires multiple: citar-insert-citation.

Is it really? You could detect if there is already a citation string at place and then add the newly selected citation to the list. But it wouldn't be so nice to remove already existing keys if one comparse with `consult-crm`.

bdarcus:
It's mainly valuable when inserting a new citation with more than one reference.

> You could detect if there is already a citation string at place and then add the newly selected citation to the list. 

Yes, that's what the `org-cite-insert` does, and the code in this PR for latex.

bdarcus:
OK, merged.

If you see any places to refine, @roshanshariff, feel free to submit a PR.

citar.el
andersjohansson: > Are you aware of "suggested change" comments? There's a plus icon on the toolbar.
Will use that next time!

> that function returns global + local
Yes, that was my expectation, but then again, I may have missed the point of these settings.

My use-case is that I only use a global json-file (exported from zotero via betterbibtex) for all citing (and only cite in org) so I have:
```
(setq org-cite-global-bibliography '("BIBFILE.json")
      citar-bibliography nil)
```
Earlier, I experienced all entries were duplicated if I had both variables set to the same global file, but this doesn't seem to be the case anymore, so perhaps that (as recommended in the readme?) would be the right thing to do now then in any case (which means my problems are solved).

bdarcus: What if you do this (or vice versa)?

```elisp
(setq citar-bibliography '("BIBFILE.json")
      org-cite-global-bibliography citar-bibliography)
```

andersjohansson:
That seems to work now, no duplicates. I think it was more than a month ago that I saw duplicates, and I guess many things can have changed since then 🙂.

roshanshariff:
@bdarcus Great work, thank you! I'm very excited to see what how far this package has come so quickly :)

@andersjohansson The duplication bug was maybe #351, now fixed.

bdarcus:
> > Alas, I just tested icomplete-vertical, and it doesn't work correctly!
> 
> What is the problem? Candidates too long or issues with the text properties?

Forgot to answer this @minad: no idea. I get an "args out of range" error when trying to run a command.

PR #363, opened 2021-11-04
feat: completing-read citar-select-ref option

bdarcus:
Three commits:

1. a new completing-read-based `citar-select-ref` to complement the existing one.
2. changing `citar-open-entry` to use 1.
3. bugfix for notes function

I bundle the first two here in part to see how well 1 will work with embark.

Turns out: well.

cc @roshanshariff - WDYT?

### DIFF
--- a/citar-file.el
+++ b/citar-file.el
@@ -43,20 +43,6 @@
   :group 'citar
   :type '(boolean))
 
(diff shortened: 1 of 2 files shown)


bdarcus:
I merged this because of the bugfix, and because I don't think the rest will be controversial.

PR #366, opened 2021-11-05
remove bibtex-completion dependency

### LINKED ISSUE #246: Remove bibtex-completion dependency
Not a high short term priority, but ...

#222 goes a long way towards this, but to finish would also need:

1. #245 
2. replace the at-point finder code, including ensuring that `embark-act` works correctly with it
3. rewrite the code that finds local bibs
4. maybe remove the "add-pdf" commands?
5. #308
6. check on whether there are any variable dependencies still

I think that's it?

### LINKED ISSUE #367: Getting citar-markdown to work
### Issue

I'm trying to set up citations using citar in markdown but I cannot get them to work.

Using the information about `citar-org` in the README I tried the following.

```
(defvar legenda '("~/LaTeX/legenda.bib"))

(use-package citar-markdown
    :custom
    (citar-markdown-local-bib-files legenda)
    (citar-markdown-insert-keys 'citar))
```

I get the `citar-org` to work but when I try to insert a citation in a markdown file I don't see the `citar` minibuffer and I get the following error:

`Symbol’s function definition is void: citar-markdown-local-bib-files`

Am I doing something wrong? Did I miss some documentation?

bdarcus:
Fix #246.

Also fix #367.

-----

Per suggestion at https://github.com/bdarcus/citar/issues/246#issuecomment-961584996, this should do it @minad @roshanshariff.

The only issue is the `citar--insert-entry` function doesn't work correctly.

Any ideas?

I could just remove that entirely, and instead use embark for that?

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -72,6 +72,7 @@
 (defvar citar-file-extensions)
 (defvar citar-file-open-prompt)
 (defvar citar-file-variable)
+(defvar org-cite-bibliography)
 
 ;;; Variables
 
@@ -85,7 +86,7 @@
   "Face used to highlight content in `citar' candidates."
   :group 'citar)
 
-(defcustom citar-bibliography nil
+(defcustom citar-bibliography org-cite-bibliography
   "A list of bibliography files."
   :group 'citar
   :type '(repeat file))
--- a/citar.el
+++ b/citar.el
@@ -8,7 +8,7 @@
 ;; License: GPL-3.0-or-later
 ;; Version: 0.8
 ;; Homepage: https://github.com/bdarcus/citar
-;; Package-Requires: ((emacs "26.3") (bibtex-completion "1.0") (parsebib "3.0"))
+;; Package-Requires: ((emacs "27.1") (s "1.12") (parsebib "3.0"))
 ;;
 ;; This file is not part of GNU Emacs.
 ;;
@@ -40,8 +40,8 @@
   (require 'cl-lib)
   (require 'subr-x))
 (require 'seq)
+(require 'browse-url)
 (require 'citar-file)
-(require 'bibtex-completion)
 (require 'parsebib)
 (require 's)
 
@@ -57,6 +57,9 @@
 (declare-function reftex-access-scan-info "reftex")
 (declare-function reftex-get-bibfile-list "reftex")
 (declare-function TeX-current-macro "tex")
+
+(declare-function bibtex-completion-add-PDF-attachment "ext:bibtex-completion")
+(declare-function bibtex-completion-add-pdf-to-library "ext:bibtex-completion")
 
 ;;; Declare variables for byte compiler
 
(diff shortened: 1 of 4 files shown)


citar.el
```diff
@@ -834,15 +833,24 @@ With prefix, rebuild the cache before offering candidates."
...
   "Attach PDF(s) associated with the KEYS-ENTRIES to email.
 With prefix, rebuild the cache before offering candidates."
   (interactive (list (citar-select-refs
                       :rebuild-cache current-prefix-arg)))
- (bibtex-completion-add-PDF-attachment
-  (citar--extract-keys
-   keys-entries)))
+  (unless (require 'bibtex-completion nil 'noerror)
+    (error "This is a deprecated command that relies on 'bibtex-completion'"))
+  (bibtex-completion-add-PDF-attachment
```
minad, line 852: I think you also need a `declare-function` statement to avoid bytecode compilation warnings.

citar.el
```diff
@@ -8,7 +8,7 @@
 ;; License: GPL-3.0-or-later
 ;; Version: 0.8
 ;; Homepage: https://github.com/bdarcus/citar
-;; Package-Requires: ((emacs "26.3") (bibtex-completion "1.0") (parsebib "3.0"))
+;; Package-Requires: ((emacs "27.2") (s "1.12") (parsebib "3.0"))
```
minad, line 11: Is there a particular reason to require 27.2? On Debian stable there is still only 27.1 available, which I use. And it works reasonably well. I consider dropping support for 26 to be justified.

citar.el
```diff
@@ -852,9 +860,10 @@ URL.
 With prefix, rebuild the cache before offering candidates."
   (interactive (list (citar-select-refs
                       :rebuild-cache current-prefix-arg)))
+  (unless (require 'bibtex-completion nil 'noerror)
+    (error "This is a deprecated command that relies on 'bibtex-completion'"))
   (bibtex-completion-add-pdf-to-library
```
minad, line 865: Same here with the declare-function.

citar.el
```diff
@@ -834,15 +833,24 @@ With prefix, rebuild the cache before offering candidates."
   (bibtex-completion-insert-bibtex
    (citar--extract-keys keys-entries)))
 
+(defun citar--insert-bibtex (key)
+  "Insert the bibtex entry for KEY at point."
+  ;; TODO: this doesn't work correctly.
+  (with-temp-buffer
+    (citar--open-entry key)
+    (bibtex-copy-entry-as-kill))
+  (bibtex-yank))
```
minad, line 842: Haven't tested but I'd suggest something like this, which avoids going via the kill ring.

~~~ elisp
(let ((beg (bibtex-beginning-of-entry))
      (end (bibtex-end-of-entry)))
  (buffer-substring-no-properties beg end))
~~~

minad: Right, the second option does not work.

bdarcus: The first one does correctly insert the entry, but then also (I think via `bibtex-find-entry` opens the bibtex file with entry at point.

I thought the `with-temp-buffer` would avoid that?

minad: Okay, this needs more thought :)

bdarcus: Hmm, [bibtex-completion](https://github.com/tmalsburg/helm-bibtex/blob/aa775340ba691d2322948bfdc6a88158568a1399/bibtex-completion.el#L683) uses a regex search and parsebib to do this.

Maybe `bibtex-find-entry` is the wrong tool here?

bdarcus: I think I know the issue. The GLOBAL arg for `bibtex-find-entry` will search the buffer only when nil.

So I think could insert the file contents in the temp-buffer and use that?

bdarcus: This works:

```elisp
(defun citar--insert-bibtex (key)
  "Insert the bibtex entry for KEY at point."
  (let ((bibtex-files
         (seq-concatenate 'list citar-bibliography (citar--local-files-to-cache))))
    (insert
     (with-temp-buffer
       (dolist (bib-file bibtex-files)
         (insert-file-contents bib-file))
       (bibtex-find-entry key)
       (let ((beg (bibtex-beginning-of-entry))
             (end (bibtex-end-of-entry)))
         (buffer-substring-no-properties beg end))))))
```

PR #368, opened 2021-11-05
Fix pcase use in citar-filenotify

andersjohansson:
Debugging why my cache wasn’t correctly invalidated, the previous pcase patterns didn’t seem to work at all. I’m constantly confused by pcase however, so I’m not sure if this is the best way, but it seems to work correctly.

### DIFF
--- a/citar-filenotify.el
+++ b/citar-filenotify.el
@@ -79,8 +79,8 @@ If it is other than 'global or 'local invalidate both"
 
 CHANGE refers to the notify argument."
   (pcase (cadr change)
-    ('(nil changed) (funcall func scope))
-    ('(created deleted renamed)
+    ((or 'nil 'changed) (funcall func scope))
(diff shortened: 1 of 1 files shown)


citar-filenotify.el
```diff
@@ -79,8 +79,8 @@ If it is other than 'global or 'local invalidate both"
 
 CHANGE refers to the notify argument."
   (pcase (cadr change)
-    ('(nil changed) (funcall func scope))
-    ('(created deleted renamed)
+    ((or (pred null) 'changed) (funcall func scope))
```
minad, line 82: (or 'nil 'change) would work too.

bdarcus: I don't understand pcase very well either @andersjohansson. 

Can you take a look at @minad's suggestion see WHYT?

andersjohansson: That looks better, pushed that!

PR #372, opened 2021-11-05
Headers

minad:
Two small corrections regarding compilation/installation. The headers should have the simpler form at least as long as the files are distributed as a single package.


bdarcus:
Thank you!

PR #373, opened 2021-11-05
A few more minor improvements

minad:
Btw, do you plan to implement all the missing major mode functions? It would be more user friendly for example if `citar-insert-citation` works in every supported mode.


minad:
Problem is that this breaks the Embark actions in the minibuffer and the other major mode functions which rely on non existing major mode function. But does Embark work in the minibuffer? I don't see any actions.

bdarcus:
> Btw, do you plan to implement all the missing major mode functions? It would be more user friendly for example if `citar-insert-citation` works in every supported mode.

WDYM? Add more major mode support, or fill out the functions supported in the existing code?

For example, markdown doesn't yet have keys-at-point or local-bib support.

Certainly yes for the latter.

But insert citation should now work in all of them, which is most important.

bdarcus:
> Problem is that this breaks the Embark actions in the minibuffer and the other major mode functions which rely on non existing major mode function. But does Embark work in the minibuffer? I don't see any actions.

It should with latex and org.

minad:
> But insert citation should now work in all of them, which is most important.

This is what I mean.

> It should with latex and org.

Maybe there is something wrong with the Embark prompter. I am investigating this now. Still fiddling around here - I will ping you later when it works on my side.

minad:
Okay, the issue was only that your keymaps don't inherit from the `embark-general-map` which unfortunately leads to C-h not working as expected. I have to discuss this with Omar I guess.

But this PR should be ready - the only relevant commit is the one about the error messages. The other commits are only stylistic and depend on preference.

citar.el
```diff
@@ -295,7 +281,7 @@ offering the selection candidates."
            (lambda (string predicate action)
              (if (eq action 'metadata)
                  `(metadata
-                   (affixation-function . citar--affixation)
+                   (affixation-function . ,#'citar--affixation)
```
bdarcus, line 284: For my future reference, @minad, what's the significance of this change?

minad:
>> @@ -295,7 +281,7 @@ offering the selection candidates."
>             (lambda (string predicate action)
>               (if (eq action 'metadata)
>                   `(metadata
> -                   (affixation-function . citar--affixation)
> +                   (affixation-function . ,#'citar--affixation)
> 
> For my future reference, @minad <https://github.com/minad>, what's the
> significance of this change?

The hashed symbol tells the bytecode compiler to check if a function
with that name exists.

bdarcus:
And how do you judge when to use it and when not? There are other places in the code where the symbol, for example, is referring to a function; most notably in `citar-major-mode-functions`.

PR #374, opened 2021-11-05
Install embark general maps as parent keymap

minad:
This change is questionable. I know that you dislike that the general Embark
actions are always there since they clutter which-key. Unfortunately without
this change Embark is badly broken. We need at least the C-h=embark-keymap-help
action. Furthermore in the minibuffer we would still like to have the
keybindings for embark-export/collect-snapshot etc.

### DIFF
--- a/citar-org.el
+++ b/citar-org.el
@@ -386,6 +386,8 @@ strings by style."
 ;; Embark configuration for org-cite
 
 (with-eval-after-load 'embark
+  (set-keymap-parent citar-org-map embark-general-map)
+  (set-keymap-parent citar-org-buffer-map embark-general-map)
   (add-to-list 'embark-target-finders 'citar-org-citation-finder)
   (add-to-list 'embark-keymap-alist '(bib-reference . citar-org-map))
   (add-to-list 'embark-keymap-alist '(oc-citation . citar-org-buffer-map))
--- a/citar.el
+++ b/citar.el
@@ -676,6 +676,8 @@ FORMAT-STRING."
 
 ;;;###autoload
 (with-eval-after-load 'embark
+  (set-keymap-parent citar-map embark-general-map)
+  (set-keymap-parent citar-buffer-map embark-general-map)
   (add-to-list 'embark-target-finders 'citar-keys-at-point)
   (add-to-list 'embark-keymap-alist '(bib-reference . citar-map))
   (add-to-list 'embark-keymap-alist '(citation-key . citar-buffer-map)))


minad:
This addresses the Embark problems I've observed in #373. However you may not like it because of which-key.

minad:
Thanks. But please check if you are still satisfied with the action menu. Alternatively we could install only the actions which are absolutely necessary (embark-keymap-help). But I like the entire embark set, for example SPC=mark allows me to mark arbitrary targets in buffers. I prefer consistency in Embark.

bdarcus:
IIRC, my biggest issue with it and which-key is at-point.

But I'll play with it again, and it does make sense to have at least some of the other actions available.

bdarcus:
> IIRC, my biggest issue with it and which-key is at-point.

This is what I'm talking about:

image

Probably I need to get rid of the menu descriptions, or change them in some way.

roshanshariff:
This might be something that needs fixing in Embark. I think it's nice to have the `embark-general-map` bindings available, but they shouldn't be displayed by the indicator. But I'm aware that there's been a lot of discussion on this already so I'm trying to catch up to avoid going in circles. I thought of some approaches to fix this problem:

### Put embark actions under a prefix key

I know you've tried something like this [before](https://github.com/oantolin/embark/issues/207#issuecomment-881476606) @bdarcus (and [proposed](https://github.com/oantolin/embark/issues/207#issuecomment-835518230) by @minad).
```elisp
(fset 'embark-general-map embark-general-map) ; This should go in embark.el
(dolist (map (list citar-map citar-buffer-map))
  ;; Do the same for citar-org-map and citar-org-buffer-map in citar-org.el
  (define-key map "G" '("embark" . embark-general-map))
  (define-key map "\C-h" '("help" . embark-keymap-help)))
```
Pros: everything is discoverable for new users. Cons: the actions are not available unprefixed.

### Filter out the extra actions

The Embark verbose indicator doesn't suffer from this problem:
* The `negative-argument` and `digit-argument` bindings are filtered out in `embark--formatted-bindings`
* The Embark general commands can be hidden by configuring `embark-verbose-indicator-excluded-actions`
 
If the functions that do the filtering can be abstracted out (so they produce keymaps with those actions removed) then the which-key indicator can do the same thing. Cons: It's kind of ugly to have all this special-case code in the indicators, and it's not possible to have the actions be visible in some keymaps but not in others. For example, the `embark-insert` should probably be shown for other target types, but not in `citar-map` because it inserts the entire formatted completion candidate string.

(**TODO:** Citar should probably have an Embark target finder for the minibuffer that turns the formatted candidate string back into its citation key...)

### Automatically set parent keymaps

This is revisiting oantolin/embark#258, but maybe the following solution avoids some of the concerns raised there:
1. `embark-act` should use `(make-composed-keymap action-keymap embark-general-map)` so that Embark actions are always available unless shadowed by bindings in the action keymap.
2. But the indicators are given the unmodified action keymap, rather than this composed map.

The pros of this approach:
* Consistency: the Embark actions are available no matter what.
* Modularity: third-party packages don't need to depend on Embark or have their keymaps inherit from `embark-general-map`.
* Control and flexibility: the keymap decides what appears in the indicator (but see the next point). If it inherits from one of the embark keymaps, then those keys will be visible in the indicator. If it doesn't, the keys will be hidden but still available to the user. Or it can do something like the first approach above, binding `embark-general-map` under a prefix key for discoverability; Embark actions will still be available unprefixed unless shadowed by another binding.
* Configurability: The indicators should be configurable to either just show the keymap given to them, or also include the Embark general actions. So, for example, maybe the verbose indicator includes the Embark actions by default, but the which-key indicator just shows the original action keymap.
* Status quo: If implemented in this way, no existing Embark behaviour will change. Existing embark keymaps inherit from `embark-general-map` anyway, so those keys will still be visible in any indicators. The only change will be that Embark actions will be available in keymaps that don't inherit from the Embark maps.
 
I personally think a combination of all of these approaches would be nice, but of course @oantolin and @minad will have a more informed opinion.

minad:
> This might be something that needs fixing in Embark. I think it's nice to have the embark-general-map bindings available, but they shouldn't be displayed by the indicator.

These are normal actions and they should be shown. As I see it, there is nothing to fix. :)
As you noted, there is already `embark-verbose-indicator-excluded-actions` to hide unwanted actions.

### Put embark actions under a prefix key

This is inconvenient. A user who prefers to have the actions under a prefix key can move the map, but this should not be the default configuration. Since Embark uses just normal keymaps (without magic automatic inheritance mechanisms!) it is easy to move things around if you disagree as a user.

### Filter out the extra actions

Maybe the which-key indicator should also use some exclusion list? It may be reasonable to clean up the verbose indicator filter logic and abstract it such that it can be reused for the which-key indicator.

----

### Automatically set parent keymaps

> embark-act should use (make-composed-keymap action-keymap embark-general-map) so that Embark actions are always available unless shadowed by bindings in the action keymap.

Yes this makes it a bit easier to enable the general/meta keymap always available for third party keymaps. However I don't see a downside in setting the keymap parent when registering the keymap in Embark as Citar is doing. Though It may be a little bit better to register a composed keymap instead to avoid the mutation (see https://github.com/bdarcus/citar/pull/374). We should find the best pattern here, but I don't see the need for introducing the magic auto inheritance again. Note that there are different parent keymaps. Therefore we cannot hardcode this to the general map and would need additional logic to determine the parent. It will not be simpler.

> But the indicators are given the unmodified action keymap, rather than this composed map.

I strongly disagree with this proposal. The general actions should be available and should be visible in the indicator. If you want to hide some actions, just add them to the exclude list.

> Consistency: the Embark actions are available no matter what.

In principle good, but we have different parent keymaps.

> Modularity: third-party packages don't need to depend on Embark or have their keymaps inherit from embark-general-map.

Not an issue. When registering the keymap, the registered keymap can be either modified to add the parent or one should register a composed keymap in the first place.

> Control and flexibility: the keymap decides what appears in the indicator (but see the next point). If it inherits from one of the embark keymaps, then those keys will be visible in the indicator. If it doesn't, the keys will be hidden but still available to the user.

This sounds confusing. From my experience users have issues understanding Embark, if we add a feature where the general actions are sometimes shown or not it will be more confusing.

> Configurability: The indicators should be configurable to either just show the keymap given to them, or also include the Embark general actions. So, for example, maybe the verbose indicator includes the Embark actions by default, but the which-key indicator just shows the original action keymap.

I partially agree with this. If we can make the indicators more flexible it is good, e.g, also hiding based on the keymap instead of only based on the command name.

> Status quo: If implemented in this way, no existing Embark behaviour will change.

This is not a pro argument - it is rather a minimal requirement ;)

----

So to summarize it seems to me that there is only an indicator problem - nothing more. The keymaps are good as they are and the current mechanism is simple and easily understood. There is no magic inheritance mechanism. Note that we removed this mechanism in an attempt to simplify the situation and in order to give more control to the user who defines the keymap. I wouldn't want us to go back to the old automatic inheritance mechanism and I wouldn't be happy with indicators differentiating between automatically inherited keymaps and explicitly configured parent keymaps, which is even more complicated than before. The indicators can be improved and I think improvements there would be welcome. The Embark verbose indicator has been introduced to make Embark more accessible to new users. The which-key indicator could profit from similar improvements. However as one gains more experience with Embark one may want to switch to the minimal indicator - the completing read prompter is still available to help. Omar is not using the verbose indicator or which-key indicator regularly as far as I know, so this dog fooding aspect is missing.

minad:
Another general thought - it is always better to enable and show all the things in the default configuration, since this helps discoverability. It is easier to disable and hide than show something hidden. You may not even be aware of the existence. I am following this in my packages mostly. Embark also follows this principle to some extent - the verbose indicator is a good example. I have it disabled, Omar has it disabled. But it is very helpful to get started.

It is a bit of a different question if you want to retain backward compatibility. If a new feature gets added in Emacs it is often disabled by default in order to not offend the old timers. Not a good idea in my opinion. It would be better to make people automatically aware of changes by just enabling them right away (Of course only features which are harmless - no magic auto delete please). Disabling the feature is just a setting away.

roshanshariff:
@minad, I hope you won't mind if I try a little harder to convince you about some aspects of this proposal :slightly_smiling_face:

1. When embark is activated, *every* Emacs key binding is a potential action. The only addition is that an action-specific keymap is added to the front of the active keymaps list. I'm proposing that *two* keymaps be activated instead --- first, the unmodified action keymap, and second an "embark keymap" that does not depend on the action. This doesn't seem very magical, to be honest, and in fact is more consistent for the user. To be clear: I'm not saying that the action keymap is modified, just that Embark chooses another keymap to activate.

2. Now, for the indicators: even though any Emacs binding is a potential action, the indicator should of course not show every single keybinding to start off with. The user probably wanted to use a target-specific action, so they show only those at first (but may expand if the user presses a prefix key that's not bound in the action map). My proposal is that the indicators should continue to do this: just show the actions that are in the (unmodified) action keymap. Again, not very magical.

How does (1) interact with (2)? Well, in the simplest case nothing special needs to happen. `embark-act` activates an action map and an "embark keymap", but the indicators can just show the action map as usual.

But, for discoverability, maybe the indicator shows the union of the action map and the "embark map", minus the unnecessary noise. The verbose indicator can show them in separate sections: the first has the bindings in the action map and the second has the bindings in the "embark map" that are not in the action map. This second section will be useful to people new to Embark and can be disabled once they're familiar with it.

If you're not only familiar with Embark but also with all the action keymaps, then you don't need any indicator at all. You and @oantolin probably fall into this category. But keep in mind that once third-party packages start supplying Embark action maps, it will help to have indicators for those new unfamiliar keymaps.

On the whole, all this seems less magical and more decoupled to me.

roshanshariff:
In general, Emacs has the convention that transient modes (like `isearch-mode`, for example) have their own keymaps that are activated when the mode is active (c.f. `isearch-mode-map`). Now Embark looks like a mode and quacks like a mode, but doesn't have an `embark-mode-map`. Now it so happens that the main function of "Embark mode" is to find targets and activate a target-specific action keymap. Because of this we've been trying to shoehorn an "embark-mode-map" into the action keymap, which is the root of all these problems. I say we just activate this extra map and get it over with :)

All this is independent of indicators, by the way. But of course Embark indicators should guide the user about the action keymap and the "embark mode map" as best they can, and depending on the user's familiarity with the action keymap and Embark.

minad:
> I hope you won't mind if I try a little harder to convince you about some aspects of this proposal

I honestly don't see the point and I don't see new arguments. The current solution is simple and consistent on the side of Embark and your proposal does not enable any new behavior which cannot be implemented with the existing code. 

As I said, we simplified a few things on the side of Embark. Activating multiple keymaps at once is not simpler, adding auto inheritance is not simpler. The current implementation activates a single keymap, which is just a normal keymap with parents. This is a common Emacs scenario, which is familiar to the user.

Furthermore your goal seems to be to solve *indicator issues*. Let's not change the keymap mechanism in order to improve the indicator situation. Instead fix the indicator!

minad:
> Now it so happens that the main function of "Embark mode" is to find targets and activate a target-specific action keymap. Because of this we've been trying to shoehorn an "embark-mode-map" into the action keymap, which is the root of all these problems.

This comes closer to an actual problem. Embark uses the embark-meta-map/embark-general-map for two different purposes. 1. As a substitute for a transient mode map as you point out. 2. As a source for general actions. 

This is indeed questionable and one could indeed separate an `embark-mode-map` from the actual action maps. This map will always be active an have the commands in it which are needed for Embark to function correctly. The actions which are of general use should stay in the `embark-general-map`.

minad:
Let's take a closer look at `embark-meta-map` and `embark-general-map`.

## embark-meta-map

- `C-h` embark-keymap-help (required)
- `-` negative-argument (required)
- `0-9` digit-argument (required)

## embark-general-map

- `i` embark-insert (general action)
- `w` kill-new (general action)
- `C-s` embark-isearch (general action)
- `SPC` mark (general action)
- `DEL` delete-region (general action)
- `C` embark-consult-search-map (general action prefix map)
- `E` embark-export (pseudo action)
- `S` embark-collect-snapshot (pseudo action)
- `L` embark-collect-live (pseudo action)
- `B` embark-become (pseudo action)

The `embark-meta-map` is indeed the `embark-mode-map` you are missing and it contains commands which are needed for Embark to function correctly. In contrast the general map contains actions (and pseudo actions) only. So I think I agree with your proposal of always activating the meta map and taking it completely out of the inheritance hierarchy. Then the indicators wouldn't also have to bother with the commands registered there.

roshanshariff:
Yes, exactly, we agree! Then third-party packages like Citar don't need their keymaps to inherit from anything. But of course users can bind some actions from `embark-general-map` in `citar-map` if they want, and Citar can also provide some of these bindings itself when it makes sense.

oantolin:
I'm coming a little late to the discussion, so forgive if I miss something:

@bdarcus: all those `digit-argument` and `negative-argument` commands are already filtered out in the most recent version of the which-key indicator on the wiki.

@roshanshariff: let me see if I understand your proposal, which seems to me to be the old auto-inheritance solution we moved away from. You propose that:

1. We don't inherit from `embark-general-map` in the rest of the action maps.

2. The keymap prompter for `embark-act` would now activate the action map for the target type and additionally `embark-general-map`.

3. The indicators would recieve only the action map for the specific target type.

So, if I understand correctly you want the same key bindings to be activated in `embark-act` as are activated now, i.e., from the point of view of which action key bindings are available, you want no change at all. Your proposal only affects indicators, then, right? So it seems to me that a simpler way to achieve the change you want is to add a configurable filter list to the `which-key` indicator. The verbose indicator already has a configurable filter list and the which-key indicator (the latest version on the wiki) also has a filter list, we just need a tiny change to make it configurable.

I like the idea of adding a filter list to the which-key indicator more than I like your proposal for several reasons:

1. It achieves the same effect without modifying Embark at all, just modifying the which-key indicator!

2. It seems conceptually simpler to say `embark-act` activates the action keymap than to say, it activates that and `embark-general-map`.

3. Plus, if we always activate `embark-general-map` someone (say @bdarcus :P) will at some point say it sucks that `embark-general-map` is hard-coded, sometimes you want to configure which additional keymap is activated! It seems much easier to always just activate one keymap, which can already be configured.

@minad: I agree that `embark-meta-map` sort of fulfills the role of an `embark-mode-map`, but I disagree with saying it contains the bindings *required* for Embark to function. The `digit-argument` and `negative-argument` commands are just a very minor convenience feature: I bound them just so I didn't have to press a modifier! I often even forget I bound them, my muscle-memory still has me mostly using the global `C-3` bindings anyway...

Maybe `embark-keymap-help` is a little closer to required, but I also wouldn't say you really need it. If you use the verbose indicator or the which-key indicator, you probably don't use `embark-keymap-help` much.

minad:
@roshanshariff 

> Yes, exactly, we agree! Then third-party packages like Citar don't need their keymaps to inherit from anything.

I am not so sure. Citar should inherit from the general map in order to make its actions available. If every package which registers an Embark keymap follows this convention, we also don't win anything of always enabling the meta map independent of the general map.

> What is the problem actually? So far it looks to me that the only problem is the indicator.

I would like to ask you again about the actual problem - since it seems that it is only about indicators since you dislike that the indicator is cluttered with the default actions. But the actions are there since they are useful!

oantolin:
> EDIT: I would only add that `embark-meta-map` should maybe use `?` for help instead of shadowing the global `help-map`. I often want to use `describe-function` etc. as Embark actions.

@roshanshariff I do that in my personal configuration: unbind `C-h` and bind `?` instead. Maybe it should be the default.

minad:
@oantolin 

> I agree that embark-meta-map sort of fulfills the role of an embark-mode-map, but I disagree with saying it contains the bindings required for Embark to function.
> Maybe embark-keymap-help is a little closer to required, but I also wouldn't say you really need it. If you use the verbose indicator or the which-key indicator, you probably don't use embark-keymap-help much.

You are right about the digits. Only `embar-keymap-help` is actually required. I insist on calling it required since if you use Embark configured with the minimal indicator and then register a keymap which does not inherit from `embark-meta/general-map` you cannot open the help!

roshanshariff:
I'll try my best to answer all your comments, but forgive me since I can't quite keep :) (But thank you both for listening!)

> The keymap prompter for embark-act would now activate the action map for the target type and additionally embark-general-map.

Not quite, it would activate the action map and `embark-meta-map`. This map should be very minimal and contain only "required" Citar commands. I agree with @minad that `embark-keymap-help` is required, for the same reason. I would say that `embark-cycle` should also go here. But if you feel strongly that you don't want `embark-keymap-help` here, you can unbind it.

Now I see `embark-general-map` more as a bunch of commands that are not *really* embark-specific, but are often useful when you're in embark. So the action keymap defined in the Embark package will probably want to inherit from it. Citar can choose to inherit from it, or can pick and choose if some actions make sense but others don't. (For example, embark-insert doesn't make sense for Citar, because it has its own citation/key insertion functions that are more clever.)

Now, as a user, if you strongly feel that some action is missing from the Citar map you can either (a) bind it in citar-map yourself, or (b) bind it in `embark-meta-map` if it makes sense for all Embark actions.

oantolin:
But if pospone the indicator discussion I'm a little confused about what is actually at stake. Is the only thing we are discussing whether it should be the responsibility of packages like citar to bind `embark-keymap-help`, possibly by inheriting from `embark-meta-map`, or if Embark should ensure itself that the binding is present?

Mmh. Embark does already treat `embark-cycle` specially. Maybe there is a point to treating `embark-keymap-help` specially too.

minad:
@roshanshariff If you disagree with the general map - then don't inherit from it and instead inherit from the meta map. This should also solve the indicator issue. But since Embark applies to many targets in Emacs it will be odd if Citar does not make the general actions available. As a user one will quickly miss the general Embark actions. I would strongly recommend against doing this.

This discussion has been around for about half a year. And the only problem was that the indicator shows too much and it is once again the same problem. And this applies in particular to the which-key indicator, which is just problematic. I stopped using which-key since it is quickly overwhelmed by large keymaps like C-x. You won't find anything. But the solution has already been presented multiple times - either use an exclude list or don't use such a verbose indicator at all and use the completing-read prompter instead to quickly find the desired action.

> Citar can choose to inherit from it, or can pick and choose if some actions make sense but others don't. (For example, embark-insert doesn't make sense for Citar, because it has its own citation/key insertion functions that are more clever.)

You can simply shadow `embark-insert` - no problem there?

> Now, as a user, if you strongly feel that some action is missing from the Citar map you can either (a) bind it in citar-map yourself, or (b) bind it in embark-meta-map if it makes sense for all Embark actions.

This would be wrong. `embark-meta-map` is not for actions. There is the general map for this purpose.

> I'm deferring discussion of the indicators for now, because it seems to be causing some confusion about what I'm proposing and why. Once we've figured out the right thing for "core" Embark to do, it should be easy to figure out how to make the indicators Do The Right Thing.

I vote to keep the status quo. As I see it core Embark does the "right thing". Deferring the indicator discussion does not seem correct since this is the problem here. The general Embark actions are useful for most targets, so it makes sense to collect them there in the general map. Having the keybindings available should never be a problem. You even made the argument that the general map should be always activated silently. Therefore I would say that the discussion is actually an indicator discussion :)

bdarcus:
> @bdarcus: all those `digit-argument` and `negative-argument` commands are already filtered out in the most recent version of the which-key indicator on the wiki.

So we want to update the doom module code for this @oantolin?

oantolin:
I'm warming up to this idea of treating `embark-keymap-help` specially. It is the escape hatch to go from the keymap prompter to the completing-read prompter. There is also an escape hatch in the opposite direction and it is handled specially: there is the `embark-keymap-prompter-key` variable which contains the key that is bound inside the completing-read prompter to be able to select actions by key binding.

Maybe it would be best to add a similar variable for the key binding used to switch to the completing-read prompter and 
remove the silly pseudo-command `embark-keymap-help` (which isn't ever run!).

And if we do always ensure `embark-keymap-help` is bound, then we don't need to discuss whether to auto-inherit from any keymap, since it seems that is the only command really "needed for Embark to function".

oantolin:
> So we want to update the doom module code for this @oantolin?

I think so, @bdarcus, but I'd hold off a bit on updating. The filtering should be configurable. Probably people don't want to see the general actions most of the time either.

minad:
@oantolin

> And if we do always ensure embark-keymap-help is bound, then we don't need to discuss whether to auto-inherit from any keymap, since it seems that is the only command really "needed for Embark to function".

I agree. Since we also treat `embark-cycle` specially that should be okay. I guess the only issue is the keybinding. As it turns out we also have `embark-cycle-key`. I think I am the one responsible for this brain fart. And I wanted Embark to be only configurable via keymaps! :-P

roshanshariff:
> You can simply shadow embark-insert - no problem there?

This is a hack. It forces Citar to rely on embark-insert being bound to a particular key, and also forces Citar to use the same key for its own insertion function. In software engineering terms, it is making the binding of `embark-insert` in `embark-general-map` part of the exported interface of Embark.

> This would be wrong. embark-meta-map is not for actions.

Why not? If it is the keymap that is active whenever embark-acting, then it surely makes sense (as a user) that should be available whenever embark-acting.

> And I wanted Embark to be only configurable via keymaps! 

You can achieve this! Just bind embark-keymap-help and embark-cycle in embark-meta-map, and always activate embark-meta-map whenever embark-act is running. Then get rid of `embark-cycle-key` (or just have it be a customization point). This way, if somebody does describe-function on `embark-cycle` or `embark-keymap-help`, Emacs will very helpfully tell them that they're bound in embark-meta-map and what key to use.

minad:
@roshanshariff 

> This is a hack.

No it is not. The Embark keys follow a convention which is not about to change. Keybindings are part of the exported interface - they are as much interface as you can get. They interface the user!

> Why not? If it is the keymap that is active whenever embark-acting, then it surely makes sense (as a user) to use it for actions that should be available whenever embark-acting.

By definition. `embark-meta-map` is used also as parent map for the become keymaps. It does not make sense to bind actions there by definition. But you can of course ignore this.

> You can achieve this! Just bind embark-keymap-help and embark-cycle in embark-meta-map, and always activate embark-meta-map whenever embark-act is running. 

I think this is what is going to happen as I understood @oantolin. All the special bindings which are needed for "Embark to function correctly" will always be available by magic. But there won't be a change to the inheritance mechanism. You can remove the `set-keymap-parent` which clutters the Citar action maps.

minad:
I've quickly put together a PR which adds a `embark-help-key` and removes the `embark-keymap-help` binding from the meta map, such that everything should function correctly even if the action keymap does not derive from any embark keymap.

See https://github.com/oantolin/embark/pull/401

roshanshariff:
I think we mostly agree now :slightly_smiling_face: In retrospect, perhaps I should have worded some things differently, but one lives and learns!

> They interface the user!

Exactly, keybindings in `embark-general-map` are part of the contract between the user and Embark, and in particular can be customized by the user. For Citar to reliably shadow `embark-insert` while still inheriting from `embark-general-map`, it would have to know where the user has decided to bind it! Alternatively, Citar should just bind its insert function to whatever makes sense for it, which is now a contract between Citar and the user to be customized as they please.

And, following the Embark ethos, `citar-map` will then remain a regular keymap that can be used outside Embark as well, bound to a prefix key for example.

But I think we may be talking past each other now, so maybe once the proposed changes have been made we can see where we are and adjust accordingly.

> I've quickly put together a PR

Thank you very much! I'm always astounded at how quickly you're able to produce working code :) I have some more thoughts on how to simplify this aspect of Embark (from my brief encounter with the keymap prompter when I was adding the prefix-help-char support) but I will collect my thoughts more carefully and discuss them there.

minad:
> I think we mostly agree now

Yes, we do! :)

> For Citar to reliably shadow embark-insert while still inheriting from embark-general-map, it would have to know where the user has decided to bind it!

Here I argue that Citar should *not* reliably shadow `embark-insert` since the users may intentionally want to unshadow it for whatever reason they have. I know it does not make much sense to still have access to `embark-insert` for the heavily formatted Citar candidates and there is always M-x if you really want to do this.

> And, following the Embark ethos, citar-map will then remain a regular keymap that can be used outside Embark as well, bound to a prefix key for example.

Yes, this is a great approach! And it should also be documented in the Citar readme.

> But I think we may be talking past each other now, so maybe once the proposed changes have been made we can see where we are and adjust accordingly.

No, I think we are not. Citar can now remove the parent keymaps. Note that I pointed out the "problematic nature" of this PR already in the comment https://github.com/bdarcus/citar/pull/374#issue-1046237824. I appreciate that you restarted the discussion.

However I would still prefer if Citar keeps the parent keymaps since I consider the general actions to be useful, also note the pseudo actions for snapshoting and exporting. On the other hand the user can also set the keymap parent in the user config, so there is no serious problem here. Independent on how @bdarcus decides about the parent keymap, the Embark help will be available.
 
> Thank you very much! I'm always astounded at how quickly you're able to produce working code :) 

You are welcome! If you are sitting in front of Emacs anyways, then you can quickly apply some changes on the fly basically. But this is simple code and there is probably some issue there for @oantolin to find ;)

> I have some more thoughts on how to simplify this aspect of Embark (from my brief encounter with the keymap prompter when I was adding the prefix-help-char support) but I will collect my thoughts more carefully and discuss them there.

From my side any kind of simplification is welcome. Often packages add and add stuff, I appreciate when things get simpler. But I see Embark as a great simplification overall. Many utilities fall out from its generality, like the consult-outline live toc we discussed recently (https://github.com/oantolin/embark/issues/399#issuecomment-962623740).

bdarcus:
> However I would still prefer if Citar keeps the parent keymaps since I consider the general actions to be useful ...

Which are useful here though, at least at point? Most seem, and so are a distraction.

For example, the delete region one isn't useful when we may more specific commands to delete or modify a citation or reference.

In the mini buffer, definitely embark-collect-snapshot and embark-become are useful, and so I definitely want those available.

minad:
> Most seem not, and so are a distraction.

No this is the misconception to I try to argue against. They are not a distraction by their mere existence in the keymap. They only become a distraction thanks to the indicator. Therefore I argue fthst the issue is actually an indicator issue - the solution is to simply hide the unwanted actions. Hiding the general actions (not by default, but in the user config) is a good idea. If everything is consistent they are always there. 

The point is that embark allows many more keybindings within the action context like normal c-x commands. These are also not a distraction by their mere existence in the keymap or availability. The only difference is that they are not shown in the indicator!

But I don't feel too strongly about this - if you drop the keymap from here one can rebind the keymap in the user config. The only thing I really care about is that we do not reintroduce another magic inheritance mechanism in Embark, beyond the most basic commands/bindings which must be always there (embark-cycle, embark-keymap-help and the default action on RET).

roshanshariff:
> They are not a distraction by their mere existence in the keymap.

I'm not sure I agree with this. If you care about the action keymap as a meaningful thing containing actions applicable to a particular target type, then it doesn't make sense for it to have actions that don't apply to that target. (Regardless of whether they're inherited from elsewhere or not, that's just an Emacs implementation detail.)

In this regard, they're more like prefix keymaps than the entire Emacs keymap hierarchy. Sure, I don't really care if some extra commands are bound somewhere in the hierarchy, but I really want the `C-x n` keymap to only contain narrowing commands, `C-x r` to only contain register/bookmark commands, `M-s` to do with searching, etc. As a user, I would not be happy if some package just put some questionably relevant bindings in various prefix maps.

I see the indicator problem as a symptom of this. I use which-key and like it quite a lot. If I know that I want to do something with frames, I hit `C-x 5` and which-key tells me what I can do, or `C-x n` for narrowing etc. @oantolin complained that which-key is not effective when if you just hit `C-x` for example, but that's not where it shines. (And, in fact, I find it useful there as well, if I linger for a bit after `C-x` that means I'm not quite sure where to go, and `which-key` shows me a compact single-page thematically-grouped index of a lot of Emacs' built-in functionality.)

Along those lines, I mentioned above that `citar-map` should ideally be something that can be bound as its own prefix keymap, and for that purpose it's not great if it contains irrelevant commands that that just happen to be bound in `embark-general-map`. Of course we can get around this by making another keymap that inherits from `citar-map` just for embark usage, but I think you'll agree that's not elegant.

bdarcus:
I'm pretty agnostic about the solution, but will note the "bug report" we just got that I linked to here.

If we do hide entries, how?

minad:
> I'm not sure I agree with this. If you care about the action keymap as a meaningful thing containing actions applicable to a particular target type, then it doesn't make sense for it to have actions that don't apply to that target. (Regardless of whether they're inherited from elsewhere or not, that's just an Emacs implementation detail.)

They do - the action keymaps contain meaningful things which apply to the target. If you disagree with some specific general actions since they are not actually that general you can propose to remove them. But note that Embark is approximating some things - it is not about creating the perfect keymap which may not have any false positives under any circumstances. See also what I wrote in https://github.com/oantolin/embark/pull/401#issuecomment-963965104 about the approximate nature of Embark.

minad:
@bdarcus If we do hide entries, how?

I think it should be sufficient to just not inherit from the Embark keymaps here, after https://github.com/oantolin/embark/pull/401#issuecomment-963965104 or some variation of it gets merged.

PR #378, opened 2021-11-06
Fix defcustom types, and other small changes

### LINKED ISSUE #371: uniitialized error/warning in customize ui
Why this, and how to fix?

image

roshanshariff:
* Having a complicated type for the `citar-major-mode-functions` defcustom was breaking the customize buffer. For now, just give it a simple alist type. Fixes #371.
* Improve the type specifications for various other defcustoms.
* Fix a typo in a keymap definition.
* Use derived-mode-p to test the keys in `citar-major-mode-functions`, and allow `t` as a key to serve as the fallback choice.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -62,6 +62,10 @@
 
 ;;; Variables
 
+(defgroup citar nil
+  "Citations and bibliography management."
+  :group 'editing)
+
 (defface citar
   '((t :inherit font-lock-doc-face))
   "Default Face for `citar' candidates."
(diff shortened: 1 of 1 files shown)


roshanshariff:
I haven't quite figured out what the problem is exactly, but this is definitely a bug in customize. It seems to break whenever you specify an `alist` type whose keys and/or values are `repeat` or `alist`. Moreover, if any variable in the customize buffer has a problem, then the widgets for all other variables also stop working.

bdarcus:
Interesting.

Thanks!

Issue #381, opened 2021-11-07
Custom completion categories and Embark keymaps


minad:
I've seen that citar defines its own completion categories `bib-reference`, `citation-key` and `oc-citation`. First I'd like to make the proposal to prefix the categories with `citar-` since they are Citar-specific as far as I understood. What is the difference between these categories? It seems you distinguish the categories mostly to provide different keymaps in Embark? This brings me to my second point - does it make sense to keep this distinction, since it goes a bit against the Embark spirit where no difference is made for the keymaps of objects coming from minibuffer or normal buffers. Of course it does not make sense to execute all actions on minibuffer candidates or buffer candidates. However Embark also allows to execute arbitrary actions via M-x on the current candidate, which makes this point less relevant.

bdarcus:
> What is the difference between these categories

@roshanshariff asked about this recently too.

I will say part of the reason is to do with the bibtex-completion dependency, and the fact that the "insert" commands work differently than does `org-cite-insert` as it works with the `org-cite-make-insert-processor` function.

In org-cite, it makes sense to include `org-cite-insert` at point, because that command behaves differently depending on context. So if point is after an existing key, for example, it appends a new one. If it's on a key, it replaces. If on the "cite" part, it prompts to edit a style.

But the bibtex-completion latex and markdown insert aren't that smart, so they don't make sense to use at point, and can cause unexpected results.

But ...

If we revised those insert citation functions to behave consistently with `org-cite-insert` (something @aikrahguzar already started to do with his latex adapter), then not only could we align the keymaps, but it _might_ also make sense to remove `citar-insert-keys`.

Only practical impediment is my current lack of understanding of buffer/point coding.

roshanshariff:
> [@bdarcus here](https://github.com/bdarcus/citar/pull/390#issuecomment-965118522) I think the direction this is headed is two embark maps, or possibly even one.

I like the idea of having only one target type, but thinking about it more perhaps there are logically two kinds of targets: citations and references.

Citations (subsuming 'oc-citation) are targets found in buffers. Their actions should probably include: open-all-[files/url/notes/bibentries], insert/edit/delete keys, change citation type, and edit pre/suffix.

References (subsuming 'bib-reference and 'citation-key) are targets appearing either in the document as components of citations, or as completion candidates. Their actions should be open [file/url/note/bibentry] and insert (into citation at point or make new citation otherwise)

[...]

Issue #262, opened 2021-08-27
Add biblio lookup?


bdarcus:
I'm not sure this is a good idea, but while I'm thinking about it ...

Here's how the `biblio-lookup` interactive function works:

```elisp
(biblio-lookup 'biblio-crossref-backend "public space")
```
In `helm-bibtex`, you can pipe a search string to that, with the "fallback" section below disclosing as you type in the prompt:

![image](https://user-images.githubusercontent.com/1134/131195366-c6181185-433b-42e8-a14b-4feb7e508385.png)

With `ivy-bibtex`, "fallback" is just an action, that opens an action submenu,

How might this work in `completing-read` though?

aikrahguzar:
This does seem difficult since we need both the search string and a selected entry.

But to me at least a better way would be to put these fallback options under an embark menu. I haven't looked at how embark targets work but can we use the search string as an emabrk target? The readme for embark says that `In the minibuffer, the target is the current top completion candidate`, so I assume the search string is not a default target and a custom one will be needed.

bdarcus:
> But to me at least a better way would be to put these fallback options under an embark menu. I haven't looked at how embark targets work but can we use the search string as an emabrk target? The readme for embark says that `In the minibuffer, the target is the current top completion candidate`, so I assume the search string is not a default target and a custom one will be needed.

Worth considering, but I'm not sure how this would work, either from a technical POV, or UX.

minad:
> I haven't looked at how embark targets work but can we use the search string as an emabrk target? 

There is `embark-become`. Is this what you are looking for?

bdarcus:
> There is `embark-become`. Is this what you are looking for?

Unless there's something misconfigured here or in doom, I don't think so.

This is what I get when I invoke it:

```console
which-key: No bindings found in Act on embark-become 'state public space'
```

Is that expected?

Basically, what we're discussing here is a workflow by which you enter filter terms for the library, but if not finding anything, or what you're looking for, instead you run a search of online databases.

minad:
`embark-become` is not a fallback, but it allows you to reuse the input string and inject it in another command. See https://github.com/oantolin/embark#switching-to-a-different-command-without-losing-what-youve-typed. I think this is approximately what you are looking for here modulo bugs? But maybe I misunderstood something. Of course one could also add a fallback action/menu item which then starts a biblio command, but I would find that pretty ugly and not in the spirit of Embark.

bdarcus:
[...]

PR #386, opened 2021-11-07
feat: add embark-become map

bdarcus:
This is the start of an `embark-become` keymap for biblio lookup (see #262).

But it seems, given how the become mappings work, it might make sense to also address #381.

In particular, I'm having a hard time accommodating so many bindings with reasonably mnemonic keys, since the become maps has to have both sets of mappings. 

So how should this work?

Another possibility is just not include this and point people to the wiki.

https://github.com/bdarcus/citar/wiki/Embark

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -31,7 +31,7 @@
 ;;  and CSL JSON bibliographic data, including LaTeX, markdown, and org-cite
 ;;  citation editing support.
 ;;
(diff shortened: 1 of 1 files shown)


bdarcus:
I'm going to close this in favor, at least for now, of:

https://github.com/bdarcus/citar/wiki/Embark

Issue #383, opened 2021-11-07
Make citar-insert-citation context aware


bdarcus:
In markdown and latex, `citar-insert-citation` should ideally work pretty similarly to how `org-cite-insert` works:

1. if point is on an existing key, prompt to replace it
2. if point is after existing key, prompt to append a new one (or new ones)
3. if point is before an existing key, prompt to push a new one (or new ones)

For LaTeX, also:

4. if prompt is on a command, prompt to replace it

With this, could simplify the keymaps, and _maybe_ remove the interactive `citar-insert-keys`.

I also am wondering if we can define `org-cite-insert`, or a wrapper for it, as the org `insert-citation` function?

See also #381

bdarcus:
Closed via #426

PR #390, opened 2021-11-09
rationalize maps; remove embark parent maps

### LINKED ISSUE #381: Custom completion categories and Embark keymaps

bdarcus:
This reverts #374, but in the context of this embark change:

https://github.com/oantolin/embark/pull/401

See https://github.com/bdarcus/citar/pull/374#issuecomment-963980484.

I'd also maybe like to figure out to fix #381; to simplify and generalize the embark maps if I can figure out how to do that.

To me the big question is how to get a "insert citation" binding to use `org-cite-insert` in org-mode, but otherwise use `citar-insert-citation` (see #383).

Edit: I am making progress on the last thing, but it requires other changes, including to `citar--local-files-to-cache`.

### DIFF
--- a/citar-org.el
+++ b/citar-org.el
@@ -95,20 +95,9 @@ Each function takes one argument, a citation."
     map)
   "Keymap for org-cite Embark minibuffer functionality.")
 
-(defvar citar-org-buffer-map
-  (let ((map (make-sparse-keymap)))
-    (define-key map (kbd "o") (cons "open source (file or link)" #'citar-open))
-    (define-key map (kbd "e") (cons "open bibtex entry" #'citar-open-entry))
-    (define-key map (kbd "f") (cons "open source file" #'citar-open-library-files))
-    (define-key map (kbd "l") (cons "open source link" #'citar-open-link))
-    (define-key map (kbd "n") (cons "open notes" #'citar-open-notes))
-    (define-key map (kbd "r") (cons "refresh" #'citar-refresh))
-    map)
-  "Keymap for org-cite Embark at-point functionality.")
-
 (defvar citar-org-citation-map
   (let ((map (make-sparse-keymap)))
-    (define-key map (kbd "<mouse-1>") (cons "default action" #'citar-dwim))
+    (define-key map (kbd "<mouse-1>") (cons "default action" #'org-open-at-point))
     (with-eval-after-load 'embark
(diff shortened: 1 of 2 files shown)


oantolin:
> To me the big question is how to get a "insert citation" binding to use `org-cite-insert` in org-mode, but otherwise use `citar-insert-citation`.

If you really never never want to use the current `citar-insert-citation` in org mode, you could just make that function check if you are in org mode and if so run `org-cite-insert`.

bdarcus:
> you could just make that function check if you are in org mode and if so run `org-cite-insert`.

Oh, right!

I suppose I need to leave room for someone wanting to use org-ref, for example, but I can figure that out.

citar.el
```diff
@@ -188,6 +188,7 @@ If you use 'org-roam' and 'org-roam-bibtex', you can use
 (defcustom citar-major-mode-functions
   '(((org-mode) .
      ((local-bib-files . citar-org-local-bibs)
+      (insert-citation . citar-org-cite-insert)
```
bdarcus, line 191: I'm not clear why this doesn't work.

bdarcus:
As I  slowly work on this, and it is related to your previous comments on the maps, @roshanshariff, any input?

I think the direction this is headed in two embark maps, or possibly even one.

I do have some local changes I'll push in a bit that gets this closer to working.

citar-org.el
```diff
@@ -178,6 +167,10 @@ With PROC list, limit to specific processor(s)."
         references
       (car references))))
 
+(defun citar-org-cite-insert (&rest _args)
+  "Wrapper for 'org-cite-insert'."
+  (org-cite-insert current-prefix-arg))
```
bdarcus, line 172: Maybe here should use `call-interactive`?

PR #405, opened 2021-11-12
fix(file): dedup default file parser results

### LINKED ISSUE #402: citar-open-library-files returns a list with duplicate files
With `citar-file-open-prompt`set to 't', citar-open-library-files opens a completing-read prompt with duplicate files.

Oddly, the number of duplicates is a factor of the number of directories set in `citar-library-paths`: when set to nil or a single directory, each file is duplicated once (2 files listed); when set to two directories, each file is duplicated twice (4 files listed); three directories, 6 files listed.

I think I've narrowed it down to some of the iterating done in `citar-file-parser-default`. But that's as far as I've gotten. 

A sample bibtex entry:
```
@article{Wilson2004,
  title = {Foucault on the "{{Question}} of the {{Author}}": A {{Critical Exegesis}}},
  shorttitle = {Foucault on the "{{Question}} of the {{Author}}"},
  author = {Wilson, Adrian},
  year = {2004},
  month = apr,
  journal = {The Modern Language Review},
  volume = {99},
  number = {2},
  pages = {339},
  issn = {00267937},
  doi = {10.2307/3738750},
  langid = {english},
  file = {/Users/localauthor/Zotero/storage/LZULTBNK/Wilson2004 - Foucault on the Question of the Author A Critic.pdf}
}
```

localauthor:
Fix #402

fixup! fix(file): dedup default file parser results

### DIFF
--- a/citar-file.el
+++ b/citar-file.el
@@ -120,15 +120,15 @@ Example: ':/path/to/test.pdf:PDF'."
            (file-field (citar-get-value
                         citar-file-variable entry))
            (results-file
-            (when file-field
-              (seq-mapcat
-               (lambda (func)
-                 (funcall
-                  func
-                  ;; Make sure this arg is non-nil.
-                  (or dirs "/")
-                  file-field))
-               citar-file-parser-functions))))
+            (delete-dups (when file-field
+                           (seq-mapcat
+                            (lambda (func)
+                              (funcall
+                               func
+                               ;; Make sure this arg is non-nil.
+                               (or dirs "/")
+                               file-field))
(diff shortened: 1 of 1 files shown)


bdarcus:
I guess that's what I get what I don't test a commit! Thanks.

bdarcus:
Though, shouldn't `delete-duplicates` go inside the `when`?

localauthor:
Right, I don't see why not, as a sort of catch-all.

bdarcus:
> Right, I don't see why not, as a sort of catch-all.

It's just more logically correct. 

Done.

PR #406, opened 2021-11-12
Add support for finding files with expanded naming convention

localauthor:
Partial implementation* of the `bibtex-completion-find-additional-pdfs` feature, from `bibtex-completion`.

Allows `citar-file` functions to find files that begin with a citekey and end with an extension listed in `citar-file-extensions`, with any amount (or kind) of text between.

Example: Citar currently finds only "Smith2012.pdf"; with this commit it will also find "Smith2012 - John Smith - Title of Example Book, Vol. 1 (2012).pdf."

*I say "partial implementation" because in `bibtex-completion` this feature is enabled by setting a defcustom to 't'. That's certainly an option, if desirable.

### DIFF
--- a/citar-file.el
+++ b/citar-file.el
@@ -70,13 +70,19 @@ will open, via `citar-file-open'."
   :type '(repeat string))
 
 (defcustom citar-file-find-additional-files nil
-  "If non-nil, all files whose base name starts with the reference key
-and whose extension is listed in `citar-file-extensions' are
-located by the functions `citar-open-library-files' and
-`citar-open-notes', not only files with the naming scheme
-\"<key>.<extension>\"."
+  "Find additional library files starting with reference key.
+If t, all files whose base name starts with the reference key and
+whose extension is listed in `citar-file-extensions' are located
+by the functions `citar-open-library-files' and
+`citar-open-notes'.  If nil, only files with the naming scheme
+\"<key>.<extension>\" are located.  Otherwise, its value is a
+regular expression specifying how the key is separated from the
+rest of the filename."
   :group 'citar
-  :type '(boolean))
+  :type '(choice (const :tag "Ignore additional files" nil)
+                 (const :tag "Find all files starting with key" t)
+                 (const :tag "Find files with space after key" "[[:space:]]")
+                 (regexp :tag "Filename separator")))
 
 (defvar citar-notes-paths)
 
@@ -118,8 +124,9 @@ Example: ':/path/to/test.pdf:PDF'."
 (defun citar-file--possible-names (key dirs extensions &optional entry find-additional)
   "Possible names for files correponding to KEY, ENTRY with EXTENSIONS in DIRS."
     (let* ((filematch (when find-additional
-                        (format "\\`%s.*\\.\\(?:%s\\)\\'"
+                        (format "\\`%s\\(?:%s.*\\)?\\.\\(?:%s\\)\\'"
                                 (regexp-quote key)
+                                (if (eq t find-additional) "" find-additional)
                                 (mapconcat #'regexp-quote extensions "\\|"))))
            (results-key (seq-mapcat
                          (lambda (dir)
@@ -185,7 +192,7 @@ with citekey as filename.
 Returns the filename whether or not the file exists, to support a
 function that will open a new file if the note is not present."
   (let* ((possible-files
-          (citar-file--possible-names key dirs extensions nil citar-file-find-additional-files))
+          (citar-file--possible-names key dirs extensions))
          (existing-files
           (seq-filter #'file-exists-p possible-files)))
     (if existing-files
--- a/citar-file.el
+++ b/citar-file.el
@@ -112,12 +112,15 @@ Example: ':/path/to/test.pdf:PDF'."
              (extension)
              (seq-map
               (lambda (directory)
-                (expand-file-name
-                 (concat key "." extension) directory))
+                (directory-files directory t
+                                         (s-concat "^" (regexp-quote key)
+                                                   ".*\\("
+                                                   extension
+                                                   "\\)$")))
               dirs)))
-    (let* ((results-key (seq-mapcat
-                         #'possible-file-names-with-extension
-                         extensions))
+    (let* ((results-key (apply #'append (seq-mapcat
+                                         #'possible-file-names-with-extension
+                                         extensions)))
            (file-field (citar-get-value
                         citar-file-variable entry))
            (results-file
--- a/citar-file.el
+++ b/citar-file.el
@@ -69,6 +69,15 @@ will open, via `citar-file-open'."
   :group 'citar
   :type '(repeat string))
 
+(defcustom citar-file-find-additional-files nil
+  "If non-nil, all files whose base name starts with the BibTeX key
+and whose extension is listed in `citar-file-extensions' are
+located by the functions `citar-open-library-files' and
+`citar-open-notes', not only files with the naming scheme
+\"<key>.<extension>\"."
+  :group 'citar
+  :type '(boolean))
+
 (defvar citar-notes-paths)
 
 ;;;; Convenience functions for files and paths
@@ -106,17 +115,20 @@ Example: ':/path/to/test.pdf:PDF'."
          (mapcar (apply-partially #'expand-file-name fn) dirs)))
      parts)))
 
-(defun citar-file--possible-names (key dirs extensions &optional entry)
+(defun citar-file--possible-names (key dirs extensions &optional entry find-additional)
   "Possible names for files correponding to KEY, ENTRY with EXTENSIONS in DIRS."
   (cl-flet ((possible-file-names-with-extension
              (extension)
              (seq-map
               (lambda (directory)
-                (directory-files directory t
-                                         (s-concat "^" (regexp-quote key)
-                                                   ".*\\("
-                                                   extension
-                                                   "\\)$")))
+                (if find-additional                    
+                    (directory-files directory t
+                                     (concat "^" (regexp-quote key)
+                                             ".*\\("
+                                             extension
+                                             "\\)$"))
+                  (list (expand-file-name
+                         (concat key "." extension) directory))))
               dirs)))
     (let* ((results-key (apply #'append (seq-mapcat
                                          #'possible-file-names-with-extension
@@ -124,21 +136,21 @@ Example: ':/path/to/test.pdf:PDF'."
            (file-field (citar-get-value
                         citar-file-variable entry))
            (results-file
-           (when file-field
-             (seq-mapcat
-              (lambda (func)
-                (funcall
-                 func
-                 ;; Make sure this arg is non-nil.
-                 (or dirs "/")
-                 file-field))
-              citar-file-parser-functions))))
+            (when file-field
+              (seq-mapcat
+               (lambda (func)
+                 (funcall
+                  func
+                  ;; Make sure this arg is non-nil.
+                  (or dirs "/")
+                  file-field))
+               citar-file-parser-functions))))
       (delete-dups (append results-key results-file)))))
 
 (defun citar-file--files-for-entry (key entry dirs extensions)
     "Find files related to KEY, ENTRY in DIRS with extension in EXTENSIONS."
     (seq-filter #'file-exists-p
-                (citar-file--possible-names key dirs extensions entry)))
+                (citar-file--possible-names key dirs extensions entry citar-file-find-additional-files)))
 
 (defun citar-file--files-for-multiple-entries (keys-entries dirs extensions)
   "Find files related to a list of KEYS-ENTRIES in DIRS with extension in EXTENSIONS."
@@ -175,7 +187,7 @@ with citekey as filename.
 Returns the filename whether or not the file exists, to support a
 function that will open a new file if the note is not present."
   (let* ((possible-files
-          (citar-file--possible-names key dirs extensions))
+          (citar-file--possible-names key dirs extensions nil citar-file-find-additional-files))
          (existing-files
           (seq-filter #'file-exists-p possible-files)))
     (if existing-files
--- a/citar-file.el
+++ b/citar-file.el
@@ -70,7 +70,7 @@ will open, via `citar-file-open'."
   :type '(repeat string))
 
 (defcustom citar-file-find-additional-files nil
-  "If non-nil, all files whose base name starts with the BibTeX key
+  "If non-nil, all files whose base name starts with the reference key
 and whose extension is listed in `citar-file-extensions' are
 located by the functions `citar-open-library-files' and
 `citar-open-notes', not only files with the naming scheme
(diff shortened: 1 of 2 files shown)


bdarcus:
Could you please review this if you time @roshanshariff?

It's a small commit.

citar-file.el
```diff
@@ -112,12 +112,15 @@ Example: ':/path/to/test.pdf:PDF'."
              (extension)
              (seq-map
               (lambda (directory)
-                (expand-file-name
-                 (concat key "." extension) directory))
+                (directory-files directory t
+                                         (s-concat "^" (regexp-quote key)
```
bdarcus, line 116:
```suggestion
                                         (concat "^" (regexp-quote key)
```

I want to avoid s unless needed.

bdarcus (commented):
Just a minor suggestion for now.

citar-file.el
```diff
@@ -112,12 +112,15 @@ Example: ':/path/to/test.pdf:PDF'."
...
+                                                   ".*\\("
+                                                   extension
+                                                   "\\)$")))
               dirs)))
-    (let* ((results-key (seq-mapcat
-                         #'possible-file-names-with-extension
-                         extensions))
+    (let* ((results-key (apply #'append (seq-mapcat
+                                         #'possible-file-names-with-extension
+                                         extensions)))
```
roshanshariff, line 123: Maybe instead of making a regexp per extension and looping, you could make a single regexp that matches all the extensions? Something like this:
```suggestion
    (let* ((results-key (seq-mapcat
                         (lambda (dir)
                           (directory-files
                            dir t
                            (format "\\`%s.*\\.\\(?:%s\\)\\'"
                                    (regexp-quote key)
                                    (mapconcat #'regexp-quote extensions "\\|"))))
                         dirs)))
```

bdarcus: WDYT about adding a defcustom for this?

I'm agnostic ATM.

localauthor: I didn’t see a reason for a defcustom in the bibtex-completion version, so I wasn’t compelled to propose one here.

roshanshariff: I'm not sure either. Maybe if someone has keys that are prefixes of other keys this could cause false positives? I would be inclined to add a defcustom if/when someone asks for it.

More likely, people might have other library organization schemes (directory per reference?) We can also think about how to accommodate those if somebody asks.

More importantly, there's the performance concern with performing a directory scan to check if a reference has any files. This could really slow down the "has:note" testing when compiling completion candidates. Bibtex-completion in fact [doesn't seem to do this scan.](https://github.com/tmalsburg/helm-bibtex/blob/aa775340ba691d2322948bfdc6a88158568a1399/bibtex-completion.el#L101)

bdarcus: > More importantly, there's the performance concern with performing a directory scan to check if a reference has any files. 

That was the main thing I was wondering about.

roshanshariff: I think the proper way to do this is to scan the relevant directories once (and add a file watch) and cache which references have files in them. Then use that cache when preparing the completion candidates. But this would need a bigger code change.

roshanshariff: Performing a potentially O(n^2) operation (in the number of entries) doesn't seem like a good idea. So we do need a defcustom to allow this, and we need to let-bind it to nil when compiling candidates.

bdarcus: I would think at minimum that function needs another optional arg so it can be turned off so that the performance-sensitive operations involved in generating a candidate list don't get bogged down.

roshanshariff: The code to do that is almost the same as adding the defcustom though, since you need to handle both cases. And the defcustom does allow people to turn off the feature if they're getting false positives. Would you be willing to implement this @localauthor?

localauthor: When I have some time I’ll give it a shot. Thanks for the review and suggestions.

localauthor: I tried to incorporate your changes, @roshanshariff , but could only get it working without them. Any thoughts/suggestions?

localauthor:
What do you think about the find-additional function also applying to notes? Any reason not to allow expanded naming convention for notes? I don't use the note functionality, so I don't know all the ins and outs there.

bdarcus:
> What do you think about the find-additional function also applying to notes? 

I don't like it :-)

> Any reason not to allow expanded naming convention for notes.

It would just add to the performance concerns and UI complexity, with unclear practical benefit.

To make explicit what may not clear ATM, the design of the notes system is primarily aimed at working well with org-roam (though without in any way requiring it), which it does.

Extending it this way may not even work there because IDs for bib notes are, in fact, the ID for the bib source (eg the key).

I do expect someone soon to ask for the opposite: being able to link to sub-file bib notes. I'm not sure if I'm willing to support that either, but that's another conversation.

citar-file.el
```diff
@@ -69,6 +69,15 @@ will open, via `citar-file-open'."
   :group 'citar
   :type '(repeat string))
 
+(defcustom citar-file-find-additional-files nil
+  "If non-nil, all files whose base name starts with the BibTeX key
```
bdarcus, line 73:
```suggestion
  "If non-nil, all files whose base name starts with the reference key
```
This package is no longer exclusively BibTeX.

citar-file.el
```diff
@@ -106,36 +115,42 @@ Example: ':/path/to/test.pdf:PDF'."
...
+                  (or dirs "/")
+                  file-field))
+               citar-file-parser-functions))))
       (delete-dups (append results-key results-file)))))
 
 (defun citar-file--files-for-entry (key entry dirs extensions)
     "Find files related to KEY, ENTRY in DIRS with extension in EXTENSIONS."
     (seq-filter #'file-exists-p
-                (citar-file--possible-names key dirs extensions entry)))
+                (citar-file--possible-names key dirs extensions entry citar-file-find-additional-files)))
```
bdarcus, line 153: If the defcustom is set to `t`, will this not run under `[citar--format-candidates](https://github.com/bdarcus/citar/blob/6e73acd559a1d60c74af58e1794f4f1e389e5df9/citar.el#L443)`?

See in particular:

https://github.com/bdarcus/citar/blob/6e73acd559a1d60c74af58e1794f4f1e389e5df9/citar.el#L457-L462

I guess we should clarify what the defcustom does.

My impulse is to say only turns on file finding for the open commands, but is not used in candidate generation.

I suppose ideally we could do some profiling to figure out how significant the performance concern is. The test would be a large file (say 2000+ entries), and multiple directories.

localauthor:
Away from keyboard right now so I can’t check, but doesn’t this need to run under `citar—format-candidates` in order for the prefix section of the candidate list to accurately show if there’s a file or not for a given candidate, ie, to have the `F` symbol appear?

It seems that this defcustom could be bypassed in `citar-file—files-for-entry` with a conditional, but how else to get accurate prefix symbols for candidates with files?
On Nov 13, 2021, 15:03 +0200, bdarcus ***@***.***>, wrote:
	@bdarcus commented on this pull request.
In citar-file.el:
>        (delete-dups (append results-key results-file)))))

(defun citar-file--files-for-entry (key entry dirs extensions)
	    "Find files related to KEY, ENTRY in DIRS with extension in EXTENSIONS."
    (seq-filter #'file-exists-p
-                (citar-file--possible-names key dirs extensions entry)))
+                (citar-file--possible-names key dirs extensions entry citar-file-find-additional-files)))
If the defcustom is set to t, will this not run under [citar--format-candidates](https://github.com/bdarcus/citar/blob/6e73acd559a1d60c74af58e1794f4f1e389e5df9/citar.el#L443)?
See in particular:
https://github.com/bdarcus/citar/blob/6e73acd559a1d60c74af58e1794f4f1e389e5df9/citar.el#L457-L462
I guess we should clarify what the defcustom does.
My impulse is to say only turns on file finding for the open commands, but is not used in candidate generation.
I suppose ideally we could do some profiling to figure out how significant the performance concern is. The test would be a large file (say 2000+ entries), and multiple directories.
—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

citar-file.el
```diff
@@ -106,36 +115,42 @@ Example: ':/path/to/test.pdf:PDF'."
          (mapcar (apply-partially #'expand-file-name fn) dirs)))
      parts)))
 
-(defun citar-file--possible-names (key dirs extensions &optional entry)
+(defun citar-file--possible-names (key dirs extensions &optional entry find-additional)
```
roshanshariff, line 118:
```suggestion
(defun citar-file--possible-names (key dirs extensions &optional entry)
```
I'm leaning towards just checking the `defcustom` inside this function to see whether to do the directory scan. Otherwise the function arguments list seems to get unwieldy for no real benefit.

Then you would just make the following change on line 468 in `citar.el` (in `citar--format-candidates`)
```elisp
       (let* ((citar-find-additional-files nil) ; add this line
              (files
               (when (citar-file--files-for-entry
```

citar-file.el
```diff
@@ -172,7 +187,7 @@ with citekey as filename.
 Returns the filename whether or not the file exists, to support a
 function that will open a new file if the note is not present."
   (let* ((possible-files
-          (citar-file--possible-names key dirs extensions))
+          (citar-file--possible-names key dirs extensions nil citar-file-find-additional-files))
```
roshanshariff, line 190:
```suggestion
          (citar-file--possible-names key dirs extensions))
```
See my first comment above; it seems simpler to not add more function arguments up the chain and instead let-bind the defcustom in `citar--format-candidates`.

roshanshariff (changes requested):
I've suggested the following changes
* Removing the extra function arguments, and instead let-binding `citar-file-find-additional-files` to `nil` inside `citar--format-candidates`
* Making a single regexp that'll match any of the given extensions, so you don't have to do multiple directory scans for each key.

bdarcus:
> It seems that this defcustom could be bypassed in `citar-file—files-for-entry` with a conditional, but how else to get accurate prefix symbols for candidates with files?

I think this is why we need to run some benchmarks. 

Here's what I get with the current main branch code, with a fairly small ~250 entry bib file, and a single library directory.

```elisp
ELISP> (benchmark-run 100 (citar--format-candidates citar-bibliography))
(13.547878455 48 3.15389796)

ELISP> (benchmark-run 10 (citar--format-candidates citar-bibliography))
(1.326038574 4 0.2686027160000002)
```

I'd like to see how it changes with this branch.

BTW, I was testing possible names, and surprised by this (only the first two should be returned), though it makes sense the regex might be tricky and hard to generalize.

```elisp
ELISP> (citar-file--possible-names "huck1971" '("~/org/pdf/" "/" "~/" "/tmp") citar-file-extensions nil t)
("/home/bruce/org/pdf/huck1971-test.pdf" "/home/bruce/org/pdf/huck1971.pdf" "/home/bruce/org/pdf/huck1971a.pdf" "/home/bruce/org/pdf/huck1971b.pdf" "/home/bruce/org/pdf/huck1971c.pdf" "/home/bruce/org/pdf/huck1971d.pdf" "/home/bruce/org/pdf/huck1971e.pdf")
```

roshanshariff:
> the right boundary needs to be from among some list of characters, that would not conflict with keys.

Unfortunately, the set of characters not allowed in keys is [apparently very small](https://tex.stackexchange.com/questions/408530/what-characters-are-allowed-to-use-as-delimiters-for-bibtex-keys).

Also, ignore what I said in my review about not having extra function arguments. Doing it that way would apply the same logic to both files and notes. Instead, let's keep the extra argument for `citar-file--possible-names` and pass in the value of `citar-file-find-additional-files` (as you have) but also let-bind that value to nil in `citar--format-candidates` (as I suggested in my review). I hope that makes sense :)

bdarcus:
> > the right boundary needs to be from among some list of characters, that would not conflict with keys.
> 
> Unfortunately, the set of characters not allowed in keys is [apparently very small](https://tex.stackexchange.com/questions/408530/what-characters-are-allowed-to-use-as-delimiters-for-bibtex-keys).

Maybe `~` is reasonable?

roshanshariff:
> Maybe ~ is reasonable?

Unfortunately, '~` in filenames are at risk of being interpreted as your home directory on *nix. (Not always, but it's a gamble to use them anyway).

How's this for an imperfect but maybe good-enough proposal? Spaces are definitely a bad idea in bibtex keys, but perfectly acceptable in filenames. So we require a space after the key and the rest of the filename. So the filename -> key associations would be
* "huck1971-test.pdf" -> "huck1971-test"
* "huck1971a.pdf" -> "huck1971a"
* "Smith2012 - John Smith - Title of Example Book, Vol. 1 (2012).pdf." -> "Smith2012"

If we think this is acceptable, the regexp format string would become (untested)
```elisp
"\\`%s\\(?:[[:space:]].*\\)?\\.\\(?:%s\\)\\'"
```

bdarcus:
> How's this for an imperfect but maybe good-enough proposal? Spaces are definitely a bad idea in bibtex keys, but perfectly acceptable in filenames. So we require a space after the key and the rest of the filename.

While I have what is in effect an aesthetic aversion to spaces in filenames, this does seem the only likely practical solution.

We could also make it configurable; so rather than a boolean on the defcustom, allow a choice of "space", "hyphen", or nil?

But maybe that's overkill at this point.

roshanshariff:
How about something like this?
```elisp
(defcustom citar-file-find-additional-files nil
  "Find additional library files starting with reference key.
If t, all files whose base name starts with the reference key and
whose extension is listed in `citar-file-extensions' are located
by the functions `citar-open-library-files' and
`citar-open-notes'.  If nil, only files with the naming scheme
\"<key>.<extension>\" are located.  Otherwise, its value is a
regular expression specifying how the key is separated from the
rest of the filename."
  :group 'citar
  :type '(choice (const :tag "Ignore additional files" nil)
                 (const :tag "Find files with space after key" " ")
                 (regexp :tag "Filename separator" nil)))
```
and `filematch` defined as follows:
```elisp
(when find-additional
  (format "\\`%s\\(?:%s.*\\)?\\.\\(?:%s\\)\\'"
          (regexp-quote key)
          (if (eq t find-additional) "" find-additional)
          (mapconcat #'regexp-quote extensions "\\|")))
```
This gives us all the configurability one could want, with no extra configuration variables and very little extra code. Setting the variable to t is the same as setting it to "", i.e. it'll accept any/no suffix after the key name. If you want it to be space-separated, set it to " ", or whatever else you like.

bdarcus:
> This gives us all the configurability one could want, with no extra configuration variables and very little extra code.

We'd probably want to include a default regex (drfconst?) if going this way, so easy for users to select that?

roshanshariff:
If you can think of another regex that's likely to be broadly useful for users, why not just add it as an option in `choice` like I did with "[[:space:]]"? That way we don't have another variable/constant that is just a default value for this variable. For example,
```elisp
(const :tag "Find files with hyphen after key" "-")
```

bdarcus:
That's OK; I think I just misunderstood the code.

bdarcus:
@localauthor - is this RTM from your POV?

I can test it again tomorrow.

Would be nice to have something added on this to the README.

localauthor:
I think it's ready to merge, yes --- except for adding something to the README. Perhaps after merging?

There still remains the issue of having accurate prefixes in the candidate list, per the thread above: https://github.com/bdarcus/citar/pull/406#issuecomment-968086462

It comes down to this line, where is `citar-file-find-additional-files` is set to `nil` before the candidates are propertized: https://github.com/localauthor/citar/blob/4536f49a13a5c45287b65b35518c100a2e836fbb/citar.el#L456

Not sure how to mitigate any performance issues attendant to this, but maybe that's outside the scope of this PR?

bdarcus:
OK, merged; can refine on subsequent commits.

PR #413, opened 2021-11-16
fix(file): remove filter on possible-names

### LINKED ISSUE #412: citar-file--possible-names broke
Something isn't working correctly with this function now, which has cascading impacts (for example, open-notes now doesn't work correctly). It now only returns existing files, but should always return a list of paths.

```elisp
ELISP> (citar-file--possible-names "low2001" citar-notes-paths '("org"))
("/home/bruce/org/roam/biblio/low2001.org")
ELISP> (citar-file--possible-names "low200X1" citar-notes-paths '("org"))
nil
```

cc @localauthor @roshanshariff

bdarcus:
Fix #412.

Also, autoload citar-org-open-notes-default.

-----
@roshanshariff @localauthor - can you two confirm this change is fine?

Basically, the function change (suggested by @roshanshariff?) filtered the list of possible names to ensure the files were present, which is not the purpose of it (checking if a file is present should happen elsewhere). 

So I simply removed the filter.

But I don't now how this will interact with the new extended search functionality. Will this cause a problem with open-notes, for example, when creating a new note?

EDIT: seems not:

```elisp
ELISP> (let ((citar-file-find-additional-files nil))(citar-file--get-note-filename "t1" citar-notes-paths '("org")))
(("/home/bruce/org/roam/biblio/t1.org" . new))

ELISP> (let ((citar-file-find-additional-files t))(citar-file--get-note-filename "t1" citar-notes-paths '("org")))
(("/home/bruce/org/roam/biblio/t1.org" . new))
```

### DIFF
--- a/citar-file.el
+++ b/citar-file.el
@@ -132,11 +132,9 @@ Example: ':/path/to/test.pdf:PDF'."
                          (lambda (dir)
                            (if filematch
                                (directory-files dir t filematch)
-                             (seq-filter
-                              #'file-exists-p
-                              (seq-map (lambda (ext)
-                                         (expand-file-name (concat key "." ext) dir))
-                                       extensions))))
+                             (seq-map (lambda (ext)
+                                        (expand-file-name (concat key "." ext) dir))
+                                      extensions)))
                          dirs))
            (file-field (citar-get-value
                         citar-file-variable entry))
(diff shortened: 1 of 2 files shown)


bdarcus:
I merged this, as fixing an obvious bug seems the first priority, and I don't see any negative consequences ATM.

PR #423, opened 2021-11-20
Add 'citar-attach-library-files'

### LINKED ISSUE #422: Attach resources to outgoing MIME message
I often need to share papers with colleagues. One of the killer features of my previous tool was the ability to attach papers directly from my library using the completion mechanism.

At it's simplest, I'd like a command like `citar-attach-library-files` which allows me to attach papers to an outgoing MIME message. A few other "nice" features would be cool but not crucial: `citar-attach-note-files` to attach associated notes and/or `citar-attach` to generically allow the attachment of either. Another "nice" feature would be to automatically open a message buffer when this command is invoked elsewhere. However, I think 95% of the use cases would be covered with `citar-attach-library-files`.

I tried to implement this myself with my limited experience. You can see I just directly invoke `mml-attach-file`.

```Emacs Lisp
;;;###autoload
(defun citar-attach-library-files (keys-entries)
  "Attach library files associated with KEYS-ENTRIES to the outgoing MIME message.

  (let ((files
          keys-entries
	  citar-library-paths
	  citar-file-extensions)))
    (if (and citar-file-open-prompt
	     (> (length files) 1))
	(let ((selected-files
	       (citar-select-files files)))
	  (dolist (file selected-files)
	    (mml-attach-file file)))
      (dolist (file files)
	(mml-attach-file file)))))
```

bdarcus:
This replaces the 'citar-add-pdf-attachment' wrapper with a new function.

Close #422.

------

@xcapaldi - I'll take a closer look tomorrow before merging. 

If you'd like me to credit you on the commit, I need the email address to use.

### DIFF
--- a/citar-file.el
+++ b/citar-file.el
@@ -69,7 +69,7 @@ will open, via `citar-file-open'."
   :group 'citar
   :type '(repeat string))
 
-(defcustom citar-note-extensions '("org" "md")
+(defcustom citar-file-note-extensions '("org" "md")
   "List of file extensions to recognize for notes.
 
 These are the extensions the 'citar-open-note-function'
--- a/citar.el
+++ b/citar.el
@@ -57,8 +57,10 @@
 (defvar embark-meta-map)
 (defvar citar-org-open-note-function)
 (defvar citar-file-extensions)
+(defvar citar-file-note-extensions)
 (defvar citar-file-open-prompt)
 (defvar citar-file-variable)
+(defvar citar-file-find-additional-files)
 
 ;;; Variables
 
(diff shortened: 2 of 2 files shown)


xcapaldi:
> This replaces the 'citar-add-pdf-attachment' wrapper with a new function.
> 
> Close #422.
> 
> @xcapaldi - I'll take a closer look tomorrow before merging.
> 
> If you'd like me to credit you on the commit, I need the email address to use.

Sure, my email is xcapaldi@scribo.biz

bdarcus:
Give how similar this attach function is to the open, I split out the logic into a more general internal function here:

https://github.com/bdarcus/citar/pull/423/commits/a5ac3d286f0ee337a2b241177efe8172fd722742

Question, @xcapaldi - I don't use emacs for email (yet). How is this supposed to work? From within a mail mode?

If yes, is it a bit weird for it to be available anywhere?

xcapaldi:
@bdarcus Yeah, I agree it is a bit weird that this works anywhere but `mml-attach-file` actually can insert the proper mml attachment format into any buffer since its just a formatted text block something like this:

```
<#part type="application/octet-stream" filename="~/Dropbox/library/references.bib" disposition=attachment>
<#/part>
```
I think the ideal usage would be that if the attachment is invoked via citar, it either attaches directly to the open message buffer, or if there is no message buffer open, it opens a new message buffer. But again, to be honest, I'm not sure how many people use email in emacs or would use this functionality. The way you've implemented it already is perfectly usable as I never invoke the function outside a message buffer anyway.

PR #425, opened 2021-11-20
Updates for test script

mcfarljm:
A few updates for the test script, including some that were needed to get it to run:
- Install other package dependencies: `parsebib` and `s`
- Update `load-path` so that requires in `citar.el` are handled
- Update location of tmp directly in message
- Add `embark-act` binding (facilitate testing embark with citar)

### DIFF
--- a/test/install.el
+++ b/test/install.el
@@ -27,6 +27,9 @@
 (package-install 'embark)
 (package-install 'marginalia)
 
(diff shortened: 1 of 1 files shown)


bdarcus:
Thanks; I've neglected this!

bdarcus:
Note to self: should probably change selectrum to vertico.

Issue #203, opened 2021-08-01
Clarify distinction between actions on keys vs on citations (lists of keys)


bdarcus:
Actually, as I'm working on this, I have a related larger UX question:

Right now, at-point functionality supports both a single reference, and a list of references.

In practical terms, that means if point is on an org-cite citation (within "cite" for example), and you run any action, it will pass the list of keys.

But in other org-cite implementations, it will ask you to choose which reference key to act on.

I can see arguments for both approaches.

I can also see arguments for it working as a list for some commands, and not for others.

WDYT @oatmealm @ilupin @titaniumbones @mclearc @salutis

Here's my thoughts; let's consider three cases:

1. MUST allow a list
2. Probably SHOULD allow a list
3. Probably SHOULD NOT allow a list

Here's what it's in the keymap:

- `org-cite-insert`: 1
- `bibtex-actions-open`: 2
- `bibtex-actions-open-entry`: 3 (because usually the entries are in the same file)
- `bibtex-actions-open-link`: 2
- `bibtex-actions-open-notes`: 2 (this is complicated though, since notes can be single-file, or multi)
- `bibtex-actions-open-pdf`: 2

Note: I plan to likely add other actions from `org-ref-cite`.

_Originally posted by @bdarcus in https://github.com/bdarcus/bibtex-actions/pull/199#issuecomment-890523820_

bdarcus:
Here's the commit that implements 1 above.

https://github.com/bdarcus/bibtex-actions/pull/199/commits/da414dfa8560abcc37b1ca68b96db80d3d689b64

salutis:
I wonder, when will `org-cite` come to Emacs' `master` branch?

bdarcus:
It's already there ;-)

https://blog.tecosaur.com/tmio/2021-07-31-citations.html

mclearc:
I haven't any really strong feelings about this, since my workflow doesn't really depend on it. I would think there is some benefit to changing to 2, since (i) simplification of code is almost always a net good, and (ii) acting more like `org-ref-cite` is also a plus, since if Kitchin is already doing this, the principle of least surprise would encourage some alignment here.

salutis:
> It's already there ;-)
> 
> https://blog.tecosaur.com/tmio/2021-07-31-citations.html

BRB, `git pull && make && make install` 😄

salutis:
@bdarcus

[...]

Issue #190, opened 2021-07-25
oc: Revise at-point behavior


bdarcus:
When point is on a `citation-reference`, default command should run.

When point is on `citation`, should open bibtex-actions, perhaps "pre-selecting" the candidates (though need to think again about how best to do the pre-selection in completing-read).

cc @ilupin

Issue #385, opened 2021-11-07
latex.markdown: Add functions to shift references


bdarcus:
As a possible companion to #383, it might be nice to implement the same sort of keymap-based citation-reference shifting as is present in the org adapter (and adapted from org-ref-cite).

It may not be as easy as in org though.

IF it made sense, perhaps the current functions in `citar-org` should be generalized?

PR #426, opened 2021-11-21
Context-sensitive insertion and edit commands

### LINKED ISSUE #278: bibtex-actions-insert-citation vs org-cite-insert

### LINKED ISSUE #387: make citar-org support embark-dwim
currently if I run `embark-dwim` on org citation, I got
```
  nil("peeters_cross-language_2021")
  #f(compiled-function () #<bytecode 0x1a7da90d012d516>)()
  embark--quit-and-run(#f(compiled-function () #<bytecode 0x1a7da90d012d516>))
  embark--act(nil (:orig-type oc-citation :orig-target "peeters_cross-language_2021" :bounds nil :type oc-citation :target "peeters_cross-language_2021") t)
  embark-dwim(nil)
  #<subr funcall-interactively>(embark-dwim nil)
  apply(#<subr funcall-interactively> (embark-dwim nil))
  funcall-interactively(embark-dwim nil)
  #<subr call-interactively>(embark-dwim nil nil)
  apply(#<subr call-interactively> (embark-dwim nil nil))
  explain-pause--wrap-call-interactively(#<subr call-interactively> embark-dwim nil nil)
  apply(explain-pause--wrap-call-interactively #<subr call-interactively> (embark-dwim nil nil))
  call-interactively(embark-dwim nil nil)
  command-execute(embark-dwim)
```
should we add default action(ENTER) in citar-org-map?

### LINKED ISSUE #393: Why not making 'citar-insert-citation' also works in org mode
I use latex, org for my writing and notes, and sometimes in markdown. So I would like to use 'citar-insert-citation' command in all these three scenarios. Also, it would be nice that the package can be lazy-loaded, triggered by 'citar-insert-citation', instead of loading after org mode.

Great package! Thanks for all you guys' contribution.

roshanshariff:
This is a continuation of #394.

- Context-aware insert-citation
  - [x] Org mode
  - [x] Latex (already implemented)
  - [x] Markdown
- Context-aware insert/edit function
  - [x] Org mode (calls `org-cite-insert`)
  - [x] Latex (calls `citar-latex-insert-citation`)
  - [x] Markdown (calls `citar-markdown-insert-citation`)
- Target finders for keys and citations
  - [x] Org mode
  - [x] Latex
  - [x] Markdown (based on #395)
- [x] Remove Org-specific Embark target (#381)
- [x] Update documentation

Partly implements #383: in Markdown, `citar-insert-citation` will now add new keys after the current citation key, or at the beginning of the citation if point is just after `[`.

Fixes #278, fixes #393: `citar-insert-citation` now does something useful in Org mode, either inserting a new org-cite citation or adding new keys to an existing citation at point. The new keys are added at the beginning of the citation if point is before the first key, or after the current key otherwise.

Relevant to #203: there are now two target finders, one for individual keys and the other for full citations. Both target types have the same keymap, offering the usual open actions. Running `embark-act` once will select and highlight the single key at point, and then running `embark-cycle` (or pressing the key bound to `embark-act` again) will select and highlight the entire citation. Depending on which target is active, the open commands will either act on a single key or all the keys in the citation.

Relevant to #190: the new `citar-insert-edit` action now runs the `insert-edit` major mode function, which is currently a thin wrapper around `org-cite-insert` in Org mode. TBD whether this behaviour should be changed.

Fixes #387.

### DIFF
--- a/citar-latex.el
+++ b/citar-latex.el
@@ -72,11 +72,6 @@ the point."
   (when (citar-latex-is-a-cite-command (TeX-current-macro))
     (split-string (thing-at-point 'list t) "," t "[{} ]+")))
 
-;;;###autoload
-(defun citar-latex-insert-keys (keys)
-  "Insert comma sperated KEYS in a latex buffer."
-  (insert (string-join keys ", ")))
-
 (defvar citar-latex-cite-command-history nil
   "Variable for history of cite commands.")
 
@@ -109,7 +104,7 @@ inserted."
         (TeX-parse-macro macro
                          (when citar-latex-prompt-for-extra-arguments
                            (cdr (citar-latex-is-a-cite-command macro))))))
-    (citar-latex-insert-keys keys)
+    (citar--insert-keys-comma-separated keys)
     (skip-chars-forward "^}") (forward-char 1)))
 
 (defun citar-latex-is-a-cite-command (command)
--- a/citar-latex.el
+++ b/citar-latex.el
@@ -34,7 +34,7 @@
 (defvar citar-major-mode-functions)
 
 (defcustom citar-latex-cite-commands
-  '((("cite" "Cite" "citet" "Citet" "parencite"
+  '((("cite" "Cite" "citet" "Citet" "citep" "Citep" "parencite"
       "Parencite" "footcite" "footcitetext" "textcite" "Textcite"
       "smartcite" "Smartcite" "cite*" "parencite*" "autocite"
       "Autocite" "autocite*" "Autocite*" "citeauthor" "Citeauthor"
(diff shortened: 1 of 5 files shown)


citar-latex.el
```diff
@@ -65,17 +65,58 @@ the point."
...
-  "Return a list of keys at point in a latex buffer."
-  (unless (fboundp 'TeX-current-macro)
-    (error "Please install AUCTeX"))
-  (when (citar-latex-is-a-cite-command (TeX-current-macro))
-    (split-string (thing-at-point 'list t) "," t "[{} ]+")))
+(defun citar-latex-key-at-point ()
+  "Return citation key at point with its bounds.
+The return value is (KEY . BOUNDS), where KEY is the citation key
+at point and BOUNDS is a pair of buffer positions.  Returns nil
+if there is no key at point."
```
bdarcus, line 72:
```suggestion
  "Return citation key at point with its bounds.
  
The return value is (KEY . BOUNDS), where KEY is the citation key
at point and BOUNDS is a pair of buffer positions.  

Return nil if there is no key at point."
```

I'm just suggesting some trivial docstring formatting etc changes.

But substantive questions here:

1. Why did this need to get more complex?
2. How does a user change an existing key?

citar-latex.el
```diff
@@ -65,17 +65,58 @@ the point."
...
-(defun citar-latex-insert-keys (keys)
-  "Insert comma sperated KEYS in a latex buffer."
-  (insert (string-join keys ", ")))
+(defun citar-latex-citation-at-point ()
+  "Find citation macro at point and extract keys.
+Finds brace-delimited strings inside the bounds of the macro,
+splits them at comma characters, and trims whitespace.
+Returns (KEYS . BOUNDS), where KEYS is a list of the found
+citation keys and BOUNDS is a pair of buffer positions indicating
+the start and end of the citation macro."
```
bdarcus, line 94:
```suggestion
(defun citar-latex-citation-at-point ()
  "Find citation macro at point and extract keys.
  
Find brace-delimited strings inside the bounds of the macro,
splits them at comma characters, and trims whitespace.

Return (KEYS . BOUNDS), where KEYS is a list of the found
citation keys and BOUNDS is a pair of buffer positions indicating
the start and end of the citation macro."
```

Changed here to use the imperative consistently "return rather than returns".

citar-latex.el
```diff
@@ -65,17 +65,58 @@ the point."
...
+          (push (split-string (match-string-no-properties 1) "," t "[[:space:]]*")
+                keylists))
+        (cons (apply #'append (nreverse keylists))
+              bounds)))))
+
+(defun citar-latex--macro-bounds ()
+  "Return the bounds of the citation macro at point.
+Returns a pair of buffer positions indicating the beginning and
+end of the enclosing citation macro, or nil if point is not
+inside a citation macro."
```
bdarcus, line 109:
```suggestion
  "Return the bounds of the citation macro at point.
  
Return a pair of buffer positions indicating the beginning and
end of the enclosing citation macro, or nil if point is not
inside a citation macro."
```

I'm not sure on this one on the whitespace. WDYT?

citar-markdown.el
```diff
@@ -39,37 +40,86 @@
...
-                    (mapconcat (lambda (k) (concat "@" k)) keys "; ")
-                    postnote))))
-
-(defconst citar-markdown-regex-citation-key
-  "\\(-?@\\([[:alnum:]_][[:alnum:]_:.#$%&+?<>~/-]*\\)\\)"
-  ;; borrowed from pandoc-mode
+  "Insert a pandoc-style citation consisting of KEYS.
+If the point is inside a citation, add new keys after the current
+key.  If point is immediately after the opening \[, add new keys
+to the beginning of the citation."
```
bdarcus, line 51:
```suggestion
  "Insert a pandoc-style citation consisting of KEYS.
  
If the point is inside a citation, add new keys after the current
key.  

If point is immediately after the opening \[, add new keys
to the beginning of the citation."
```

Also, same substantive question as with latex: how does a user edit an existing key?

bdarcus (commented):
Nice work @roshanshariff!

I just added a few minor line comments/questions, and suggestions which are mostly about docstring formatting.

EDIT: I committed the suggestions and merged. You can followup on the questions.

PS - Perhaps worth mentioning: 

I've gone back-and-forth about whether it's the right approach for citation insertion and editing, but the solid at-point functionality might provide a foundation on which to add capf functionality in the future if it makes sense.

citar-latex.el
roshanshariff: > Why did this function need to get more complex? Just curious.

The previous version had false positives with cases like this:
```latex
\citet[see][p.~99]{smith90}
```
If point was in the affixes, it would consider "see" or "p.~99" to be citation keys. In the new code, a citation key is a comma-delimited item inside a pair of curly braces within the citation macro. I checked that this holds for the Natbib and Biblatex macros. Now, if point is on the affixes, `key-at-point` will return nil and Embark will fall back to acting on the entire macro.

There is still a corner case: in the Biblatex "multi-volume" commands, the volume number also appears within braces. I figured we could deal with this later when we revisit more complicated citation macros.

This code can probably be simplified a little using `syntax-ppss`, similar to the Markdown citation parser, but it wasn't immediately clear to me how to do that.

roshanshariff:
The `insert-edit` functions for Markdown and Latex are pretty much the minimum viable implementations, and they don't let you change the citation macro or delete/replace keys. The next step is to make them more useful: either
1. try to imitate the logic of `org-cite-make-insert-processor`, or
2. design what we think the correct `insert-edit` behaviour should be and use it everywhere.

If we choose (2), we can also use it in Org mode; insert processors don't have to be made using `org-cite-make-insert-processor`. Either way, we should have a coherent story for all the [editing uses cases](https://github.com/bdarcus/citar/pull/394#issuecomment-967096778) (ref. #385).

bdarcus:
> insert processors don't have to be made using `org-cite-make-insert-processor`.

Correct. I asked the developer to add that because didn't want to figure out to do all that coding myself :-)

Also possible: if we find some good tweaks, contribute them to org. I'd started to submit small patches along those lines.

PR #428, opened 2021-11-21
Minor edits

localauthor:
Combining these commits, for convenience.

The first two seem uncontroversial: remove a duplicate function and include `citar-file-note-extensions` in `citar-open`. 

The third and fourth might be handled differently:

The third adds a "No associated file" message to `citar--library-files-action`, for cases where there is no associated file.

The fourth adds a "Key not found" message to `citar-select-ref/s`, which is returned if one of the many citar-open functions (open, library-files, notes, link, entry) is invoked at-point on a key that is malformed or simply not in the bibliography. Currently in this case these functions return an unhelpful error ("Wrong type argument: stringp, nil") because `citar-select-refs` has passed them `nil`.

### DIFF
--- a/citar-filenotify.el
+++ b/citar-filenotify.el
@@ -76,7 +76,7 @@ The callback is called without it when
 If it is other than 'global or 'local invalidate both"
   (unless (eq 'local scope)
     (setq citar--candidates-cache 'uninitialized))
-  (unless (eq 'glocal scope)
+  (unless (eq 'global scope)
     (setq citar--local-candidates-cache 'uninitialized)))
 
 (defun citar-filenotify--make-default-callback (func scope &optional change)
--- a/citar.el
+++ b/citar.el
@@ -302,28 +302,29 @@ Includes the following optional argument:
 
 'REBUILD-CACHE' if t, forces rebuilding the cache before
 offering the selection candidates."
-  (let* ((crm-separator "\\s-*&\\s-*")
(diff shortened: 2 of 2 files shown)


bdarcus:
I think your if-let changes need adjustment in indentation in a few places?

localauthor:
Maybe that's better? Not sure if I've handled it correctly...

bdarcus:
I think so.

BTW, I pushed a commit that removes the duplicate defcustom, as I wanted to avoid having that merged in a doom PR that may be merged today.

bdarcus:
If you have a few moments, can you review @roshanshariff?

bdarcus:
Can you drop the defcustom commit @localauthor, so I can do a rebase merge with the others?

roshanshariff:
Looks good to me! While you're doing this, would you mind adding another small typo fix? In `citar-filenotify.el`, in the function `citar-filenotify--invalidate-cache`, `'global` is misspelled as `'glocal`. This will save me from having to roll this into another PR.

bdarcus:
> Looks good to me! While you're doing this, would you mind adding another small typo fix? In `citar-filenotify.el`, in the function `citar-filenotify--invalidate-cache`, `'global` is misspelled as `'glocal`. This will save me from having to roll this into another PR.

Done.

I wanted to get this merged today, as the Doom PR is likely to be merged in the next few hours.

PR #434, opened 2021-11-23
feat(org): add separate format-note function

### LINKED ISSUE #420: Add "cursor here" marker to note templates
I would like to specify where to put the cursor when creating new notes.

Consider the following `note` template:

```
#+title: ${title}
#+subtitle: Bibliographic Notes
#+author: Rudolf Adamkovič
#+email: salutis@me.com
#+latex_header: \setlength{\parindent}{4em}
#+latex_header: \setlength{\parskip}{1em}
#+property: header-args+ :comments link
#+cite_export: csl apa.csl

| ← PUT THE CURSOR HERE, INSTEAD OF TWO LINES BELOW "print_bibliography"

* References

#+print_bibliography:
```

bdarcus:
Enhance note functionality.

- Split open-note into open and format functions defcustoms
- Simplify and enhance default org note format function

Also fix #420.

-------------------

Remaining questions/issues:

1. [X] hook up defcustom
2. [X] should the note template get its own defcustom? I think not ATM.
3. [X] do these changes break org-roam-bibtex integration? If yes (almost certainly the case), fix.
3. [X] need to adjust open-note
2. [X] should I move the remaining mode-specific details (like `#+title`) from template to function? I tend to think not.
3. [X] doc update

On org-roam-bibtex, see the key [orb-edit-note function](https://github.com/org-roam/org-roam-bibtex/blob/d65b70e9d19efc5001e01a61c36cc3e57e198131/org-roam-bibtex.el#L584), which [orb-citar-edit-note](https://github.com/org-roam/org-roam-bibtex/blob/d65b70e9d19efc5001e01a61c36cc3e57e198131/org-roam-bibtex.el#L1035) wraps. The problem I see ATM is the former in fact does both open and edit.

### DIFF
--- a/citar-org.el
+++ b/citar-org.el
@@ -45,6 +45,10 @@
 (declare-function org-cite-make-insert-processor "oc")
 (declare-function org-cite-get-references "oc")
 (declare-function embark-act "ext:embark")
+(declare-function evil-insert "ext:evil")
+(declare-function org-roam-ref-add "ext:org-roam-node")
+(defvar citar-open-note-function)
+(defvar citar-file-open-function)
 (defvar embark-target-finders)
 (defvar embark-keymap-alist)
 (defvar embark-pre-action-hooks)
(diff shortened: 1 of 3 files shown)


roshanshariff:
I also looked into `insert-skeleton` and it should be pretty easy to expose the bibliography fields as variables by let-binding `skeleton-further-elements`. Skeletons also let you specify where point should be left after insertion, and support including arbitrary other information. `autoinsert.el` has lots of examples of using skeletons to pre-populate various kinds of files.

I like the idea of a more powerful templating system, but not so much having to use that same system for mode-specific metadata like org IDs or front matter. Instead, I think users should use templates to define the actual content they want in the note, without having to also include the text representing (for example) the properties drawer containing an org ID.

Instead, ideally, we would use `org-id-get` to handle the creation of a new ID in org-mode, or similar functionality in other modes. That way we decouple that mode-specific functionality from Citar (so don't have to generate UUIDs, for example). I can work on implementing this when I have some time.

bdarcus:
> I like the idea of a more powerful templating system, but not so much having to use that same system for mode-specific metadata like org IDs or front matter. Instead, I think users should use templates to define the actual content they want in the note, without having to also include the text representing (for example) the properties drawer containing an org ID.

I agree. I just can't see how to do that cleanly, in part because of the overlap of "content" and mode-specific markup.

I'll update this later with the example I'm thinking of.

roshanshariff:
I've been experimenting with `skeleton-insert` and it beautifully handles that use case and others too. In fact, with the help of a utility function it can do things like inserting org IDs and Roam refs, if appropriate. I'll submit a PR in a day or two.

bdarcus:
OK, I'll wait on that to figure out what to do with this.

citar-org.el
```diff
@@ -254,27 +256,46 @@ strings by style."
...
+      (and (derived-mode-p 'org-mode)
+           (fboundp 'org-roam-buffer-p)
+           (org-roam-buffer-p))
+      (ignore-errors (citar-org-id-get-create))
+      (ignore-errors (org-roam-ref-add (concat "@" key)))
+      (when template (insert note-meta))
+      (search-backward "|")
+      (delete-char 1)
+      (when (boundp 'evil-state)
+        (evil-insert 1))))
```
bdarcus, line 298: Little thing (but details matter!), but this doesn't work ATM.

bdarcus:
I've simplified this so the note template is just this:

```elisp
 "Notes on ${author editor}, ${title}"
```

And the default function just this:

```elisp

(defun citar-org-format-note-default (key entry filepath)
  "Format a note FILEPATH from KEY and ENTRY."
    (let* ((template (citar-get-template 'note))
           (note-meta
            (when template
              (citar--format-entry-no-widths
               entry
               template))))
      (funcall citar-file-open-function filepath)
      ;; This just overrides other template insertion.
      (erase-buffer)
      (citar-org-make-preamble key)
      (insert "#+title: ")
      (when template (insert note-meta))
      (insert "\n\n|\n\n#+print_bibliography:")
      (search-backward "|")
      (delete-char 1)
      (when (boundp 'evil-state)
        (evil-insert 1))))
```

Issue #438, opened 2021-11-23
Review public/internal function names, docstrings


bdarcus:
There's maybe some inconsistency ATM that would be nice to resolve as we approach 1.0/stable.

See, for example:

![image](https://user-images.githubusercontent.com/1134/143724141-498c8af8-f749-4852-9a8a-7032d727719a.png)

My hope is that release can provide a stable foundation for future growth and iteration.

salutis:
I think the following two variables should have the same prefix:

```emacs-lisp
(setq-default citar-library-paths (list my/bibtex-directory)
              citar-file-extensions '("pdf"
                                      "html"
                                      "epub"
                                      "txt"
                                      "png"
                                      "jpg"))
```

Perhaps rename `citar-file-extensions` to `citar-library-file-extensions`? 

The same applies to all the other APIs that say "file" and nothing else. We should always say "bibliography-file" or "library-file" because "file" can mean either.

Other ideas:

- `citar-bibliography` could become `citar-bibliography-files` to match others
- `citar-org-local-bib-files` needs a better name (Org file, Org document, or Org buffer instead of "Org-local"; also "bib" vs "bibliography" elsewhere)

salutis:
I also find `citar-open-entry` vs `citar-open` slightly unclear because `citar-open` also opens some "entry".

bdarcus:
> I also find `citar-open-entry` vs `citar-open` slightly unclear because `citar-open` also opens some "entry".

I agree, but haven't thought of a better name that's also concise. There's other discussions here about these name issues.

localauthor:
The [official guidance](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html) on private functions is pretty brief: "Use two hyphens to separate prefix and name if the symbol is not meant to be used by other packages."

### Internal functions (should probably have double hypen):
- citar.el
  - citar-get-value
  - citar-has-a-value
  - citar-display-value
  - citar-has-file
  - citar-has-note
  - citar-get-template
  - citar-get-link
  - citar-open-multi
  - citar-select-group-related-resources

- citar-file.el
  - citar-file-parser-default
  - citar-file-parser-triplet

- citar-org.el
  - citar-org-id-get-create

- citar-latex.el
  - citar-latex-select-command
  - citar-latex-is-a-cite-command

### Not sure about these:
- from citar.el; any thoughts one way or the other?
  - citar-select-ref
  - citar-select-refs
  - citar-select-resource

[...]

Issue #431, opened 2021-11-22
Add org-roam template note function


bdarcus:
The default org note function works well enough with org-roam. But it might be nice to have a dedicated option that uses it's template system.

I'm not very familiar with programming there, and likely don't have time to pick that up. So PR welcome.

mbosley:
I imagine that you can pull heavily from the template structure from org-roam-bibtex?

bdarcus:
I thought about that, but it's awfully complex.

bdarcus:
Jethro shows a pretty simple function here:

https://jethrokuan.github.io/org-roam-guide/

bdarcus:
Closing this; can be handled in:

https://github.com/emacs-citar/citar-org-roam

PR #439, opened 2021-11-24
feat: Use skeleton.el for formatting new notes

### LINKED ISSUE #420: Add "cursor here" marker to note templates

roshanshariff:
Allows for new features, like running Elisp code during note setup, deciding where point should be placed, etc.

- [ ] Update README
- [ ] Decide on a reasonable default for `citar-note-skeleton`
- [ ] Should we split the note functions into a separate `citar-notes.el` file?
- [ ] Do we need more customization points?

Fixes #420.

### DIFF
--- a/citar-org.el
+++ b/citar-org.el
@@ -77,6 +77,19 @@ Each function takes one argument, a citation."
   :group 'citar-org
   :type '(repeat function))
 
+(defcustom citar-org-note-setup-id 'org-roam-only
+  "Whether `citar-org-note-setup` adds an Org ID."
+  :group 'citar-org
+  :type '(choice (const :tag "Always" t)
+                 (const :tag "Only in Org-roam buffers" org-roam-only)
+                 (const :tag "Never" nil)))
+
+(defcustom citar-org-note-setup-roam-ref t
+  "Whether `citar-org-note-setup` adds an Org-roam reference."
+  :group 'citar-org
+  :type '(choice (const :tag "In Org-roam buffers" t)
+                 (const :tag "Never" nil)))
+
 ;;; Keymaps
 
 (defvar citar-org-citation-map
@@ -253,6 +266,21 @@ of a new ID."
         (org-id-get-create force)
       (goto-char point)
       (set-marker point nil))))
+
+;;;###autoload
+(defun citar-org-note-setup (key)
+  "Add Org Id and/or Org-roam reference to element at point.
+See the customization options `citar-org-note-setup-id` and
+`citar-org-note-setup-roam-ref`."
+  (when (derived-mode-p 'org-mode)
+    (let ((roam (and (fboundp 'org-roam-buffer-p) (org-roam-buffer-p))))
+      (when (and citar-org-note-setup-id
+                 (or roam (not (eq 'org-roam-only citar-org-note-setup-id))))
+        (ignore-errors (citar-org-id-get-create)))
+      (when (and roam citar-org-note-setup-roam-ref)
+        (ignore-errors (org-roam-ref-add (concat "@" key))))))
+  ;; Return nil so nothing is inserted if used in skeleton
+  nil)
 
 ;;; Embark target finder
 
--- a/citar-org.el
+++ b/citar-org.el
@@ -62,12 +62,6 @@
           (const long)
           (const short)))
 
-(defcustom citar-file-note-org-include nil
-  "The org note type."
-  :group 'citar
-  :type '(repeat (const :tag "Org ID" 'org-id)
-                 (const :tag "Org-Roam :ROAM_REF:" 'org-roam-ref)))
-
 (defcustom citar-org-style-targets nil
   "Export processor targets to include in styles list.
 
@@ -243,38 +237,22 @@ strings by style."
   (seq-difference (org-cite-list-bibliography-files)
                   org-cite-global-bibliography))
 
-;;; Org note function
+;;; Org note functions
 
 ;;;###autoload
-(defun citar-org-open-notes-default (key entry)
-  "Open a note file from KEY and ENTRY."
-  (if-let* ((file
-             (caar (citar-file--get-note-filename
-                    key
-                    citar-notes-paths '("org"))))
-            (file-exists (file-exists-p file)))
-      (funcall citar-file-open-function file)
-    (let* ((uuid (org-id-new))
-           (template (citar-get-template 'note))
-           (note-meta
-            (when template
-              (citar--format-entry-no-widths
-               entry
-               template)))
-           (org-id (when (member 'org-id citar-file-note-org-include)
-                     (concat "\n:ID:   " uuid)))
-           (org-roam-key (when (member 'org-roam-ref citar-file-note-org-include)
-                           (concat "\n:ROAM_REFS: @" key)))
-           (prop-drawer (or org-id org-roam-key))
-           (content
-            (concat (when prop-drawer ":PROPERTIES:")
-                    org-roam-key org-id
-                    (when prop-drawer "\n:END:\n")
-                    note-meta "\n")))
-      (funcall citar-file-open-function file)
-      ;; This just overrides other template insertion.
-      (erase-buffer)
-      (when template (insert content)))))
+(defun citar-org-id-get-create (&optional force)
+  "Call `org-id-get-create` while maintaining point.
+If point is at the beginning of the buffer and a new properties
+drawer is created, move point after the drawer.  More generally,
+if `org-id-get-create` inserts text at point, move point after
+the insertion.  With optional argument FORCE, force the creation
+of a new ID."
+  (let ((point (point-marker)))
+    (set-marker-insertion-type point t)
+    (unwind-protect
+        (org-id-get-create force)
+      (goto-char point)
+      (set-marker point nil))))
 
 ;;; Embark target finder
 
--- a/citar.el
+++ b/citar.el
@@ -107,16 +107,11 @@ for inserting formatted references."
 
 (defcustom citar-note-skeleton
   '(nil
-    \n                                  ; new line if not at beginning of line
-    '(and (derived-mode-p 'org-mode)
-          (fboundp 'org-roam-buffer-p)
-          (org-roam-buffer-p)
-          (ignore-errors (citar-org-id-get-create))
-          (ignore-errors (org-roam-ref-add (concat "@" &=key=))))
-    (when (derived-mode-p 'org-mode) "#+TITLE: ") | "# "
+    (when (derived-mode-p 'org-mode) (citar-org-note-setup))
+    (if (derived-mode-p 'org-mode) "#+TITLE: " "# ")
     "Notes on "
-    (when &title (format "\"%s\"" (citar-clean-string &title))) | (concat "@" &=key=)
-    \n \n _                             ; place point here
+    (if &title (format "\"%s\"" (citar-clean-string &title)) (concat "@" &=key=))
+    "\n\n" _
     (when (derived-mode-p 'org-mode) "\n\n#+print_bibliography:"))
   "Skeleton for newly created notes.
 See `skeleton-insert` for valid values.  Fields of the
@@ -124,7 +119,17 @@ bibliography entry are bound with names prefixed by '&'; for
 example the \"=key=\" field can be accessed using &=key=, the
 title using &title, etc.  The entry itself is bound to &=entry=."
   :group 'citar
-  :type '(repeat sexp))
+  :type '(list (choice :tag "Interactor"
+                       (const :tag "None" nil)
+                       (string :tag "Prompt")
+                       (sexp :tag "Expression"))
+               (repeat :inline t :tag "Elements"
+                       (choice :value "" :tag ""
+                               (const :tag "New line and indent" \n)
+                               (const :tag "Indent" >)
+                               (string :tag "Text")
+                               (const :tag "Cursor" _)
+                               (sexp :tag "Eval")))))
 
 (defcustom citar-insert-reference-function
   #'citar--insert-reference
--- a/citar.el
+++ b/citar.el
@@ -45,6 +45,7 @@
 (require 'parsebib)
 (require 's)
 (require 'crm)
+(require 'skeleton)
 
 ;;; Declare variables for byte compiler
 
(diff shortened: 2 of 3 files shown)


bdarcus:
@salutis - this isn't done, but you might have some thoughts?

bdarcus:
I updated my PR with some of your code, to compare; still have some more todo tomorrow.

bdarcus:
So I'm skeptical this makes sense. Let me explain and see WYT.

So you have some smart enhancements, which I've borrowed for #431.

With those enhancements, we have a consistent and simple template system that works for UI and new notes, and which includes @salutis' request for being able to define where point is upon completion.

We also have the function configurable, so if one wants to do other things, that's simple enough.

But I don't see what adding a different template system does for us. The template defcustom itself, I think, is going to be more difficult for the average user, and not much or any simpler than just writing a new function.

Unless ... you  can adapt it so that it provides all the advantages of the current templates system, including simplicity and "or" variable choices), but allows us to remove the s dependency?

It does occur to me, however, that the template system itself is somewhat orthogonal to the different design approaches (more logic in template vs less and configurable function, etc).

roshanshariff:
> I'm skeptical skeleton makes sense here.

Yeah, my main concern was whether skeleton is useful and simple enough to justify. I'll outline the pros and cons I considered, and what the UX can be. Of course, users can bypass the skeleton entirely by setting `citar-note-setup-functions` to a different function. But if you're mostly happy with the default and just want to modify it slightly, editing the skeleton defcustom is much easier than copying/overriding a whole function.

> the template system itself is somewhat orthogonal to the different design approaches

I also considered just making `citar-skeleton-insert` a separate add-on, but I realized it gives us several features that are nice to have by default. For example, the default skeleton can adapt to both Org and Markdown notes; then just changing `citar-file-note-extensions` to "md" gives you nicely formatted Markdown notes.

> But I don't see what adding a different template system does for us.

In fact, this whole idea was largely driven by not tying notes functionality too closely with Org mode, but still having good UX. I realized that a skeleton is mostly just a list of text to insert and functions to evaluate. With a slightly better type specification, it's also pretty easy to edit in a Customize buffer. Here are some use cases that become very easy:

### Variables

Including variables (and `or` choices) is not much harder with a skeleton than with `s-format`. This works: `(or @author @editor)`. And, of course, this is just elisp rather than a special string-based template syntax.



Now, in the current PR, missing fields other than "author", "editor", and "title" will be unbound, rather than being bound to nil. To fix this, I just need to fill out the list of valid bibliography fields in the code.

### Rearranging elements

Now suppose somebody wanted to use an Org heading instead of `#+TITLE:`, and attach the Org ID and Org-roam reference to that heading instead of the whole file. They could simply move the `(citar-org-note-setup)` element further down:

```elisp
(setq citar-note-skeleton
      '(nil
        "* Notes on \""
        (citar-clean-string &title) "\"\n"
        (citar-org-note-setup)))
```


This opens the door to people who want to store all their notes in one file; they'll just need a function that opens that file and returns the buffer with point placed appropriately. You get most of the functionality of org-capture even without Org mode and in a much simpler way.

### Including text produced by functions

To include the text produced by some function, you could just make the skeleton something like
```elisp
(setq citar-note-skeleton '(nil (my-note-text &=entry=))
```
A simple skeleton that shows this and also recreates the current org note functionality looks like this in the Customize buffer:



This allows the use case that @salutis [mentioned](https://github.com/bdarcus/citar/issues/420#issuecomment-974090038). I've also wanted to include an outline extracted from the attached PDF, or turn the authors field into a list of links to notes for each author.

### Adapting to different major modes

This was my main motivation for trying to extract as much of the notes functionality out of `citar-org` as possible. A skeleton that's only a little more complicated than the examples above can work well in Org and Markdown:
```elisp
'(nil
  (when (derived-mode-p 'org-mode) (citar-org-note-setup))
  (if (derived-mode-p 'org-mode) "#+TITLE: " "# ")
  "Notes on "
  (if &title (format "\"%s\"" (citar-clean-string &title)) (concat "@" &=key=))
  "\n\n" _
  (when (derived-mode-p 'org-mode) "\n\n#+print_bibliography:"))
```
I'm pushing a new commit that uses this as the default skeleton (and adds the `citar-org-note-setup` helper function). It avoids using any of the more esoteric skeleton elements. The `derived-mode-p` guard is to prevent autoloading `citar-org` unless you're creating an Org-mode note. The helper function uses the "correct" Org-mode and Org-roam commands to insert IDs and add references, rather than having to write out an appropriate properties drawer.

---

Anyway, I realize there are trade-offs to be made, though I think in this case the pros outweigh the cons. But I understand if you still have concerns about complexity, UX, or other downsides.

citar.el
```diff
@@ -94,17 +95,42 @@
...
+for inserting formatted references."
     :group 'citar
     :type  '(alist :key-type symbol
                    :value-type string
                    :options (main suffix preview note)))
 
+(defcustom citar-note-skeleton
+  '(nil
+    (when (derived-mode-p 'org-mode) (citar-org-note-setup))
+    (if (derived-mode-p 'org-mode) "#+TITLE: " "# ")
```
bdarcus, line 111:
```suggestion
    (if (derived-mode-p 'org-mode) "#+title: " "# ")
```

Minor thing: org these days recommends downcasing these.

bdarcus:
> Anyway, I realize there are trade-offs to be made, though I think in this case the pros outweigh the cons. But I understand if you still have concerns about complexity, UX, or other downsides.

Let's see if we get any feedback from others.

Do note the org-roam-bibtex breakage thing I ran into in #434, which I've resolved with an additional defcustom.

cc @localauthor

citar-org.el
```diff
@@ -83,6 +77,19 @@ Each function takes one argument, a citation."
...
+  :type '(choice (const :tag "Always" t)
+                 (const :tag "Only in Org-roam buffers" org-roam-only)
+                 (const :tag "Never" nil)))
+
+(defcustom citar-org-note-setup-roam-ref t
+  "Whether `citar-org-note-setup` adds an Org-roam reference."
+  :group 'citar-org
+  :type '(choice (const :tag "In Org-roam buffers" t)
+                 (const :tag "Never" nil)))
+
```
bdarcus, line 92: Are these really needed?

If you're using org-roam v2, these (at least the ROAM_REF) aren't really optional, so it seems can just add it if in an org-roam buffer?

Id is a little different though, since it's not specific to org-roam.

roshanshariff: Yeah, that makes sense. I can remove them and implement the behaviour you describe (always add ID, and in Org-roam also add roam_ref). If somebody doesn't even want the ID, then they can just remove that function from the skeleton.

localauthor:
My two cents, since you ping'ed:

Having looked into skeleton.el a bit (I hadn't heard of it before), if I'm honest, I'm not really compelled to take the time to learn the nuances of a new package just for one use-case, which in my case, admittedly, is fairly simple. Unless the UX is dead simple, though, I can imagine new users being similarly wary of wading into the details of a package that, despite shipping with emacs, they may never have heard of.

I'm quite satisfied with open-notes function that I've already written, which builds on what the current citar-template offers with `insert`s and `goto-char`s, so I can't see any great reason to translate it into another language. Even if I were coming at Citar afresh, though, and faced the option of writing a standard function or a skeleton function, I would almost certainly choose the former --- the devil you know, sort of thing.

roshanshariff:
> Unless the UX is dead simple, though, I can imagine new users being similarly wary

Fair enough. I figured the UX was fairly simple, enough that you shouldn't need to know what a "skeleton" is, but I understand that's a matter of opinion.

bdarcus:
> Also consider that the existing `s-format` based approach is not completely intuitive either. 

Maybe, but it's consistent. There's one simple template language, rather than two.

> You have to learn its syntax, and you have to read the documentation to learn that the cursor will jump to the "|" character. And if you want to have another "|" character in your template (perhaps unlikely, I know) then you're out of luck.
> 
> Also, if you would prefer to write your own function, you can just put it in `citar-note-setup-functions` instead of `citar-insert-skeleton`, and just use `insert` and `goto-char` as you did.

Yes, but that could be used in #434 also. And even with that, one could still write a function that uses skeleton instead. I, for example, had contemplated in #431 adding an option to use org-roam template system. I don't necessarily think it's a good idea, but it could be, and in any case is possible already.

It would also be possible to modify #434 to handle markdown and org in one function, and maybe even with one template, but I'm not sure we really need that, given that a) most users will use org, and b) writing a markdown-specific function is pretty trivial.

roshanshariff:
> There's one simple template language, rather than two.

I was under the impression that you wanted to migrate away from `s-format`. For formatting references, I thought the plan was to switch to using `citeproc-el`, which is undoubtedly more suitable for that purpose. That would leave `s-format` only being used for minibuffer candidates, which is a very limited use case (packing fields into fixed-width columns) and is very easy to replace with something simple.

Now, I do think that something like `skeleton` is a better templating system for initializing content in buffers, but I accept that we have different opinions on whether the templates are simple enough. :shrug:

bdarcus:
> > There's one simple template language, rather than two.
> 
> I was under the impression that you wanted to migrate away from `s-format`.

That is more about the backend code and potentially removing a dependency than what the user sees. And it's not critical anyway.

> For formatting references, I thought the plan was to switch to using `citeproc-el`, which is undoubtedly more suitable for that purpose. 

This is TBD; another option is to remove that command entirely. 

But in any case, I don't think this strengthens your point, as that would in effect add a third (if widely supported and dedicated) template language.

> That would leave `s-format` only being used for minibuffer candidates, which is a very limited use case (packing fields into fixed-width columns) and is very easy to replace with something simple.
 
Yes.

> Now, I do think that something like `skeleton` is a better templating system for initializing content in buffers, but I accept that we have different opinions on whether the templates are simple enough. 🤷

Right, but also whether we can in effect demand users learn another template language; one that isn't widely known or used.

We also seem, I think, to have a different set of requirements in mind as we evaluate what is "better".

I'll take another look at #434 in the next day or two and see  how much I can refine it, so maybe we have a better basis for comparison.

roshanshariff:
> It seems with your setup code, creating new function is pretty similar to the skeleton template.

Yeah, that was basically my reason for switching to the skeleton. It is a list of elisp functions, except if you just want a string you don't have to create a little function/lambda that returns it (and your functions don't have to call `insert`, just return strings). But just from looking at it, you know how to change it. There's no mental barrier of having to create or modify a new function.

As motivation, I'll recall some frustrations I had when I first started using Citar but wasn't very comfortable with elisp:

I wanted the authors to show up in my note, so I found the note template and added the author field, but for some reason only their last names appeared even though the bibliography had the full names. I dug around for a bit but gave up (now I understand why this was, but at the time it seemed very mysterious). If I had seen a template like the examples I gave above, I would've understood (even without knowing elisp) how to change them to make the authors show up, and it would've done exactly what I expected with no magic.

Later, I wanted to include the current date/time in my note, but even though I'd figured out which function could do that, I couldn't incorporate it into the template (I looked at the `s-format` documentation too). Again, the complexity of creating a new open function from scratch was very high, and I gave up. I suppose I could have made a function that called the old function, put some special marker in the note template where I wanted the date to appear, then gone back and replaced the marker with the current date, but this required knowledge of elisp buffer editing I didn't have.

Solving these issues doesn't require skeleton, but it does make things easier. Even something like this is easier to customize:
```
(defcustom citar-note-format '("Notes on " title "\n\n") ...
```
The code would loop through the elements of that list, eval them, and insert the results. You'd have to bind the bibliography fields to make them available, and maybe you'd add the cursor positioning feature, and now you've recreated skeletons.

bdarcus:
> EDIT: To clarify, my point wasn't that fewer template systems are better, but to use the right one for the job. Shortening author names and using `s-format` is fine for minibuffer candidates, `citeproc-el` is good for formatting references, and for notes neither of those is very good (that you have to intermix concerns and specially handle org IDs in notes shows that its capabilities are being stretched).

But the org-ids and such, at least, can be easily handled in either approach. It comes down how much power you insist on having in the templates.

salutis:
@bdarcus 

> @salutis - this isn't done, but you might have some thoughts?

From the perspective of a user, I can say the following: The `citar-open-note-function` worked well, once I learned about it. 
I faced one problem, namely that Citar does not expose`citar-file--get-note-filename` as a public/stable API. (From what I know, two dashes denote private/unstable API.) In the end, I hard-coded `my/bibtex-directory` instead, as I have only one directory where to look. Other than that, I had no issues.

bdarcus:
> I faced _one_ problem, namely that Citar does not expose`citar-file--get-note-filename` as public/stable API. (From what I understand, two dashes denote private/unstable API.) In the end, I hard-coded `my/bibtex-directory`.

The two dashes is basically: don't rely on this, as it may or may not change.

But the entire code base hasnt been exactly stables pre-1.0.

Why I added #438. We need to review all this make the changes soon.

bdarcus:
> But the org-ids and such, at least, can be easily handled in either approach. It comes down how much power you insist on having in the templates.

I've updated #434 along these lines, so the template is even simpler, but the default note behavior enhanced.

The downside compared to this approach is that @salutis would still need his own custom function if wants to add the BibTeX entry to the note.

My impulse is to merge that, since it's a minor iteration of existing functionality, and spend more time thinking about this.

bdarcus:
> One possibly would be to merge a version of this as an "experimental" feature, and see what feedback we get. So, for example, provide a "skeleton" function, but don't have it default initially.

No rush, but thoughts on this @roshanshariff?

roshanshariff:
Yeah, when I have some time I will refactor this pull request to make it available as an alternative note creation function, on top of the interfaces added in #434. Even if it's not default, I can see it being useful for people who have the use cases I outlined above.

bdarcus:
Maybe we end up with the `citar-open-note-function` (I'll come back to this below) modified to have options like:

1. simple (not sure what to cal it, but the current default)
2. skeleton (this)
3. org-roam-bibtex (`orb-citar-edit-note`; not sure about this, but some citar users are using this, so we could make it easy and obvious)
4. custom

And it could be in time that 3 is replaced by a org-roam-template solution here.

We'll need to settle `citar-open-note-function`  vs `citar-format-note-function`, and how it aligns with the above list.

bdarcus:
Ok to close this @roshanshariff?

PR #443, opened 2021-11-26
feat: transient/hydra experiments

### LINKED ISSUE #437: Add transient menu
I realize I never put an issue here, but there is the never-completed #186 PR.

image

I'm not actually convinced we should do this, or that it should necessarily be included here, but we should make sure the function signatures and such here at least make this easy to do. Maybe a transient-embark adapter?

It may be on some of the main commands we can split the functions to make this easier.

bdarcus:
This is partially to provide users an alternative at-point interface, but more to ensure the code base is sufficiently flexible. Most of this can be used with hydra too, for example.

But ... with the small number of items currently, and recent enhancements to embark, I'm not sure this is worth the hassle ATM.

The balance may change if those numbers expand significantly though, where having groups can help make the UI more clear.

Can also be a good exercise to think through what that UX should be. For example, here's a (not very good) variation on the below:

image

FWIW, key addition required to get either of these working was new interactive functions that shadow/wrap the main citar functions, but take zero arguments.

Close #437.

-----

## Pretty hydra

EDIT: after too much time trying unsuccessfully to track down the loading error I mention below, I had better success with a pretty-hydra, which fully works as expected (though by default doesn't look quite as nice as the transient).

image

----------

## Transient

image

This works, but:

- [ ] need to fix: in general, loading problems; currently I get an error from org-cite unless I explicitly load `citar-transient` from the buffer; I haven't been able to figure out why
- [ ] I did some basic testing with latex, and it works, but I'm unsure how we would better integrate it; how a user would access it, for example, beyond the obvious `M-x citar-transient-menu`; I suppose this is a general issue in modes outside org. I guess use `citar-dwim`?
- [ ] if I merge this, might be nice to use a macro to create the embark keymaps and transient menu from the same source
- [ ] fix the headers
- [ ] update docs

### DIFF
--- a/citar-hydra.el
+++ b/citar-hydra.el
@@ -1,0 +1,64 @@
+;;; citar-hydra.el --- Pretty hydra menu for citar -*- lexical-binding: t; -*-
+;;
+;; Copyright (C) 2021 Bruce D'Arcus
+;;
+;; Author: Bruce D'Arcus <https://github.com/bdarcus>
+;; Maintainer: Bruce D'Arcus <bdarcus@gmail.com>
+;; Created: November 26, 2021
+;; Modified: November 26, 2021
+;; Version: 0.0.1
+;; Keywords: abbrev bib c calendar comm convenience data docs emulations extensions faces files frames games hardware help hypermedia i18n internal languages lisp local maint mail matching mouse multimedia news outlines processes terminals tex tools unix vc wp
+;; Homepage: https://github.com/bdarcus/citar-menu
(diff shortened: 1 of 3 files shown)


bdarcus:
I moved this code to [a gist](https://gist.github.com/bdarcus/09dff264a75ae78330d8ee1a1ee5e1b2), and may later add it to the wiki, but will close the PR.

FYI @roshanshariff ; not sure if we would want to do any subtle refactors to make this simpler, though aside from having to create new functions for hydra or transient, it's admittedly pretty simple.

Issue #450, opened 2021-11-30
Refactor citar-major-mode-functions


bdarcus:
Am not sure, but wonder if we should refactor the nested alist here into something more like org-cite uses for its processors?

```elisp
(cl-defstruct (org-cite-processor (:constructor org-cite--make-processor)
				  (:copier nil))
  (name nil :read-only t)
  (activate nil :read-only t)
  (cite-styles nil :read-only t)
  (export-bibliography nil :read-only t)
  (export-citation nil :read-only t)
  (export-finalizer nil :read-only t)
  (follow nil :read-only t)
  (insert nil :read-only t))
```

https://git.savannah.gnu.org/cgit/emacs/org-mode.git/tree/lisp/oc.el#n310

Maybe something like this?

```elisp
(cl-defstruct (citar-adapter (:constructor citar--make-adapter)
				  (:copier nil))
  (mode nil :read-only t)
  (insert-citation nil :read-only t)
  (citation-at-point nil :read-only t)
  (insert-edit nil :read-only t))
```

So then:

```elisp
(citar--make-adapter
  :mode 'org
  :insert-citation 'citar-whatever)
```

https://nullprogram.com/blog/2018/02/14/

PR #451, opened 2021-11-30
refactor major-mode-functions to use cl-defstruct

bdarcus:
This is the start of an implementation of #450.

You can see what the impact would be by looking at the org example changes here.

Any feedback on this direction @oatmealm @localauthor @roshanshariff?

Would this be more elegant and clear?

I think if the `citar-register-adapter` blocks are autoloaded the functionality should be the same?

OTOH, might just be overkill, since it's not new adapters aren't likely common.

### DIFF
--- a/citar-org.el
+++ b/citar-org.el
@@ -438,5 +438,12 @@ Argument CITATION is an org-element holding the references."
     :follow #'citar-org-follow
     :activate #'citar-org-activate))
 
+(citar-register-adapter 'org
+  :local-bib-files #'citar-org-local-bib-files
+  :insert-citation #'citar-org-insert-citation
+  :insert-edit #'citar-org-insert-edit
+  :key-at-point #'citar-org-key-at-point
+  :citation-at-point #'citar-org-citation-at-point)
+
 (provide 'citar-org)
 ;;; citar-org.el ends here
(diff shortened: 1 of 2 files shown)


roshanshariff:
I think the design goal here is slightly different from that of the org-cite processors. There, you have "bundles" of functionality provided by different backends, and choosing the backend (processor) is up to the user based on their use case. So it makes sense to create structs representing each processor, give them names, put them in a registry, and have defcustoms to choose the correct bundle.

In Citar, on the other hand, you have some generic operations, and the implementation of the operations depends on the context (major mode, in this case). But, given the context, there's really only one implementation of a particular operation.

I think the more appropriate pattern would be what @minad [outlined](https://github.com/bdarcus/citar/pull/357#issuecomment-958749127), using `cl-defgeneric` to define the generic operations and `cl-defmethod` to define their implementations. The code would call the generic function, and the correct implementation gets chosen based on the major mode.

That being said, however, I'm not sure I see the advantage of either of those compared to what we already have now. We're essentially recreating just the major-mode-dispatch feature of `cl-defgeneric` in a fairly straightforward way. Unless we need to dispatch based on things other than major mode, I don't see a compelling reason to add the extra complexity.

PR #453, opened 2021-12-03
Add back embark-general-map parent for citar-map

### LINKED ISSUE #433: embark act menu
I'm not really sure where all the pieces stand, but ATM, in Doom, I don't have access to useful embark functions like `embark-collect-snapshot`. Do we need to set the parent map again, at least in the minibuffer?

bdarcus:
Particularly in the minibuffer, access to some of the embark commands
like `embark-collect-snapshot` is really useful.

Close #433

-----------
Which-key still is a bit confusing when we do this, however, since it has no grouping. 

image

Maybe should just remove the kb descriptions, so the command names show up in the different indicator menus?

Any thoughts @localauthor @roshanshariff?


localauthor:
I'm in favor of this. But I don't use which-key, so for me an abundance of functions is always preferable.

When you say "remove the kb descriptions," do you mean, for example, "open source document" and "open source URL or DOI"? If so, I'm in favor of removing those. I think the function names themselves are sufficiently descriptive. And I have seen no other cases where embark shows a custom string rather than just the function name.

bdarcus:
> I'm in favor of removing those.

Ok, I'll do that as part of this PR.

The menu descriptions are really best for transient or hydra menus. But they don't use these keymaps anyway.

bdarcus:
Here's the result of what I merged.

image

PR #455, opened 2021-12-03
Refactor reference functions, add citar-copy-reference, citar-citeproc

localauthor:
In response to https://github.com/bdarcus/citar/issues/448#issuecomment-981833659 and in light of #245.

This refactors the handling of references to better facilitate the addition of `citar-copy-reference`.

- `citar-insert-reference-function` is now `citar-format-reference-function`

- `citar--insert-reference` is now `citar--format-reference`

The function set in `citar-format-reference-function` must return formatted references as a string, which is either inserted by `citar-insert-reference` or added to the kill-ring by `citar-copy-reference`.

To consider:
- What do you think about the way `citar--format-reference` formats multiple references (ie, no spacing b/t references; refs not sorted)? This is how `citar--insert-reference` worked too, but maybe now is a good time to tweak things, if desired.

Also added here, in its own commit: `citar-citeproc-format-reference`, which requires the `citeproc-el` library.

Some details for consideration:
- When/where to check for, or ensure, that `citeproc-el` is loaded?
- Where should `citar-citeproc` functions/variables be located?
-  The user needs to set the path to the desired CSL file. This is done here with a new defcustom, `citar-citeproc-csl-style`. Thoughts?
- The function needs to point to a csl-locales directory. This is done here by pointing to `org-cite-csl-locales-dir`, because that's what worked with my config. There must be a more general way, but I'm not sure what would be best.

### DIFF
--- a/citar-citeproc.el
+++ b/citar-citeproc.el
@@ -68,7 +68,7 @@
   "Select CSL style to be used with 'citar-citeproc-format-reference'."
   (interactive)
   (unless citar-citeproc-csl-styles-dir
-    (error "Be sure to set 'cite-citeproc-csl-styles-dir' to your CSL styles directory"))
+    (error "Be sure to set 'citar-citeproc-csl-styles-dir' to your CSL styles directory"))
   (let* ((files (directory-files citar-citeproc-csl-styles-dir t "csl"))
          (list (mapcar
                 (lambda (file)
@@ -87,7 +87,7 @@ With prefix-argument, select CSL style."
             current-prefix-arg)
     (citar-citeproc-select-csl-style))
   (unless citar-citeproc-csl-locales-dir
-    (error "Be sure to set 'cite-citeproc-csl-locales-dir' to your CSL locales directory"))
+    (error "Be sure to set 'citar-citeproc-csl-locales-dir' to your CSL locales directory"))
   (let* ((itemids (mapcar (lambda (x) (car x)) keys-entries))
          (proc (citeproc-create citar-citeproc-csl-style
 			        (citeproc-hash-itemgetter-from-any citar-bibliography)
--- a/citar-citeproc.el
+++ b/citar-citeproc.el
@@ -40,7 +40,7 @@
 ;;; Code:
 
 (require 'citar)
-(require 'citeproc-el)
+(require 'citeproc)
 
 (defcustom citar-citeproc-csl-style-dirs nil
   "List of CSL style directories."
@@ -55,20 +55,39 @@
 (defvar citar-citeproc-csl-style nil
   "Path to CSL style file to be used with 'citar-citeproc-format-reference'.")
 
+(defun citar-citeproc-csl-metadata (file)
+  "Return metadata value from csl FILE."
+  (let* ((parse-tree (xml-parse-file file))
+         (_style-node (assq 'style parse-tree))
+         (_info (car (xml-get-children _style-node 'info)))
+         (title (caddr (car (xml-get-children _info 'title)))))
+    title))
+
 (defun citar-citeproc-select-csl-style ()
   "Select CSL style to be used with 'citar-citeproc-format-reference'."
   (interactive)
+  (when (and citar-citeproc-csl-style-dirs
+             (stringp citar-citeproc-csl-style-dirs))
+    (error "Make sure 'citar-citeproc-csl-style-dirs' is a list of paths"))
   (let* ((files (delete-dups (apply #'append (mapcar
                                               (lambda (dir)
-                                                (directory-files dir nil "csl"))
+                                                (directory-files dir t "csl"))
                                               citar-citeproc-csl-style-dirs))))
-                (style (completing-read "Select CSL style file: " files nil t))))
-    (setq citar-citeproc-csl-style (concat citar-citeproc-csl-style-dirs style)))
+         (list (mapcar
+                (lambda (file)
+                  (cons (citar-citeproc-csl-metadata file) file))
+               files))
+         (style (completing-read "Select CSL style file: " list nil t))
+         (file (cdr (assoc style list))))
+    (setq citar-citeproc-csl-style file)))
 
 (defun citar-citeproc-format-reference (keys-entries)
   "Return formatted reference(s) for KEYS-ENTRIES via 'citeproc-el'.
   Formatting follows CSL style set in 'citar-citeproc-csl-style'.
   With prefix-argument, select CSL style."
+  (when (and citar-citeproc-csl-style-dirs
+             (stringp citar-citeproc-csl-style-dirs))
+    (error "Make sure 'citar-citeproc-csl-style-dirs' is a list of paths"))
   (when (or (eq citar-citeproc-csl-style nil)
             current-prefix-arg)
     (citar-citeproc-select-csl-style))
--- a/citar-citeproc.el
+++ b/citar-citeproc.el
@@ -21,7 +21,7 @@
 ;;  CSL styles, using 'citeproc-el'.
 
 ;;  To use: load this file, set the required directory paths
-;;  'citar-citeproc-csl-locales-dir' and 'citar-citeproc-csl-style-dirs', set
+;;  'citar-citeproc-csl-locales-dir' and 'citar-citeproc-csl-styles-dir', set
 ;;  'citar-format-reference-function' to 'citar-citeproc-format-reference',
 ;;  and call one of the general reference functions, either
 ;;  'citar-insert-reference' or 'citar-copy-reference'.
@@ -29,7 +29,7 @@
 ;;  To set a CSL style, either set 'cite-citeproc-csl-style' manually to the
 ;;  path to the desired CSL style file or call
 ;;  'citar-citeproc-select-csl-style' to choose from a style file located in
-;;  'citar-citeproc-csl-style-dirs'.
+;;  'citar-citeproc-csl-styles-dir'.
 
 ;;  If a CSL style is not set before running 'citar-citeproc-format-reference',
 ;;  the user will be prompted to set a style.
@@ -38,14 +38,14 @@
 ;;  'citar-copy-reference' with a prefix-argument.
 
 ;;; Code:
-
+(require 'xml)
 (require 'citar)
 (require 'citeproc)
 
-(defcustom citar-citeproc-csl-style-dirs nil
+(defcustom citar-citeproc-csl-styles-dir nil
   "List of CSL style directories."
   :group 'citar
-  :type '(repeat string))
+  :type ' string)
 
 (defcustom citar-citeproc-csl-locales-dir nil
   "Path to CSL locales dir, required for 'citar-citeproc-format-reference'."
@@ -58,36 +58,29 @@
 (defun citar-citeproc-csl-metadata (file)
   "Return metadata value from csl FILE."
   (let* ((parse-tree (xml-parse-file file))
-         (_style-node (assq 'style parse-tree))
-         (_info (car (xml-get-children _style-node 'info)))
-         (title (caddr (car (xml-get-children _info 'title)))))
+         (style-node (assq 'style parse-tree))
+         (info (car (xml-get-children style-node 'info)))
+         (title (caddr (car (xml-get-children info 'title)))))
     title))
 
+;;;###autoload
 (defun citar-citeproc-select-csl-style ()
   "Select CSL style to be used with 'citar-citeproc-format-reference'."
   (interactive)
-  (when (and citar-citeproc-csl-style-dirs
-             (stringp citar-citeproc-csl-style-dirs))
-    (error "Make sure 'citar-citeproc-csl-style-dirs' is a list of paths"))
-  (let* ((files (delete-dups (apply #'append (mapcar
-                                              (lambda (dir)
-                                                (directory-files dir t "csl"))
-                                              citar-citeproc-csl-style-dirs))))
+  (let* ((files (directory-files citar-citeproc-csl-styles-dir t "csl"))
          (list (mapcar
                 (lambda (file)
                   (cons (citar-citeproc-csl-metadata file) file))
-               files))
+                files))
          (style (completing-read "Select CSL style file: " list nil t))
          (file (cdr (assoc style list))))
     (setq citar-citeproc-csl-style file)))
 
+;;;###autoload
 (defun citar-citeproc-format-reference (keys-entries)
   "Return formatted reference(s) for KEYS-ENTRIES via 'citeproc-el'.
-  Formatting follows CSL style set in 'citar-citeproc-csl-style'.
-  With prefix-argument, select CSL style."
-  (when (and citar-citeproc-csl-style-dirs
-             (stringp citar-citeproc-csl-style-dirs))
-    (error "Make sure 'citar-citeproc-csl-style-dirs' is a list of paths"))
+Formatting follows CSL style set in 'citar-citeproc-csl-style'.
+With prefix-argument, select CSL style."
   (when (or (eq citar-citeproc-csl-style nil)
             current-prefix-arg)
     (citar-citeproc-select-csl-style))
--- a/citar-citeproc.el
+++ b/citar-citeproc.el
@@ -21,7 +21,7 @@
 ;;  CSL styles, using 'citeproc-el'.
 
 ;;  To use: load this file, set the required directory paths
-;;  'citar-citeproc-csl-locales-dir' and 'citar-citeproc-csl-styles-dir', set
+;;  'citar-citeproc-csl-locales-dir' and 'citar-citeproc-csl-style-dirs', set
 ;;  'citar-format-reference-function' to 'citar-citeproc-format-reference',
 ;;  and call one of the general reference functions, either
 ;;  'citar-insert-reference' or 'citar-copy-reference'.
@@ -29,7 +29,7 @@
 ;;  To set a CSL style, either set 'cite-citeproc-csl-style' manually to the
 ;;  path to the desired CSL style file or call
 ;;  'citar-citeproc-select-csl-style' to choose from a style file located in
-;;  'citar-citeproc-csl-style-dir'.
+;;  'citar-citeproc-csl-style-dirs'.
 
 ;;  If a CSL style is not set before running 'citar-citeproc-format-reference',
 ;;  the user will be prompted to set a style.
@@ -42,10 +42,10 @@
 (require 'citar)
 (require 'citeproc-el)
 
-(defcustom citar-citeproc-csl-styles-dir nil
-  "Path to CSL styles dir."
+(defcustom citar-citeproc-csl-style-dirs nil
+  "List of CSL style directories."
   :group 'citar
-  :type 'string)
+  :type '(repeat string))
 
 (defcustom citar-citeproc-csl-locales-dir nil
   "Path to CSL locales dir, required for 'citar-citeproc-format-reference'."
@@ -53,13 +53,17 @@
   :type 'string)
 
 (defvar citar-citeproc-csl-style nil
-  "Path to CSL style dir, for use with 'citar-citeproc-format-reference'.")
+  "Path to CSL style file to be used with 'citar-citeproc-format-reference'.")
 
 (defun citar-citeproc-select-csl-style ()
+  "Select CSL style to be used with 'citar-citeproc-format-reference'."
   (interactive)
-  (let* ((list (directory-files citar-citeproc-csl-styles-dir nil "csl"))
-         (style (completing-read "Select CSL style file: " list nil t)))
-    (setq citar-citeproc-csl-style (concat citar-citeproc-csl-styles-dir style))))
+  (let* ((files (delete-dups (apply #'append (mapcar
+                                              (lambda (dir)
+                                                (directory-files dir nil "csl"))
+                                              citar-citeproc-csl-style-dirs))))
+                (style (completing-read "Select CSL style file: " files nil t))))
+    (setq citar-citeproc-csl-style (concat citar-citeproc-csl-style-dirs style)))
 
 (defun citar-citeproc-format-reference (keys-entries)
   "Return formatted reference(s) for KEYS-ENTRIES via 'citeproc-el'.
--- a/citar-citeproc.el
+++ b/citar-citeproc.el
@@ -1,0 +1,83 @@
+;;; citar-citeproc.el --- Citeproc reference support for citar -*- lexical-binding: t; -*-
+;;
+;; This file is not part of GNU Emacs.
+;;
+;; This program is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;;  Provides functions for formatting bibliographic references according to
+;;  CSL styles, using 'citeproc-el'.
+
+;;  To use: load this file, set the required directory paths
+;;  'citar-citeproc-csl-locales-dir' and 'citar-citeproc-csl-styles-dir', set
+;;  'citar-format-reference-function' to 'citar-citeproc-format-reference',
+;;  and call one of the general reference functions, either
+;;  'citar-insert-reference' or 'citar-copy-reference'.
+
+;;  To set a CSL style, either set 'cite-citeproc-csl-style' manually to the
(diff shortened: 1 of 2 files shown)


bdarcus:
I'll take a closer look tomorrow, but ...

I was wondering if we should perhaps add `citar-citeproc.el` (or `citar-csl.el`?), where we put the citeproc-related stuff?

cc @andersjohansson @roshanshariff

localauthor:
> WDYM? A standard location or something?

Yes. Otherwise, another defcustom?

bdarcus:
> Yes. Otherwise, another defcustom?

I guess the only sensible place to look for this would be citeproc-el, which doesn't have such a defcustom?

And so I guess it make sense to add one here.

On default location, I have previously wondered if the XDG spec might provide an idea, but I'm a little unclear on where the csl example would fit.

bdarcus:
I guess another possibility is just have a defcustom for the style, and people can then concat the org-cite styles dir if they like?

citar.el
```diff
@@ -921,14 +927,48 @@ With prefix, rebuild the cache before offering candidates."
...
+  (interactive (list (citar-select-refs
+                      :rebuild-cache current-prefix-arg)))
+  (let ((references (apply citar-format-reference-function (list keys-entries))))
+    (if (not (equal "" references))
+        (progn
+          (kill-new references)
+          (message (format "Copied:\n  %s" references)))
+      (message "Key not found."))))
+
+(defun citar--format-reference (keys-entries)
```
bdarcus, line 945: Minor thing: but why is this private, and the next public?

localauthor: Ah, that just carried over from citar--insert-reference. Changed to `citar-format-reference` in latest commit.

citar-citeproc.el
```diff
@@ -0,0 +1,83 @@
...
+(defvar citar-citeproc-csl-style nil
+  "Path to CSL style dir, for use with 'citar-citeproc-format-reference'.")
+
+(defun citar-citeproc-select-csl-style ()
+  (interactive)
+  (let* ((list (directory-files citar-citeproc-csl-styles-dir nil "csl"))
+         (style (completing-read "Select CSL style file: " list nil t)))
+    (setq citar-citeproc-csl-style (concat citar-citeproc-csl-styles-dir style))))
+
+(defun citar-citeproc-format-reference (keys-entries)
```
bdarcus, line 64: I realize this was my mistake, but should we change this to plural?

localauthor: You mean change 'citar-citeproc-format-reference' to 'citar-citeproc-format-references'?

If so, should all commands with the '-reference' suffix be plural? Or just '-format-reference(s)'?

For some reason I feel that the two commands, 'insert-reference' and 'copy-reference', should be singular. Hard to say why I feel that way though...

andras-simonyi:
> I guess the only sensible place to look for this would be citeproc-el, which doesn't have such a defcustom?

`citeproc-el` doesn't have any defcustoms by design because it is meant to be a "low-level" library which is not user-facing in any sense. Ideally, users should be able to use applications utilizing `citeproc-el` without being aware of its existence. Maybe in the context of Emacs this separation of applications  and libraries is a bit problematic but I still think that the `no defcustoms for libs` design principle has some merits, the top-level groups in the customization interface also point in this direction IMHO.

bdarcus:
> `citeproc-el` doesn't have any defcustoms by design because it is meant to be a "low-level" library which is not user-facing in any sense.

Right; makes sense.

So we should add one.

localauthor:
> So we should add one.

I've added two: `citar-citeproc-csl-style-dirs` and `citar-citeproc-csl-locales-dir`

localauthor:
> I was wondering, for example, if we want to make it easier for users to select styles, not only for this purpose, but also in the document.

See what you think of the function `citar-citeproc-select-csl-style` in the latest commit.

bdarcus:
> See what you think of the function `citar-citeproc-select-csl-style` in the latest commit.

I wonder: should we pull the completion metadata from the styles?

I was curious how one would even do this in Emacs, and it's not too hard; for example:

```elisp
(defun citar-citeproc-csl-metadata (file)
  "Return metadata value from csl FILE."
  (let* ((parse-tree (xml-parse-file file))
         (_style-node (assq 'style parse-tree))
         (_info (car (xml-get-children _style-node 'info)))
         (title (caddr (car (xml-get-children _info 'title)))))
    title))
```

But we can worry about that later.

citar-citeproc.el
bdarcus: In part this is the general confusion about single vs multiple throughout.

IDK either.

This is related to discussion in the wider ecosystem about, for example, possibly adding multi-actions to embark.

But I digress. Maybe just ignore this for now.

bdarcus:
I'm pretty busy this week @localauthor, but plan to find time to try this out.

Just mark it ready to review when you think it's ready for me to do that?

citar-citeproc.el
```diff
@@ -0,0 +1,106 @@
...
+
+;;  A CSL style can also be set by calling 'citar-insert-reference' or
+;;  'citar-copy-reference' with a prefix-argument.
+
+;;; Code:
+
+(require 'citar)
+(require 'citeproc)
+
+(defcustom citar-citeproc-csl-style-dirs nil
```
bdarcus, line 45: Why a list?

I get that it provides more flexibility, but is that really necessary?

Consider that `org-cite-csl-styles-dir` is a string.

citar.el
```diff
@@ -115,9 +115,10 @@ for the title field for new notes."
                    :value-type string
                    :options (main suffix preview note)))
 
-(defcustom citar-insert-reference-function
-  #'citar--insert-reference
-  "A function that takes a list of (KEY . ENTRY), and returns formatted references."
+(defcustom citar-format-reference-function
+  #'citar-format-reference
+  "A function that takes a list of (KEY . ENTRY), and returns
+formatted references as a string."
```
bdarcus, line 121: Do I have to set this to use the citeproc rendering? If yes, to what value?

And should this provide a choice to make this explicit?

citar-citeproc.el
```diff
@@ -0,0 +1,106 @@
...
+                                              citar-citeproc-csl-style-dirs))))
+         (list (mapcar
+                (lambda (file)
+                  (cons (citar-citeproc-csl-metadata file) file))
+               files))
+         (style (completing-read "Select CSL style file: " list nil t))
+         (file (cdr (assoc style list))))
+    (setq citar-citeproc-csl-style file)))
+
+(defun citar-citeproc-format-reference (keys-entries)
```
bdarcus, line 84: Don't you need to autoload this?

bdarcus (commented):
Looking good, but I've left a few comments to resolve, in part so I can test it :-)

Initial try didn't work for me.

For example, I did this:

```elisp
(setq citar-format-note-function 'citar-citeproc-format-reference)
```
... but `citar-insert-reference` doesn't appear to use that function (no select-style prompt, and output is wrong).

Also, this is my fault, but getting this compiler warning we should figure out how to fix.

```elisp
Compiling file /home/bruce/Code/citar/citar-citeproc.el at Tue Dec  7 08:00:15 2021
citar-citeproc.el:58:1: Warning: variable ‘_info’ not left unused
citar-citeproc.el:58:1: Warning: variable ‘_style-node’ not left unused
```

Also, seems we need to tell people they must include the locale file, as citeproc will return an error otherwise.

bdarcus:
> Also, this is my fault, but getting this compiler warning we should figure out how to fix.

Oh, duh.

To fix this, just remove the underscores.

Also, you need to add this to the top:

```elisp
(require 'xml)
```

citar-citeproc.el
```diff
@@ -0,0 +1,106 @@
...
+;;  'citar-citeproc-csl-style-dirs'.
+
+;;  If a CSL style is not set before running 'citar-citeproc-format-reference',
+;;  the user will be prompted to set a style.
+
+;;  A CSL style can also be set by calling 'citar-insert-reference' or
+;;  'citar-copy-reference' with a prefix-argument.
+
+;;; Code:
+
```
bdarcus, line 41:
```suggestion
(require 'xml)
```

citar-citeproc.el
```diff
@@ -0,0 +1,106 @@
...
+(defvar citar-citeproc-csl-style nil
+  "Path to CSL style file to be used with 'citar-citeproc-format-reference'.")
+
+(defun citar-citeproc-csl-metadata (file)
+  "Return metadata value from csl FILE."
+  (let* ((parse-tree (xml-parse-file file))
+         (_style-node (assq 'style parse-tree))
+         (_info (car (xml-get-children _style-node 'info)))
+         (title (caddr (car (xml-get-children _info 'title)))))
```
bdarcus, line 63:
```suggestion
         (style-node (assq 'style parse-tree))
         (info (car (xml-get-children style-node 'info)))
         (title (caddr (car (xml-get-children info 'title)))))
```

citar.el
localauthor: I added some instructions to the docstring. Hopefully makes things clearer.

bdarcus: I still can't get `citar-citeproc` to autoload correctly, despite this:

```elisp
citar-format-note-function is a variable defined in ‘citar.el’.

Its value is ‘citar-citeproc-format-reference’
Original value was 
‘citar-org-format-note-default’
```

citar.el
```diff
@@ -926,18 +937,31 @@ With prefix, rebuild the cache before offering candidates."
...
-      (when template
-        (insert (citar--format-entry-no-widths (cdr key-entry) template))))))
+  (interactive (list (citar-select-refs)))
+  (insert (apply citar-format-reference-function (list keys-entries))))
+
+;;;###autoload
+(defun citar-copy-reference (keys-entries)
+  "Copy formatted reference(s) associated with the KEYS-ENTRIES."
+  (interactive (list (citar-select-refs)))
+  (let ((references (apply citar-format-reference-function (list keys-entries))))
```
roshanshariff, line 948:
```suggestion
  (let ((references (funcall citar-format-reference-function keys-entries)))
```

citar.el
```diff
@@ -926,18 +937,31 @@ With prefix, rebuild the cache before offering candidates."
...
-  (apply citar-insert-reference-function (list keys-entries)))
-
-(defun citar--insert-reference (keys-entries)
   "Insert formatted reference(s) associated with the KEYS-ENTRIES."
-  (let ((template (citar-get-template 'preview)))
-    (dolist (key-entry keys-entries)
-      (when template
-        (insert (citar--format-entry-no-widths (cdr key-entry) template))))))
+  (interactive (list (citar-select-refs)))
+  (insert (apply citar-format-reference-function (list keys-entries))))
```
roshanshariff, line 942:
```suggestion
  (insert (funcall citar-format-reference-function keys-entries)))
```

citar-citeproc.el
```diff
@@ -0,0 +1,103 @@
...
+;;;###autoload
+(defun citar-citeproc-format-reference (keys-entries)
+  "Return formatted reference(s) for KEYS-ENTRIES via 'citeproc-el'.
+Formatting follows CSL style set in 'citar-citeproc-csl-style'.
+With prefix-argument, select CSL style."
+  (when (or (eq citar-citeproc-csl-style nil)
+            current-prefix-arg)
+    (citar-citeproc-select-csl-style))
+  (unless citar-citeproc-csl-locales-dir
+    (error "Be sure to set 'cite-citeproc-csl-locales-dir' to your CSL locales directory"))
```
roshanshariff, line 90:
```suggestion
    (error "Be sure to set 'citar-citeproc-csl-locales-dir' to your CSL locales directory"))
```

citar-citeproc.el
```diff
@@ -0,0 +1,103 @@
...
+         (info (car (xml-get-children style-node 'info)))
+         (title (caddr (car (xml-get-children info 'title)))))
+    title))
+
+;;;###autoload
+(defun citar-citeproc-select-csl-style ()
+  "Select CSL style to be used with 'citar-citeproc-format-reference'."
+  (interactive)
+  (unless citar-citeproc-csl-styles-dir
+    (error "Be sure to set 'cite-citeproc-csl-styles-dir' to your CSL styles directory"))
```
roshanshariff, line 71:
```suggestion
    (error "Be sure to set 'citar-citeproc-csl-styles-dir' to your CSL styles directory"))
```

roshanshariff (commented):
Just suggesting a few small typos/changes that I noticed.

localauthor:
> Just suggesting a few small typos/changes that I noticed.

Thanks for the edits. Any chance you’re able to get this running properly?

bdarcus:
> Thanks for the edits. Any chance you’re able to get this running properly?

I'm curious too. I still can't get it to work, and I can't see why.

I thought maybe I had something wonky with my install, but wiped the citar build directory and rebuilt it, and still things just aren't working:

```elisp
ELISP> (citar-citeproc-format-reference '(("kohn2005" (citar--get-entry "kohn2005"))))
"Kohn, M. (2005). Securing the Spectacular City: the politics of revitalization and homelessness in downtown seattle. Pacific Historical Review, 74(1), 137."

ELISP> (citar-insert-reference '(("kohn2005" (citar--get-entry "kohn2005"))))
nil
ELISP>  () , .
```

When I open a new session:

- the two autoloaded functions show up.
- the variables are correctly set
- `citar-citeproc-format-reference` works correctly when I call it directly, per above

I'm confused.

localauthor:
> when I run it interactively, I get a result that is clearly the default function.

The only possibility I can think of is that the variable `citar-format-reference-function` is being reset to the default at some point, somehow.

Have you tried using `edebug` to step through `citar-insert-reference` to see what is happening?

I've successfully run this feature starting from `emacs -Q` and loading only this minimal config:

```
;; first pull citar-citeproc feature branch ("https://github.com/localauthor/citar/tree/copy-reference") into citar build directory

(require 'package)
(add-to-list 'package-archives '("melpa" . "https://melpa.org/packages/"))
(package-refresh-contents)

(package-install 'vertico)
(vertico-mode)

(package-install 'citeproc)

;; load citar-citeproc
(add-to-list 'load-path "~/.emacs.d/straight/build/citar/")
(load-file "~/.emacs.d/straight/build/citar/citar-autoloads.el")
(load-file "~/.emacs.d/straight/build/citar/citar.el")
(load-file "~/.emacs.d/straight/build/citar/citar-citeproc.el")

;; minimal citar settings

(setq citar-bibliography "~/Dropbox/gr-bibliography.bib")

;; minimal citeproc settings

(setq citar-format-reference-function #'citar-citeproc-format-reference)

(setq citar-citeproc-csl-locales-dir "~/.csl/locales/")
(setq citar-citeproc-csl-styles-dir "~/.csl/")
(setq citar-citeproc-csl-style "~/.csl/chicago-fullnote-bibliography.csl")
```

bdarcus:
Sigh ... user error. I had the wrong variable set for that ;-(

> Have you tried using edebug to step through citar-insert-reference to see what is happening?

How do you do this?

roshanshariff:
It worked for me too :)

> How do you do this?

Put the point in the function you want to be able to debug and run `edebug-eval-defun` (or `eval-defun` with a prefix argument). Then run the code, and it'll stop as soon as execution goes into the function. Then `SPC` steps through each function, `g` resumes evaluation, etc. (You can look at the available edebug commands under the `C-x X` prefix).

citar.el
```diff
@@ -110,11 +110,22 @@ for the title field for new notes."
...
+'citar-format-reference', formats references using the 'preview'
+template set in 'citar-template'. To use 'citeproc-el' to format
+references according to CSL styles, set the value to
+'citar-citeproc-format-reference'. Alternatively, set to a custom
+function that takes a list of (KEY . ENTRY) and returns formatted
+references as a string."
   :group 'citar
-  :type 'function)
+  :type '(choice (const :tag "Use 'citar-template'" citar-format-reference)
+                 (const :tag "Use 'citeproc-el'" citar-citeproc-format-reference)
```
roshanshariff, line 127:
```suggestion
  :type '(choice (function-item :tag "Use 'citar-template'" citar-format-reference)
                 (function-item :tag "Use 'citeproc-el'" citar-citeproc-format-reference)
```

andras-simonyi:
For the record, and to reiterate what I wrote in a comment on #245,  it is most probably unnecessary to use citeproc-el's built-in itemgetter which reparses the bibliography file since the required data is already in the Citar cache. The problem with the attempts I saw was simply that they didn't parse the name and data fields into the format required by citeproc-el, but this would be trivial to do if the cache contains those in a standardized format.

bdarcus:
> The problem with the attempts I saw was simply that they didn't parse the name and data fields into the format required by citeproc-el, but this would be trivial to do if the cache contains those in a standardized format (BibTeX?).

The cache just holds the data that `parsebib-parse` returns.

```elisp
  ("title" . "New Retail Capital and Neighborhood Change")
  ("url" . "http://onlinelibrary.wiley.com/doi/10.1111/j.1540-6040.2009.01269.x/abstract")
  ("doi" . "10.1111/j.1540-6040.2009.01269.x")
  ("date" . "2009-03-01")
  ("author" . "Zukin, Sharon and Trujillo, Valerie and Frase, Peter and Jackson, Danielle and Recuber, Tim and Walker, Abraham")
  ("=type=" . "article")
  ("=key=" . "zukin2009"))
```

roshanshariff (commented):
Some more small quality-of-life improvements. Another needed change is to add `(citeproc "0.9")` to the `Package-Requires` line in `citar.el`.

roshanshariff:
> The cache just holds the data that parsebib-parse returns

If I'm not mistaken, Citar's cached entries only contain the fields returned by `citar--fields-to-parse`, which is only those used in templates and a few extras (file, doi, and url). If the CSL style needs other fields, it seems to me that re-parsing would be needed anyway.

andras-simonyi:
> If I'm not mistaken, Citar's cached entries only contain the fields returned by `citar--fields-to-parse`, which is only those used in templates and a few extras (file, doi, and url). If the CSL style needs other fields, it seems to me that re-parsing would be needed anyway.

Yes, this is the main issue -- parsing the cached data into the format citeproc-el needs is basically one function call.

bdarcus:
> Seems like a genuine trade-off.

Right; citeproc needs access to more data, and more structured data, than does the rest of the functionality here.

So what do you suggest?

andras-simonyi:
> > Seems like a genuine trade-off.
> 
> Right; citeproc needs access to more data, and more structured data, than does the rest of the functionality here.
> 
> So what do you suggest?

Well, for a while I've been thinking about  somehow unifying all these in-memory bibliographic databases  -- there seem to be a lot of parallels between `citeproc-el`-s itemgetters and `citar`'s,`helm-bibtex`'s and `ebib's` etc. in-memory structures. Maybe it would be useful to come up with a common API and data model for them which would support querying and composing these objects (perhaps using EIEO?), but this would require a longer discussion and careful planning, of course.

bdarcus:
> > > Seems like a genuine trade-off.
> > 
> > 
> > Right; citeproc needs access to more data, and more structured data, than does the rest of the functionality here.
> > So what do you suggest?
> 
> Well, for a while I've been thinking about somehow unifying all these in-memory bibliographic databases -- there seem to be a lot of parallels between `citeproc-el`-s itemgetters and `citar`'s,`helm-bibtex`'s and `ebib's` etc. in-memory structures. Maybe it would be useful to come up with a common API and data model for them which would support querying and composing these objects (perhaps using EIEO?), but this would require a longer discussion and careful planning, of course.

@andras-simonyi I was just remembering this. You might want to look at the new cache design @roshanshariff implemented? See citar-cache.el.

Not sure how suitable it would be for your unified idea, but it could be close?

joostkremers:
> 


> > Well, for a while I've been thinking about somehow unifying all these in-memory bibliographic databases -- there seem to be a lot of parallels between `citeproc-el`-s itemgetters and `citar`'s,`helm-bibtex`'s and `ebib's` etc. in-memory structures. Maybe it would be useful to come up with a common API and data model for them which would support querying and composing these objects (perhaps using EIEO?), but this would require a longer discussion and careful planning, of course.

I guess the first question is what exactly would you like to unify? To store the actual entries, Ebib uses a hash table where each entry is an alist of field-value pairs. The alist is essentially the same structure that `bibtex.el` uses, which is a deliberate choice on my part, because it allows me to use some functions from `bibtex.el`, such as `bibtex-generate-autokey`.

`parsebib.el` of course uses the same structure, and therefore (I assume) every package that depends on `parsebib.el` does, as well. So that part wouldn't be difficult to unify. (Though keep in mind that Ebib stores the actual contents of the `.bib` files, while most other packages probably use the prettification functionality of `parsebib.el`, so the actual contents of the in-memory hash table will be different, even for the same `.bib` file.)

When it comes to the metadata, however, Ebib's requirements are very different from e.g., `citar.el`'s, at least judging by a quick glance at `citar-cache.el`. First, because Ebib not only reads but also writes `.bib` files, it needs metadata about the file's modification status. It also keeps track of `@String` and `@Preamble` definitions, `@Comments` and even a local variable block. It keeps track of marked entries, of the current entry, the current sort field and direction, and of the current search filter.

Furthermore, possibly the greatest stumbling block is that Ebib's in-memory storage is set up to allow for dependent databases, which share their data with a main database such that if you edit an entry in one, the change is reflected in the other. This is done by a special access function which, when the database being accessed is a dependent one and if the data being retrieved or modified is shared, performs the request on the main database instead.

Because of this requirement, I abandoned `cl-defstruct` (which I used originally) and started using simple alists to represent each database. (The details can be seen [here](https://github.com/joostkremers/ebib/blob/master/ebib-db.el), BTW.)


> Something like that might be appropriate for parsebib,

Well, everything I said about Ebib doesn't apply to `parsebib.el`, of course. But OTOH `parsebib.el` right now does not provide any metadata about a `.bib` file. If there's a clear benefit if moving certain code into `parsebib.el`, I'd definitely support it. Right now, though, I'm not sure what sort of functionality could be moved to `parsebib.el` and what the benefit would be.

>  particularly if @joostkremers could address Ihor's request about putting it on ELPA.

I'll start working on that soon.

bdarcus:
Not discounting the possibility such an idea from @andras-simonyi wouldn't work, but on this ...

> ... because Ebib not only reads but also writes `.bib` files, it needs metadata about the file's modification status.

The citar cache tracks things like size, checksum, modification time.

> I'd definitely support it. Right now, though, I'm not sure what sort of functionality could be moved to parsebib.el and what the benefit would be.

I'm not sure on the details either. I think the origin of the idea is really just the high-level recognition that, for example, currently when using citar and citeproc with org, we have this same bib file(s) parsed into caches three different times; one for each package, both by parsebib, and then oc-basic (which currently does fontification and such). That seems less than ideal

joostkremers:
> The citar cache tracks things like size, checksum, modification time.

Oh, cool! I didn't notice. It was a *very* quick glance. :smile:

> I'm not sure on the details either. I think the origin of the idea is really just the high-level recognition that, for example, currently when using citar and citeproc with org, we have the same bib file(s) parsed into caches three different times; one for each package, both by parsebib, and then oc-basic (which currently does fontification and such). That seems less than ideal

Yes, it does.

I guess one option might be to move the cache to `parsebib.el`. Or have a separate library that keeps the cache and calls out to `parsebib.el` whenever it notices that one of the underlying `.bib` files has been modified or hasn't been loaded yet. (Which I guess is what `citar-cache.el` is doing right now, so it might be just a matter of making that package more general, so other packages can use it as a back-end as well.)

bdarcus:
> I guess one option might be to move the cache to `parsebib.el`. Or have a separate library that keeps the cache and calls out to `parsebib.el` whenever it notices that one of the underlying `.bib` files has been modified or hasn't been loaded yet. (Which I guess is what `citar-cache.el` is doing right now ...

Yes, it does. I also suggested ways we could build on this to optimize performance later.

#654 

> so it might be just a matter of making that package more general, so other packages can use it as a back-end as well.)

Right:

I think the two cache hash-tables we have now do reflect the priorities of our use cases (which are the same as ivy/helm-bibtex and such.

- `:entries` we parse the files for display, as you suggested makes sense
- `:preformatted` we format (most of) it for display and store it

I'm guessing, based on what you say above, that changing the first to use the more structured data (needed by citeproc and ebib) would add code on our end, and probably be slower, in a place where performance is really important.

Which is maybe the key roadblock to this idea.

bdarcus:
PS - look at citar-citeproc.el for a possible ready-made preview.

andras-simonyi:
Hm. the stored information for entries from CSL-JSON files seems to be a strange hybrid of CSL and bib(la)tex... is this documented somewhere? Are the CSL-JSON name (and date?) fields converted to bib(la)tex fields?

bdarcus:
See https://github.com/joostkremers/parsebib/issues/13

The solution we use is not to do any translation, but encourage users to include both field names when needed, since our template system will output the first match.

That's why you see this in the default template:

```elisp
${date year issued:4}
```

.. and then `citar-get-value` is written to handle both symbols and strings.

https://github.com/emacs-citar/citar/blob/c5ed78e69ab1ddbf19f6b5793c8356678466caf8/citar.el#L797-L805

Per the issue I linked above, it's possible there's a more general for parsebib, but not clear.

bdarcus:
Is it faster as a consequence?

andras-simonyi:
> Is it faster as a consequence?

I don't really see a difference but I experimented with Org files where the rendering task dominates (~1200 citations). There would probably be a speedup with large, already cached bibliography files (~10,000 or more entries).

PR #457, opened 2021-12-04
refactor find-additional

### LINKED ISSUE #456: Use file-name-all-completions for find-additional?
@localauthor @roshanshariff - I just stumbled on `file-name-all-completions`. I wonder if this would be at least theoretically faster for checking for file variants, since it's a built-in function written in C?

Example from the docs:

```elisp
(file-name-all-completions "vim" "/usr/bin/")

("vimtutor" "vimdiff" "vim")
```

Maybe something like this?

```elisp
(defun citar-file--find-additional-multi-dirs (key dirs extensions)
  "Return a list of files and variants for KEY from list of DIRS for EXTENSIONS."
  (seq-mapcat (lambda (dir) (citar-file--find-additional key dir extensions)) dirs))

(defun citar-file--find-additional (key dir extensions)
  "Return a list of files and variants from DIR for KEY for EXTENSIONS."
  (seq-filter
    (lambda (fn)
      (member (file-name-extension fn) extensions)) (file-name-all-completions key dir)))
```

Any reason not to do this?

bdarcus:
Add new functions for this, and use the C-based `file-name-all-completions`.

Close #456.

-------

I initially forgot to check the `citar-file-find-additional-files ` defcustom, which I forgot was more flexible than the initial commit. 

I have kept it, but simplified it to allow either space or hyphen to be appended to the key (mostly because I couldn't figure out how to get customize to allow one to enter a random string).

WYDT @localauthor @roshanshariff?

### DIFF
--- a/citar-file.el
+++ b/citar-file.el
@@ -74,18 +74,22 @@ will open, via `citar-open-notes'."
 
 (defcustom citar-file-find-additional-files nil
   "Find additional library files starting with reference key.
+
 If t, all files whose base name starts with the reference key and
 whose extension is listed in `citar-file-extensions' are located
 by the functions `citar-open-library-files' and
-`citar-open-notes'.  If nil, only files with the naming scheme
-\"<key>.<extension>\" are located.  Otherwise, its value is a
-regular expression specifying how the key is separated from the
-rest of the filename."
+`citar-open-notes'.
+
+If nil, only files with the naming scheme \"<key>.<extension>\"
+are located.
+
+Otherwise, you may specify to separate the key from rest of
+filename with either a space or a hyphen."
   :group 'citar
   :type '(choice (const :tag "Ignore additional files" nil)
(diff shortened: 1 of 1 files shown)


citar-file.el
```diff
@@ -77,15 +77,14 @@ will open, via `citar-open-notes'."
...
+`citar-open-notes'. If nil, only files with the naming scheme
+\"<key>.<extension>\" are located. Otherwise, its value is a
+string to be appended to the key, to separate it from the rest of
+the filename."
   :group 'citar
   :type '(choice (const :tag "Ignore additional files" nil)
                  (const :tag "Find all files starting with key" t)
-                 (const :tag "Find files with space after key" "[[:space:]]")
-                 (regexp :tag "Filename separator")))
+                 (const :tag "Find files with string after key")))
```
bdarcus, line 87: This doesn't present a text field to add the string in the customize interface though.

Issue #452, opened 2021-12-02
(user-customizable) Defaultable citation commands


Hugo-Heagren:
**Is your feature request related to a problem? Please describe.**
I often (almost always) use the same command for citations -- `\autocite`. It is frustrating to have to select this (especially as it is not the default) every time I insert a citation into a TeX buffer.

**Describe the solution you'd like**
1) Have a customizable variable, set to the default command to use for citations (in my case, the string `autocite`, `\autocite{%s}` or some such).
2) Allow users to customize (probably through some other variable, or perhaps the same one if it were handled cleverly) whether they would normally like to be prompted for a citation command by `citar-insert-citation`, or to always just use the default without prompting.
3) Add a prefix arg to `citar-insert-citation`, which would reverse the behaviour specified in 2 on a per-invocation basis.

**Describe alternatives you've considered**
No alternatives really, though this is a small QOL improvement, so that isn't particularly surprising.

**Additional context**
This is similar to a feature I suggest at karthink/consult-reftex#3, a somewhat similar project. As there, I would be happy to have a stab at implementing this and submit a PR.

bdarcus:
Makes sense.

I know @aikrahguzar, who wrote the new latex stuff, typically only uses `\cite`, so would like appreciate.

>  I would be happy to have a stab at implementing this and submit a PR.

Please do.

The only suggestion I would make as you work on that is to keep in mind this project support multiple major modes. So you'll want to consider how the approach should work outside latex.

Also, if you can, you might play with the org support, which doesn't prompt for the style, though you can edit it by placing point in the right place.

Should the latex mode work the same?

bdarcus:
BTW, on a [tangentially-related](https://github.com/bdarcus/oxr) point, can you post a screenshot of the consult-reftex ref insertation UI?

Hugo-Heagren:
> BTW, on a tangentially-related point, can you post a screenshot of the consult-reftex ref insertion UI?

Sure.

![2021-12-02--19:01:34](https://user-images.githubusercontent.com/62905215/144486289-20a681ef-2d2a-4066-8bb6-7cd10145464f.jpg)

(I have [an issue](https://github.com/karthink/consult-reftex/issues/4) open about customising the categories, so that it recognises `ex` for example and so forth)

Hugo-Heagren:
Gotta love a maintainer who replies fast!

> Please do.

Awesome. 

[...]

PR #458, opened 2021-12-04
Default citation style

Hugo-Heagren:
Here's a draft PR to get started with, fixing #452. I've only implemented LaTeX support so far. Some notes:

Basically what we're doing is having a variable which controls whether to prompt for a citation style, and then the behaviour that variable controls is inverted by a prefix arg. The only entry point is `citar-insert-citation`, so it'll be that commands prefix arg that controls things. There are thus two approaches:
1) have a single variable (e.g. `'citar-prompt-for-cite-style`), do the inversion logic with in `citar-insert-citation`, then pass the result to the specific major mode functions as an argument which tells them whether or not to prompt. They don't need to worry about handling prefix args or doing logic with the variable. This has the advantage that there is only one variable and one piece of logic.
2) have a separate variable for each major mode, pass the prefix arg itself to each major mode function, and let them handle it. This has more going on (and therefore, more to go *wrong*). But it's what I ended up with. The three modes we support are quite different, at least at the moment. LaTeX has good support for prompting for citation styles, org less so, pandoc doesn't *have* citation styles, though it has a binary option to suppress author's names. I decided that trying to handle the logic in `citar-insert-citation` was more generalistic than the functions themselves would support.

For similar reasons, all the variables controlling this behaviour are mode-specific: one for LaTeX, one for markdown etc. Do you think this is a good call, or would you rather a more generalistic approach?

Because these are just variables, they can be set file-/dir-(project-?)locally, so that should satisfy @roshanshariff's idea (though I don't use or know much about local variables, so I haven't actually tested this yet).

I put `invert-prompt` in front of `command` in `citar-latex-insert-citation` because `citar-insert-citation` passes a prefix arg to every major mode function, so the argument which inherits that has to come first. This will apply to the others as well.

I'll have a go at implementing similar things for markdown and org:
- Do you like the idea of using a prefix arg as a switch for suppressing the author in markdown? 
- Would you like me to try to implement initial citaton-style-choice for org mode (like we currently have for latex) and make the behaviour consistent? I would be happy to try, but maybe you think that's beyond the scope of this PR?

(This is a draft because there's so much unfinished. I'll promote to full when it's a bit more tied up)

### DIFF
--- a/citar-latex.el
+++ b/citar-latex.el
@@ -53,6 +53,23 @@ the point."
   :type '(alist :key-type (repeat string)
                 :value-type sexp))
 
+(defcustom citar-latex-prompt-for-cite-style t
+  "Whether to prompt for a citation command when inserting."
+  :group 'citar
+  :type '(radio (const :tag "Prompt for a command" t)
+                (const :tag "Do not prompt for a command" nil))
+  :safe 'always)
+
+(defcustom citar-latex-default-cite-command "cite"
+  "Default command for citations.
+
+Must be in `citar-latex-cite-commands'. Used when as a cite
+command when prompting for one is disabled, and as the default
+entry when it is enabled."
+  :group 'citar
+  :type 'string
+  :safe 'always)
+
 (defcustom citar-latex-prompt-for-extra-arguments t
   "Whether to prompt for additional arguments when inserting a citation."
   :group 'citar-latex
@@ -127,13 +144,16 @@ inside a citation macro."
   "Variable for history of cite commands.")
 
 ;;;###autoload
-(defun citar-latex-insert-citation (keys &optional command)
+(defun citar-latex-insert-citation (keys &optional invert-prompt command)
   "Insert a citation consisting of KEYS.
 
 If the command is inside a citation command keys are added to it. Otherwise
 a new command is started.
 
-If the optional COMMAND is provided use it, otherwise prompt for one.
+If the optional COMMAND is provided use it (ignoring INVERT-PROMPT).
+Otherwise prompt for a citation command, depending on the value of
+`citar-latex-prompt-for-cite-style'. If INVERT-PROMPT is non-nil, invert
+whether or not to prompt.
 
 The availiable commands and how to provide them arguments are configured
 by `citar-latex-cite-commands'.
@@ -147,11 +167,15 @@ inserted."
     (if (citar-latex-is-a-cite-command (TeX-current-macro))
         (progn (skip-chars-forward "^,}")
                (unless (equal ?} (preceding-char)) (insert ", ")))
-      (let ((macro (or command
-                       (completing-read "Cite command: "
-                                        (seq-mapcat #'car citar-latex-cite-commands)
-                                        nil nil nil
-                                        'citar-latex-cite-command-history nil nil))))
+      (let ((macro
+	     (or command
+		 (if (xor invert-prompt citar-latex-prompt-for-cite-style)
+                     (completing-read "Cite command: "
+                                      (seq-mapcat #'car citar-latex-cite-commands)
+                                      nil nil nil
+                                      'citar-latex-cite-command-history
+				      citar-latex-default-cite-command nil))
+	       citar-latex-default-cite-command)))
         (TeX-parse-macro macro
                          (when citar-latex-prompt-for-extra-arguments
                            (cdr (citar-latex-is-a-cite-command macro))))))
--- a/citar-latex.el
+++ b/citar-latex.el
@@ -170,12 +170,8 @@ inserted."
       (let ((macro
 	     (or command
 		 (if (xor invert-prompt citar-latex-prompt-for-cite-style)
-                     (completing-read "Cite command: "
-                                      (seq-mapcat #'car citar-latex-cite-commands)
-                                      nil nil nil
-                                      'citar-latex-cite-command-history
-				      citar-latex-default-cite-command nil))
-	       citar-latex-default-cite-command)))
+                     (citar-latex-select-command)
+		   citar-latex-default-cite-command))))
         (TeX-parse-macro macro
                          (when citar-latex-prompt-for-extra-arguments
                            (cdr (citar-latex-is-a-cite-command macro))))))
@@ -188,6 +184,14 @@ inserted."
 With ARG non-nil, rebuild the cache before offering candidates."
   (citar-latex-insert-citation
    (citar--extract-keys (citar-select-refs :rebuild-cache arg))))
+
+(defun citar-latex-select-command ()
+  "Complete a citation command for LaTeX."
+  (completing-read "Cite command: "
+                   (seq-mapcat #'car citar-latex-cite-commands)
+                   nil nil nil
+                   'citar-latex-cite-command-history
+		   citar-latex-default-cite-command nil))
 
 (defun citar-latex-is-a-cite-command (command)
   "Return element of `citar-latex-cite-commands` containing COMMAND."
--- a/citar.el
+++ b/citar.el
@@ -896,17 +896,22 @@ With prefix, rebuild the cache before offering candidates."
           (browse-url-default-browser link)
         (message "No link found for %s" key-entry)))))
 
+
 ;;;###autoload
-(defun citar-insert-citation (keys-entries)
+(defun citar-insert-citation (keys-entries &optional arg)
   "Insert citation for the KEYS-ENTRIES.
-With prefix, rebuild the cache before offering candidates."
-  (interactive (list (citar-select-refs
-                      :rebuild-cache current-prefix-arg)))
+
+Prefix ARG is passed to the mode-specific insertion function. It
+should invert the default behaviour for that mode with respect to
+citation styles. See specific functions for more detail."
+  (interactive (list (citar-select-refs) ;; key-entries
+		     current-prefix-arg)) ;; arg
   (citar--major-mode-function
    'insert-citation
    (lambda (&rest _)
      (message "Citation insertion is not supported for %s" major-mode))
-   (citar--extract-keys keys-entries)))
+   (citar--extract-keys keys-entries)
+    arg))
 
 (defun citar-insert-edit (&optional arg)
   "Edit the citation at point."


roshanshariff:
Instead of having two variables, `citar-latex-prompt-for-cite-style` and `citar-latex-default-cite-command`, does it make sense to have only one? Then if it is nil, you would always prompt and otherwise only prompt when there's a prefix argument?

Also, to make it a local variable you should declare it using `defvar-local` instead of `defcustom`. This way it can have a different value in each buffer, possibly specified as a file-local or dir-local variable.

bdarcus:
> Instead of having two variables, `citar-latex-prompt-for-cite-style` and `citar-latex-default-cite-command`, does it make sense to have only one? Then if it is nil, you would always prompt and otherwise only prompt when there's a prefix argument?
> 
> Also, to make it a local variable you should declare it using `defvar-local` instead of `defcustom`. This way it can have a different value in each buffer, possibly specified as a file-local or dir-local variable.

This raises a couple of related questions for me:

1. if `defvar-local`, then how does the user set the default command for the local buffer? Does `citar-insert-citation` prompt the first time? Always, or only if with a prefix arg?
2. In general, including here, should we distinguish bibtex, natbib, biblatex? Or if 1 is the right solution, does that not matter?

roshanshariff:
> In general, including here, should we distinguish bibtex, natbib, biblatex?

I think we should have a completing-read based `citar-latex-select-style` just like we have in org mode, which can offer the citation macros that make sense with the current packages. For now it can just offer all citation commands like `citar-latex-insert-citation` already does.

> how does the user set the default command for the local buffer?

That's a good question, and there are a few different options for the UI. Maybe users will want some combination of these?
1. Prompt always (current behaviour of `citar-latex-insert-citation`)
2. Prompt only with prefix arg, and otherwise use default.
   1. Have a separate command to set the per-buffer default (which can also offer to add that default as a file-local variable so you don't have to use it every time).
   2. Set the default every time a citation is inserted, so without a prefix arg it just uses the last-used citation command in that buffer.
   3. With repeated prefix arg, prompt for the citation command and set it as the default. (I'm wary of this because it seems to overload the prefix args too much).
   4. Optionally, have a separate global (but mode-specific) default that is used if there's no per-buffer default. This can be set to "cite" by default.

What do you think? Can you think of other interactions that might make sense? Based on what the UI should be, we can figure out the appropriate combination of global and buffer-local variables we need.

I do think we should try to keep the behaviour as similar as possible in (at least) org and latex, and ideally also have code commonality between them if possible.

bdarcus:
> > In general, including here, should we distinguish bibtex, natbib, biblatex?
> I think we should have a completing-read based `citar-latex-select-style` just like we have in org mode, which can offer the citation macros that make sense with the current packages. 

+1

> > how does the user set the default command for the local buffer?
> 
> That's a good question, and there are a few different options for the UI. Maybe users will want some combination of these?
> 
> 1. Prompt always (current behaviour of `citar-latex-insert-citation`)
> 2. Prompt only with prefix arg, and otherwise use default.
>    
>    1. Have a separate command to set the per-buffer default (which can also offer to add that default as a file-local variable so you don't have to use it every time).
>    2. Set the default every time a citation is inserted, so without a prefix arg it just uses the last-used citation command in that buffer.
>    3. With repeated prefix arg, prompt for the citation command and set it as the default. (I'm wary of this because it seems to overload the prefix args too much).
>    4. Optionally, have a separate global (but mode-specific) default that is used if there's no per-buffer default. This can be set to "cite" by default.
> 
> What do you think? Can you think of other interactions that might make sense? Based on what the UI should be, we can figure out the appropriate combination of global and buffer-local variables we need.

I'm not really sure ATM; need to think on it more, and see what @Hugo-Heagren has to say.

> I do think we should try to keep the behaviour as similar as possible in (at least) org and latex, and ideally also have code commonality between them if possible.

Yes, with the caveat that there are differences among default latex, natbib and biblatex, and org.

The first one has two commands, for example, the last has a default (no style) which will do the reasonable thing across export backends.

Hugo-Heagren:
> Instead of having two variables, `citar-latex-prompt-for-cite-style` and `citar-latex-default-cite-command`, does it make sense to have only one? Then if it is nil, you would always prompt and otherwise only prompt when there's a prefix argument?

I don't think so. One of the uses of `citar-latex-default-cite-command` is to set the default choice (that is, the `def` argument in completing-read) when prompting for a cite command. Users who always want to be prompted might still want to control the default here. At least that was my rationale, does anyone else have opinions?

> Also, to make it a local variable you should declare it using defvar-local instead of defcustom. This way it can have a different value in each buffer, possibly specified as a file-local or dir-local variable.

Thanks for flagging this, I don't know much about local variables. Though for what it's worth, I think `defcustom` is usually the best way to go for variables, and [there seems](https://www.gnu.org/software/emacs/manual/html_node/elisp/File-Local-Variables.html) to be a way to allow custom variables to be set locally. I'll have at that.

bdarcus:
> Assuming this is how people use citar, it would be better to have a single variable which is set globally and can be _re_set locally, than expecting it to be set locally for every file.

That's probably a reasonable POV.

Hugo-Heagren:
Just pushed some new work. Minor points:
- make all the config variables file-local-safe, though still defined with `defcustom`. I've tested this, and they do change the behaviour file locally.
- Implement prefix arg as a switch for author suppression in markdown. This makes sense to me and I know I would use it if I wrote much in markdown -- I asked and nobody seemed to object to the idea, so there's an implementation.
- separate the functionality for selecting a citation command in latex, like it is separated in org. Refactor as required.

The state of the main functionality is this. Setting the relevant variable will make make `citar-insert-citation` prompt or not, by default in LaTeX, and setting the default command will make insertion default to that command when it doesn't prompt (and use that command as the default candidate when it *does* prompt). Passing a prefix arg simply inverts the behavior for that invocation.

I have used very similar code for the org code, hoping to replicate the functionality, but it doesn't work for me. It might be a problem with my machine. At the moment, org mode consistently doesn't prompt for a style by default, and consistently does prompt with a prefix arg, regardless of the values of the variables. It also seems to ignore the default style variable. I just don't get what's going on here, could someone with better elisp have a look and tell me what's going wrong?

roshanshariff:
Excellent! I didn't know that `defcustoms` could be used as file-local variables, and agree that having two variables makes sense in the way you've envisioned it.

For `citar-org-insert-edit` I think the cleanest approach would be to modify the code of `org-cite-make-insert-processor` very slightly so it can insert a different citation style. Maybe it makes sense to defer this for now, and we can figure out the best way to upstream that change in Org?

It should be fairly easy to implement this for `citar-org-insert-citation` though, by changing the `style` variable from a function parameter to a let-binding and using `citar-org-select-style` to prompt for its value when inserting a new citation.

bdarcus:
> * Implement prefix arg as a switch for author suppression in markdown. This makes sense to me and I know I would use it if I wrote much in markdown -- I asked and nobody seemed to object to the idea, so there's an implementation.

I'd have to test it to have a firmer view, but my impulse is it's not worth worrying about this for pandoc/markdown, given how trivial the syntax there is. I have a hunch it's just easier to add a`-` when needed.

Maybe that's my slight anti-prefix arg POV though :shrug:.

bdarcus:
> For `citar-org-insert-edit` I think the cleanest approach would be to modify the code of `org-cite-make-insert-processor` very slightly so it can insert a different citation style.

I've not followed this closely today, but WDYM?

Hugo-Heagren:
> For citar-org-insert-edit I think the cleanest approach would be to modify the code of org-cite-make-insert-processor very slightly so it can insert a different citation style. Maybe it makes sense to defer this for now, and we can figure out the best way to upstream that change in Org?

What version of org are you currently on? I've just checked and the docstring of `org-cite-make-insert-processor` in my emacs says this, which seems to be what you're suggesting:

```
org-cite-make-insert-processor is a compiled Lisp function in ‘oc.el’.

(org-cite-make-insert-processor SELECT-KEY SELECT-STYLE)

Build a function appropriate as an insert processor.

SELECT-KEY is a function called with one argument.  When it is nil, the function
should return a citation key as a string, or nil.  Otherwise, the function
should return a list of such keys, or nil.  The keys should not have any leading
"@" character.

SELECT-STYLE is a function called without any argument.  It should return a
style string, or nil.

The return value is a function of two arguments: CONTEXT and ARG.  CONTEXT is
either a citation reference, a citation object, or nil.  ARG is a prefix
argument.

The generated function inserts or edit a citation at point.  More specifically,

  On a citation reference:

    - on the prefix or right before th "@" character, insert a new reference
      before the current one,
    - on the suffix, insert it after the reference,
    - otherwise, update the cite key, preserving both affixes.

    When ARG is non-nil, remove the reference, possibly removing the whole
    citation if it contains a single reference.

  On a citation object:

    - on the style part, offer to update it,
    - on the global prefix, add a new reference before the first one,
    - on the global suffix, add a new reference after the last one,

  Elsewhere, insert a citation at point.  When ARG is non-nil, offer to complete
  style in addition to references.
```

Hugo-Heagren:
> I have used very similar code for the org code, hoping to replicate the functionality, but it doesn't work for me. It might be a problem with my machine. At the moment, org mode consistently doesn't prompt for a style by default, and consistently does prompt with a prefix arg, regardless of the values of the variables. It also seems to ignore the default style variable. I just don't get what's going on here, could someone with better elisp have a look and tell me what's going wrong?

(I think I actually forgot to push this, so it's all a bit moot) I have discovered the problem. `org-cite-make-insert-processor`

bdarcus:
For context, @Hugo-Heagren, when we've discussed possibly rewriting the org insert processor code to bypass `org-cite-make-insert-processor`, I've simply noted that another option is that we submit a patch upstream.

That may or may not apply here. I don't entirely understand the state of this part of the discussion :-)

Hugo-Heagren:
I think this is finished. The state at the moment:
- My original intention with regard to citar-latex has been fulfilled: there are two variables controlling:
  1. whether to prompt
  2. what to use as a default when prompting, or to insert *by* default when *not* prompting
  
  Using a prefix argument with `citar-insert-citation` in a (La)TeX buffer inverts the behaviour of the first variable.
- Using a prefix argument to control whether or not to prompt for a citation style is something *already* controlled natively by org mode, and (though I have had a look), citar cannot and should not override this cleanly. Much better to just submit a patch to org, so this PR doesn't touch the org functionality any more.
- Markdown doesn't support styles in individual citations, so don't touch markdown either.

bdarcus:
Thanks!

We can iterate on this.

PR #460, opened 2021-12-05
citar-file.el: Refactor file-finding functions

roshanshariff:
Implements the changes discussed in #456:

Split the functionality of `citar-file--possible-names` into two new functions:
- `citar-file--parse-file-field` to extract file names from an entry field
- `citar-file--directory-files` to find files in directories whose names start with keys.

Change `citar-file--files-for-entry`, `citar-file--files-for-multiple-entries`, and `citar-file--get-note-filename` to use these functions.

Add the function `citar-file--has-file-p`, which scans the relevant directories once and returns a predicate for the existence of associated files. Use this function in `citar--format-candidates` to efficiently add "has:files" and "has:notes" indicators.

With this implementation, the "has:file" indicator can be incorrect when `citar-file-find-additional-files` is set to `t` or some separator that is present in the citation key. In particular, setting it to t will only find files that don't have extra text after the key, just as if it was set to nil. If it is set to a separator that is present in a citation key, then it can cause false positives or negatives for filenames that contain additional text after the key. These issues only affect the indicator display and don't affect actually opening files or notes.

### DIFF
--- a/citar-file.el
+++ b/citar-file.el
@@ -74,13 +74,18 @@ will open, via `citar-open-notes'."
 
 (defcustom citar-file-find-additional-files nil
   "Find additional library files starting with reference key.
+
 If t, all files whose base name starts with the reference key and
 whose extension is listed in `citar-file-extensions' are located
 by the functions `citar-open-library-files' and
 `citar-open-notes'.  If nil, only files with the naming scheme
 \"<key>.<extension>\" are located.  Otherwise, its value is a
 regular expression specifying how the key is separated from the
-rest of the filename."
+rest of the filename.
+
+Note: setting this variable to t or a separator that is present
+in citation keys can lead to incorrect \"has:note\" indicators in
+the displayed citation key candidates."
   :group 'citar
   :type '(choice (const :tag "Ignore additional files" nil)
                  (const :tag "Find all files starting with key" t)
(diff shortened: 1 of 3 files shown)


roshanshariff:
I couldn't reproduce the issue you had with `customize-variable` for `citar-file-find-additional-files`. Choosing "Filename separator" in the value menu lets me type in an arbitrary string as the separator regexp.

bdarcus:
This is confusing. Since it's important, can you revise it to clarify?

> With this implementation, the "has:file" indicator can be incorrect when citar-file-find-additional-files is set to t or some separator that is present in the citation key. In particular, setting it to t will only find files that don't have extra text after the key ...

And, I guess, need a bit on it in the README?

citar-file.el
```diff
@@ -74,13 +74,18 @@ will open, via `citar-open-notes'."
...
 by the functions `citar-open-library-files' and
 `citar-open-notes'.  If nil, only files with the naming scheme
 \"<key>.<extension>\" are located.  Otherwise, its value is a
 regular expression specifying how the key is separated from the
-rest of the filename."
+rest of the filename.
+
+Note: setting this variable to t or a separator that is present
+in citation keys can lead to incorrect \"has:note\" indicators in
+the displayed citation key candidates."
```
bdarcus, line 88:
```suggestion
Note: setting this variable to t or a separator that is present
in citation keys can lead to incorrect \"has:note\" indicators in
the displayed citation key candidates; for example when ..."
```

Not sure exactly what you want to say here, but the way you've written this sounds like it just isn't practical for notes.

Maybe "the specified separator (for example a hyphen) is included in the key."?

citar-file.el
```diff
@@ -124,46 +129,118 @@ Example: ':/path/to/test.pdf:PDF'."
...
+(defun citar-file--parse-file-field (entry fieldname &optional dirs)
+  "Return files listed in FIELDNAME of ENTRY.
+File names are expanded relative to the elements of DIRS."
+  (unless dirs (setq dirs (list "/")))  ; make sure DIRS is non-nil
+  (when-let ((filefield (citar-get-value fieldname entry)))
+    (seq-mapcat (lambda (parser)
+                  (funcall parser dirs filefield))
+                citar-file-parser-functions)))
+
+(defun citar-file--filename-regexp (keys extensions &optional find-additional)
```
bdarcus, line 141: So this is a function which _creates_ a regexp? Perhaps the below would be more clear?

```suggestion
(defun citar-file--filename-make-regexp (keys extensions &optional find-additional)
```

bdarcus (commented):
I haven't had a chance to test it yet, but left a couple of comments, with one bigger question.

citar-file.el
roshanshariff: Sorry, that was a typo from late-night coding. That should be "has:files" and I'll make the text more comprehensive about what the potential issue can be. The issue also applies to "has:notes" but it's likely to be a problem because people probably just have one file named after the cite key.

bdarcus: But am I correct it's only an issue in situations where you have a key like `foo-bar`, where you specify a hyphen as  separator?

Isn't it not a non-issue if space is the separator?

roshanshariff: Yes, exactly, there's only a problem if you use (say) a hyphen as a separator and also have hyphens in your cite key. So, for example, if you have a key called `foo-bar` and a file called `foo-bar-document.pdf`, then it'll think that document belongs to key `foo` and won't show a "has:note" on key `foo`.

And even then, if the document is called `foo-bar.pdf` (without any additional text) it will show "has:note" on both keys, "foo" and "foo-bar".

bdarcus: Yeah, so the language of that docstring currently is misleadingly bleak :-)

Really we should default (well, not sure it's possible, but) to space if non-nil, and urge caution when using other alternatives (hyphens, colons, etc.).

And this line needs to be removed, doesn't it, since there's no way to match in this case? E.g there _must_ be a separator for it to work?

```elisp
```

roshanshariff: I think using space as a default is fine, but maybe it's okay to leave the "t" option in as non-default? After all, it works fine when you're actually opening files or notes, and only sometimes has problems with the "has:files" indicator. If somebody is fine with that, then the option is available to them.

bdarcus: The discrepancy between displayed info and behavior, which is part of what we are wanting to fix here, seems a bit hackish to me.

I don't know how strongly I feel about this, but my impulse is to remove it and see what input we get. It just seems like this would be odd to have to explain in the docs.

You have any thoughts on this @localauthor?

localauthor: I haven't had a chance to test this out, let alone parse all the changes. Hash tables are a bit opaque to me.

One thing I've just found, though, is that `citar--open-note` makes a call to the value of `citar-file-open-function`, which I have set to `citar-file-open-external`, to open pdfs externally. The problem with calling this for notes is that it takes a while for Emacs to call the external application set to open notes which is.... Emacs. So opening a note in Emacs from Emacs takes a painful 7 seconds. Yikes!

roshanshariff: I've pushed a new commit with those changes.

bdarcus: @localauthor - the main reason I tagged you is on the question of whether, with this enhancement, it makes any sense to allow `citar-file-find-additional-files` to be `t`, which by definition would mean the display and the behavior could be inconsistent.

Seems to me either we need to require people to use a separator if they want to use additional-files, or we need a better way to associate key with files to remove that limitation (which AFAIK doesn't exist).

WDYT?

localauthor: Ok, just catching up now. It seems right to remove `t` as an option and default to space. Keeping `t` suggests its a boolean, which it really isn't anymore --- so `t` just seems to cause unnecessary problems/complications.

And, yes, a warning about using other separators seems warranted as well.

To make this function a bit clearer to the user, might it make sense to change the name of the defcustom to something more accurate/precise, something that would indicate that its value is a separator string? I'm thinking something like `citar-file-extended-filename-separator`?

The use of `additional-files` as a descriptor for this feature is just a carry-over from bibtex-completion anyway, and it doesn't really describe what's being done here at this point.

roshanshariff: Sounds good, I'll make those changes. It occurred to me that if someone *really* wants they can get the effect of `t` by setting the separator to the empty string, but there's no reason to suggest that or document it.

I had an idea for making some of these separator-related downsides go away, but I need to do some experimenting to see what the performance impact is like. The idea is to construct a giant regexp that matches all the possible citation keys using the `regexp-opt` function. Emacs should be able to compile that regexp to something that has better-than-linear-time performance. If this works out, it'll be only a small change to this PR.

bdarcus: > ... if someone really wants they can get the effect of t by setting the separator to the empty string, but there's no reason to suggest that or document it.

Right.

> I had an idea for making some of these separator-related downsides go away, but I need to do some experimenting to see what the performance impact is like. The idea is to construct a giant regexp that matches all the possible citation keys using the regexp-opt function. Emacs should be able to compile that regexp to something that has better-than-linear-time performance. If this works out, it'll be only a small change to this PR.

So to be clear, you suggest merging this now, and will experiment with that separately, as possible future small PR?

roshanshariff: I've pushed a new commit with the changes we've discussed so far, and I think it's ready to merge (I've tested it, but it's probably good for someone else to verify basic functionality). And yeah, we can defer the other potential improvement to another PR, which I'll probably get to on the weekend.

I've also properly fixed #465 now, using `find-file` for both new and existing notes.

bdarcus: OK, cool!

Note it looks like you'll have some conflicts you'll need to resolve. Sorry about that.

But it looks like you maybe able to just do `git rebase main` and resolve them.

bdarcus: Actually, I just resolved the one conflict.

roshanshariff: It looks like that conflict resolution reverted the fix for #465. I'll push a new version with the correct resolution.

bdarcus: Ah right; we have conflicting fixes! Thanks.

bdarcus:
Merged. Thanks!

oatmealm:
How would I go about using this to find files where the citekey is prefixed, i.e. @citekey.org? Is this what this custom setting ment for?

roshanshariff:
@oatmealm, if you remove the @ at the beginning of the filenames so they're named like `citekey.org`. It should just work. If you have something like `citekey - some other text.org` then you could customize `citar-file-additional-files-separator`.

oatmealm:
thanks @roshanshariff ! I have both... the ones prefixed with @ contain some additional information from the main file, which is citekey.org. So the @ files are sort of a sidecar...

roshanshariff:
If you're okay renaming the sidecar files so they also start with just the cite key, they should be picked up by Citar. You could use suffixes to distinguish them from the main file.

Otherwise, it might be possible to advise the regexp function to make it ignore leading @ characters, but that's probably more brittle.

Issue #459, opened 2021-12-04
Export buffer citations to a bib or csl json file


pemocarlo:
**Is your feature request related to a problem? Please describe.**
I normally use a global bibliography file, but for some documents I prefer to only use a local citations source with the entries used there (e.g to share a standalone .tex + .bib with others)

**Describe the solution you'd like**
An interactive command to create a citations file containing only the entries in the current buffer

**Describe alternatives you've considered**
Keep a local bibliography file for specific documents, which is a hassle if you are already using a global one

**Additional context**

bdarcus:
> An interactive command to create a citations file containing only the entries in the current buffer.

How would one set the output format with such a command? Would it prompt for that, for example? Also for a filename, perhaps with a reasonable default?

And presumably it'd work across the major-modes?

pemocarlo:
Thank you @bdarcus for your reply!

> How would one set the output format with such a command? Would it prompt for that, for example? If yes, would that include the ability to convert among formats; for example, from bibtex to json?
Or would it simply copy from the global bib in whatever format that is (easier obviously)?

I would go with the second (easier) option, that would cover my use case.
Conversion is a nice utility, but maybe out of the scope of this package?

> Also for a filename, perhaps with a reasonable default?

Yes, great! :-)

bdarcus:
> Conversion is a nice utility, but maybe out of the scope of this package?

Yes.

Note for me, or whoever implements this:

Functions to get the keys in a buffer:

1. latex: `reftex-all-used-citation-keys`
2. org: doesn't currently have one; `org-cite-list-keys` is limited to the export process (though I might ask about this)
3. markdown: doesn't exist

localauthor:
Unless I'm missing something, it seems that `citar-markdown-list-keys` also accurately lists org-cite keys.

Maybe someone can confirm by testing out with the latest commit?  If you're using `citar-org`, you'll have to make sure `citar-markdown.el` is loaded first.

On that last point, if all works as expected in org, some complications arise with regard to ensuring the availability of necessary functions and variables. `citar-markdown-list-keys` and the defconst `citar-markdown-citation-key-regexp` are in `citar-markdown.el`, and users of `citar-org` functions might not have that file loaded.

Any elegant solutions come to mind?

bdarcus:
The commit I just merged should address this.

I have not tested it yet (!), so let us know.

localauthor:
[...]

PR #466, opened 2021-12-07
Add citar-export-buffer-citations-to-file

localauthor:
Adds `citar-export-buffer-citations-to-file` as suggested in issue #459, @pemocarlo.

Also adds `citar-markdown-list-keys`.

EDIT: (see comment below)
~~Currently only available `latex-mode` and `markdown-mode`.~~

~~Support for `org-mode` would require a function that lists all org-citations in buffer, which is more than I can undertake at this point.~~

Thoughts on function name? Other issues?

### DIFF
--- a/citar-markdown.el
+++ b/citar-markdown.el
@@ -124,5 +124,19 @@ citation."
              (cons (nreverse keys) (cons startpos endpos))))))
      (reverse (nth 9 (syntax-ppss))))))
 
+;;;###autoload
+(defun citar-markdown-list-keys ()
+  "Returns a list of all keys from markdown citations in buffer."
+  (interactive)
+  (save-match-data
+    (let ((pos 0)
+          matches)
+      (while (string-match citar-markdown-citation-key-regexp (buffer-string) pos)
+        (setq pos (match-end 0))
+        (push (car (split-string (match-string-no-properties 0 (buffer-string))
+                                 nil t ".*@\\|\\,.*\\]"))
+              matches))
+      (delete-dups (nreverse matches)))))
+
 (provide 'citar-markdown)
 ;;; citar-markdown.el ends here
--- a/citar-org.el
+++ b/citar-org.el
@@ -331,6 +331,14 @@ With optional argument FORCE, force the creation of a new ID."
                  (<= (point) (cdr bounds)))
         element))))
 
+(defun citar-org-list-keys ()
+  "List citation keys in the org buffer."
+  (let ((org-tree (org-element-parse-buffer)))
+    (delete-dups
+     (org-element-map org-tree 'citation-reference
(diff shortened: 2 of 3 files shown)


localauthor:
Unless I'm missing something, it seems that citar-markdown-list-keys also accurately lists org-cite keys.

Maybe someone can confirm by testing out with the latest commit? If you're using citar-org, you'll have to make sure citar-markdown.el is loaded first.

On that last point, if all works as expected in org, some complications arise with regard to ensuring the availability of necessary functions and variables. citar-markdown-list-keys and the defconst citar-markdown-citation-key-regexp are in citar-markdown.el, and users of citar-org functions might not have that file loaded.

Any elegant solutions come to mind?

bdarcus:
I was wondering if this should rely on the major mode function infrastructure?

As for org, I can provide a function for this.

bdarcus:
> As for org, I can provide a function for this.

```elisp
(defun citar-org-list-keys ()
  "List citation keys in the org buffer."
  (let ((org-tree (org-element-parse-buffer)))
    (delete-dups
     (org-element-map org-tree 'citation-reference
       (lambda (r) (org-element-property :key r))
       org-tree))))
```

localauthor:
> So could add `list-keys` to the major mode functions, etc?

Just pushed these changes.

Only questions I have is about the function name. Some alts:
- `citar-export-buffer-citations`
- `citar-export-local-bib-file`

Or some variation.

citar-markdown.el
```diff
@@ -124,5 +124,17 @@ citation."
...
+;;;###autoload
+(defun citar-markdown-list-keys ()
+  "Returns a list of all keys from markdown citations in buffer."
+  (interactive)
+  (save-match-data
+    (let ((pos 0)
+          matches)
+      (while (string-match citar-markdown-citation-key-regexp (buffer-string) pos)
+        (setq pos (match-end 0))
+        (push (car (split-string (match-string-no-properties 0 (buffer-string)) nil t ".*@\\|\\,.*\\]")) matches))
```
bdarcus, line 136: This line is way too long; try to keep it under 80 characters.

bdarcus (commented):
> Only questions I have is about the function name. Some alts:
>
> citar-export-buffer-citations
> citar-export-local-bib-file

I _think_ I prefer the second one above, since that's what the user literally asked for.

bdarcus:
Also, on the commit message, it says "see also #459", which will just end up linking.

But if you use a word like "close" instead, github with auto close the issue.

bdarcus:
So the function name question?

https://github.com/bdarcus/citar/pull/466#pullrequestreview-825186269

bdarcus:
Thanks!

PR #467, opened 2021-12-07
fix: citar--insert-bibtex and citar-export-local-bib-file

localauthor:
Sets bibtex-dialect before running functions from bibtex.el

### DIFF
--- a/citar-latex.el
+++ b/citar-latex.el
@@ -30,6 +30,7 @@
 (require 'citar)
 (require 'tex nil t)
 (require 'reftex-parse)
+(require 'reftex-cite)
 
 (defvar citar-major-mode-functions)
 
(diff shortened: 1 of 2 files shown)


bdarcus:
> Sets bibtex-dialect before running functions from bibtex.el

Should you check if `bibtex-dialect` is set first, so that a user is not prompted if it is?

Or is that not necessary?

localauthor:
> Should you check if `bibtex-dialect` is set first, so that you don't prompt if it is?
> 
> Or is that not necessary?
> 
> And should we suggest people consider setting this globally in their config?

Both seem unnecessary, at least from my testing. Hard to say why though. It's a somewhat labyrinthine codebase.

> Finally, does that variable only impact parsing (like, this is why I got an empty file, because dialect was set to bibtex but my file is biblatex?), or does it enable conversion between them? Just curious ...

Again, not entirely sure about this. I found that the empty file was a result of another variable, `bibtex-entry-type`, being `nil` prior to 'bibtex-set-dialect' being called. This prevented _any_ entries from being found, not just entries of a certain type.

PR #470, opened 2021-12-08
Clarify citar-citeproc settings and minor adjustments

localauthor:
Expand commentary in citar-citeproc.el
- Clarify instructions for use
- Add info about CSL style and locale files

Allow 'citar-citeproc-csl-style' to be set to filename only, not full path
- This requires the file to be located in 'citar-citeproc-csl-styles-dir'
- Full path also accepted

Add changes suggested by roshanshariff in #455

### DIFF
--- a/citar-citeproc.el
+++ b/citar-citeproc.el
@@ -18,24 +18,36 @@
 ;;; Commentary:
 
 ;;  Provides functions for formatting bibliographic references according to
-;;  CSL styles, using 'citeproc-el'.
+;;  Citation Style Language (CSL) styles, using 'citeproc-el'. For
+;;  information on using CSL, see <http://www.citationstyles.org>.
 
-;;  To use: load this file, set the required directory paths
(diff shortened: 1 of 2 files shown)


bdarcus:
Thanks. I just squash merged so I could shorten the first line of the commit message to 50 characters or less.

localauthor:
I also realize I forgot about adding 'citar-copy-reference' to the 'citar-map' and 'citar-citation-map' embark keymaps.

But the question of a sensible binding leaves me stumped.

In my personal config, I use "r" for `citar-copy-reference`, replacing `citar-refresh`, since I usually call that through M-x rather than with embark. In my mind, it makes sense to have the reference functions use "r" and "R". But then what for `citar-refresh`?

Thoughts?

bdarcus:
I forgot to check that too.

I was actually thinking the other day that maybe we don't need a binding for `citar-refresh`, particularly given we have the filenotify stuff?

And now that I think about it, do we need `citar-insert-citation` at point?

PR #474, opened 2021-12-09
Add back `citar-insert-citation` to Embark keymaps

roshanshariff:
Accidentally removed in #473.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -304,6 +304,7 @@ of all citations in the current buffer."
 (defvar citar-citation-map
   (let ((map (make-sparse-keymap)))
     (define-key map (kbd "i") #'citar-insert-edit)
+    (define-key map (kbd "c") #'citar-insert-citation)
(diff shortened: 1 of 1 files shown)


roshanshariff:
Ah, I just noticed your 425301d. Updated this PR to add back the binding to `citar-citation-map` as well.

bdarcus:
What value is it there though?

The menu is only available on in an existing citation, where citar-insert-edit is more appropriate?

roshanshariff:
`citar-insert-edit` tries to change the thing at point, and only if the cursor is positioned very carefully it will add one key to an existing citation. If I just want to add multiple keys to the citation and not edit what's already there, `citar-insert-citation` is a lot easier and less trouble.

roshanshariff:
Never mind, this doesn't work currently because running `citar-insert-citation` as an Embark action enters the existing keys into the `citar-select-ref` prompt. So yeah, I'll close this PR.

PR #482, opened 2021-12-12
Add :filter argument to `citar-select-ref`

### LINKED ISSUE #272: Add `:filter` arg to select-keys
Maybe something like ...

```elisp
(seq-filter 
  (lambda (cand) 
    (string-match filter (car cand))) 
  candidates)
```

... so could do:

```elisp
(bibtex-actions-select-keys :filter "has:notes")
```

Performance may not be ideal with large collections, but flexibility would be high.

I don't think I would do this by default in this package, but it might be convenient to have.

roshanshariff:
For example, this can be used to select a reference that has a note:
``` emacs-lisp
(citar-select-ref :filter (citar-make-note-predicate))
```
or to select a reference that has an associated file:
``` emacs-lisp
(citar-select-ref :filter (citar-make-file-predicate))
```
or to select a reference that has a "keywords" field containing "ethics":
```emacs-lisp
(citar-select-ref :filter (lambda (_key entry)
                            (when-let ((keywords (assoc-default "keywords" entry)))
                              (string-match-p "ethics" keywords))))
```
Also simplify and update documentation for functions in `citar-file.el`

Closes #272.

### DIFF
--- a/citar-file.el
+++ b/citar-file.el
@@ -206,16 +206,21 @@ need to scan the contents of DIRS in this case."
                  (puthash key (nreverse filelist) files))
                files))))
 
-(defun citar-file--has-file-p (dirs extensions &optional additional-sep entry-field)
+(defun citar-file--make-file-predicate (dirs extensions &optional entry-field)
   "Return predicate testing whether a key and entry have associated files.
 
 Files are found in two ways:
 
-- In DIRS using `citar-file--directory-files`; see its
-documentation for the meaning of EXTENSIONS and ADDITIONAL-SEP.
+- By scanning DIRS for files with EXTENSIONS using
+  `citar-file--directory-files`, which see.  Its ADDITIONAL-SEP
+  argument is taken from `citar-file-additional-files-separator`.
 
-- In the entry field ENTRY-FIELD, when it is non-nil."
-  (let ((files (citar-file--directory-files dirs nil extensions additional-sep)))
+- When ENTRY-FIELD is non-nil, by parsing the entry field it
+  names using `citar-file--parse-file-field`; see its
+  documentation.  DIRS is used to resolve relative paths and
+  non-existent files are ignored."
+  (let ((files (citar-file--directory-files dirs nil extensions
+                                            citar-file-additional-files-separator)))
     (lambda (key entry)
       (or (car (gethash key files))
           (when entry-field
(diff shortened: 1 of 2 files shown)


citar-file.el
```diff
@@ -206,16 +206,21 @@ need to scan the contents of DIRS in this case."
                  (puthash key (nreverse filelist) files))
                files))))
 
-(defun citar-file--has-file-p (dirs extensions &optional additional-sep entry-field)
+(defun citar-file--make-file-predicate (dirs extensions &optional entry-field)
```
bdarcus, line 209: Why this change in name?

The former seems more consistent to emacs predicate naming conventions?

bdarcus (commented):
Thanks!

I just left one question/comment.

This also reminds me we need to address #438 soon. Should the predicate functions be public, for example?

citar-file.el
roshanshariff: I wanted to make it clear that this function isn't a predicate, but that it *returns* a predicate. Usually the `foo-p` convention is used for the predicate functions themselves.

roshanshariff:
I realized that `citar-file--make-file-predicate` can potentially do a whole bunch of file existence checks. I'm adding a couple of lines of code to cache those, which might be especially important if it's used as a predicate for `completing-read` rather than just to format the candidates.

I'll remove the "draft" status of this PR when I've pushed that change.

citar-file.el
bdarcus: Been a long day grading, but ...

Why the need to generate the predicates?

Wouldn't the resulting code be more clear to simply write them?

roshanshariff: This function does the job of scanning the library directories for all possible files and caching them in a hash table. The predicate it returns is very efficient, since it just checks that hash table instead of actually scanning the filesystem every time it's called.

bdarcus: Ok,. but I'm talking about the example you gave; this:

``` elisp
(citar-select-ref :filter (citar-file--make-file-predicate citar-library-paths
                                                           citar-file-extensions
                                                           citar-file-variable))
```

... vs something like:

``` elisp
(citar-select-refs :filter #'citar-has-notes-p)
```

So can we somehownhabe both: the efficiency benefit + ease-of-use and clarity?

roshanshariff: Sure, I could add some small functions like:
```emacs-lisp
(defun citar--make-file-predicate ()
  (citar-file--make-file-predicate citar-library-paths
                                   citar-file-extensions
                                   citar-file-variable))

(defun citar--make-note-predicate ()
  (citar-file--make-file-predicate citar-notes-paths
                                   citar-file-note-extensions))
```
These functions would have to go in `citar.el`, rather than `citar-file.el`, because they use the `citar-library-paths` and `citar-notes-paths` variables. Then the example code would become
```emacs-lisp
(citar-select-refs :filter (citar-make-note-predicate))
```
In any case, the :filter function will have to be the result of a function call, rather than just the name of a function.

roshanshariff: I've pushed a new commit that adds these utility functions, and updated the PR opening comment to use them.

bdarcus: I'm still not thrilled with the `make-*-predicste` names.

Maybe I'm misunderstanding some subtlety, but WDYT about simply `citar-note-filter`?

It seems more explanatory, and is consistent, for example, with the parser functions.

roshanshariff: How about `citar-has-file` and `citar-has-note`? I prefer these to `citar-note-filter` because that sounds like a function that performs filtering, like `seq-filter`.

I'm pushing a new commit with these names; it also updates the docstrings to give usage guidance.

And yes, this also works with `citar-select-refs`, since that's now just a wrapper calling `(citar-select-ref :multiple t ...)`.

bdarcus: That works; merged!

roshanshariff:
Oops, that last commit didn't change the name of `citar-file--make-file-predicate` in one place. Pushing a fix now.

PR #483, opened 2021-12-12
citar-markdown-list-keys: Use `citar-markdown-citation-key-regexp`

roshanshariff:
Correctly handle Pandoc citation keys with braces:
- In `@{Foo_Bar}:`, the citation key is `Foo_Bar`, not `{Foo_Bar}:`
- Also respect Pandoc's punctuation rule: in `@Foo_bar.baz.`, the citation key is `Foo_bar.baz`.

See https://pandoc.org/MANUAL.html#citation-syntax for more examples.

### DIFF
--- a/citar-markdown.el
+++ b/citar-markdown.el
@@ -130,10 +130,8 @@ citation."
   (save-excursion
     (let (matches)
       (goto-char (point-min))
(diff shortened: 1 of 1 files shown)


localauthor:
Good edit! Didn’t know how to do that ;)

PR #487, opened 2021-12-14
Add support for Embark multi-actions

### LINKED ISSUE #486: Support embark multitarget actions
Omar just merged support for this, discussed in #477:

https://github.com/oantolin/embark/commit/b17af024d92e5c8532ed94f157e65aa7f4cda334

So we need to test this and get him feedback, but also make sure citar works with it. Once we get it working as expected, we can also add the pdf-search function, which would rely on this.

I'm not sure why, but this is the result I get.

```elisp
Debugger entered--Lisp error: (wrong-type-argument listp "huck1975a")
  #f(compiled-function (x) #<bytecode -0x1433a9947753cda6>)("huck1975a")
  mapcar(#f(compiled-function (x) #<bytecode -0x1433a9947753cda6>) ("huck1975a" "huck1975b" "huck1975c" "huck1975d" "huck1975e" "huck1975f" "huck1975" "huck1975g"))
  citar-citeproc-format-reference(("huck1975a" "huck1975b" "huck1975c" "huck1975d" "huck1975e" "huck1975f" "huck1975" "huck1975g"))
  citar-copy-reference(("huck1975a" "huck1975b" "huck1975c" "huck1975d" "huck1975e" "huck1975f" "huck1975" "huck1975g"))
  #f(compiled-function () #<bytecode -0x1bcb76d099fa68a4>)()
  apply(#f(compiled-function () #<bytecode -0x1bcb76d099fa68a4>) nil)
  #f(compiled-function () #<bytecode -0x1ebf6128d325eb1a>)()
  #f(compiled-function () #<bytecode -0x14e4eee3a110c9d2>)()
```

Any ideas @localauthor @roshanshariff?

See https://github.com/oantolin/embark/issues/433

roshanshariff:
Introduces the following naming convention for function arguments:
 - KEYS-ENTRIES is a list whose elements are either KEY or (KEY . ENTRY) pairs.
 - KEY-ENTRY-ALIST is a list whose elements are (KEY . ENTRY) pairs
 - KEYS is a list whose elements are keys.

The existing function `citar--extract-keys` will transform KEYS-ENTRIES into KEYS.

The new function `citar--ensure-entries` will transform KEYS or KEYS-ENTRIES into a KEY-ENTRY-ALIST.

Almost all the user-facing Citar commands (and the non-interactive function `citar-run-default-action`) take KEYS-ENTRIES arguments and have been added to `embark-multitarget-actions`. These arguments are transformed into `KEY-ENTRY-ALIST` in the few places that need them.

Note: I've tested all the functions in as many conditions as I could think of, but some more testing couldn't hurt before merging.

Closes #486.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -169,7 +169,9 @@ manager like Zotero or JabRef."
   :type 'hook)
 
 (defcustom citar-default-action #'citar-open
-  "The default action for the `citar-at-point' command."
+  "The default action for the `citar-at-point' command.
+Should be a function that takes one argument, a list with each
+entry being either a citation KEY or a (KEY . ENTRY) pair."
   :group 'citar
   :type 'function)
 
@@ -706,7 +708,31 @@ If FORCE-REBUILD-CACHE is t, force reload the cache."
 
 (defun citar--extract-keys (keys-entries)
   "Extract list of keys from KEYS-ENTRIES alist."
-  (seq-map #'car keys-entries))
+  (seq-map (lambda (key-entry)
+             (if (consp key-entry) (car key-entry) key-entry))
+           keys-entries))
+
+(defun citar--ensure-entries (keys-entries)
+  "Return copy of KEYS-ENTRIES with every element a (KEY . ENTRY) pair.
+Each element of KEYS-ENTRIES should be either a KEY or a (KEY
+. ENTRY) pair.  If it is the former, look up the corresponding
+ENTRY and transform the element to a (KEY . ENTRY) pair."
+  (let ((candidates 'uninitialized))
+    ;; Get candidates only if some key has a missing entry, to avoid nasty
+    ;; recursion issues like https://github.com/bdarcus/citar/issues/286. Also
+    ;; avoids lots of memory allocation in the common case when all entries are
+    ;; present.
+    (seq-map
+     (lambda (key-entry)
+       (if (consp key-entry)
+           key-entry
+         (when (eq candidates 'uninitialized)
+           (setq candidates (citar--get-candidates)))
+         (cons key-entry
+               (cddr (seq-find (lambda (cand-key-entry)
+                                 (string= key-entry (cadr cand-key-entry)))
+                               candidates)))))
+     keys-entries)))
 
 ;;;###autoload
 (defun citar-insert-preset ()
@@ -802,8 +828,15 @@ into the corresponding reference key.  Return
 ;;;###autoload
 (with-eval-after-load 'embark
   (add-to-list 'embark-target-finders 'citar-citation-finder)
-  (add-to-list 'embark-target-finders 'citar-key-finder))
-
+  (add-to-list 'embark-target-finders 'citar-key-finder)
+  (when (boundp 'embark-multitarget-actions)
+    (dolist (command (list #'citar-open #'citar-open-notes
+                           #'citar-open-entry #'citar-open-link
+                           #'citar-open-library-files #'citar-attach-library-files
+                           #'citar-insert-bibtex #'citar-insert-citation
+                           #'citar-insert-reference #'citar-copy-reference
+                           #'citar-insert-keys #'citar-run-default-action))
+      (add-to-list 'embark-multitarget-actions command))))
 
 (with-eval-after-load 'embark
   (set-keymap-parent citar-map embark-general-map)
@@ -822,17 +855,18 @@ into the corresponding reference key.  Return
   (when (and citar-library-paths
              (stringp citar-library-paths))
     (message "Make sure 'citar-library-paths' is a list of paths"))
-  (let* ((files
-         (citar-file--files-for-multiple-entries
-          keys-entries
-          (append citar-library-paths citar-notes-paths)
-          ;; find files with any extension:
-          nil))
+  (let* ((key-entry-alist (citar--ensure-entries keys-entries))
+         (files
+          (citar-file--files-for-multiple-entries
+           key-entry-alist
+           (append citar-library-paths citar-notes-paths)
+           ;; find files with any extension:
+           nil))
          (links
           (seq-map
            (lambda (key-entry)
              (citar-get-link (cdr key-entry)))
-           keys-entries))
+           key-entry-alist))
          (resource-candidates (delete-dups (append files (remq nil links))))
          (resources
           (when resource-candidates
@@ -851,7 +885,7 @@ into the corresponding reference key.  Return
                  ('attach 'mml-attach-file)))
            (files
             (citar-file--files-for-multiple-entries
-             keys-entries
+             (citar--ensure-entries keys-entries)
              citar-library-paths
              citar-file-extensions)))
       (if (and citar-file-open-prompt
@@ -889,7 +923,7 @@ With prefix, rebuild the cache before offering candidates."
              (equal citar-format-note-function
                     'citar-org-format-note-default))
     (error "You must set 'citar-notes-paths' to open notes with default notes function"))
-  (dolist (key-entry keys-entries)
+  (dolist (key-entry (citar--ensure-entries keys-entries))
     (funcall citar-open-note-function (car key-entry) (cdr key-entry))))
 
 (defun citar--open-note (key entry)
@@ -902,13 +936,13 @@ With prefix, rebuild the cache before offering candidates."
     (funcall citar-format-note-function key entry file)))
 
 ;;;###autoload
-(defun citar-open-entry (key-entry)
-  "Open bibliographic entry associated with the KEY-ENTRY.
+(defun citar-open-entry (keys-entries)
+  "Open bibliographic entry associated with the first of KEYS-ENTRIES.
 With prefix, rebuild the cache before offering candidates."
   (interactive (list (citar-select-refs
                       :rebuild-cache current-prefix-arg)))
-  (let ((key (citar--extract-keys key-entry)))
-    (citar--open-entry (car key))))
+  (when-let ((key (car (citar--extract-keys keys-entries))))
+    (citar--open-entry key)))
 
 (defun citar--open-entry (key)
   "Open bibliographic entry asociated with the KEY."
@@ -922,9 +956,8 @@ With prefix, rebuild the cache before offering candidates."
 With prefix, rebuild the cache before offering candidates."
   (interactive (list (citar-select-refs
                       :rebuild-cache current-prefix-arg)))
-  (let ((keys (citar--extract-keys keys-entries)))
-    (dolist (key keys)
-      (citar--insert-bibtex key))))
+  (dolist (key (citar--extract-keys keys-entries))
+    (citar--insert-bibtex key)))
 
 (defun citar--insert-bibtex (key)
   "Insert the bibtex entry for KEY at point."
@@ -965,7 +998,7 @@ With prefix, rebuild the cache before offering candidates."
   ;;      (browse-url-default-browser "https://google.com")
   (interactive (list (citar-select-refs
                       :rebuild-cache current-prefix-arg)))
-  (dolist (key-entry keys-entries)
+  (dolist (key-entry (citar--ensure-entries keys-entries))
     (let ((link (citar-get-link (cdr key-entry))))
       (if link
           (browse-url-default-browser link)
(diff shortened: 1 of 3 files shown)


bdarcus:
I see you have code that sets `embark-multitarget-actions`, and ti looks to me that it should work. 

Any idea why mine would be set to nil when I restart?

roshanshariff:
> Any idea why mine would be set to nil when I restart?

You probably need to rebuild citar to regenerate the autoloads file (and maybe restart emacs for good measure?)

Actually, now that I think about it, it probably doesn't need to be autoloaded. I'll push a new commit making that change. It seems to me that by the time `embark-act-all` is running on citar candidates, `citar.el` has definitely been loaded.

bdarcus:
I did rebuild and restart.

roshanshariff:
Hmm, that's very strange. Maybe try again with the latest commit?

If the problem persists, you could start emacs, run `(require 'embark)`, verify that the variable exists but is nil, then load citar, and see if it's still nil?

And if that doesn't help, try running `M-x debug-on-variable-change` with `embark-multitarget-actions` and check if the debugger runs.

bdarcus:
> If the problem persists, you could start emacs, run `(require 'embark)`, verify that the variable exists but is nil, then load citar, and see if it's still nil?

Still nil.

> And if that doesn't help, try running `M-x debug-on-variable-change` with `embark-multitarget-actions` and check if the debugger runs.

I wasn't sure how to do this last part, so I just evaluated the buffer on citar.el, which resulted in:

```elisp
Debugger entered--setting embark-multitarget-actions to (citar-open): 
  debug--implement-debug-watch(embark-multitarget-actions (citar-open) set nil)
  add-to-list(embark-multitarget-actions citar-open)
  (let ((command (car --dolist-tail--))) (add-to-list 'embark-multitarget-actions command) (setq --dolist-tail-- (cdr --dolist-tail--)))
  (while --dolist-tail-- (let ((command (car --dolist-tail--))) (add-to-list 'embark-multitarget-actions command) (setq --dolist-tail-- (cdr --dolist-tail--))))
  (let ((--dolist-tail-- (list #'citar-open #'citar-open-notes #'citar-open-entry #'citar-open-link #'citar-open-library-files #'citar-attach-library-files #'citar-insert-bibtex #'citar-insert-citation #'citar-insert-reference #'citar-copy-reference #'citar-insert-keys #'citar-run-default-action))) (while --dolist-tail-- (let ((command (car --dolist-tail--))) (add-to-list 'embark-multitarget-actions command) (setq --dolist-tail-- (cdr --dolist-tail--)))))
  (progn (let ((--dolist-tail-- (list #'citar-open #'citar-open-notes #'citar-open-entry #'citar-open-link #'citar-open-library-files #'citar-attach-library-files #'citar-insert-bibtex #'citar-insert-citation #'citar-insert-reference #'citar-copy-reference #'citar-insert-keys #'citar-run-default-action))) (while --dolist-tail-- (let ((command (car --dolist-tail--))) (add-to-list 'embark-multitarget-actions command) (setq --dolist-tail-- (cdr --dolist-tail--))))))
  (if (boundp 'embark-multitarget-actions) (progn (let ((--dolist-tail-- (list #'citar-open #'citar-open-notes #'citar-open-entry #'citar-open-link #'citar-open-library-files #'citar-attach-library-files #'citar-insert-bibtex #'citar-insert-citation #'citar-insert-reference #'citar-copy-reference #'citar-insert-keys #'citar-run-default-action))) (while --dolist-tail-- (let ((command (car --dolist-tail--))) (add-to-list 'embark-multitarget-actions command) (setq --dolist-tail-- (cdr --dolist-tail--)))))))
  (closure (citar-org-open-note-function t) nil (set-keymap-parent citar-map embark-general-map) (add-to-list 'embark-keymap-alist '(citar-reference . citar-map)) (add-to-list 'embark-keymap-alist '(citar-key . citar-citation-map)) (add-to-list 'embark-keymap-alist '(citar-citation . citar-citation-map)) (add-to-list 'embark-pre-action-hooks '(citar-insert-edit embark--ignore-target)) (if (boundp 'embark-multitarget-actions) (progn (let ((--dolist-tail-- (list #'citar-open #'citar-open-notes #'citar-open-entry #'citar-open-link #'citar-open-library-files #'citar-attach-library-files #'citar-insert-bibtex #'citar-insert-citation #'citar-insert-reference #'citar-copy-reference #'citar-insert-keys #'citar-run-default-action))) (while --dolist-tail-- (let ((command ...)) (add-to-list 'embark-multitarget-actions command) (setq --dolist-tail-- (cdr --dolist-tail--))))))))()
  eval-after-load(embark (closure (citar-org-open-note-function t) nil (set-keymap-parent citar-map embark-general-map) (add-to-list 'embark-keymap-alist '(citar-reference . citar-map)) (add-to-list 'embark-keymap-alist '(citar-key . citar-citation-map)) (add-to-list 'embark-keymap-alist '(citar-citation . citar-citation-map)) (add-to-list 'embark-pre-action-hooks '(citar-insert-edit embark--ignore-target)) (if (boundp 'embark-multitarget-actions) (progn (let ((--dolist-tail-- (list ... ... ... ... ... ... ... ... ... ... ... ...))) (while --dolist-tail-- (let (...) (add-to-list ... command) (setq --dolist-tail-- ...))))))))
  eval-buffer()  ; Reading at buffer position 33254
  funcall-interactively(eval-buffer)
  command-execute(eval-buffer record)
  execute-extended-command(nil "eval-buffer" #("eval" 0 4 (ws-butler-chg chg)))
  funcall-interactively(execute-extended-command nil "eval-buffer" #("eval" 0 4 (ws-butler-chg chg)))
  command-execute(execute-extended-command)
```

Toggling it off and re-evaluating it then seems to work. But no clue why it doesn't work to begin with.

citar.el
```diff
@@ -706,7 +708,31 @@ If FORCE-REBUILD-CACHE is t, force reload the cache."
 
 (defun citar--extract-keys (keys-entries)
   "Extract list of keys from KEYS-ENTRIES alist."
-  (seq-map #'car keys-entries))
+  (seq-map (lambda (key-entry)
+             (if (consp key-entry) (car key-entry) key-entry))
+           keys-entries))
+
+(defun citar--ensure-entries (keys-entries)
```
bdarcus, line 715:
```suggestion
(defun citar--ensure-entries (key-entry-alist)
```

Right?

There are a few remaining examples.

roshanshariff: No, the argument to `citar--ensure-entries` can contain both KEY and (KEY . VALUE) items, so it should in fact be called `keys-entries`. The return value would be a `key-entry-alist` if it had a name. I think the same is true of all the other arguments that are called `keys-entries`, but it's possible I missed some instances. There are only a very few places where there actually is a `key-entry-alist` variable.

roshanshariff:
Would you please try it again with the latest commit? I added a defvar for `embark-multitarget-actions`, thinking that the problem might have something to do with byte compilation.

bdarcus:
> Would you please try it again with the latest commit? I added a defvar for `embark-multitarget-actions`, thinking that the problem might have something to do with byte compilation.

Still no dice.

roshanshariff:
I'm running out of ideas, but so far the symptoms seem a lot like there's an older compiled version of `citar.el` that is taking precedence. That would explain why running `eval-buffer` gives you the fresh code, but other ways of loading citar don't. Is that possible? Maybe you could delete any .elc and .eln files and try again?

bdarcus:
I got it working by completing deleting the straight build directly and rebuilding.

I'll test more today then.

bdarcus:
So far, no issues.

While it's not part of this issue, I do want to figure out how to test this working with `embark-act-all`, and possibly add it.

```elisp
(defun bd/search-pdf-contents ()
  "Search pdfs."
  (interactive)
  (let* ((refs (citar-select-refs))
         (files
           refs citar-library-paths '("pdf")))
         (string (read-string "Search string: ")))
    (pdf-occur-search files string t)))
```

Do you know why I could be getting a "wrong number of arguments" error? I don't see that anything changes with those functions.

citar.el
```diff
@@ -706,7 +709,31 @@ If FORCE-REBUILD-CACHE is t, force reload the cache."
...
+  "Return copy of KEYS-ENTRIES with every element a (KEY . ENTRY) pair.
+Each element of KEYS-ENTRIES should be either a KEY or a (KEY
+. ENTRY) pair.  If it is the former, look up the corresponding
+ENTRY and transform the element to a (KEY . ENTRY) pair."
+  (let ((candidates 'uninitialized))
+    ;; Get candidates only if some key has a missing entry, to avoid nasty
```
bdarcus, line 722: Under what condition might this happen?

roshanshariff: If a function is being called interactively, then `keys-entries` will come from `citar-select-ref` and so all the entries will be present; there's no need to get candidates. Right now this is just a performance thing, because there's no situation in which `citar--ensure-entries` would be called during candidate generation. But I figured it's good to code defensively anyway.

roshanshariff:
Try this:

```elisp
(defun bd/search-pdf-contents (keys-entries &optional string)
  "Search pdfs."
  (interactive (list (citar-select-refs)))
  (let ((files (seq-filter
                (lambda (file) (string-match-p "\\.pdf$" file))
                 (citar--ensure-entries keys-entries)
                 citar-library-paths
                 '("pdf"))))
        (string (or string (read-string "Search string: "))))
    (pdf-occur-search files string t)))

(add-to-list 'embark-multitarget-actions #'bd/search-pdf-contents)
```

bdarcus:
> Try this:

So what's the broader lessen, for documentation and such? Why that optional arg there, for example?

bdarcus:
> When you add a function to `embark-multitarget-actions` it's always called non-interactively with a single function argument, which is a list. So the function can't have zero arguments.

Right, but it also didn't work when I added the one arg.

roshanshariff:
The optional string argument is just for convenience, so you can call the function directly from elisp for a particular search string (it won't prompt in this case). It won't be present when called via `embark-act[-all]`.

> Right, but it also didn't work when I added the one arg.

That I can't explain. Maybe you forgot the `citar--ensure-entries`? When called through `embark-act-all` the argument will just be a list of keys without entries, whereas `citar-file--files-for-multiple-entries` takes a KEY-ENTRY-ALIST because it needs both keys and entries.

> In any case, I can't see any harm in merging this now.

Yeah, I agree. I've tested it more in the meantime and haven't found any problems so far either.

bdarcus:
> That I can't explain. Maybe you forgot the `citar--ensure-entries`?

Yes; more that I didn't realize it was needed, and didn't think about it.

bdarcus:
Would it make sense to move that to `citar-file--files-for-multiple-entries` etc?

roshanshariff:
> would it make sense to move that to citar-file--files-for-multiple-entries etc

Currently, `citar-file` doesn't depend on `citar` and I think it's good to keep it that way (among other things, to avoid a circular dependency). So functions in `citar-file` can't do things like looking up entries in the cache.

I think in general this is a reasonable design. The user-facing functions deal with KEYS-ENTRIES with possibly missing entries, but when passing them on to deeper plumbing commands that require entries, they should call `citar--ensure-entries` if needed. That way all the logic to do that is contained in high-level functions in `citar.el`, rather than being scattered through all layers of the code.

bdarcus:
For now, I've updated the wiki example:

https://github.com/bdarcus/citar/wiki/Example-functions#search-contents-of-pdfs

@salutis - you might be interested in this; exploits a new embark feature.

bdarcus:
Why this string match, when the extension is already passed below?

> ```emacs-lisp
>                 (lambda (file) (string-match-p "\\.pdf$" file))
> ```

roshanshariff:
`citar-file--files-for-multiple-entries` uses the extension list to decide which files to return from the library directory, but it includes all the files listed in the bibtex file field regardless of extension. I happen to use a bibtex file produced by Zotero where the file field contains not just PDFs, but also saved HTML snapshots, etc., so I need to filter those out.

We can decide whether this is the right behaviour. I think `citar-open-library-files` should show all the files that are attached to the entry if they're explicitly listed in the file field. And from an API design point of view, it seems to make sense to return all the files so you can filter out the ones you don't want. If you filtered out files at a lower level, you wouldn't be able to get them back.

bdarcus:
Right, but it's inconsistent (which I missed earlier), and I don't think that inconsistency makes logical sense; certainly not in this case?

PR #489, opened 2021-12-15
Add optional ext arg to parse-file-field

### LINKED ISSUE #488: Add optional extensions arg to  citar-file--parse-file-field etc
Seems like we should add an optional `extensions` argument to this function, so that file-handling behavior can be aligned between field vs directory-based.

-----

Right, but it's inconsistent (which I missed earlier), and I don't think that inconsistency makes logical sense; certainly not in this case?

Should probably take another look at those "files" functions.

_Originally posted by @bdarcus in https://github.com/bdarcus/citar/issues/487#issuecomment-994797158_

bdarcus:
Close #488

--------

WDYT @roshanshariff @localauthor?

### DIFF
--- a/citar-file.el
+++ b/citar-file.el
@@ -57,16 +57,18 @@
   :group 'citar
   :type '(function))
 
-(defcustom citar-file-extensions '("pdf")
-  "List of file extensions to recognize for related files.
+(defcustom citar-file-extensions nil
+  "List of file extensions to filter for related files.
 
 These are the extensions the 'citar-file-open-function'
-will open, via `citar-file-open'."
+will open, via `citar-file-open'.
+
+When nil, the function will not filter the list of files."
   :group 'citar
   :type '(repeat string))
 
 (defcustom citar-file-note-extensions '("org" "md")
-  "List of file extensions to recognize for notes.
+  "List of file extensions to filter for notes.
 
 These are the extensions the 'citar-open-note-function'
 will open, via `citar-open-notes'."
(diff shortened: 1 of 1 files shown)


citar-file.el
```diff
@@ -128,14 +128,23 @@ Example: ':/path/to/test.pdf:PDF'."
...
+
+Filter by EXTENSIONS when present."
   (unless dirs (setq dirs (list "/")))  ; make sure DIRS is non-nil
-  (when-let ((filefield (citar-get-value fieldname entry)))
-    (seq-mapcat (lambda (parser)
-                  (funcall parser dirs filefield))
-                citar-file-parser-functions)))
+  (let* ((filefield (citar-get-value fieldname entry))
+         (files
+          (delete-dups
```
bdarcus, line 139: I'm not clear why I needed to add this.

roshanshariff: Probably because each of the file parsers returns the same list of files, you'll end up with two copies of everything. Deleting the duplicates seems like a reasonable way of handling that.

roshanshariff:
Looks good to me! Do you think it also makes sense to change the default for `citar-file-extensions` to nil? We couldn't do this earlier because the library file functions needed a list of extensions, but the new directory scanning code will find files with any extension.

Otherwise, people who use other kinds of files may be surprised that they don't show up in citar by default.

bdarcus:
> Do you think it also makes sense to change the default for `citar-file-extensions` to nil? ... Otherwise, people who use other kinds of files may be surprised that they don't show up in citar by default.

I worry about surprise on the other side, if it's nil.

Maybe it would be better to expand the extensions instead?

I guess part of the problem is this PR still doesn't fully align the behavior; if that variable is `nil`, then:

- the file parsing will not filter; and so will return all paths in the file field
- the directory search will return nothing

They should behave the same; I'm just not sure which.

roshanshariff:
With the new directory parsing code, when `extensions` is nil it will return files with any extension starting with that key. So the behaviour is actually aligned now.

bdarcus:
OK, I just added some changes.

PR #491, opened 2021-12-17
Add groups to citar-open functions

localauthor:
Changes:
- Adds `completing-read` to `citar--open-note`, for when multiple notes exist

- Changes `citar-select-files` to `citar-select-resources`, and calls the new function in both `citar-open` and `citar-open-library-files`

- Adds group-function, `citar-select-group-related-resources`, to `citar-select-resources`, to group candidates by type

The result of calling `citar-open` on an entry with associated files, notes, and a link (the following shows `vertico` with the `consult-completing-read-multiple` interface):
citar-open file-link-note

Selecting multiple file types, a note and a file:
citar-open file-link-note selected

Todo:
~- Somehow set proper category metadata for links in `citar-select-resources`. I'm actually not sure if this is possible.
    - Note that prior to this PR, `citar-open` did not add any metadata to candidates, so `embark-act`'ing on a file candidate did not make embark file functions available. Now embark file functions are available, but on both files _and_ links. Neither case seems optimal, if it's not possible (or easy) to set category metadata by group.~

Question:
~- Does this work as expected with `embark-act-all`? I've not used that enough to properly test it.~

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -421,9 +421,6 @@ REBUILD-CACHE and FILTER."
 
 (defun citar-select-resources (files &optional links)
   "Select resource(s) from a list of FILES, and optionally LINKS."
-  ;; DONE Add links to candidates
-  ;; DONE Add group-function
-  ;; DONE: set category metadata depending on type (differentiate links and files)
   (let* ((files (mapcar
                  (lambda (cand)
                    (abbreviate-file-name cand))
@@ -436,7 +433,7 @@ REBUILD-CACHE and FILTER."
              (add-text-properties 0 (length item) `(consult-multi (file . ,item)) item)))
       (push item resources))
     (completing-read
-     "Select resources: "
+     "Select resource: "
      (lambda (string predicate action)
        (if (eq action 'metadata)
            `(metadata
@@ -452,7 +449,7 @@ REBUILD-CACHE and FILTER."
         (cond
          ((member extension citar-file-note-extensions) "Notes")
          ((string-match "http" resource 0) "Links")
-         (t "Files")))))
+         (t "Library Files")))))
 
 (defun citar--get-major-mode-function (key &optional default)
   "Return KEY from 'major-mode-functions'."
@@ -899,7 +896,7 @@ into the corresponding reference key.  Return
 
 ;;;###autoload
 (defun citar-open (keys-entries)
-  "Open related resources (links or files) for KEYS-ENTRIES."
+  "Open any related resources (links, files, or notes) for KEYS-ENTRIES."
   (interactive (list (citar-select-refs
                       :rebuild-cache current-prefix-arg)))
   (when (and citar-library-paths
@@ -917,14 +914,11 @@ into the corresponding reference key.  Return
            (lambda (key-entry)
              (citar-get-link (cdr key-entry)))
            key-entry-alist))
-         (resources))
+         (selection (citar-select-resources files links)))
     (if files
-        (progn
-          (setq resources (citar-select-resources files links))
-          (dolist (resource resources)
-            (cond ((string-match "http" resource 0)
-                   (browse-url resource))
-                  (t (citar-file-open resource)))))
+        (cond ((string-match "http" selection 0)
+               (browse-url selection))
+              (t (citar-file-open selection)))
       (message "No associated resources"))))
 
 (defun citar--library-files-action (keys-entries action)
@@ -939,12 +933,9 @@ into the corresponding reference key.  Return
              citar-file-extensions)))
       (if (and citar-file-open-prompt
                (> (length files) 1))
-          (let ((selected-files
-                 (citar-select-resources files)))
-            (dolist (file selected-files)
-              (funcall fn file)))
-        (dolist (file files)
-          (funcall fn file)))
+          (let ((selection (citar-select-resources files)))
+            (funcall fn selection))
+        (funcall fn files))
     (message "No associated file")))
 
 ;;;###autoload
@@ -987,10 +978,10 @@ With prefix, rebuild the cache before offering candidates."
               raw-files))
             (file (when (= (length files) 1)
                     (car files))))
-      (if  (file-exists-p file)
+      (if (file-exists-p file)
           (find-file (expand-file-name file))
         (funcall citar-format-note-function key entry (expand-file-name file)))
-    (find-file (expand-file-name (completing-read "Open note: " files nil t)))))
+    (find-file (citar-select-resources files))))
 
 ;;;###autoload
 (defun citar-open-entry (keys-entries)
--- a/citar.el
+++ b/citar.el
@@ -55,6 +55,7 @@
 (defvar embark-meta-map)
 (defvar embark-transformer-alist)
 (defvar embark-multitarget-actions)
+(defvar embark-default-action-overrides)
 (defvar citar-org-open-note-function)
 
 ;;; Variables
--- a/citar.el
+++ b/citar.el
@@ -993,7 +993,8 @@ With prefix, rebuild the cache before offering candidates."
                     (car files))))
       (if (file-exists-p file)
           (find-file (expand-file-name file))
-        (funcall citar-format-note-function key entry (expand-file-name file)))
+        (when (yes-or-no-p (format "No note associated with %s. Create one?" key))
+          (funcall citar-format-note-function key entry (expand-file-name file))))
     (find-file (citar-select-resource files))))
 
 ;;;###autoload
@@ -1062,7 +1063,7 @@ With prefix, rebuild the cache before offering candidates."
     (dolist (key-entry (citar--ensure-entries keys-entries))
       (let ((link (citar-get-link (cdr key-entry))))
 	(if link
-            (browse-url-default-browser link)
+            (browse-url link)
           (message "No link found for %s" (car key-entry)))))))
 
 ;;;###autoload
--- a/citar.el
+++ b/citar.el
@@ -419,8 +419,8 @@ documentation for the return value and the meaning of
 REBUILD-CACHE and FILTER."
   (citar-select-ref :rebuild-cache rebuild-cache :multiple t :filter filter))
 
-(defun citar-select-resources (files &optional links)
-  "Select resource(s) from a list of FILES, and optionally LINKS."
+(defun citar-select-resource (files &optional links)
+  "Select resource from a list of FILES, and optionally LINKS."
   (let* ((files (mapcar
                  (lambda (cand)
                    (abbreviate-file-name cand))
@@ -914,7 +914,7 @@ into the corresponding reference key.  Return
            (lambda (key-entry)
              (citar-get-link (cdr key-entry)))
            key-entry-alist))
-         (selection (citar-select-resources files links)))
+         (selection (citar-select-resource files links)))
     (if files
         (cond ((string-match "http" selection 0)
                (browse-url selection))
@@ -933,9 +933,9 @@ into the corresponding reference key.  Return
              citar-file-extensions)))
       (if (and citar-file-open-prompt
                (> (length files) 1))
-          (let ((selection (citar-select-resources files)))
+          (let ((selection (citar-select-resource files)))
             (funcall fn selection))
-        (funcall fn files))
+        (funcall fn (car files)))
     (message "No associated file")))
 
 ;;;###autoload
@@ -981,7 +981,7 @@ With prefix, rebuild the cache before offering candidates."
       (if (file-exists-p file)
           (find-file (expand-file-name file))
         (funcall citar-format-note-function key entry (expand-file-name file)))
-    (find-file (citar-select-resources files))))
+    (find-file (citar-select-resource files))))
 
 ;;;###autoload
 (defun citar-open-entry (keys-entries)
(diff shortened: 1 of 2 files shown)


citar.el
```diff
@@ -413,26 +413,36 @@ documentation for the return value and the meaning of
...
-(defun citar-select-group-related-sources (file transform)
-  "Group by FILE by source, TRANSFORM."
-    (let ((extension (file-name-extension file)))
-      (when transform file
-        ;; Transform for grouping and group title display.
+(defun citar-select-resources (files &optional links)
+  "Select resource(s) from a list of FILES, and optionally LINKS."
+  ;; DONE Add links to candidates
+  ;; DONE Add group-function
+  ;; FIX: set category metadata depending on type (differentiate links and files); possible?
```
bdarcus, line 420: Whenever I've contemplated something like this (for example with `citar-open`, this has been the hangup.

It is possible and easy to use `consult--multi`, which will do this. But then it's `completing-read` based, so only a single selected candidate.

I simply never figured out how to do it otherwise.

Maybe you can include a screenshot or two on the main post when you get a chance?

localauthor: > It is possible and easy to use `consult--multi`, which will do this. But then it's `completing-read` based, so only a single selected candidate. And it's not explicit dependency.

This is where I landed. Bit of an impasse.

bdarcus:
@localauthor - I set this to merge to the `no-crm` branch, but before doing that, I'll put comments here:

1. I think the "Files" group maybe should be "LIbrary files"? WDYT?
2. I realize this when I was playing with this idea earlier: links have no annotations. Should they? If yes, what should they be?
3. What should happen to the existing more specific "open" commands? I guess we still need `citar-open-notes` at least?

Let me know your thoughts, and if you want to make any changes here before I merge?

bdarcus:
> Does this work as expected with `embark-act-all`? I've not used that enough to properly test it.

I am just starting to test this branch. Here's what you see after running `embark-act-all`:

image

Two things:

1. I guess notes should also have a `file` category, but remain a separate group? If yes, not sure how to do that.
2. I'm actually not sure how one would then open all of these results from there using `embark-act-all`.

Any thoughts on the above @minad?

localauthor:
Ok, I just rebased onto main. 

I changed `citar--open-notes` to use `citar-select-resources`, so now notes are properly categorized as files and grouped as notes.

As for the question of where to merge this (either to `main` directly or to `no-crm`) , I suppose it depends on your plans for `no-crm`. :)

This branch changes CRM to CR in all the open functions anyway.

bdarcus:
Let's keep it here for now. There are still some subtleties with `embark-act-all` to work out.

bdarcus:
> Per my point 2 [above](https://github.com/bdarcus/citar/pull/491#issuecomment-997980185), I can't see how to run `embark-act-all` to open all of the resulting resources, because the default action is then `embark-act`. I wonder if that needs to be the case, or what a better alternative might be.

There does seem to be a bug here though. If I only select one candidate, and then do `embark-act-all` on this results, where the default action is `citar-open`, I get a "no results found".

Even if I only select one resource, I get a "wrong type" error (it's expecting a list).

citar.el
```diff
@@ -885,7 +896,7 @@ into the corresponding reference key.  Return
 
 ;;;###autoload
 (defun citar-open (keys-entries)
-  "Open related resources (links or files) for KEYS-ENTRIES."
+  "Open any related resources (links, files, or notes) for KEYS-ENTRIES."
   (interactive (list (citar-select-refs
```
bdarcus, line 900: isn't this still CRM?

citar.el
```diff
@@ -419,26 +419,37 @@ documentation for the return value and the meaning of
...
-        ; (group-function . citar-select-group-related-sources)
-           (category . file))
-       (complete-with-action action files string predicate)))))
-
-(defun citar-select-group-related-sources (file transform)
-  "Group by FILE by source, TRANSFORM."
-    (let ((extension (file-name-extension file)))
-      (when transform file
-        ;; Transform for grouping and group title display.
+(defun citar-select-resources (files &optional links)
```
bdarcus, line 422: if this is now CR, shouldn't this be singular?
```suggestion
(defun citar-select-resource (files &optional links)
```

citar.el
```diff
@@ -885,7 +896,7 @@ into the corresponding reference key.  Return
 
 ;;;###autoload
 (defun citar-open (keys-entries)
-  "Open related resources (links or files) for KEYS-ENTRIES."
+  "Open any related resources (links, files, or notes) for KEYS-ENTRIES."
```
bdarcus, line 899:
```suggestion
  "Open any related resource (links, files, or notes) for KEYS-ENTRIES."
```

citar.el
localauthor: Only if `citar-select-multiple` is set to `t`, I think?

localauthor:
All the `open` functions should be working as expected now (hopefully!), with the exception of `embark-act-all`, which I'm just starting to work through.

bdarcus:
> All the `open` functions should be working as expected now (hopefully!), with the exception of `embark-act-all`, which I'm just starting to work through.

When I use `citar-open`, I still get a type error when trying to open any resource.

localauthor:
Latest commit make `embark-act-all` work appropriately in all contexts.

As the commit message says, each command now let-binds `embark-default-action-overrides` to change the default action from the original commands (ie, `citar-open`, `citar-open-notes`, etc.) which cannot act on files or urls, to functions that _can_ act on files and urls, namely, `citar-open-file`, `find-file` and `browse-url`, where appropriate.

The edge case here is with `citar-open` and acting on candidates of multiple types (files or urls). I've added a function `citar-open-multi` to deal with candidates that are of the type `consult-multi`. So now running `embark-act-all` on candidates that include links and files should open all of them as expected.

Hopefully it works for you as it does for me.

localauthor:
One issue with using `consult-multi` is that it requires `embark-consult.el` to be loaded and, specifically, `embark-consult--multi-transform` to be defined.

If only embark.el is loaded, these are not available.

bdarcus:
> One issue with using `consult-multi` is that it requires `embark-consult.el` to be loaded and, specifically, `embark-consult--multi-transform` to be defined.
> 
> If only `embark.el` is loaded, these are not available.

What if you define the variable? Do you get an error, or just not the behavior?

localauthor:
`embark-consult--multi-transform` is a function, so it really needs to loaded for everything to fully work.

I say "fully" because `citar-open` still functions reasonably without it, ie, there are no errors and files/links can be opened with "RET", according to type --- even when using `embark-act-all` on multiple types.

Since without `embark-consult--multi-transform`, files and urls remain typed as 'consult-multi', they are handled by the default action associated with that type, namley, the new 'citar-open-multi'. There aren't file- or url-specific actions available, but each is opened appropriately, and there are no errors.

bdarcus:
> Each `open` command now includes a let-binding for `embark-default-action-overrides` in order to override the default action depending on target type.

These result in "unused lexical variable" warnings.

On a separate matter, it's getting closer. But if I have this:

image

... and then do `embark-act-all`, I get this:

image

minad:
@localauthor Yes of course. We should switch to `multi` from `consult-multi` to make this general and to avoid the loading issue. But I will wait for a few weeks to let this settle here, since Citar does not yet seem to have converged fully. Then we can change it across the different packages and tag new releases. Probably embark and marginalia should support both `multi` and `consult-multi` for some time and treat them as aliases.

localauthor:
> ... and then do `embark-act-all`, I get this:

Before pressing "RET", what does the embark-indicator-menu look like for you? That is: what is the default action you are calling when do `embark-act-all`? If it is `citar-open`, then I would expect to see exactly what you're screenshot shows (which is a symptom of another issue I'm trying to figure out, about which more soon).

Here is what I see _before_ pressing "RET" in a similar series of steps:
Screen Shot 2021-12-21 at 15 33 20

localauthor:
> These result in "unused lexical variable" warnings.

I'm afraid I don't know how to avoid such warnings, or what they mean really.

bdarcus:
> Followed by opened files, notes, and browser.

I see the same, except for this last step.

Maybe there's something wrong with my install of the branch. Running `eval-buffer` on `citar.el` does result in the correct behavior!

bdarcus:
> > These result in "unused lexical variable" warnings.
> 
> I'm afraid I don't know how to avoid such warnings, or what they mean really.

@minad - any tips? He's let-binding `embark-default-action-overrides` per https://github.com/bdarcus/citar/pull/491#issuecomment-998660728.

minad:
Oh. These default action overrides should not be necessary. The default file and url actions should work ootb as should all the other actions. If not something else is wrong and should be fixed.

The correct approach is to define your own category if you want to overwrite or define your own actions.

bdarcus:
> Probably you miss a defvar to fix the lexical error?

Yes; I just committed that change.

citar.el
```diff
@@ -903,17 +918,21 @@ into the corresponding reference key.  Return
...
-                 (browse-url resource))
-                (t (citar-file-open resource))))
+         (selection (citar-select-resource files links)))
+    (if files
+        (cond ((string-match "http" selection 0)
+               (browse-url selection))
+              (t (citar-file-open selection)))
       (message "No associated resources"))))
 
+(defun citar-open-multi (selection)
```
minad, line 928: This command is unnecessary and should be removed. Embark already dispatches consult-multi to file and url. That's the purpose of consult-multi.

citar.el
```diff
@@ -885,13 +897,16 @@ into the corresponding reference key.  Return
...
+  "Open any related resource (links, files, or notes) for KEYS-ENTRIES."
   (interactive (list (citar-select-refs
                       :rebuild-cache current-prefix-arg)))
   (when (and citar-library-paths
              (stringp citar-library-paths))
     (message "Make sure 'citar-library-paths' is a list of paths"))
-  (let* ((key-entry-alist (citar--ensure-entries keys-entries))
+  (let* ((embark-default-action-overrides '((consult-multi . citar-open-multi)
+					    (file . citar-file-open)
+                                            (url . browse-url)))
```
minad, line 908: Here the consult-multi and url overrides should probably go away.

localauthor: Thanks for looking at this, and for the comments.

Embark and consult-multi work just as you describe when it comes to categorizing files and urls, but (as I describe [here](https://github.com/bdarcus/citar/pull/491#issuecomment-998660728)) because the source of the target is minibuffer completion, Embark sets the default action for all targets, regardless of type, to the command that opened the minibuffer in the first place (per the docstring of `embark-default-action-overrides`). That means the default action for any target type is set to `citar open`, which is a function that does not take files or urls as arguments. Hence the overrides. It's a bit of a hail mary, I guess, but I'm afraid I'm running up against the limits of my lisp/emacs knowledge here. ;)

As for the `citar-open-multi` function, it was only added to allow `embark-act-all` to work on properly on targets of multiple types. When using `embark-act` on single candidates, consult-multi works just as expected and this function isn't called. When calling `embark-act-all` on targets of multiple types, they haven't been transformed by the candidate transformer yet ---  so they remain typed as `consult-multi`. Hence, another override. Again, bit of a hail mary -- but functional!

Thanks again for your comments. Really appreciate your taking the time.

minad: Okay, maybe I missed something here but this seems wrong. The default action is supposed to be *the command that started the completion*. Overriding it with something else is counterintuitive and goes against the usual Embark behavior.

Note that Embark has a minibuffer argument injection mechanism. I suggest you dig deeper into the Embark documentation.

minad: > That means that, in this case, the default action for any target type is set to citar open, which is a function that does not take files or urls as arguments. Hence the overrides. 

I am specifically referring to this. This is wrong. There is the minibuffer injection mechanism.

> When calling embark-act-all on targets of multiple types, they haven't been transformed by the candidate transformer yet --- so they remain typed as consult-multi.

I will check this, but if this is the case it is a bug or issue in Embark. 

I know that oantolin added some target type checking to embark-act-all and there is indeed a problem when acting on candidates of multiple categories at once. Anyways this should be handled properly by  Embark itself.

bdarcus: I experimented a bit with removing these lines, and things seem to work fine, until I run `embark-act-all` on the resulting resources. So to recap:

1. `citar-open`
2. narrow candidates and then `embark-act-all`
3. from there I get a list of resources, if I run `embark-act-all`, II then get the error below:

```elisp
embark PCH: (user-error "No target found")
```

... which is I think why he introduced the `citar-open-multi` function.

image

minad: Did you load embark-consult?

bdarcus: > Did you load embark-consult? 

I did require it now, and now get:

```elisp
(execute-extended-command nil "citar-open" nil)
```

> I don't know what is wrong here exactly, but the override should not be needed due to the design of embark. consult-buffer also works without override.

I'll take another look.

minad: @bdarcus What's the status here? Did you figure out the issues?

bdarcus: > Did you figure out the issues?

Not yet. Likely in the next few days.

minad:
@bdarcus As a small demonstrator, this works for me:

~~~ elisp
(require 'embark-consult)
(defun open-example ()
  (interactive)
  (let* ((candidates
          (list (propertize "init.el" 'consult-multi
                            '(file . "~/.config/emacs/init.el"))
                (propertize "emacs.org" 'consult-multi
                            '(url . "https://www.gnu.org/software/emacs/"))))
         (result (completing-read "Open: "
                                  (lambda (str pred action)
                                    (if (eq action 'metadata)
                                        '(metadata (category . consult-multi))
                                      (complete-with-action action candidates str pred)))
                                  nil t)))
    (pcase (get-text-property 0 'consult-multi (car (member result candidates)))
      (`(url . ,url) (eww url))
      (`(file . ,file) (find-file file)))))
~~~

PR #493, opened 2021-12-17
Add `citar-select-multiple` option

### LINKED ISSUE #492: Add an option to disable multiple-candidate selection
I use Citar every day and have done so for quite a while now. Yet, I have *never* needed to select multiple items.  Not once.  Thus, all these `RET RET` confirmations at every step do nothing for me.  Thus, I would like to ask for a configuration option to disable multiple-candidate selection in Citar.

P.S. Today I decided to try using MCT (Minibuffer and Completions in Tandem) instead of Vertico, and the Citar workflow became even more painful due to multiple-candidate selections.

roshanshariff:
When nil, all citar commands will use `completing-read` instead of `completing-read-multiple`.

Closes #492

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -319,6 +319,12 @@ of all citations in the current buffer."
 
 ;;; Completion functions
 
+(defcustom citar-select-multiple t
+  "Use `completing-read-multiple' for selecting citation keys.
+  When nil, all citar commands will use `completing-read`."
+  :type 'boolean
+  :group 'citar)
+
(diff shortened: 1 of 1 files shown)


bdarcus:
I should have asked this before merging, but ...

... this doesn't have any implications for the embark stuff; does it?

roshanshariff:
Not with the new Embark version anyway. With older versions, when that defcustom is false `embark-act` won't correctly work on multi-key citations.

PR #494, opened 2021-12-18
Correctly handle `embark-act` on multi-key citations (fix bug in #487)

roshanshariff:
When running `embark-act` on multi-key citations, the target string is of the form "key1 & key2 & key3". Fix `citar--extract-keys` and `citar--ensure-entries` to correctly handle KEYS-ENTRIES lists containing strings of this form.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -714,32 +714,61 @@ If FORCE-REBUILD-CACHE is t, force reload the cache."
       (concat base-url (citar-get-value field entry)))))
 
 (defun citar--extract-keys (keys-entries)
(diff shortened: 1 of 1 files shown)


bdarcus:
Is this fixing an actual bug (with latest embark commit), or is this more for backward compatibility (the issue you mentioned earlier)?

roshanshariff:
It's fixing an actual bug. With #487, the Citar commands no longer work properly on multi-key citations.

bdarcus:
OK, I merged it, but reworded the first line of the commit message to be under 50 characters.

Issue #495, opened 2021-12-18
Remove CRM, use CR instead


bdarcus:
EDIT: this needs to be resolved before 1.0, since minad removed consult-crm from consult. Vanilla `completing-read-multiple` is just not going to work.

-----------------------

Pulling this out into a separate issue ...

> ... I also wonder if some commands should just not use CRM anyway:
> `citar-insert-citation` could be modified to position point at the end of the citation, and when called again to just add new keys to the citation before point. Then you could insert multi-key citations by repeating the command, using the `embark-quit-after-action` functionality.

To me, this is the most critical command to get right, since it can either enhance the writing process, or impede it.

I'm going to focus on org, since that's my priority.

Part of the problem is that I think single vs multi citing varies quite a bit by field, and even by scholar.

Some do it all the time, and others not at all.

I personally do it less often, but still do it, and so care about getting it right.

So ideally both work really well.

The current CRM-based approach does work well with both IF you use `consult-completing-read-multiple` AND you [modify the keybindings to use TAB to select and RET to exit](https://github.com/bdarcus/citar/issues/492#issuecomment-997024450).

To ditch CRM and rely on CR only, I'm unclear how well the solution @roshanshariff suggests above would work. It may or may not be tedious.

FWIW, here's how org-ref-cite handled it (though I'm not thrilled with it)

https://github.com/jkitchin/org-ref-cite#multiple-insertions

The `oc-basic` insert processor works like this:

![image](https://user-images.githubusercontent.com/1134/146678680-16a7056c-1db3-4914-9fed-9a7418549b98.png)

... which is OK, but tends to confuse people at first ("how do I exit?").

> Similarly for `citar-open-notes` and `citar-open-entry`. You can just repeat the command if you want to open more than one.

Yes. I also think for the first `embark-act-all` is likely a better, or additional, path. I'm less concerned about the latter.

_Originally posted by @roshanshariff in https://github.com/bdarcus/citar/issues/492#issuecomment-997004886_

jsilve24:
Did the recent changes remove the ability to select multiple citations at once when using  `citar-insert-citation`? If so, is there a way to get back the prior behavior which I had grown used to?

Thank you!

bdarcus:
I guess it did, though wonder if that was an oversight.

Do either of you know, @roshanshariff @localauthor?

Regardless, you can set `citar-select-multiple` to `t` to get the behavior back now.

bdarcus:
[...]

PR #496, opened 2021-12-19
refactor: crm -> cr

bdarcus:
This is incomplete, but is a branch to test the ideas in #495, which you should do in conjunction with 'embark-act-all'. The devil will be in the details here.

## Notes

There are two categories of commands in citar:

### Single action

That is, which take a single reference candidate, and do something with it. With `embark-act-all`, the command would be run once per reference.

All of the "open" commands fit here. This and #491 updates these to use CR instead of CRM. 

### Multi action

That is, which ideally take a list of reference candidates and do something with them as one action. With `embark-act-all`, the command would be run once, on the list of references.

Many of the "insert" (notably `citar-insert-citation`) and "copy" (`citar-copy-reference`) commands fit in this category, as does the [search-pdf-contents](https://github.com/bdarcus/citar/wiki/Example-functions#search-contents-of-pdfs) example, which I'd like to merge at some point. 

I do think it'd be nice to keep `embark-multitarget-actions` as an option for these sorts of commands, to replace with the use of CRM.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -55,7 +55,7 @@
 (defvar embark-meta-map)
 (defvar embark-transformer-alist)
 (defvar embark-multitarget-actions)
-(defvar citar-org-open-note-function)
+;(defvar citar-org-open-note-function)
 
 ;;; Variables
 
@@ -319,7 +319,7 @@ of all citations in the current buffer."
 
 ;;; Completion functions
 
-(defcustom citar-select-multiple t
+(defcustom citar-select-multiple nil
   "Use `completing-read-multiple' for selecting citation keys.
   When nil, all citar commands will use `completing-read`."
   :type 'boolean
@@ -419,11 +419,10 @@ documentation for the return value and the meaning of
 REBUILD-CACHE and FILTER."
   (citar-select-ref :rebuild-cache rebuild-cache :multiple t :filter filter))
 
-(defun citar-select-files (files)
-  "Select file(s) from a list of FILES."
-  ;; TODO add links to candidates
-  (completing-read-multiple
-   "Open related file(s): "
+(defun citar-select-file (files)
+  "Select file from a list of FILES."
+  (completing-read
+   "Related file: "
    (lambda (string predicate action)
      (if (eq action 'metadata)
          `(metadata
@@ -760,7 +759,7 @@ Return a list containing only (KEY . ENTRY) pairs."
        (lambda (key-entry)
          (if (consp key-entry)
              (list key-entry)
-           (seq-remove      ; remove keys not found in CANDIDATES
+           (seq-remove    ; remove keys not found in CANDIDATES
             #'null
             (seq-map
              (lambda (key)
@@ -873,9 +872,7 @@ into the corresponding reference key.  Return
   (add-to-list 'embark-keymap-alist '(citar-citation . citar-citation-map))
   (add-to-list 'embark-pre-action-hooks '(citar-insert-edit embark--ignore-target))
   (when (boundp 'embark-multitarget-actions)
-    (dolist (command (list #'citar-open #'citar-open-notes
-                           #'citar-open-entry #'citar-open-link
-                           #'citar-open-library-files #'citar-attach-library-files
+    (dolist (command (list #'citar-open-library-files #'citar-attach-library-files
                            #'citar-insert-bibtex #'citar-insert-citation
                            #'citar-insert-reference #'citar-copy-reference
                            #'citar-insert-keys #'citar-run-default-action))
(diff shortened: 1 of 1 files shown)


minad:
Thank you for working on this. If it turns out that the approach is viable, it would be great if you could remove the reliance on the `embark-multitarget-actions`. Then we can reconsider this addition to Embark. It would be great if we can maintain a simpler injection protocol where single candidates are the elementary unit of operation.

https://github.com/bdarcus/citar/blob/9da28efdaa024a65165a24189306943acb4bfc4f/citar.el#L875-L882

minad:
Since Embark only supports acting on multiple candidates via `embark-act-all` the multi selection feature is not as convenient as with crm. The user has to follow narrow down to all candidates. However I expect that we can improve this by adding an `embark-select` command which can be used to first select/collect candidates and then act only on the selected ones. So this regression in behavior should not hold you back from trying the CR-only approach.

bdarcus:
The main question I have is how one would do something like this with only single selections and actions:

https://github.com/bdarcus/citar/wiki/Example-functions#search-contents-of-pdfs

minad:
Good argument! Commands like thus are indeed (almost) impossible to implement without embark-multitarget-actions. One could hack something since nothing is impossible in Emacs ;)

I think this is reasonable justification to keep embark-multitarget-actions. Nevertheless I see some value in exploring a CR-only solution.

salutis:
> The main question I have is how one would do something like this with only single selections and actions:
> 
> https://github.com/bdarcus/citar/wiki/Example-functions#search-contents-of-pdfs

In this case, I would expect multiple-item selection because it makes sense. In the cases such as `citar-open`, I would expect single-item selection by default and multi-item selection with e.g. `C-u` (and an option to enable it by default).

bdarcus:
> In the cases such as `citar-open`, I would expect single-item selection by default and multi-item selection with e.g. `C-u` (and an option to enable it by default).

The idea is one would use `embark-act-all` for that. Currently that solution doesn't allow marking specific candidates, but that is planned.

It would keep this code simpler.

localauthor:
> Beyond the obvious, an advantage here is we could then use `consult--multi` to address how to mix different candidates category in "open" commands (see #491).

Does this mean you're open to adding `consult` as a dependency?

If so, I would refactor the changes in #491, push those here, and then closing that PR.

minad:
Hmm, to be honest I would avoid Consult as a dependency. Which Consult feature would you like to have in Citar? You can use the group-function.

bdarcus:
> If I recall correctly there has been a prior discussion of this approach where I already proposed this?

We have discussed consult-multi-like functionality before, but I don't recall you proposing this specific approach before. I actually didn't realize the category was/could be independent of consult!

> (In a second step I can change the name of consult-multi to multi since it is not consult specific anymore and change it in marginalia, embark and consult.)

+1

> Does this mean you're open to adding consult as a dependency?
> If so, I would refactor the changes in #491, push those here, and then close that PR.

That would likely make sense.

Just to be clear, though, I'm not 100% sure yet I'll merge this. But I guess you won't solve #491 as you're hoping without it :-)

localauthor:
Added the consult-multi change to #491. Seems to work, but I haven't put it through much testing yet.

bdarcus:
> Added the consult-multi change to #491. Seems to work, but I haven't put it through much testing yet.

I've update the other "open" commands here to use CR instead of CRM.

bdarcus:
@salutis - you want to give this a try? It changes almost all of the "open" commands, save for `citar-open` (see #491) to `completing-read`. If you do get a chance to test, please test also using them in conjunction with `embark-act-all`.

salutis:
Just tried it, including `embark-act-all`, and it works great. Sadly, I now use`citar-open` exclusively (and always with to open a _single_ item). That said, onward and upward!

bdarcus:
> Sadly, I now use`citar-open` exclusively

That'll be addressed in #491, which also adds notes and links to the results.

salutis:
We should also rename `citar-open-library-files` to `citar-open-library-file`.

bdarcus:
> We should also rename `citar-open-library-files` to `citar-open-library-file`.

That was already done. Do you really see otherwise?

bdarcus:
I merged this, so we'll have to rebase #491 on main before resolving the remaining issues there.

The main post summarizes my conclusions on this, which is that most commands will work fine with CR + `embark-act-all`, but that there are couple/few that would benefit from remaining as part of `embark-multitarget-actions`. 

The main one, that could benefit from experimentation, is `citar-insert-citation` (and the `org-cite-insert` processor).

salutis:
> That was already done. Do you really see otherwise?

All good now. I think `package.el`reverted the package under me.

PR #498, opened 2021-12-19
Use seq-remove in place of cl-delete-if

### LINKED ISSUE #497: Use `seq-remove` rather than `cl-delete-if`?
@roshanshariff - am noticing that this line is generating a warning.

https://github.com/bdarcus/citar/blob/9da28efdaa024a65165a24189306943acb4bfc4f/citar.el#L763

To get rid of the warning requires adding that library to the `eval-when-compile` block at the top. 

But would it not be appropriate to just use `seq-remove` instead?

bdarcus:
I see no issue with merging this @roshanshariff, but let me know ....

------------

Close #497

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -760,7 +760,7 @@ Return a list containing only (KEY . ENTRY) pairs."
        (lambda (key-entry)
          (if (consp key-entry)
              (list key-entry)
-           (cl-delete-if                ; remove keys not found in CANDIDATES
+           (seq-remove      ; remove keys not found in CANDIDATES
             #'null
             (seq-map
              (lambda (key)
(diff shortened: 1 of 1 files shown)


okamsn:
Just so you know, `seq-drop-while` only operates on the front of the sequence while the condition is true.  The closer alternative to `cl-delete-if` is `seq-remove`, which will remove all elements for which the condition is true. I don't know whether that matters for you.

bdarcus:
> Just so you know, `seq-drop-while` only operates on the front of the sequence while the condition is true. The closer alternative to `cl-delete-if` is `seq-remove`, which will remove all elements for which the condition is true. I don't know whether that matters for you.

Yes, thank you! I corrected the commit.

PR #500, opened 2021-12-19
Add invert-prompt argument to citar-markdown-insert-citation

### LINKED ISSUE #499: Error inserting markdown citation
**Describe the bug**
When I try to insert a citation in a markdown document, I get the error `Wrong number of arguments: (1 . 1), 2`

**To Reproduce**
1. Visit a buffer in markdown-mode
2. Call `citar-insert-citation` through the `C-c b` binding
3. Select one or more citations, then hit return (<- I'm a new user of this package, is this what I should do here?)

When repeating the above but for latex or org buffers, I get more informative errors about my incomplete setup and other packages that I should install.

**Expected behavior**
Insert a citation like `[@foo]`

**Screenshots**
n/a

**Emacs version:**
27.2, macOS

**Additional context**
n/a

roshanshariff:
When calling `citar-insert-citation` with a prefix argument, the meaning of `citar-markdown-prompt-for-extra-arguments` is now inverted.

Fixes #499.

### DIFF
--- a/citar-markdown.el
+++ b/citar-markdown.el
@@ -56,7 +56,7 @@ specified at URL
   (insert (mapconcat (lambda (k) (concat "@" k)) keys "; ")))
 
 ;;;###autoload
-(defun citar-markdown-insert-citation (keys)
+(defun citar-markdown-insert-citation (keys &optional invert-prompt)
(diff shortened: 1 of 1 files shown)


bdarcus:
Thanks!

PR #509, opened 2021-12-27
fix: type error

### LINKED ISSUE #508: `citar-open` stopped working?
I have just updated Citar and now, when I call `citar-open`, select some reference, and then select some related resource, I get:

```
funcall-interactively: Wrong type argument: listp, "/Users/salutis/bib/driscoll+2022.bib"
```

and 

```
funcall-interactively: Wrong type argument: listp, "/Users/salutis/bib/driscoll+2022.pdf"
```

Does `open-citar` work for you, folks?

bdarcus:
Fix #508

May also re-implement #491 if I can find the time.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -299,7 +299,7 @@ of all citations in the current buffer."
     (define-key map (kbd "e") #'citar-open-entry)
     (define-key map (kbd "l") #'citar-open-link)
     (define-key map (kbd "n") #'citar-open-notes)
-    (define-key map (kbd "f") #'citar-open-library-file)
+    (define-key map (kbd "f") #'citar-open-library-files)
     (define-key map (kbd "RET") #'citar-run-default-action)
     map)
   "Keymap for Embark minibuffer actions.")
@@ -311,7 +311,7 @@ of all citations in the current buffer."
     (define-key map (kbd "e") #'citar-open-entry)
     (define-key map (kbd "l") #'citar-open-link)
     (define-key map (kbd "n") #'citar-open-notes)
-    (define-key map (kbd "f") #'citar-open-library-file)
+    (define-key map (kbd "f") #'citar-open-library-files)
     (define-key map (kbd "r") #'citar-copy-reference)
     (define-key map (kbd "RET") #'citar-run-default-action)
     map)
@@ -319,7 +319,7 @@ of all citations in the current buffer."
 
 ;;; Completion functions
 
-(defcustom citar-select-multiple t
+(defcustom citar-select-multiple nil
   "Use `completing-read-multiple' for selecting citation keys.
   When nil, all citar commands will use `completing-read`."
   :type 'boolean
@@ -881,18 +881,17 @@ into the corresponding reference key.  Return
 ;;; Commands
 
 ;;;###autoload
-(defun citar-open (key-entry)
-  "Open related resources (links or files) for KEY-ENTRY."
-  (interactive (list (citar-select-ref
+(defun citar-open (keys-entries)
+  "Open related resources (links or files) for KEYS-ENTRIES."
+  (interactive (list (citar-select-refs
                       :rebuild-cache current-prefix-arg)))
   (when (and citar-library-paths
              (stringp citar-library-paths))
     (message "Make sure 'citar-library-paths' is a list of paths"))
-  (let* ((key-entry-alist (citar--ensure-entries key-entry))
+  (let* ((key-entry-alist (citar--ensure-entries keys-entries))
          (files
-          (citar-file--files-for-entry
-           (caar key-entry-alist)
-           (cdar key-entry-alist)
+          (citar-file--files-for-multiple-entries
+           key-entry-alist
            (append citar-library-paths citar-notes-paths)
            ;; find files with any extension:
            nil))
@@ -902,14 +901,15 @@ into the corresponding reference key.  Return
              (citar-get-link (cdr key-entry)))
            key-entry-alist))
          (resource-candidates (delete-dups (append files (remq nil links))))
-         (resource
+         (resources
           (when resource-candidates
             (completing-read "Related resources: " resource-candidates))))
     (if resource-candidates
-        (cond ((string-match "http" resource 0)
-               (browse-url resource))
-              (t (citar-file-open resource))))
-    (message "No associated resources")))
+        (dolist (resource resources)
+          (cond ((string-match "http" resource 0)
+                 (browse-url resource))
+                (t (citar-file-open resource))))
+      (message "No associated resources"))))
 
 (defun citar--library-file-action (key-entry action)
   "Run ACTION on file associated with KEY-ENTRY."
--- a/citar.el
+++ b/citar.el
@@ -299,7 +299,7 @@ of all citations in the current buffer."
     (define-key map (kbd "e") #'citar-open-entry)
     (define-key map (kbd "l") #'citar-open-link)
     (define-key map (kbd "n") #'citar-open-notes)
-    (define-key map (kbd "f") #'citar-open-library-files)
+    (define-key map (kbd "f") #'citar-open-library-file)
     (define-key map (kbd "RET") #'citar-run-default-action)
     map)
   "Keymap for Embark minibuffer actions.")
@@ -311,7 +311,7 @@ of all citations in the current buffer."
     (define-key map (kbd "e") #'citar-open-entry)
     (define-key map (kbd "l") #'citar-open-link)
     (define-key map (kbd "n") #'citar-open-notes)
-    (define-key map (kbd "f") #'citar-open-library-files)
+    (define-key map (kbd "f") #'citar-open-library-file)
     (define-key map (kbd "r") #'citar-copy-reference)
     (define-key map (kbd "RET") #'citar-run-default-action)
     map)
@@ -319,7 +319,7 @@ of all citations in the current buffer."
 
 ;;; Completion functions
 
-(defcustom citar-select-multiple nil
+(defcustom citar-select-multiple t
   "Use `completing-read-multiple' for selecting citation keys.
   When nil, all citar commands will use `completing-read`."
   :type 'boolean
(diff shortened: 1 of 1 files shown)


bdarcus:
Any input on this @roshanshariff?

Basically, I'm moving in the direction of ditching CRM for CR here, which involves other little changes.

roshanshariff:
I think the function arguments should have remained `keys-entries` instead of being changed to `key-entry` in commit 606f63b32; among other things, this allows the function to be called on multiple entries through `embark-act`. I think the only needed change would be in the `interactive` specifications, which would become something like
```elisp
(interactive (list (list (citar-select-ref :rebuild-cache current-prefix-arg))))
```
(Note the extra `list` around the call to `citar-select-ref`. This ensures that the function argument is a single-element list containing a `(KEY . ENTRY)` pair when called interactively.)

Then the calling convention of all the functions would remain the same, meaning their behaviour with `embark-act` and `embark-act-all` doesn't change. The only difference would be the use of CR instead of CRM when called interactively.

bdarcus:
> among other things, this allows the function to be called on multiple entries through `embark-act`

Are you 100% sure of that?

The idea of the changes I've been making is these functions just becom pure CR, and that `embark-act-all` will simply calling the function sequentially.

roshanshariff:
Hmm, what should the functions do when called through `embark-act` on a multi-key citation?

bdarcus:
> Hmm, what should the functions do when called through `embark-act` on a multi-key citation?

I'm not sure, but I guess I was assuming these functions would only act on citation-references.

I thought you had already made that change earlier.

WDYT?

roshanshariff:
The change I made was to allow these functions to operate on both citations and citation-references (or rather, single and multi-key citations) which was the status quo before the Embark changes. If you like, we could add a new Embark keymap for multi-key citations, along with separate functions that act on them?

bdarcus:
> we could add a new Embark keymap for multi-key citations, along with separate functions that act on them?

Which commands need to be a part of that separate map?

Only `citar-insert-edit`?

roshanshariff:
I was thinking you could also have the `open` functions work on multi-key citations, since they already support multi-keys when called through elisp. So you could just have them use `citar-select-ref` when called interactively, but provide a wrapper function that lets them work on " & "-separated strings. Something like this:

```elisp
(defun citar-open-library-files (keys-entries)
  (interactive (list (list (citar-select-ref))))
  ...
)

(defun citar-open-library-files-multi (keys-string)
  (citar-open-library-files (split-string keys-string " & "))
```

Then the first would be an interactive command that uses CR and gets called for single keys, whereas the second would be a non-interactive function that is only used for multi-key citations. Of course, this depends on how much you want to keep the multi-key citation functionality.

bdarcus:
> Of course, this depends on how much you want to keep the multi-key citation functionality.

I need to think more on it, but my impulse is to say it's not worth the hassle of the additional complexity.

This is the place in the code that has always been prone to hard-to-fix bugs, I think precisely because of this complexity. In this case, for example, I have no idea where the bug reported in the linked issue came from.

I thought you were in favor of ditching CRM?

roshanshariff:
> I thought you were in favor of ditching CRM?

Yeah, I'm in favour of ditching CRM, but using multi-keys doesn't require CRM. In the snippet I gave above, there is no CRM but you can still operate on multi-key citations. It just means that the only user-facing interface to them is via `embark-act`.

From what I can tell, the linked issue isn't because of CR vs CRM on citation keys, right? The bug seems to be in `citar-open` here:
```elisp
         (resources
          (when resource-candidates
            (completing-read "Related resources: " resource-candidates))))
    (if resource-candidates
        (dolist (resource resources)
          (cond ((string-match "http" resource 0)
                 (browse-url resource))
                (t (citar-file-open resource))))
      (message "No associated resources"))))
```
This should be:
```elisp
         (resource
          (when resource-candidates
            (completing-read "Related resources: " resource-candidates))))
    (if resource-candidates
        (cond ((string-match "http" resource 0)
               (browse-url resource))
              (t (citar-file-open resource)))
      (message "No associated resources"))))
```

bdarcus:
> I see you made this change in this PR.

Yes, but to support configurable CRM and CR would require a different solution. It's why I just elected to make it CR only.

roshanshariff:
Yeah, I agree that making it configurable is too complicated and CR-only is the way to go.

But of course the choice of UI (CR vs CRM) is orthogonal to the programmatic interface (whether the function argument is a list `keys-entries` or a single `key-entry`). I would suggest that the UI use CR, but the programmatic interface allow for lists (which is the status quo).

If you change the functions so they only take a single key/entry, then you make it impossible to write wrapper functions that operate on multiple keys. But if the function takes a list of key-entry pairs, then it's easy to call it on a singleton list.

bdarcus:
Can you make line comment suggestions for how you suggest changing this PR, then, so I get this fix merged? 

We can then iterate ( I still want to figure out #491).

roshanshariff:
It's hard to make line suggestions, but I would just revert all the changes to `citar-open` except the ones dealing with `resource` (lines 904-912).

bdarcus:
> I would just revert all the changes to `citar-open` except the ones dealing with `resource` (lines 904-912).

I don't understand how I do that if I'm also changing to CR.

If I revert the restriction to only allow a single candidate, then doesn't everything else I changed also become wrong if more than one candidate is passed to the function?

roshanshariff:
Currently, the only place that CRM shows up is in the call to `citar-select-refs`. If you change that to `citar-select-ref`, then it'll use CR instead. But this changes its return value from a list to a single `(KEY . ENTRY)` pair. So the only thing you need to do is to wrap the result in a `list`:
```elisp
(interactive (list (list (citar-select-ref :rebuild-cache current-prefix-arg))))
```

The rest of the function can remain unchanged, because the type of the argument is the same as it was before. Does that make sense?

bdarcus:
I still get the same error:

```elisp
let: Wrong type argument: listp, "/home/bruce/org/pdf/toly2017.pdf"
```

bdarcus:
Actually, fixed it by wrapping that inner CR call in the list.

roshanshariff:
On the `interactive` call, you do need the extra `list` with `citar-select-ref`.

In the definition of `resource`, you can get rid of the wrapping `list` as long as you remove the `dolist` after that. That's what you had before, and it was correct. Sorry for the confusion; I meant to say you should keep that. I can't suggest it in code review because it'll only allow me to edit lines that are in the commit.

bdarcus:
I'm just really confused :-)

I thought I'm going through these gymnastics to continue to allow one to pass a list to the function non-interactively?

But:

1. why is that important for this function, when it's designed to be used exclusively interactively currently (by virtue of that `resource` variable)?
2. if it is, doesn't ditching the `dolist` conflict with that goal?

roshanshariff:
Haha, yeah, it's gotten a little confusing because of trying to do this in comments. Let me paste what I think the function should look like in its entirety and then I'll say why:
```elisp
(defun citar-open (keys-entries)
  "Open related resources (links or files) for KEYS-ENTRIES."
  (interactive (list
                (list (citar-select-ref
                       :rebuild-cache current-prefix-arg))))
  (let* ((key-entry-alist (citar--ensure-entries keys-entries))
         (files
           key-entry-alist
           (append citar-library-paths citar-notes-paths)
           ;; find files with any extension:
           nil))
         (links
          (seq-map
           (lambda (key-entry)
             (citar-get-link (cdr key-entry)))
           key-entry-alist))
         (resource-candidates (delete-dups (append files (remq nil links))))
         (resource
          (when resource-candidates
            (completing-read "Related resources: " resource-candidates))))
    (if resource-candidates
        (cond ((string-match "http" resource 0)
               (browse-url resource))
              (t (citar-file-open resource)))
      (message "No associated resources"))))
```
(The only difference from your latest commit is the definition of `resource` at the bottom and the removed `dolist` after that.)

First, we need to separate two things: (1) can you pass multiple keys/entries to the function, and (2) can you open multiple files/URLs.

The use of `completing-read` vs `completing-read-multiple` to choose a *resource* only has to do with (2), not (1). And that's also what determines whether you have a `dolist` at the end of the function or not. My proposal (which is also the status quo of this PR) is to use `completing-read` to select a single resource to open.

Does that make sense as far as (2) is concerned?

bdarcus:
> Does that make sense as far as (2) is concerned?

It does.

roshanshariff:
Okay, so then for (1) I like the code in its current state: you can call the `citar-open` with a list of key-entry pairs through elisp, but when called interactively it uses CR (via `citar-select-ref`) and you select a single key/entry.

The nice thing about this is that, if you want, you can write a wrapper function like what I had above:
```elisp
(defun citar-open-multi (keys) ; Name TBD
  (citar-open (split-string keys " & "))
```
Then you can use this wrapper function for `embark-act` on multi-key citations (while still using the original `citar-open` for single-key citations. If you remove `citar-open`'s ability to act on multiple keys, then it becomes impossible to write such a wrapper function.

bdarcus:
OK, I follow that, and am fine leaving that part status quo. 

But I don't ATM see the compelling argument for this. We're maintaining complexity (demonstrated by the length of this back-and-forth)  for unclear practical benefit in my current view (which may change).

I should also point out I already changed the other functions to remove that capability in #496, without fully realizing the implications.

This probably just means we need to think about this some more before we settle things more firmly, which I'm hoping to do soon-ish.

roshanshariff:
Yeah, that seems like a reasonable approach. It seems to me that #496 was conflating the interactive interface to a function (which is also what Embark cares about) with its non-interactive elisp interface. And, as we did here, I think in some cases it's good for the elisp interface to be powerful (i.e. allow lists of keys) but the interactive/Embark interface to be simple (i.e. use CR and wrap the result in a single-element list).

In particular, this is a good fit for any functions where operating collectively on multiple keys has different behaviour from acting on each key individually. Like in this PR, where `citar-open` on multiple keys collects all the resources and presents them in one list, which is different from looping over each key and repeatedly asking the user to select a resource for each.

PR #513, opened 2021-12-29
Use consult-multi for citar-open, include notes

bdarcus:
Working on #491 on this new branch.

Just need to adapt https://github.com/bdarcus/citar/pull/491#issuecomment-1001700408, as `embark-act-all` isn't currently working.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -55,6 +55,7 @@
 (defvar embark-meta-map)
 (defvar embark-transformer-alist)
 (defvar embark-multitarget-actions)
+(defvar embark-default-action-overrides)
 ;(defvar citar-org-open-note-function)
 
 ;;; Variables
@@ -895,8 +896,7 @@ into the corresponding reference key.  Return
   (add-to-list 'embark-keymap-alist '(citar-citation . citar-citation-map))
   (add-to-list 'embark-pre-action-hooks '(citar-insert-edit embark--ignore-target))
   (when (boundp 'embark-multitarget-actions)
-    (dolist (command (list #'citar-open-library-files #'citar-attach-library-files
-                           #'citar-insert-bibtex #'citar-insert-citation
+    (dolist (command (list #'citar-insert-bibtex #'citar-insert-citation
                            #'citar-insert-reference #'citar-copy-reference
                            #'citar-insert-keys #'citar-run-default-action))
       (add-to-list 'embark-multitarget-actions command))))
@@ -912,7 +912,9 @@ into the corresponding reference key.  Return
   (when (and citar-library-paths
              (stringp citar-library-paths))
     (message "Make sure 'citar-library-paths' is a list of paths"))
-  (let* ((key-entry-alist (citar--ensure-entries keys-entries))
+  (let* ((embark-default-action-overrides
+          '((consult-multi . citar-open-multi)))
+         (key-entry-alist (citar--ensure-entries keys-entries))
          (files
           (citar-file--files-for-multiple-entries
            key-entry-alist
@@ -928,11 +930,14 @@ into the corresponding reference key.  Return
          (resource
           (when resource-candidates
             (citar-select-resource files links))))
-    (if resource-candidates
-        (cond ((string-match "http" resource 0)
-               (browse-url resource))
-              (t (citar-file-open resource))))
-    (message "No associated resources")))
+    (citar-open-multi resource)))
+
+(defun citar-open-multi (selection)
+  "Act appropriately on SELECTION when type is 'consult-multi'.
+For use with 'embark-act-all'."
+  (cond ((string-match "http" selection 0)
+         (browse-url selection))
+        (t (citar-file-open selection))))
 
 (defun citar--library-file-action (key-entry action)
   "Run ACTION on file associated with KEY-ENTRY."
--- a/citar.el
+++ b/citar.el
@@ -430,14 +430,37 @@ REBUILD-CACHE and FILTER."
            (category . file))
        (complete-with-action action files string predicate)))))
 
-(defun citar-select-group-related-sources (file transform)
-  "Group by FILE by source, TRANSFORM."
-    (let ((extension (file-name-extension file)))
-      (when transform file
-        ;; Transform for grouping and group title display.
+(defun citar-select-resource (files &optional links)
+  "Select resource from a list of FILES, and optionally LINKS."
+  (let* ((files (mapcar
+                 (lambda (cand)
+                   (abbreviate-file-name cand))
+                 files))
+         (resources (append files (remq nil links))))
+    (dolist (item resources)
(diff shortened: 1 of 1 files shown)


citar.el
```diff
@@ -430,14 +430,39 @@ REBUILD-CACHE and FILTER."
...
+                   (abbreviate-file-name cand))
+                 files))
+         (resources (append files (remq nil links))))
+    (dolist (item resources)
+      (cond ((string-match "http" item 0)
+             (add-text-properties
+              0 (length item) `(consult-multi (url . ,item)) item))
+            (t
+             (add-text-properties
+              0 (length item) `(consult-multi (file . ,item)) item)))
```
minad, line 446: Careful, this modifies the strings! I recommend propertize which keeps the original strings intact.

bdarcus:
> As a small demonstrator, this works for me:

In your example, @minad, this seems the key line:

```elisp
(pcase (get-text-property 0 'consult-multi (car (member result candidates)))
```

But the current `citar-open` function, adapted from @localauthor's work, calls a separate `citar-select-resource` function.

Am I correct that doing that will mean the resulting candidate will only be a string, with properties stripped?

If yes, does that mean we need to remove that separate function, and do the work inside `citar-open`?

Does my question make sense?

minad:
No, you don't need any kind of separate function nor any kind of embark override. This is by design of Embark that you don't need anything. The way I wrote the command in https://github.com/bdarcus/citar/pull/491#issuecomment-1001700408 uses `member` to lookup the original candidate. Then the property is extracted from that candidate. But you could also distinguish the candidates purely based on their string content (http prefix for example). The `consult-multi` properties are only needed to enable Embark and Marginalia.

citar.el
bdarcus: Yeah, I had just copied over the code from #491. Fixing this now.

bdarcus:
> But you could also distinguish the candidates purely based on their string content (http prefix for example).

Which is what the current code does.

I fixed those lines to use `propertize`, but I still get this error when running `embark-act-all` on the list of resources (see screenshot below).

```elisp
Keys not found: ~/org/pdf/toly2017.pdf
embark PCH: (wrong-type-argument stringp nil)
```

Any ideas?

image

minad:
Not really, but the issue is somewhere in Citar - your code throws the wrong-type-argument error. My example code works for you?

bdarcus:
> Not really, but the issue is somewhere in Citar - your code throws the wrong-type-argument error. 

OK.

> My example code works for you?

Yes.

roshanshariff:
@bdarcus,

To clarify, you're getting the error when you:
1. Run `citar-open`
2. Select an entry and press `RET`
3. In the list of files/URLs, run `embark-act-all`
4. Press 'RET` to run `citar-open`

If so, it seems the reason is simple: `citar-open` prompts for and expects a citation key, but is being run with a file/URL. The error message is actually correct: `~/org/pdf/toly2017.pdf` is not a valid key and is not found. When you add `citar-open` to `embark-multitarget-actions`, it's taking a different code path in which invalid keys get ignored (by `citar--ensure-entries`, since it is run non-interactively).

I'm not sure what the best way to fix this problem is, but it comes down to the fact that `citar-open`:
1. Prompts for a key
2. Then prompts for a file/URL associated with the key
So it's going to work as an Embark action for keys but not for files/URLs, since Embark injects the target as the first string the function prompts for.

bdarcus:
Yes, your analysis is correct.

roshanshariff:
Another alternative is to `let` bind `this-command` to the correct command. I'll take a stab at implementing this.

minad:
Okay, this is a reasonable solution - overriding the default action when you read the second time from the minibuffer.

roshanshariff:
Ah, yes, you're right. I see that `citar-open-multi` isn't an interactive command, so it seems like a bad idea to set `this-command` to it.

minad:
@bdarcus I prepared PRs to generalize `consult-multi` to `multi-category` in case you want to proceed with this.

* Consult: https://github.com/minad/consult/pull/488
* Marginalia: https://github.com/minad/marginalia/pull/119
* Embark: https://github.com/oantolin/embark/pull/440

bdarcus:
Sounds good @minad. I'll merge this once I get it fully working, and update the category once the related PRs get merged.

This should almost work, @roshanshariff, but I think a hint on the remaining problem is here:

image

I think "generals" should be "consult-multi".

I do think this relates to the conversation yesterday (see #509), in the sense to keep this as simple as possible, we don't really need or want to add `citar-open` to `embark-multitarget-actions`.

bdarcus:
Nevermind; I just needed to recompile and quit and reopen Emacs. This works correctly ATM!

citar.el
```diff
@@ -912,7 +912,9 @@ into the corresponding reference key.  Return
   (when (and citar-library-paths
              (stringp citar-library-paths))
     (message "Make sure 'citar-library-paths' is a list of paths"))
-  (let* ((key-entry-alist (citar--ensure-entries keys-entries))
+  (let* ((embark-default-action-overrides
+          '((consult-multi . citar-open-multi)))
```
minad, line 916: I think @localauthor has been right that we also need the file and url overrides then in case you act on single candidates. Otherwise you will still get the injection problem. Did you test acting on single candidates with RET and acting on all candidates with the default action?

bdarcus: > Did you test acting on single candidates with RET and acting on all candidates with the default action?

Yes; worked fine.

Should I add them anyway?

minad: Hmm. I am kind of confused right now. I have to think about it.

minad: Did you load `embark-consult` in your tests?

bdarcus: > Did you load `embark-consult` in your tests?

I had forgot to check that, but just now added `(require 'embark-consult)`; still no errors.

Should I require `embark-consult` then?

bdarcus: Note I do get the embark error on `citar-open-library-files`, I just realized. That's because I haven't added the override there.

minad: > Should I require embark-consult then?

No, we will merge the multi-category renaming soon.

localauthor: So pleased to return from holiday to find all of this merged! :)

A few small things:

It seems the resources are not being propertized correctly in `citar-select-resources`, leading to all resources being categorized as "general." This means the embark-file and embark-url maps are not available when `embark-act`-ing on individual resources. 

Also, `citar-file-open` needs to `expand-file-name`, since they are abbreviated in `citar-select-resources`.

I'm working up a small PR with these changes. I'll also include the fresh `multi-category` type as well.

bdarcus: Make sure you branch or rebase off latest main, as I've resolved some of these since I merged this.

PR #514, opened 2021-12-29
fixes to multi-category implementation

localauthor:
per comment in this thread:
https://github.com/bdarcus/citar/pull/513#discussion_r776470085

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -440,9 +440,9 @@ REBUILD-CACHE and FILTER."
          (resources (append files (remq nil links))))
     (dolist (item resources)
       (cond ((string-match "http" item 0)
-             (propertize item `consult-multi '(url . ,item)))
+             (add-text-properties 0 (length item) `(multi-category (url . ,item)) item))
             (t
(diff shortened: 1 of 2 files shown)


citar.el
```diff
@@ -440,9 +440,9 @@ REBUILD-CACHE and FILTER."
          (resources (append files (remq nil links))))
     (dolist (item resources)
       (cond ((string-match "http" item 0)
-             (propertize item `consult-multi '(url . ,item)))
+             (add-text-properties 0 (length item) `(multi-category (url . ,item)) item))
             (t
-             (propertize item `consult-multi '(file . ,item))))
+             (add-text-properties 0 (length item) `(multi-category (file . ,item)) item)))
```
minad, line 445: This change is not good since it mutates the items. Use propertize and push the propertized item instead.

bdarcus: I should have caught this.

Do you have time to do a PR @localauthor?

localauthor: Done. Out of curiosity, are there implications of mutating vs. not mutating in this context, or is it just generally better to not?

PR #519, opened 2022-01-04
Add error handling to citar-insert-citation and citar-open

### LINKED ISSUE #441: citar-insert-citation not exiting early on invalid major-modes
**Describe the bug**
A minor UX issue: `citar-insert-citation` follows through with the interactive selection, even though the major mode isn't supported.

**To Reproduce**
1. Open an emacs-lisp file (or any file not in a supported major-mode)
2. Run `citar-insert-citation`
3. Interactive prompts shows, and upon selection, we see:

```
Citation insertion is not supported for emacs-lisp-mode
```

**Expected behavior**
Running `citar-insert-citation` should exit immediately in unsupported major-modes.

**Emacs version:**
GNU Emacs 29.0.50 (build 1, x86_64-apple-darwin20.6.0, NS appkit-2022.60 Version 11.5.2 (Build 20G95)) of 2021-11-22

localauthor:
Adds a major-mode check to `citar-insert-citation`
Fixes #441

~Adds a defvar `citar-major-modes`; a list of currently supported major modes. (This list could be made programatically, but that seemed like overkill.)~

Adds "No associated resource" message to `citar-open`, sidestepping a "Wrong type argument" error.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -932,8 +932,9 @@ into the corresponding reference key.  Return
            key-entry-alist))
          (resource-candidates (delete-dups (append files (remq nil links))))
          (resource
-          (when resource-candidates
-            (citar-select-resource files links))))
+          (if resource-candidates
+              (citar-select-resource files links)
+            (error "No associated resources"))))
     (citar-open-multi resource)))
 
 (defun citar-open-multi (selection)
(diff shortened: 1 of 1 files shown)


bdarcus:
> Adds a defvar `citar-major-modes`; a list of currently supported major modes.

I don't see this.

localauthor:
> I don't see this.

Oversight! Sorry about that.

citar.el
```diff
@@ -277,6 +277,9 @@ of all citations in the current buffer."
   :group 'citar
   :type 'alist)
 
+(defvar citar-major-modes '(org-mode latex-mode markdown-mode)
```
bdarcus, line 280: What about text-mode? And would that cover emails?

localauthor: The variable is used in interactive specification of `citar-insert-citation`, to prevent the issue raised in #441. So it only lists modes in which citations can be inserted.

Thoughts on a different name and/or docstring?

localauthor: In other words, it wouldn't have any bearing on the use of Citar in text-mode or in emails.

I gave it such a general name since it is just a list of modes derived from 'citar-major-mode-functions'. Perhaps it would make sense to mention that in the docstring?

Something like:
"List of major modes for which there is major mode specific functionality, per 'citar-major-mode-functions'."

The other option --- which I suggested was overkill, but maybe isn't --- would be to skip adding this defvar and make the list programmatically in 'citar-insert-citation' itself, based on `citar-major-mode-function`.

bdarcus: I'm unclear why we need the defvar to begin with. The relevant function only works if there's a corresponding major-mode function.

Right?

So what am I missing?

localauthor: https://github.com/bdarcus/citar/issues/441#issuecomment-979170594

It's to avoid the UX issue mentioned here. Currently, the user can call `insert-citation` in an unsupported mode and go through with selecting a reference, only to be told afterwards that citation-insertion is not allowed in that mode.

PR #520, opened 2022-01-05
Update README to also set citar-bibliography

save-buffer:
The current org-cite example doesn't seem to work for me unless I set `citar-bibliography` (I get "Make sure to set citar-bibliography" if I try to insert a citation). This change just updates the readme to set this variable.


bdarcus:
Thanks!

@localauthor @roshanshariff - this reminds me. If a user only uses local bibs, should't setting `citar-bibliography` be optional?

roshanshariff:
As far as I can tell, setting `citar-bibliography` is indeed optional. Currently that message is printed when it is unset and there is no local bibliography either. Maybe it should be rephrased?

PR #534, opened 2022-01-19
fix(latex): duplicate comma

### LINKED ISSUE #533: Calling `citar-insert-citation` in LaTeX files should not result in multiple commas
Start with point after the comma following the citation key: 

```tex
\cite{entry1,}
```

Call `citar-insert-citation` and select entry with key `entry2`. The expected result (and what we get using `reftex`) is

```tex
\cite{entry1,entry2}
```

Instead, what we get is 

```tex
\cite{entry1,, entry2}
```

I do not think the space after the comma is necessary nor ideal, but I do not know whether this is inconsistent with best practices. The double comma, though, seems like a bug. 

From what I can tell, the behavior is due to the way `citar-latex-insert-citation` is defined. Changing [this line](https://github.com/bdarcus/citar/blob/b24b5b94b7d5f6e7e96331b48a22e977e0013de1/citar-latex.el#L170) to 

```el
(unless (or (equal ?} (preceding-char)) (equal ?, (preceding-char))) (insert ", ")))
```

seems to deal with the problem, but I do not know whether this introduces unintended issues.

Running Emacs 28.0.90 on macOS.

bdarcus:
Fix #533

-----

I've not tested this yet (I don't really use or work with the latex code), but can you take a look @apc @Hugo-Heagren @roshanshariff?

### DIFF
--- a/citar-latex.el
+++ b/citar-latex.el
@@ -167,7 +167,9 @@ inserted."
   (when keys
     (if (citar-latex-is-a-cite-command (TeX-current-macro))
         (progn (skip-chars-forward "^,}")
-               (unless (equal ?} (preceding-char)) (insert ", ")))
+               (unless (or (equal ?} (preceding-char))
+                           (equal ?, (preceding-char)))
+                 (insert ", ")))
       (let ((macro
 	     (or command
 		 (if (xor invert-prompt citar-latex-prompt-for-cite-style)


apc:
Will have a look when I’m back at my computer tomorrow.

roshanshariff:
This code is fine as a quick fix, but it only works in a very narrow situation, where you have something like `\cite{key1,}` and point is between the `,` and `}`. It won't work if point is before the comma, or if there is a space between the comma and closing brace. Perhaps it's fine to merge for now, but we need to revisit it and make sure the logic is sound. (I think the existing code isn't quite right either; why is it skipping the comma if the previous character is `}`?)

apc:
So, following up on @roshanshariff's concerns, I'm not sure what the best way to separate 'linting' issues, broadly speaking, from how to think of the behavior of this particular function. One certainly wouldn't want calling `citar-insert-citation` to introduce syntax errors, but would we want `citar-insert-citation` to also cleanup previous errors? 

For instance, if you have this in your code

```tex
\cite{key, ,, ,|}
```

where `|` stands for point, would you want calling `citar-insert-citation` to remove the 'junk' before point? I probably wouldn't expect it to. 

The case I brought up in the original issue seemed different in kind from this, in that a user can reasonably expect that inserting the comma is necessary if they want to add another citation. In other words, from the user's perspective, it seems naturally to worry that if you have point after "key1" here

```tex
\cite{key1|}
```

and you call `citar-insert-citation` in order to cite the entry with key key2, you will end up with

```tex
\cite{key1key2}
```

rather than the desired

```tex
\cite{key1,key2}
```

Does that make sense?

bdarcus:
So where do we stand on this?

roshanshariff:
I've opened #541 with a somewhat more comprehensive fix:
* Keys are added after the comma following point, if there is one. A following comma is inserted only if necessary.
* Otherwise, keys are added before the closing `}`, inserting preceding comma if necessary.
* If point is before the opening `{`, keys are added immediately after that. This matches the Org mode behaviour.
* If there is a bare citation command with no argument list (like `\cite`), insert keys enclosed in `{ }` after the macro.

bdarcus:
Closing in favor of #541 then.

apc:
Will do. Thanks for the heads up!

PR #537, opened 2022-01-19
Update default value of `citar-templates`

apc:
Biblatex prefers 'journaltitle' to 'journal', and at least prior to Emacs 28, `bibtex.el` wouldn't even recognize 'journal' as an alias for 'journaltitle' if you had `bibtex-dialect` set to `biblatex`. This small change is just to reflect the fact that many entries biblatex is happy with lack a 'journal' field but have a 'journaltitle' field.


bdarcus:
Merged.

I rewrote the commit message to simplify it, and confirm to the 72 character line length.

PR #538, opened 2022-01-19
Add error to citar-get-template

### LINKED ISSUE #536: Getting "Key not found" error when calling `citar-copy-reference`
Call `citar-copy-reference`, highlight an entry, and press RET. Expected a message of the form "Copied: ..." and a list of formatted references either in the kill-ring or the system clipboard. Instead, what I get is a "Key not found" error. I'm assuming this is a bug, but maybe I've misunderstood something?

Emacs version: 28.0.90. Citar version: v0.9-66-gb24b5b9

localauthor:
Fix #536

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -698,7 +698,11 @@ are refreshed."
 
 (defun citar-get-template (template-name)
   "Return template string for TEMPLATE-NAME."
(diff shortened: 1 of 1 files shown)


bdarcus:
Thanks!

Issue #518, opened 2022-01-01
Support multiple notes in one org file


bdarcus:
## Summary

Given that org-roam has removed the file-per-note restriction (see also #517), we might do the same.

## Details

I'm just not sure how best to do that ATM, as I don't have a ton of experience coding and working with org structure.

There are also, I think, some design nuances to consider; for example:

If we remove this restriction, seems it should be more flexible than only a single org file. For example, maybe I want to mix single-per-file and multi-per-file (though not sure that's a great idea in org-roam given its robust linking, along with [transclusion](https://github.com/nobiot/org-transclusion))? 

Regardless, how to best configure what UX here?

OTOH, maybe if people want that sort of flexibility, they should store the note file paths in the actual record?

## Counterpoint

One could reasonably argue this is not worth the hassle, and that we have a design now that works well for one-file-per-note/reference use cases, that this is increasingly what we're seeing in diverse new note-taking packages, and that users here aren't actually requesting it. For that reason, I've marked the milestone as "future" (post 1.0).

ademberbic:
Hi @bdarcus, thank you so much for your work on this package, it's been a joy to use so far! For whatever it's worth, I recently slapped together a few functions to add this functionality to citar à la ivy-bibtex, which I'd been using before. Needless to say that it's very quick-and-dirty code, and you may well have had a completely different implementation in mind, but [here it is](https://gist.github.com/ademberbic/f99a15dc4a38d1dedd3235b26ae6828c) nonetheless. Hope it proves useful, and thanks again.

bdarcus:
Thanks @ademberbic. Based on this, do you have any inputs on my design questions above?

I don't actually recall how ivy-bibtex handles this.

ademberbic:
Good question @bdarcus—I'm not at all familiar with org-roam myself, so I'm afraid I can only be of so much use here.

With ivy-bibtex and helm/bibtex, it's an either/or choice: if `bibtex-completion-notes-path` is a filename, it assumes one-file-for-notes; if it's a directory, it assumes one-file-per-note. Their implementation is [here](https://github.com/tmalsburg/helm-bibtex/blob/ce8c17690ddad73d01531084b282f221f8eb6669/bibtex-completion.el#L1398): if multiple keys are selected (which, out of laziness, my code assumes will never happen), it either opens all the notes files or opens an org-capture-style narrowed buffer for each, depending on the system in use.

[...]

PR #539, opened 2022-01-21
Add initial-input argument to citar-select-ref

andersjohansson:
This can be useful when invoking it from another function, like
`org-roam-capture`.

My use case is for creating a new reading note with org-roam. If `org-roam-node-find` doesn’t find the author/title searched for, this is passed to `org-roam-capture`, where the citar search is initiated from the entered search string in roam ("${title}" in the capture template below).

I bypass the citar machinery in for file-notes here, because it seemed to assume a one note per file-system (the old org-roam way) which I don’t like (I currently use yearly files with date-trees for my notes).

``` emacs-lisp
(setq org-roam-capture-templates
'(
("l"
"Reading notes"
entry
"* Read %(aj/org-return-reference-author-year-title \"${title}\") :#reading%^g
:PROPERTIES:
:ID: %(org-id-new)
:ROAM_REFS: %(aj/org-last-returned-citation-key)
:END:
%U
%i
%?"
:empty-lines-before 1
:target (file+datetree "~/notes/notes-%<%Y>.org" month))))


(defvar aj/org-last-returned-citation-key nil)
(defun aj/org-last-returned-citation-key ()
  aj/org-last-returned-citation-key)

(defun aj/org-return-reference-author-year-title (&optional initial-input)
  "Fetch a simply formatted reference and return it as a string.
Possibly passss INITIAL-INPUT"
  (let* ((ref (citar-select-ref :initial-input initial-input))
         (key (car ref)))
    (setq aj/org-last-returned-citation-key key)
    (citar--format-entry-no-widths ref "${author editor} (${year issued date}) /${title}/")))
```

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -352,7 +352,7 @@ and other completion functions."
                           (or (null predicate) (funcall predicate cand))))))))
           (complete-with-action action candidates string predicate))))))
 
-(cl-defun citar-select-ref (&optional &key rebuild-cache multiple filter)
+(cl-defun citar-select-ref (&optional &key rebuild-cache multiple filter initial-input)
   "Select bibliographic references.
 
 A wrapper around 'completing-read' that returns (KEY . ENTRY),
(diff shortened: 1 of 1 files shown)


andersjohansson:
Oh yes, you already pointed to the one note per file-issue here: #518

bdarcus:
Awhile back, I removed initial-input support in favor of pre-filtering the candidates.

andersjohansson:
I see! Indeed, it doesn't work so well with CRM or `consult-completing-read-multiple`, as discussed earlier. (my use case here was for single ref selection, so I didn't realize that)

 In my usage, I believe it is a little different from pre-filtering because there is no way to go back and change a bad search without quitting the command then. I see it as somewhat different than the aim to pre-filter by presence of pdfs etc, like the difference between the predicate and initial-input arguments to completing-read.

Perhaps it would be suitable only when the `:multiple` argument is not given to `citar-select-ref` then. I also saw the discussions on moving away from CRM in #492, so I guess this could be reconsidered if you choose to do that everywhere.

bdarcus:
I just mentioned the context, not entirely sure what you were trying to do ;-)

andersjohansson:
Yeah, the point is to go on from a failed search for an existing roam node to immediately start searching with the same terms in the citar-database in order to create a roam node for the reference. So I do think it is a valid use case for `initial-input`.

PR #540, opened 2022-01-21
Use format-spec instead of s-format

### LINKED ISSUE #335: remove s, s-format dependency
In the interest of further removing dependencies, it would be nice to replace the current template formatting code, perhaps with `replace-regexp-in-string`.

Probably the best approach would just be to adapt the two relevant s functions; for example:

https://github.com/bdarcus/bibtex-actions/pull/336/commits/30ed7f61d067fc52aa721b56db11282a1a450829

localauthor:
Use format-spec for display templates, rather than s-format.

This could do with some testing.

Closes #335
(Unless there are other s functions I overlooked.)

### DIFF
--- a/README.org
+++ b/README.org
@@ -160,21 +160,44 @@ For the prefix, you can filter for associated files or notes using =has:file= or
 *** Templates
 
 The =citar-templates= variable configures formatting for these sections, as well as the default note function.
-Here's the defaults:
+Here are the defaults:
 
 #+BEGIN_SRC emacs-lisp
 (setq citar-templates
-      '((main . "${author editor:30}     ${date year issued:4}     ${title:48}")
-        (suffix . "          ${=key= id:15}    ${=type=:12}    ${tags keywords:*}")
-        (note . "Notes on ${author editor}, ${title}")))
+      '((main . "%->30a     %>4d     %->48t")
+        (suffix . "          %>15k    %>12T   %K")
+        (preview . "%>a (%>4d) %t, %p.\n")
+        (note . "Notes on %a, %t")))
 #+END_SRC
+
+The formatting is done using Emacs's built-in =format-spec= library. For
+details about setting the templates, see that library and the GNU Emacs Lisp
+Reference Manual =(info "(elisp) Custom Format Strings")=.
+
+For purposes specific to Citar, each alphabetic character in the templates
+above (e.g., "a", "d", "t", etc.) is keyed to one or more bibliographic
+fields as specified in the variable =citar-template-fields=:
+
+#+BEGIN_SRC emacs-lisp
+(setq citar-template-fields
+      '((?a . ("author" "editor"))
+        (?d . ("date" "year" "issued"))
+        (?t . ("title"))
+        (?k . ("=key="))
+        (?T . ("=type="))
+        (?K . ("tags" "keywords"))
+        (?p . ("journal" "journaltitle"
+               "publisher" "containter-title"
+               "collection-title")))
+#+END_SRC
+
+You may associated multiple field names with a given letter, as in the case
+of "a" above; the formatter will print the first one it finds.
 
 Note:
 
-1. You may include multiple variables in a field; the formatter will print the first one it finds.
-2. If you plan to use CSL JSON at all, you can and should include CSL JSON variables names where appropriate as such options. The default main template dates field demonstrates this.
-3. The asterisk signals to the formatter to use available space for the column.
-4. The note template does not take widths, as formatting is inline there rather than columnar.
+1. If you plan to use CSL JSON at all, you can and should include CSL JSON variables names where appropriate as such options. The default main template dates field demonstrates this.
+2. The note template does not take widths, as formatting is inline there rather than columnar.
 
 *** Icons
 
--- a/README.org
+++ b/README.org
@@ -1,4 +1,9 @@
+
 [[https://melpa.org/#/citar][file:https://melpa.org/packages/citar-badge.svg]]
+
+* Breaking Changes
+
+- Templating is now handled with Emacs's built-in =format-spec= library, so any customizations to the variable =citar-templates= will need to be updated to conform to that specification. See [[#templates][Templates]] for more details.
 
 * Citar
   :PROPERTIES:
@@ -194,10 +199,7 @@ fields as specified in the variable =citar-template-fields=:
 You may associated multiple field names with a given letter, as in the case
 of "a" above; the formatter will print the first one it finds.
 
-Note:
-
-1. If you plan to use CSL JSON at all, you can and should include CSL JSON variables names where appropriate as such options. The default main template dates field demonstrates this.
-2. The note template does not take widths, as formatting is inline there rather than columnar.
+Note: If you plan to use CSL JSON at all, you can and should include CSL JSON variables names where appropriate as such options. The default main template dates field demonstrates this.
 
 *** Icons
 
--- a/README.org
+++ b/README.org
@@ -174,8 +174,8 @@ The formatting is done using Emacs's built-in =format-spec= library. For
 details about setting the templates, see that library and the GNU Emacs Lisp
 Reference Manual =(info "(elisp) Custom Format Strings")=.
 
-For purposes specific to Citar, each alphabetic character in the templates
-above (e.g., "a", "d", "t", etc.) is keyed to one or more bibliographic
+For purposes specific to Citar, the alphabetic characters in the templates
+above (e.g., "a", "d", "t", etc.) are keyed to one or more bibliographic
 fields as specified in the variable =citar-template-fields=:
 
 #+BEGIN_SRC emacs-lisp
--- a/citar.el
+++ b/citar.el
@@ -98,7 +98,7 @@ to include."
   :type '(repeat string))
 
 (defcustom citar-templates
-  '((main . "%->30a     %4d     %->48t")
+  '((main . "%->30a     %>4d     %->48t")
     (suffix . "          %>15k    %>12T   %K")
     (preview . "%>a (%d) %t, %p.\n")
     (note . "Notes on %a, %t"))
--- a/citar.el
+++ b/citar.el
@@ -100,7 +100,7 @@ to include."
 (defcustom citar-templates
   '((main . "%->30a     %>4d     %->48t")
     (suffix . "          %>15k    %>12T   %K")
-    (preview . "%>a (%d) %t, %p.\n")
+    (preview . "%>a (%>4d) %t, %p.\n")
     (note . "Notes on %a, %t"))
   "Configures formatting for the bibliographic entry.
 
--- a/citar.el
+++ b/citar.el
@@ -138,7 +138,7 @@ the list will be the replacement."
                 :value-type (repeat string)))
 
 (defcustom citar-format-reference-function
-#'citar-format-reference
+  #'citar-format-reference
   "Function used to render formatted references.
 
 This function is called by 'citar-insert-reference' and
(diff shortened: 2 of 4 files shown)


bdarcus:
Did you see this comment? https://github.com/bdarcus/citar/issues/335#issuecomment-1018881966

localauthor:
> Did you see this comment? [#335 (comment)](https://github.com/bdarcus/citar/issues/335#issuecomment-1018881966)

Yes, I responded there, but will restate my response here:

It seems like truncation would generally be a problem specific to titles, and perhaps authors/editors.

So, would it be unreasonable to include those fields in full in the "candidate-hidden" variable of `citar--format-candidates`?

bdarcus:
Yeah, that might be a solution.

I expect titles to be most common, but the original bug report was relating to authors.

bdarcus:
First test, and I see two issues:

1. alignment problems
2. biblatex dates

image

localauthor:
> First test, and I see two (I guess related) issues:
> 
> 1. alignment problems
> 2. biblatex dates

Could you post the template you're using for this?

bdarcus:
It's just the default template, which I don't understand, since it looks to me the dates should be truncated: `%4d`.

localauthor:
Right, the language of the manual is not the clearest. Try `%.4d` instead.

citar.el
```diff
@@ -99,21 +98,46 @@ to include."
   :type '(repeat string))
 
 (defcustom citar-templates
-  '((main . "${author editor:30}     ${date year issued:4}     ${title:48}")
-    (suffix . "          ${=key= id:15}    ${=type=:12}    ${tags keywords keywords:*}")
-    (preview . "${author editor} (${year issued date}) ${title}, ${journal journaltitle publisher container-title collection-title}.\n")
-    (note . "Notes on ${author editor}, ${title}"))
+  '((main . "%->30a     %4d     %->48t")
```
bdarcus, line 101:
```suggestion
  '((main . "%->30a     %.4d     %->48t")
```

localauthor: Actually, this isn't right either: if there is no date, then it doesn't leave 4 blank spaces, therefore, alignment problems.

I think it should be `%>4d`

bdarcus: > I think it should be `%>4d`

Correct.

citar.el
```diff
@@ -99,21 +98,46 @@ to include."
...
+    (suffix . "          %>15k    %>12T   %K")
+    (preview . "%>a (%d) %t, %p.\n")
```
bdarcus, line 103: What should the date syntax be here?

localauthor: Should be `%>4d` also, to restrict it to the year.

bdarcus: So IUC, still:

- [ ] this
- [ ] the truncation issue above
- [ ] docs update

localauthor: Yes. I'll commit this template fix with the truncation solution I suggested above (adding `(full-fields (citar--format-entry entry "%a %t")` to `citar-format-candidates`.

Working on the docs now.

bdarcus: > Wondering if there should be some kind of "breaking change" notification up top, for anyone who has customized `citar-templates`?

Yeah, I was wondering about that too. We do have a changelog, but never update it!

README.org
```diff
@@ -160,21 +160,44 @@ For the prefix, you can filter for associated files or notes using =has:file= or
...
+        (?t . ("title"))
+        (?k . ("=key="))
+        (?T . ("=type="))
+        (?K . ("tags" "keywords"))
+        (?p . ("journal" "journaltitle"
+               "publisher" "containter-title"
+               "collection-title")))
+#+END_SRC
+
+You may associated multiple field names with a given letter, as in the case
```
bdarcus, line 194:
```suggestion
You may associate multiple field names with a given letter, as in the case
```

README.org
```diff
@@ -160,21 +160,44 @@ For the prefix, you can filter for associated files or notes using =has:file= or
...
+of "a" above; the formatter will print the first one it finds.
+
 Note:
 
-1. You may include multiple variables in a field; the formatter will print the first one it finds.
-2. If you plan to use CSL JSON at all, you can and should include CSL JSON variables names where appropriate as such options. The default main template dates field demonstrates this.
-3. The asterisk signals to the formatter to use available space for the column.
-4. The note template does not take widths, as formatting is inline there rather than columnar.
+1. If you plan to use CSL JSON at all, you can and should include CSL JSON variables names where appropriate as such options. The default main template dates field demonstrates this.
+2. The note template does not take widths, as formatting is inline there rather than columnar.
```
bdarcus, line 200: Is this true, given the date formatting?

citar.el
```diff
@@ -99,21 +99,46 @@ to include."
...
+it. When there is more than one field, the non-nil field will be
+the replacement. If multiple fields have a value, the earliest in
+the list will be the replacement."
+  :group 'citar
+  :type '(alist :key-type character
+                :value-type (repeat string)))
+
 (defcustom citar-format-reference-function
-  #'citar-format-reference
+#'citar-format-reference
```
bdarcus, line 141: Formatting?

bdarcus (commented):
Looks good. I left a few minor suggestions.

If you have time, can you take a look @apc?

README.org
localauthor: Wondered about this too. It's true that note template isn't columnar, and the date width isn't specified for alignment purposes. But, it seems pretty needless to have a note on this.

localauthor: What about just:
"Note: If you plan to use CSL JSON at all, you can and should include CSL JSON variables names where appropriate as such options. The default main template dates field demonstrates this."

citar.el
bdarcus: Actually, can you please just add the "Breaking Changes" section towards the top?

Ideally we also get back to updating the changelog, where we can put more fine-grained details, but to avoid needless panic and frustration, would be good to add this to the README.

localauthor: Maybe the **!!! BREAKING CHANGE NOTICE !!!** is overdoing it a bit? 🤷‍♂️

README.org
```diff
@@ -1,5 +1,13 @@
+
 [[https://melpa.org/#/citar][file:https://melpa.org/packages/citar-badge.svg]]
 
+* *!!! BREAKING CHANGE NOTICE !!!*
```
bdarcus, line 4: I don't feel strongly about it, but you could just have it "Breaking Changes", and then have an itemized list (of one currently).

aikrahguzar:
Please consider not merging this or at asking opinion of people who are not very familiar with elisp. I think it will make it significantly harder for them to change the template and I would imagine it is one variable people are most likely to change.

It also seems to me that `format-spec` is not buying us much here beyond truncation in a sub-optimal way and inserting  whitespace. 

This moves us closer to a template which is structured and in that case I can suggest an alternative: a template is a list like
```elisp
'((20 . ("author" "editor"))
  "   "
  (4 . ("date" "year" "issued"))
  "   "
  (30 . ("title")))
```
i.e an element of the list is either a string in which it will be inserted literally or else it can be a cons consisting of a number for the length and a list for the fields in which case the fields can be looked up and then the length can be used to add invisibility properties. This seems more easily understandable to me than the `format-spec` specification and can be implemented without using either `s-format` and `format-spec`.

bdarcus:
I'm not really following you @aikrahguzar, or why you think your example is more clear and understandable.

If we're not going to use a built-in template language like format-spec, but still want to remove the s dependency, maybe the templates should stay the same?

But doing that requires us creating and maintaining more code.

roshanshariff:
I'm rather sympathetic to @aikrahguzar's view. But I also wanted to ask how strongly we're attached to using a templating system to format the completion strings.

Instead, maybe it's enough to just give a list of column widths and a list of fields to show in each column? The display function would just concatenate the values of those fields, truncated/padded to fit in the columns and separated by a space or two. Do we need more flexibility than this?

localauthor:
It's definitely true that the `format-spec` templates are less readable than the `s-format` templates. It's good to hear your concerns about that, @aikrahguzar. As you say, I would appreciate hearing from others who have customized the templates, and see about finding some acceptable balance of user experience and maintenance prospects, whether with `format-spec` or without. I'm absolutely open to either.

In the case of moving forward with `format-spec`, I had considered posting either in the readme or on the wiki some examples of `format-spec` templates, as a kind of configuration guide for new and existing users. But, again, it would be helpful for me to see others' customizations, to see what kind of examples would be most helpful. Do you have any examples, @aikrahguzar, or anyone following along?

To move forward without `format-spec` or 's-format`, it seems that keeping `s-format` style templates would require writing into citar some version of the function `s-format` --- effectively 'citar-s-format`. Is that right?

Finally, if we only need a display function, as @roshanshariff suggests, what would that look like?

I guess, to sum up, it would be helpful to see some workable alternatives to `format-spec` and `s-format` in a competing/complementary PR, to get a better sense of the feasible options. I'm absolutely ok with forgoing `format-spec` in favor of a good alternative.

bdarcus:
I agree we could have some examples to compare, and potentially for the wiki/README.

But can someone address this, just to make sure it's not lost?

> .. maybe we could have a macro or something that transforms the current syntax into format-spec?

I'm not saying we should do it, but wouldn't be it be fairly trivial, and allow retaining the current syntax?

aikrahguzar:
> I'm not really following you @aikrahguzar, or why you think your example is more clear and understandable.

That was maybe my own lack of experience with templating system combined with have some experience with lists. I do find it somewhat clear than `format-spec` because to me `format-spec` seems somewhat cryptic with all the flag it allows and which don't seem to be needed for our case. But I actually find the `s-format` version of the template much clearer than both of those.

And with that example, I was trying to get to what @roshanshariff  said better here,
> Instead, maybe it's enough to just give a list of column widths and a list of fields to show in each column? The display function would just concatenate the values of those fields, truncated/padded to fit in the columns and separated by a space or two. Do we need more flexibility than this?


> I agree we could have some examples to compare, and potentially for the wiki/README.
> 
> But can someone address this, just to make sure it's not lost?
> 
> > .. maybe we could have a macro or something that transforms the current syntax into format-spec?
> 
> I'm not saying we should do it, but wouldn't be it be fairly trivial, and allow retaining the current syntax?
> 
> In effect, that would allow us to have two separate questions:
> 
>     1. which template syntax?
> 
>     2. which template backend code?
> 
> 
> EDIT: there's also the [org-roam function](https://github.com/org-roam/org-roam/blob/c0c240b975c82a4e9b978a9e55f3af415aa91de6/org-roam-utils.el#L165-L195) that minad referenced.

That function doesn't seem like a lot of code!

aikrahguzar:
> In the case of moving forward with `format-spec`, I had considered posting either in the readme or on the wiki some examples of `format-spec` templates, as a kind of configuration guide for new and existing users. But, again, it would be helpful for me to see others' customizations, to see what kind of examples would be most helpful. Do you have any examples, @aikrahguzar, or anyone following along?

I think examples can go a long way if we do go the `format-spec` way. In my config, I have the title as the `*` field so that it can take whatever the space is left over after the rest of the fields. It is not clear to me if that is still possible after it but if it is that would be a nice example for me.

bdarcus:
Thoughts on next steps here?

bdarcus:
Just bumping this again.

aikrahguzar:
> Just bumping this again.

My preference would be not changing the format and going the `org-roam` route. That function seems like it should be enough for what we need here. 

But I feel I might have expressed myself too strongly the first time :) so I am fine with this moving forward too.

bdarcus:
> My preference would be not changing the format and going the `org-roam` route. That function seems like it should be enough for what we need here.

I looked at it again, and I think not.

Unless I'm missing something, it doesn't have support for column rendering.

I still wonder about a macro or function which wraps this code here, so the template syntax stays the same, but the backend code changes.

bdarcus:
> It is true that this doesn't have direct support for column rendering but the code for that is already in `citar` and this doesn't add that much.

Thanks. I implemented it in #570.

PR #541, opened 2022-01-23
citar-latex-insert-citation: improve editing of existing citations

### LINKED ISSUE #533: Calling `citar-insert-citation` in LaTeX files should not result in multiple commas

roshanshariff:
When inserting keys into an existing citation, improve the handling of cases like `\cite{key,key2,}` (with an extra comma).  Now adds a comma only if necessary to insert new keys into the desired position.

Fix #533.

### DIFF
--- a/citar-latex.el
+++ b/citar-latex.el
@@ -162,12 +162,22 @@ by `citar-latex-cite-commands'.
 If `citar-latex-prompt-for-extra-arguments' is `nil`, every
 command is assumed to have a single argument into which keys are
 inserted."
-  (unless (fboundp 'TeX-current-macro)
-    (error "Please install AUCTeX"))
   (when keys
-    (if (citar-latex-is-a-cite-command (TeX-current-macro))
-        (progn (skip-chars-forward "^,}")
-               (unless (equal ?} (preceding-char)) (insert ", ")))
+    (if-let ((bounds (citar-latex--macro-bounds)))
+        (pcase-exhaustive (progn (skip-chars-forward "^,{}" (cdr bounds))
+                                 (following-char))
+          ((guard (= (point) (cdr bounds))) ; couldn't find any of ",{}"
+           (insert "{}")
+           (backward-char))
+          ((or ?{ ?,)                   ; insert after following "{" or ","
+           (forward-char)
+           (unless (looking-at-p "[[:space:]]*[},]")
+             (insert ",")
+             (backward-char)))
+          (?}                           ; insert before "}"
+           (skip-chars-backward "[[:space:]]")
+           (unless (member (preceding-char) '(?{ ?,))
+             (insert ","))))
       (let ((macro
 	     (or command
 		 (if (xor invert-prompt citar-latex-prompt-for-cite-style)
@@ -176,7 +186,7 @@ inserted."
         (TeX-parse-macro macro
                          (when citar-latex-prompt-for-extra-arguments
                            (cdr (citar-latex-is-a-cite-command macro))))))
-    (citar--insert-keys-comma-separated keys)
+    (insert (string-join keys ","))
     (skip-chars-forward "^}") (forward-char 1)))
 
 ;;;###autoload


apc:
Works great so far. I thought of another marginal use cases which aren't quite handled by this, but I suspect this may not be worth spending any time on:  If you are inserting a cite command without something like `smartparens-mode` enabled, you may end up with unbalanced braces after calling `citar-insert-citation`, e.g. if point is after "\cite" but before the closing brace of a footnote environment, you will go from this

```tex
\footnote{For more see \cite|}
```
to this

```tex
\footnote{For more see \cite{key}|
```

For a number of reasons I don't think this is worth worrying about, but I thought I'd bring this to your attention in case you think it is worth worrying about. (And perhaps, if you do, a warning when calling `citar-insert-citation` when point is before a closing brace after a citation command could be a quick fix?)

One observation, which may have been already made: unlike the original definition, `citar-latex-insert-citations` no longer inserts a space after the comma and before inserting a new reference. I actually prefer it this way, largely because it's what I'm used to. But if @bdarcus feels otherwise, I thought I'd point this out again. 

Aside from these minor points, which I'm inclined to ignore, it works great!

roshanshariff:
I'm not sure why, but I'm seeing different behaviour on your example, @apc. When I run `citar-insert-citation` on
```latex
\footnote{For more see \cite|}
```
it doesn't detect the preceding `\cite` macro and inserts a new citation. We're limited in our ability to fix this because we're relying on AUCTeX code to parse macros, and it doesn't seem to handle this case very well.

I'm surprised that you're able to end up with unbalanced braces though, because the code always inserts braces in balanced pairs or not at all.

apc:
Sorry, I messed up in my description. I get the same behavior you describe if I start with this: 

```tex
\footnote{For more see \cite|}
```

What I should have said in my original comment is that the unbalanced parentheses occur if the user enters "\cite{" before a closing brace (from an enclosing footnote environment, say) and calls `citar-insert-citation` starting with this

```tex
\footnote{For more see \cite{|}
```

The more I think about this, though, the more I think you/we should ignore this case, since it's hard telling apart the following two cases: 

**Case 1**: User starts with a well-formed footnote, wants to insert a citation, but calls `citar-insert-citation` from a place where braces are unbalanced (you start entering the citation manually, stop after inserting the opening brace but before the closing brace, and call `citar-insert-citation` from there). Say, the user starts here: 

```tex

\footnote{For more see |}

```

goes to 

```tex
\footnote{For more see \cite{|}
```

and calls `citar-insert-citation`. 

**Case 2**: User starts with an ill-formed footnote, wants to insert a citation, but calls `citar-insert-citation` after just entering the citation command before inserting an opening brace. Say, the user starts here: 

```tex
\footnote{For more see|
```

goes to

```tex
\footnote{For more see \cite|
```

and calls `citar-insert-citation`. 

I don't know whether one can distinguish between these two cases reliably enough to make `citar-insert-citation` sensitive to the distinction (we'd want `citar-insert-citation` to only enter a closing brace in the first case, but to enter both opening and closing braces in the second one). Does that make sense?

bdarcus:
So where do we stand with this?

roshanshariff:
I think this might be ready to merge. There are a few edge cases that don't work, but they're either impractical to correctly handle or can be changed in the future. This patch adds the ability to insert new keys _before_ any existing keys, which was not possible before.

One case that is currently broken is the following:
```latex
\citep[|see, e.g.]{foo,bar}
```
where the point is before the comma in the prefix text. Both the current and new code will insert new keys after that comma. Solving this will require some more sophisticated parsing. I have ideas on how to deal with this that I will implement later when I can.

bdarcus:
OK, I'll merge; thanks.

I wonder if a capf, for key completion, is a better solution long-term?

roshanshariff:
> I wonder if a capf, for key completion, is a better solution long-term?

Potentially, yes, but it will still encounter these parsing issues to decide when it is appropriate to suggest citation keys. So I don't think it'll be any easier to implement.

bdarcus:
Yes, I considered that right after posting that!

Issue #465, opened 2021-12-07
`citar-file-open-function` should ignore `citar-file-open-external`


Yevgnen:
**Describe the bug**
`citar-file-open-function` is used to open notes and library files like PDFs. When it's set to `citar-file-open-external`, [this line](https://github.com/bdarcus/citar/blob/main/citar.el#L842) returns `nil` and the file buffer is not opened. And the note template is inserted in current buffer incorrectly.

The `open` command for mac gives 

```
 $ open "/Users/User/org/bibliography/papers/teach-2020-old-dog-new-tricks.org"
The file /Users/User/org/bibliography/papers/teach-2020-old-dog-new-tricks.org does not exist.
Perhaps you meant 'http:///Users/User/org/bibliography/papers/teach-2020-old-dog-new-tricks.org'?
```

I personally think `find-file` is almost enough for `org`, `md` notes for Emacs users. 

**To Reproduce**
1. Set `citar-file-open-function` to `citar-file-open-external`.
2. Open a *new* note.

**Emacs version:**
`GNU Emacs 27.2 (build 1, x86_64-apple-darwin21.1.0, Carbon Version 165 AppKit 2113) of 2021-11-18`

**Additional context**
Tested on macOS, not sure if it's reproducable on other platforms.

bdarcus:
So let me clarify:

Is the real issue that `citar-open-notes` is using `citar-file-open-external`, rather than only `find-file`?

I think this may be an unintended bug with some recent changes.

bdarcus:
@roshanshariff - if you can fix this easily in #460, can you please do that?

Otherwise, I'll do it after I merge that.

We just need open notes always to use `find-file`.

bdarcus:
Should be fixed via #460.

PR #543, opened 2022-01-29
Fix citar-open-multi to open notes with find-file

localauthor:
Incorporates into the new 'citar-open-multi' a fix from #460, #465

Issue:
- when using citar-open, notes are not distinguished from files, and are therefore opened with 'citar-file-open-function'
- when 'citar-file-open-function' is set to 'citar-file-open-external', notes are opened by the system, which takes some time to find emacs
- presumably notes should be opened in emacs directly, using find-file; previously discussed in #465

Fix:
'citar-multi-open' checks directory of selection against 'citar-notes-paths'

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -946,6 +946,10 @@ into the corresponding reference key.  Return
 For use with 'embark-act-all'."
   (cond ((string-match "http" selection 0)
          (browse-url selection))
+        ((member t (mapcar (lambda (x)
+                             (file-in-directory-p selection x))
+                           citar-notes-paths))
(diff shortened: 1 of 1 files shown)


bdarcus:
Thanks!

Will this be easily extended, do you think, if and when we address #518?

localauthor:
I would think so, since 'citar-multi-open' is just a cond, which can bear quite a lot of specificity. But I'm not familiar enough (at all) with org-roam to have a sense of what exactly that would entail.

PR #546, opened 2022-01-31
Add option for citar-open to forego selection prompt

### LINKED ISSUE #544: Automatically select sole candidate during multiple completion
citar provides `citar-open` to open links and files from a citation. This relies on multiple completion, but all of my bibliographic entries are associated with a single resource: a link or a file. Each time I invoke these commands, I have to `C-n` down from the input prompt to the sole resource to select it before confirming the selection. In contrast, `citar-open-library-files` can open a single file resource without multiple selection prompt, but is unable to handle URL resources. 

An ideal configuration for me would open both links and files without a prompt, or else would automatically select the sole candidate in the multiple completion prompt.

For context, I use vertico, consult, and embark. I could not find information on default candidate selection vis a vis consult's repository.

localauthor:
Closes #544

When citar-open-prompt is t, the default, citar-open will prompt for resource
selection even when there is only one resource. When nil, single resources
will open immediately, without any prompt.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -425,17 +425,6 @@ documentation for the return value and the meaning of
 REBUILD-CACHE and FILTER."
   (citar-select-ref :rebuild-cache rebuild-cache :multiple t :filter filter))
 
-(defun citar-select-file (files)
-  "Select file from a list of FILES."
-  (completing-read
-   "Related file: "
-   (lambda (string predicate action)
-     (if (eq action 'metadata)
-         `(metadata
-        ; (group-function . citar-select-group-related-sources)
-           (category . file))
-       (complete-with-action action files string predicate)))))
-
 (defun citar-select-resource (files &optional links)
   "Select resource from a list of FILES, and optionally LINKS."
   (let* ((files (mapcar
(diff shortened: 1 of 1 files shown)


bdarcus:
Is there a reason this is specific to `citar-open`?

localauthor:
The two other open functions, 'citar-open-notes' and 'citar-open-link', aren't set up to handle multiples.

I'm not sure why, in the case of 'open-notes'. It seems that the function 'citar-file--get-note-filename' is capable of locating multiple notes, but it only returns the car of the list.

If we change it to return a whole list, then (unless I'm wrong) it effectively does the same thing as `citar-file--files-for-entry`.

Is there any reason not to use `citar-file--files-for-entry` in `citar-open-notes`? Then we could add a similar conditional, to open without a prompt if there's only one file.

What do you think?

As for links: is it possible to have more the one link associated with an entry?

bdarcus:
I'll look at this more closely later, but in the meantime will just note that we're sort of in this in-between space between previous heavy reliance on CRM, and now migrating more to CR, with the idea to exploit `embark-all` instead.

PR #548, opened 2022-02-03
Add preview template to README

JonathanReeve:
The previously stated default value in the README would make citar choke, since it didn't include `preview`. So I'm just updating this to reflect the actual default value.


bdarcus:
Thanks!

PR #550, opened 2022-02-08
Fix minimal configuration example

andreas-bauer:
The minimal configuration code example in the `README.org` missing a closing bracket `)`.
This causes an `End of file during parsing` error if the code example is copy-pasted into the Emacs config.


bdarcus:
Thank you!

PR #551, opened 2022-02-08
Change default value of `citar-additional-fields`

apc:
Include "crossref" in the default value of `citar-additional-fields`, to make use of `parsebib`'s ability to make use of crossref information.


bdarcus:
Thanks.

PR #561, opened 2022-03-24
Change bindings on org buffer map

### LINKED ISSUE #503: Do not rebind C-d and C-k
In Emacs, `DEL` deletes character backward and `C-DEL` deletes them forward. Why do we rebind `C-DEL` default to `citar-org-delete-citation`? Does Org bring this (bad) idea or does Citar bring it?

bdarcus:
Close #503


salutis:
@bdarcus Thank you so much!

PR #562, opened 2022-03-24
Add citar-add-file-to-library

### LINKED ISSUE #528: Add citar-add-library-file
I added a modified, more general, version of this to the wiki. I may add it back to citar; I just wasn't sure if anyone used this ;-)

https://github.com/bdarcus/citar/wiki/Example-functions#add-file-to-library

There's still a little bug in it that I'll fix.

It uses `completing-read` rather than `completing-read-multiple`, and works on any file.

Note: recent changes in many of these commands are designed to make them compatible with `embark-act-all`. I'm not sure the UI on the current version of this command actually is, seems it seems to prompt you to rename the file in one case. Worth testing this.

It also occurs to me that we might want to include an option to add from a DOI, which maybe does that without prompting if the entry has a DOI?

_Originally posted by @bdarcus in https://github.com/bdarcus/citar/discussions/522#discussioncomment-1924933_

bdarcus:
Replace the original pdf-oriented function with a more general one.

Close #528

--------

This currently assumes one file per key.

If a file already exists, Emacs will prompt to confirm whether you want to overwrite it.

I'm thinking it also might confirm the file name/path before saving?

But can worry about those later, I think.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -1167,6 +1167,46 @@ With prefix, rebuild the cache before offering candidates."
        (error "Make sure 'citar-library-paths' is a list of paths"))
      (citar--library-file-action key-entry 'attach)))
 
+(defun citar--add-file-to-library (key)
+  "Add a file to the library for KEY.
+The FILE can be added from an open buffer, a file path, or a
+URL."
+  (let* ((source
+          (char-to-string
+           (read-char-choice
+            "Add file from [b]uffer, [f]ile, or [u]rl? " '(?b ?f ?u))))
+         (directory (if (cdr citar-library-paths)
+                        (completing-read "Directory: " citar-library-paths)
+                      (car citar-library-paths)))
+         (file-path
+          ;; Create the path without extension here.
+          (expand-file-name key directory)))
+    (pcase source
+      ("b"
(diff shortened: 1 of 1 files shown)


bdarcus:
@apc - you figured out an earlier version of this; right?

I've been waiting to merge this because I wanted to figure out a couple of refinements before doing so:

1. supporting any file extension
2. confirming before overwriting existing files

If you have time, could you please test this?

citar.el
```diff
@@ -1167,6 +1167,46 @@ With prefix, rebuild the cache before offering candidates."
...
+      ("f"
+       (let* ((file (read-file-name "Add file: " nil nil t))
+              (extension (file-name-extension file)))
+        (copy-file file
+         (concat file-path "." extension) 1)))
+      ("u"
+       (let* ((url (read-string "Add file URL: "))
+              (extension (url-file-extension url)))
+         (when (< 1 extension)
+           ;; TODO what if there is no extension?
```
bdarcus, line 1198: This is one question that probably needs to be resolved now. Right now, it should just not do anything if given a URL like "https://test.org/one".

PR #568, opened 2022-04-09
File notify tweaks

### LINKED ISSUE #567: Error running hook "citar-filenotify-local-watches" after calling `org-edit-special` on a LaTeX fragment
**Describe the bug**
Editing a LaTeX fragment in an org mode document with `C-c '` when citar-filenotify-setup is enabled for LaTeX buffers results in the following bug:

    Error (LaTeX-mode-hook): Error running hook "citar-filenotify-local-watches" because: (wrong-type-argument stringp nil)

**Expected behavior**
`citar-filenotify-local-watches` should check if the buffer is backed by a file before trying to monitor for changes

**Emacs version:**
29.0.50 ([deathwish](https://copr.fedorainfracloud.org/coprs/deathwish/)/[emacs-pgtk-nativecomp](https://copr.fedorainfracloud.org/coprs/deathwish/emacs-pgtk-nativecomp/))

bdarcus:
Fix #567.

@aikrahguzar - any feedback on this, including the docstring?

cc @gkowzan

### DIFF
--- a/citar-filenotify.el
+++ b/citar-filenotify.el
@@ -120,6 +120,11 @@ CHANGE refers to the filenotify argument."
                   (citar-filenotify--callback 'local x)))))
            (citar--local-files-to-cache)))))
 
+(defun  citar-filenotify-rm-local-watches ()
+  "Delete the filenotify watches for the local bib files."
+  (mapc #'file-notify-rm-watch citar-filenotify--local-watches)
+  (setq citar-filenotify--local-watches 'uninitialized))
+
 (defun citar-filenotify-local-watches ()
   "Hook to add and remove watches on local bib files.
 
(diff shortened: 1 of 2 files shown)


aikrahguzar (commented):
Looks good to me. But I think we should use a different name for the function and I suggested an alternate docstring but somehow github is not giving me an option to make it a suggestion.

bdarcus:
> Looks good to me. But I think we should use a different name for the function and I suggested an alternate docstring but somehow github is not giving me an option to make it a suggestion.

OK; I fixed those two issues.

citar-filenotify.el
```diff
@@ -129,11 +134,7 @@ watches aren't added.  This means a mode hook containing this
 function can run several times without adding duplicate watches."
   (when (eq 'uninitialized citar-filenotify--local-watches)
     (citar-filenotify--add-local-watches))
-  (add-hook 'kill-buffer-hook
-            (lambda ()
-              (mapc #'file-notify-rm-watch citar-filenotify--local-watches)
-              (setq citar-filenotify--local-watches 'uninitialized))
-            nil t))
+  (add-hook 'kill-buffer-hook #' citar-filenotify-rm-local-watches nil t))
```
aikrahguzar, line 137:
```suggestion
  (add-hook 'kill-buffer-hook #'citar-filenotify-rm-local-watches nil t))
```
There is space between the quote and function name.

bdarcus: Oops; fixed!

PR #570, opened 2022-04-13
Remove s dependency

### LINKED ISSUE #335: remove s, s-format dependency

bdarcus:
This is an alternative to #540, which keeps the current template format.

I still need to test more thoroughly.

Also, need to replace `s-match-strings-all` with a built-in in order to remove the s dependency. But initial checks there is no such thing, so would need to add another function here, and/or change how `citar--fields-for-format` works.

https://github.com/magnars/s.el/blob/08661efb075d1c6b4fa812184c1e5e90c08795a9/s.el#L452

Thanks to @aikrahguzar, who I credited on the commit.

Close #335.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -8,7 +8,7 @@
 ;; License: GPL-3.0-or-later
 ;; Version: 0.9
 ;; Homepage: https://github.com/bdarcus/citar
-;; Package-Requires: ((emacs "27.1") (s "1.12") (parsebib "3.0") (org "9.5") (citeproc "0.9"))
+;; Package-Requires: ((emacs "27.1") (parsebib "3.0") (org "9.5") (citeproc "0.9"))
 
 ;; This file is not part of GNU Emacs.
 
@@ -43,7 +43,6 @@
 (require 'browse-url)
 (require 'citar-file)
 (require 'parsebib)
-(require 's)
 (require 'crm)
 
 ;;; Declare variables for byte compiler
@@ -327,7 +326,7 @@ of all citations in the current buffer."
 
 (defcustom citar-select-multiple nil
   "Use `completing-read-multiple' for selecting citation keys.
-  When nil, all citar commands will use `completing-read`."
+When nil, all citar commands will use `completing-read`."
   :type 'boolean
   :group 'citar)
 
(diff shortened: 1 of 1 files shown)


citar.el
```diff
@@ -820,12 +819,27 @@ Return a list containing only (KEY . ENTRY) pairs."
                                           'invisible t))
       display-value)))
 
+(defun citar-format (template replacer)
+  "Format TEMPLATE with the function REPLACER.
+The templates are of form ${foo} for variable foo.
+REPLACER takes an argument of the format variable.
+Adapted from `citar-format'."
```
bdarcus, line 826: Is this adapted from org-roam, or from s-format, @aikrahguzar?

aikrahguzar: > Is this adapted from org-roam, or from s-format, @aikrahguzar?

`org-roam`. I didn't change the doc string apart from deleting the part about defaults.

Also this probably should be an internal function so `citar--format` maybe?

bdarcus: > Also this probably should be an internal function so `citar--format` maybe?

Yes, that makes sense.

bdarcus:
While I'm at this, @roshanshariff, can you tell me what the docsting for this "default" parameter should be?

https://github.com/bdarcus/citar/blob/c94799e9f33e0645b5ac8696e19bdc9328d7cb7e/citar.el#L459-L460

aikrahguzar:
This works for `citar--fields-for-format`.
```elisp
(defun citar--fields-for-format (template)
  "Return list of fields for TEMPLATE."
  (let* ((regexp "\\(?:\\`\\|}\\|:\\)[^{]*\\(?:\\${\\|\\'\\)\\|[[:space:]]+"))
         ;; The readable version of regexp is (rx (or (seq (or bos "}" ":") (0+ (not "{")) (or "${" eos)) (1+ space)))
    (split-string template regexp t)))
```
I have finally learned to live with regexes by just going for `rx` and `M-x regexp-builder`.

bdarcus:
> This works for `citar--fields-for-format`.

Thanks!

So this should now be ready to merge. I'll plan to merge it tomorrow.

bdarcus:
Actually, I spoke too soon; I have a git problem I need to fix.

citar.el
```diff
@@ -807,7 +801,7 @@ Return a list containing only (KEY . ENTRY) pairs."
   (let ((content-width (apply #'+
                               (seq-map #'string-to-number
                                        (split-string format-string ":"))))
-        (whitespace-width (string-width (s-format format-string
+        (whitespace-width (string-width (citar-format format-string
```
aikrahguzar, line 805:
```suggestion
        (whitespace-width (string-width (citar--format format-string
```

aikrahguzar:
> So this should now be ready to merge. I'll plan to merge it tomorrow.

The regexp in `citar--fields-for-format` can use some testing though the `rx` version seems correct to me and the string is just evaluating the `rx` version/ I did check with my own and the default templates and it seemed correct to fingers crossed.

bdarcus:
Do you think I should use the rx version in the code, for easier maintenance?

aikrahguzar:
> Do you think I should use the rx version in the code, for easier maintenance?

I have been using rx if I need to write a regular expression since I can't read the string ones at all. The difference is a macro expansion at compile time so I don't think it should affect anything. But I haven't see it used much in code I have come across and my elisp fu is limited so I am hesitant to opine but if it had a bug I would fix the rx version and the evaluate it to the string.

On the other hand there is `rxt-elisp-to-rx` which seems to convert strings to rx form pretty nicely so in the end it is probably not a big deal either way.

PR #577, opened 2022-04-20
Use browse-url instead of browse-url-default-browser

kngwyu:
Hi, thank you for your work on this great package.

Does this use of `browse-url-default-browser` come with some intention?
If not, I think `browse-url` is better because:
- it's consistent with other codes. `citar-open` and `citar-open-multi` use it.
- `browse-url-default-browser` doesn't care about `browse-url-browser-function` that users can set. This is actually a problem for me.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -1082,7 +1082,7 @@ With prefix, rebuild the cache before offering candidates."
                       :rebuild-cache current-prefix-arg)))
   (let ((link (citar-get-link (cdr key-entry))))
     (if link
-        (browse-url-default-browser link)
+        (browse-url link)
       (message "No link found for %s" (car key-entry)))))
 
 ;;;###autoload


kngwyu:
Or just calling `browse-url-browser-function` is better if we are sure that the argument is URL? 🤔

bdarcus:
> Does this use of `browse-url-default-browser` come with some intention?

It's a good question.

I can't see any obvious reason for it, and I can't tell who to blame for that code, and so who to ask.

> Or just calling browse-url-browser-function is better if we are sure that the argument is URL?

Why wouldn't we be?

Might be worth looking at how bibtex-completion handles this?

kngwyu:
> Might be worth looking at how bibtex-completion handles this?

They use their custom variable or `browse-url-browser-function`
https://github.com/tmalsburg/helm-bibtex/blob/ce8c17690ddad73d01531084b282f221f8eb6669/bibtex-completion.el#L940

Should I revise this PR to use `browse-url-browser-function`?

localauthor:
I recall looking into this a while back and thinking it should be `browse-url`, for consistency and user-configurability.

bdarcus:
> Should I revise this PR to use `browse-url-browser-function`?

I don't think so. It's just a variable; right?

bdarcus:
Thanks!

kngwyu:
Thank you guys for pretty speedy responses 🙂

PR #589, opened 2022-05-08
Include local-bib-files for citeproc

### LINKED ISSUE #588: citar-citeproc: doesn't work with local bib sources
If I run `citar-copy-reference` with items in a local bib file, citeproc can't find them.

I discovered this when testing #587.

Looking at the code, here's the problem:

https://github.com/bdarcus/citar/blob/a80eaf22a1e789edcd953df09624f34cee337001/citar-citeproc.el#L113

It should be fixed on #589.

cc @localauthor

bdarcus:
The citeproc formatting code only passes the global bibliography file(s).

This adds the local, if present.

Fix #588


bdarcus:
Closing in favor of #590.

PR #595, opened 2022-05-17
Add file/note data to candidates

### LINKED ISSUE #594: How to get a list of keys for references that have notes?
I thought this would be straightforward, and it's possible I'm missing something, but I can't figure it out.

From a discussion on a related project.

Thoughts on this @roshanshariff @aikrahguzar?

EDIT: I linked a simple PR to make this easier.

------

So you need a list keys for all entries that have associated notes?

E.g.:

```elisp
("doe1" "doe2" "doe3")
```

That's all?

If so, could you not just take the code you reference above to filter the candidates, and from there extract the keys?

EDIT: maybe not. I'll dig a little.

_Originally posted by @bdarcus in https://github.com/Vidianos-Giannitsis/zetteldesk.el/issues/5#issuecomment-1129300695_

bdarcus:
Currently, this PR has two commits, both to facilitate programmatic filtering etc of candidates.

## Add cons cells to candidate entries

First, when creating candidates, it adds cons cells for files or notes, I can drop this commit if we prefer the more general solution below, but I don't believe the operations here (pushing cons to a list) are expensive.

```elisp
(("has-note" . t) ; this PR adds this and next line
 ("has-file" . t)
 ("doi" . "10.1111/j.1540-6040.2009.01269.x")
 ("journaltitle" . "City \\& Community")
 ("date" . "2009-03-01")
 ("author" . "Zukin, Sharon and Trujillo, Valerie and Frase, Peter and Jackson, Danielle and Recuber, Tim and Walker, Abraham")
 ("title" . "New Retail Capital and Neighborhood Change: Boutiques and Gentrification in {{New York City}}")
 ("=type=" . "article")
 ("=key=" . "zukin2009"))
``` 

To filter:

```elisp
(seq-filter 
  (lambda (entry) 
    (assoc "has-note" entry)) 
  (citar--get-candidates))
```
## Add filter arg to `citar--get-candidates`

Second, more generally, it adds an optional filter to `citar--get-candidates`:

```elisp
(citar--get-candidates t (citar-has-note))
```

FWIW, this is actually slower than the first option.

```elisp
ELISP> (benchmark-run-compiled 10 (citar--get-candidates nil (citar-has-note)))
(0.036507296 0 0.0)

ELISP> (benchmark-run-compiled 10 (seq-filter 
    (lambda (e) 
      (assoc "has-note" e)) 
  (citar--get-candidates)))
(0.004006728 0 0.0)
```

See https://github.com/Vidianos-Giannitsis/zetteldesk.el/issues/5

Close #594

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -697,12 +697,14 @@ are refreshed."
       (error "No template for \"%s\" - check variable 'citar-templates'" template-name))
     template))
 
-(defun citar--get-candidates (&optional force-rebuild-cache)
+(defun citar--get-candidates (&optional force-rebuild-cache filter)
   "Get the cached candidates.
 
 If the cache is unintialized, this will load the cache.
 
-If FORCE-REBUILD-CACHE is t, force reload the cache."
+If FORCE-REBUILD-CACHE is t, force reload the cache.
+
+If FILTER, use the function to filter the candidate list."
   (unless (or citar-bibliography (citar--local-files-to-cache))
     (error "Make sure to set citar-bibliography and related paths"))
   (when force-rebuild-cache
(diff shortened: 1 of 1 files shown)


bdarcus:
cc @localauthor this may or may not be useful for zk.

PR #598, opened 2022-05-19
API cleanup

### LINKED ISSUE #438: Review public/internal function names, docstrings

bdarcus:
## Preface

If you object to these changes after they are merged (for example, think certain functions should be public), please file an issue with your request(s), and link to this.

## Questions/issues

The last two are the most important:

1. salutis' suggestions https://github.com/bdarcus/citar/issues/438#issuecomment-1001245176
2. related, I guess private variables should also have `--`? If yes, fix that too.
3. ~Should we rename the predicate functions to end with `-p`, per elisp convention, and actually keep them public (`citar-has-note-p`)?~ (no; they aren't predicate functions exactly)
4. ~Is `citar-has-a-value` function a predicate function? It sounds like it, but docstring says no. Maybe it should be something like `citar-field-with-value`?~ 
5. ~Am I correct we're inconsistently using `ITEM` and `ENTRY` to refer to the same thing in arg lists and docstrings?~
6. Make citar--get-candidates public? 
7. Do we really want the other get-* functions to be private? If the answer to 6 is yes, then we probably want all these public too?

---------------------

Rename a number of functions for internal consistency, and to on balance
make more of the API private.

Close #438

### DIFF
--- a/citar-citeproc.el
+++ b/citar-citeproc.el
@@ -109,13 +109,13 @@ With prefix-argument, select CSL style."
                     citar-citeproc-csl-style
                   (expand-file-name citar-citeproc-csl-style citar-citeproc-csl-styles-dir)))
          (keys (citar--extract-keys keys-entries))
-	 (bibs (flatten-list
-		(list citar-bibliography
-		      (citar--major-mode-function 'local-bib-files #'ignore))))
+         (bibs (flatten-list
+                (list citar-bibliography
+                      (citar--major-mode-function 'local-bib-files #'ignore))))
          (proc (citeproc-create style
-			        (citeproc-hash-itemgetter-from-any bibs)
-			        (citeproc-locale-getter-from-dir citar-citeproc-csl-locales-dir)
-			        "en-US"))
+                                (citeproc-hash-itemgetter-from-any bibs)
+                                (citeproc-locale-getter-from-dir citar-citeproc-csl-locales-dir)
+                                "en-US"))
          (references (car (progn
                             (citeproc-add-uncited keys proc)
                             (citeproc-render-bib proc 'plain)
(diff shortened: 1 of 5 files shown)


bdarcus:
@aikrahguzar - I think my questions 4 and 5 are you. Any thoughts?

aikrahguzar:
> @aikrahguzar - I think my questions 4 and 5 are you. Any thoughts?

4)  `citar-field-with-value` is a much better name. Lets go with that.
5) It does seem to be the case from a quick look at the occurrences.

bdarcus:
Thanks @aikrahguzar; I fixed both.

bdarcus:
> I think we discussed this in the original PR, but `citar-has-note` and `citar-has-file` aren't predicate functions, but rather _return_ predicate functions. Naming them with `-p` might be more confusing than not.

OK; thanks. I did miss that.

bdarcus:
I may have forgot to fix it farther down.

BTW, why do the functions need to be compiled, rather than just standard predicate functions?

roshanshariff:
> BTW, why do the functions need to be compiled, rather than just standard predicate functions?

See https://github.com/bdarcus/citar/pull/482#discussion_r767209134

bdarcus:
In retrospect, then, should we use your original function name there, with "make"? 

Or just leave as is?

roshanshariff:
I think leaving as-is makes sense.

bdarcus:
I don't want to dwell on this too much, so will merge this later today if there is no other input.

We can entertain opening up a few of these functions later.

PR #603, opened 2022-05-22
Generalize note API

### LINKED ISSUE #601: Make has-note configurable
... Seems one obvious needed change here would be something like a configurable `citar-has-notes-function` variable. One could then swap one that queried org-roam instead to create the candidate data.

_Originally posted by @bdarcus in https://github.com/bdarcus/citar/issues/518#issuecomment-1133216770_

-----------
@roshanshariff - in the context of discussing #518, it occurred to me (and I think you may have mentioned this earlier) we'd want to do something like the below:

```elisp
(defcustom citar-has-note-function (citar-has-note)
  "The function to return notes associated with a reference.

Function must take two arguments: KEY and ENTRY."
  :group 'citar
    :type '(choice
            (function-item :tag "Default, one-note-per-file" (citar-has-note))
            (function-item :tag "Org-Roam v2" citar-org-roam-has-note)
            (function :tag "Other")))

(defun citar-org-roam-has-note (key _entry)
  "Query org-roam for presence of bibliographic notes for KEY."
  (let ((ref-node (org-roam-node-from-ref (concat "@" key)))) 
    (list (org-roam-node-file ref-node))))
```

Does that seem right to you? Any suggestions?

BTW, with org-roam v2, you can have notes on a reference in different places; example, where first "toly2017" is a within-file note, and the second is a file:

image

... and here's what the first one looks like internally:

```elisp
#s(org-roam-node :file "/home/bruce/org/roam/20200307105756-covid_19.org" :file-title "COVID-19" :file-hash nil :file-atime
                  (25226 5062 342907 390000)
                  :file-mtime
                  (25226 5062 337907 382000)
                  :id "aff1a98d-8b4c-4771-95ea-163090b248f4" :level 1 :point 1262 :todo nil :priority nil :scheduled nil :deadline nil :title "China Outbreak and Response" :properties
                  (("CATEGORY" . "20200307105756-covid_19")
                   ("ROAM_REFS" . "@toly2017")
                   ("ID" . "aff1a98d-8b4c-4771-95ea-163090b248f4")
                   ("BLOCKED" . "")
                   ("FILE" . "/home/bruce/org/roam/20200307105756-covid_19.org")
                   ("PRIORITY" . "B")
                   ("ITEM" . "China Outbreak and Response"))
                  :olp nil :tags nil :aliases nil :refs
                  ("toly2017"))
```

bdarcus:
This PR attempts to answer two questions:

1. How would we #518?
2. How, in doing so, would that include ensuring that the citar UI correctly updates the status of "has notes" display, when multiple bib notes may be contained within a single file?

So it adds a `citar-keys-with-notes-functions` defcustom, with the idea that, for example, a `citar-org-roam` package could provide a function to query the org-roam db, and provide that list of keys.

```elisp
(defun citar-org-roam--keys-with-note ()
  "Return a list of keys with associated note(s)."
  (mapcar #'car (org-roam-db-query
                 [:select ref :from refs :where (= type "cite")])))
```

And of course, that it could provide an "open-function" to have org-roam handle that.

Finally, it also, per suggestion from @roshanshariff, turns the `open-notes` variable into a list, to provide more flexibility.

Aside: this PR originally included that org-roam integration (an answer to the first question above), but I removed it to simplify. I did confirm it basically worked though, and will add a separate PR once I merge this.

-----

## Issues

I created a few little convenience functions along with this, with this unexpected result (note the first two items in the list):

```elisp
ELISP> (citar--keys-with-library-files)
("" "." "abaza2013" ...
```

I also need to figure out if this is a problem.

```elisp
ELISP> (length (citar-org-roam--keys-with-notes))
17 (#o21, #x11, ?\C-q)
ELISP> (hash-table-count (citar-file--directory-files citar-notes-paths nil citar-file-note-extensions
                                            citar-file-additional-files-separator))
33 (#o41, #x21, ?!)
```

Close #601

### DIFF
--- a/citar-file.el
+++ b/citar-file.el
@@ -27,6 +27,7 @@
   (require 'cl-lib)
   (require 'subr-x))
 (require 'seq)
+(make-obsolete 'citar-format-note-function 'citar-create-note-function "1.0")
 
 ;;; pre-1.0 API cleanup
 
@@ -93,6 +94,9 @@ separator that does not otherwise occur in citation keys."
                  (regexp :tag "Filename separator")))
 
 (defvar citar-notes-paths)
+(defvar citar-create-note-function)
+(defvar citar-library-paths)
+(defvar citar-library-file-extensions)
 
 ;;;; Convenience functions for files and paths
 
@@ -229,6 +233,26 @@ need to scan the contents of DIRS in this case."
                  (puthash key (nreverse filelist) files))
                files))))
 
+(defun citar-file--has-file-notes-hash ()
+  "Return a hash of keys and file paths for notes."
+  (citar-file--directory-files
+   citar-notes-paths nil citar-file-note-extensions
+   citar-file-additional-files-separator))
+
+(defun citar-file--has-library-files-hash ()
+  "Return a hash of keys and file paths for library files."
+  (citar-file--directory-files
+   citar-library-paths nil citar-library-file-extensions
+   citar-file-additional-files-separator))
+
+(defun citar-file--keys-with-file-notes ()
+  "Return a list of keys with file notes."
+  (hash-table-keys (citar-file--has-file-notes-hash)))
+
+(defun citar-file--keys-with-library-files ()
+  "Return a list of keys with file notes."
+  (hash-table-keys (citar-file--has-library-files-hash)))
+
 (defun citar-file--has-file (dirs extensions &optional entry-field)
   "Return predicate testing whether a key and entry have associated files.
 
@@ -321,6 +345,19 @@ of files found in two ways:
                     (_ "xdg-open"))
                   nil 0 nil
                   file)))
+
+(defun citar-file--open-note (key entry)
+  "Open a note file from KEY and ENTRY."
+  (if-let* ((file (citar-file--get-note-filename key
+                                                 citar-notes-paths
+                                                 citar-file-note-extensions))
+            (file-exists (file-exists-p file)))
+      (find-file file)
+    (if (and (null citar-notes-paths)
+             (equal citar-create-note-function
+                    'citar-org-format-note-default))
+      (error "You must set 'citar-notes-paths'")
+      (funcall citar-create-note-function key entry file))))
 
 (defun citar-file--get-note-filename (key dirs extensions)
   "Return existing or new filename for KEY in DIRS with extension in EXTENSIONS.
--- a/citar.el
+++ b/citar.el
@@ -62,8 +62,10 @@
 
 ;; also rename
 (make-obsolete 'citar-has-a-value 'citar--field-with-value "1.0")
+(make-obsolete 'citar--open-note 'citar-file--open-note "1.0")
 
-
+(make-obsolete-variable
+ 'citar-format-note-function 'citar-create-note-function "1.0")
 
 ;;; Declare variables and functions for byte compiler
 
@@ -222,6 +224,20 @@ If nil, single resources will open without prompting."
   :group 'citar
   :type '(boolean))
 
+;;; Note-handling setup
+
+(defcustom citar-open-note-functions
+  '(citar-file--open-note)
+  "List of functions to open a note."
+  :group 'citar
+  :type '(function))
+
+(defcustom citar-keys-with-notes-functions
+  '(citar-file--keys-with-file-notes)
+  "Functions, without argument, to return a list of keys with notes."
+  :group 'citar
+  :type '(function))
+
 (defcustom citar-open-note-function
   'citar--open-note
   "Function to open a new or existing note.
@@ -233,9 +249,9 @@ ENTRY: an alist with the structured data (title, author, etc.)"
   :group 'citar
   :type 'function)
 
-(defcustom citar-format-note-function
+(defcustom citar-create-note-function
   'citar-org-format-note-default
-  "Function to format a new note.
+  "Function to create a new note.
 
 A note function must take three arguments:
 
@@ -475,6 +491,14 @@ REBUILD-CACHE and FILTER."
              (category . multi-category))
          (complete-with-action action (delete-dups resources) string predicate))))))
 
+(defun citar--keys-with-notes ()
+  "Return a list of keys with associated notes."
+  (seq-uniq
+   (seq-mapcat
+    (lambda (fn)
+      (funcall fn))
+    citar-keys-with-notes-functions)))
+
 (defun citar--select-group-related-resources (resource transform)
   "Group RESOURCE by type or TRANSFORM."
     (let ((extension (file-name-extension resource)))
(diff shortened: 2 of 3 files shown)


citar.el
```diff
@@ -222,6 +222,23 @@ If nil, single resources will open without prompting."
   :group 'citar
   :type '(boolean))
 
+(defcustom citar-open-note-functions
+  '(citar-file--open-note citar--open-note-field citar-org-roam-open-note)
+  "List of functions to open a note."
+  :group 'citar
+  :type '(function))
```
bdarcus, line 229: Note we already have a defcustom below for this, which takes a single function. It's used now, for example, by org-roam users.

citar.el
```diff
@@ -222,6 +222,23 @@ If nil, single resources will open without prompting."
...
+
+(defcustom citar-create-note-function nil
+  "Function to create a new note."
+  :group 'citar
+  :type 'function)
```
bdarcus, line 234: Arguably we don't need this; this behavior can be handled in the actual note open-note function?

bdarcus:
This is getting closer to done. Any input on this @localauthor?

Really my biggest concern here ATM is getting the note API right; e.g. the `citar.el` changes.

citar.el
```diff
@@ -605,6 +622,24 @@ repeatedly."
   (citar-file--has-file citar-notes-paths
                         citar-file-note-extensions))
 
+(defun citar--has-file-notes-hash ()
```
bdarcus, line 625: These are convenience functions. Not sure; should they go in `citar-file.el`?

bdarcus:
@publicimageltd - since you work with org-roam, any feedback on this?

I'm currently thinking to scale this PR back to only focus on the citar changes, and will probably have a separate repo for citar-org-roam.

bdarcus:
> I'm currently thinking to split this PR so this only focuses on the citar changes, and one for citar-org-roam.

Done, and ready for review. I hope to merge this sometime today or tomorrow.

localauthor:
I’ve meant to have a look at this but I’ve been a bit swamped. (End of term…) I’ll try tonight.

publicimageltd:
> @publicimageltd - since you work with org-roam, any feedback on this?

Hullo, no time no see! I always love commenting stuff  :grin: , but I actually don't quite get what it is all about. Could you give a little pointer?

publicimageltd:
Thanks, that does help! Your proposed example for an Org Roam function returns a list of ROAM-REFS property values. I just tried it in my DB and was surprised to actually find some refs, I don't use them (should add handling them to Delve, I guess...) 

So I can't say much, it sounds all fine. I had a quick look at the  Org Roam manual, however, and it says:

>   You may assign multiple refs to a single node, for example when you
> want multiple papers in a series to share the same note, or an article
> has a citation key and a URL at the same time.

Maybe this could cause a problem? There might be multiple refs of the same type per node.

bdarcus:
> Thanks, that does help! Your proposed example for an Org Roam function returns a list of ROAM-REFS property values. I just tried it in my DB and was surprised to actually find some refs, I don't use them (should add handling them to Delve, I guess...)

Would that not be the only way to distinguish bibliographic notes?
 
> So I can't say much, it sounds all fine. I had a quick look at the Org Roam manual, however, and it says:
> 
> > You may assign multiple refs to a single node, for example when you
> > want multiple papers in a series to share the same note, or an article
> > has a citation key and a URL at the same time.
> 
> Maybe this could cause a problem? There might be multiple refs of the same type per node.

I had thought about that, but hadn't yet tested it. I had _hoped_ org-roam could correctly handle this.

E.g. from a citar POV, all we need is:

1. to know: does this reference have associated notes?
2. a way to open them

So I had thought we could just let org-roam handle the second.

But I don't know how it's UI deals with that scenario.

This is partly why I split this PR; I think it will take more time to figure out those details, test, etc.

publicimageltd:
The first case is dealt with. `org-roam-ref-find` offers completion where each ref is associated with a node. I've created a node with two refs, they are listed separately as it should be, so finding the node per ref works well. Internally, an alist is created mapping each ref to a freshly created node object. The public API only offers completion right now, however, so a `citar-org-roam` package would have to use some private functions of Org Roam or roll its own to identify the node. Seems to be no problem.

I think these refs are a nice thing to have. They didn't work well for me in v1, however, so I got into the habit of simply linking references directly instead of using the bibtex key. Turns out that this is not the optimal way, it looks great how it works and that it can be used in citar this way.

bdarcus:
I assume for my case 2 `org-roam-ref-find` will just present multiple candidates per key.

citar-file.el
```diff
@@ -229,6 +233,25 @@ need to scan the contents of DIRS in this case."
                  (puthash key (nreverse filelist) files))
                files))))
 
+(defun citar-file--has-file-notes-hash ()
+  "Return a hash of keys and file paths for notes."
+  (citar-file--directory-files
+   citar-notes-paths nil citar-file-note-extensions))
```
localauthor, line 239: Any reason not to include `citar-file-additional-files-separator` as an argument here?

bdarcus: I think I had mistakenly assumed that was only for library files?

bdarcus: Fixed.

bdarcus:
OK, merged. If anything comes up, we can fix in subsequent commits.

bdarcus:
> I will add a separate PR once I merge this.

https://github.com/bdarcus/citar/pull/609

localauthor:
Not sure if this should be incorporated somehow, or where, but this adds the "N" prefix to candidates whose crossref has a note. 

```elisp
(defun citar-xref-notes ()
  (let ((all-keys (mapcar (lambda (x) (nth 1 x)) citar--candidates-cache))
        (note-keys (citar-file--keys-with-file-notes)))
    (remq nil      
          (mapcar (lambda (key)
                    (let* ((entry (citar--get-entry key))
                           (xref (citar--get-value "crossref" entry)))
                      (when (member xref note-keys)
                        key)))
                  all-keys))))
  
(setq citar-keys-with-notes-functions '(citar-xref-notes citar-file--keys-with-file-notes))
```

(As an aside, is there a better way to get a list of all keys?)

bdarcus:
> Not sure if this should be incorporated somehow, or where, but this adds the "N" prefix to candidates whose crossref has a note.

What are those notes? Do you have an example?

> (As an aside, is there a better way to get a list of all keys?)

Maybe access the cache via `(citar--get-candidates)` instead?

localauthor:
I have an edited volume "Anderson2020" that contains an essay "Bernhard2020," (that is, "Bernhard2020" lists "Anderson2020" as a crossref in my bib file).

The predicate for library files,`citar-file--has-file`, checks for these crossrefs, so library file named "Anderson2020.pdf" will be considered a library file for "Bernhard2020," and the "F" prefix will be added to it in the candidate list.

The previouis predicate for notes, `citar-has-note`, was built on `citar-file--has-file`, so it inherited the crossref checking. Now that the predicate has changed, it has lost crossref checking.

Maybe it's not desirable to add this back, in which case I'll just keep the above in my own config.

bdarcus:
OIC; I just missed that. I thought by "crossref" you meant the web service!

It seems like we should be consistent (e.g. it was an oversight on my part); no? Which would support adding it back in.

bdarcus:
I just created a new issue for this.

PR #605, opened 2022-05-23
Strip file-paths from resource candidates

localauthor:
Maybe something to consider after 1.0.

This transforms resource candidates (notes and library files) to only filenames, removing directory paths. Much cleaner display, imo. :)

I've been using it in my fork for a while and haven't had any issues. Embark still works as expected, as the target doesn't change --- the target is still the full file path.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -479,7 +479,9 @@ REBUILD-CACHE and FILTER."
   "Group RESOURCE by type or TRANSFORM."
     (let ((extension (file-name-extension resource)))
       (if transform
-          resource
+          (if (file-regular-p resource)
+              (file-name-nondirectory resource)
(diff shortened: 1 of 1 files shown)


bdarcus:
> Maybe something to consider after 1.0.

It's a small enough change I see no problem merging now.

I'm just trying to think of there's anyplace where it could present a problem. I think I'll just merged and see what feedback we get :-)

localauthor:
> > Maybe something to consider after 1.0.
> 
> It's a small enough change I see no problem merging now.
> 
> I'm just trying to think of there's anyplace where it could present a problem. I think I'll just merged and see what feedback we get :-)

One way to do it 😄

PR #609, opened 2022-05-26
Add citar-org-roam

### LINKED ISSUE #602: Better integrate org-roam and citar?
> Which makes me also wonder about something more tightly coupled to OR (v2); say a citar-org-roam.

_Originally posted by @bdarcus in https://github.com/bdarcus/citar/issues/518#issuecomment-1133112053_

## Preface

This is an issue to collect notes on what changes might happen in these two packages to allow easier and better integration.

Obviously, while this focuses on `org-roam`, it should be general enough to work with any note-taking package.

## For citar

1. #601
2. need to confirm `citar-open-note-function` is sufficient, including for within-file nodes; I suspect it is.

I think that's it.

## For org-roam

1. https://github.com/org-roam/org-roam/issues/2202
2. ??

Also, why `@` is included in the citation refs, even internally?

E.g.

```elisp
(org-roam-node-from-ref "@doe")
```

… returns a result, but this does not:

```elisp
(org-roam-node-from-ref "doe")
```

Does that mean they’re somehow tied to what citation system you’re using? So the value is different for org-ref than it is for org-cite? If yes, is that necessary?

EDIT: it may be to distinguish `org-cite` from `org-ref`. The fragmentation here is just not good.

### LINKED ISSUE #612: Generalize the "notes" part of `citar-open`
We need to adjust `citar-open` to #603 (or drop the command altogether?).

The problem currently is the candidate list here is file paths, and typed as such for Embark.

And `citar--select-resource` takes a list of files as input.

But for org-roam v2, for example, with its more general concept of note, the candidates would be "org-roam-ref" "nodes".

I need to look, but I suspect we need to add another defcustom for the candidates.

-------------

## Notes

- `org-roam-ref-read` is the completion function
- it has a customizable annotation function, to display the metadata
- it uses `org-roam-ref-read--completions`; its candidates are a key and a list of org-roam nodes, which look like this:

```elisp
(#("agnew2003" 0 9
   (node #s(org-roam-node :file "/home/bruce/org/roam/biblio/agnew2003.org" :file-title nil :file-hash nil :file-atime nil :file-mtime nil :id "fd1e8dfa-8bfd-43fb-884a-b350a7a0850a" :level nil :point 1 :todo nil :priority nil :scheduled nil :deadline nil :title "Notes on Agnew, From Megalopolis to Global City-Region?" :properties nil :olp nil :tags nil :aliases nil :refs nil)
          type "cite"))
  . #s(org-roam-node :file "/home/bruce/org/roam/biblio/agnew2003.org" :file-title nil :file-hash nil :file-atime nil :file-mtime nil :id "fd1e8dfa-8bfd-43fb-884a-b350a7a0850a" :level nil :point 1 :todo nil :priority nil :scheduled nil :deadline nil :title "Notes on Agnew, From Megalopolis to Global City-Region?" :properties nil :olp nil :tags nil :aliases nil :refs nil))
```

bdarcus:
Better integrate citar and org-roam.

The goal is a package you can install via MELPA, that will integrate the full flexibility of org-roam v2, and support the following additional use cases:

1. multiple refs/citekeys per bibliographic note
2. multiple bibliographic notes per ref/citekey (though per notes below, this may not be possible/wise)
3. multiple bibliographic notes per file (#518)

To be clear, however, the core package will still work with org-roam without this package. It's just in that scenario, you are limited to one bib-note-per-file.

Close #602 #518

--------

## Notes

- This currently assumes a separate MELPA package.
- The one function I know doesn't work ATM is `citar-org-roam--create-note`. I am instead just using the default create-note function ATM. I don't really understand capture programming, so could use help here.
- Also, `citar-open` assumes files for notes, so need to fix #612 
- There are some questions in the code.
- Needs documentation.
- worth nothing, too, that the org-roam v2 db includes a `citations` table, that tracks what note nodes cite which references; so an opportunity there, and there's start of a command to use this

## org-roam doesn't now support case 2

Technically, org-roam does not now support the second case I note above. See https://github.com/org-roam/org-roam/issues/2207.

Jethro is open to changing org-roam on this, but says he can't do it (doesn't have time), and that if I want to submit PRs:

> ... this is going to be a large-scale change across the code-base.

... which of course sounds scary.

### DIFF
--- a/citar-org-roam.el
+++ b/citar-org-roam.el
@@ -1,0 +1,128 @@
+;;; citar-org-roam.el --- Citar/org-roam integration -*- lexical-binding: t; -*-
+;;
+;; Copyright (C) 2022 Bruce D'Arcus
+;;
+;; Author: Bruce D'Arcus <bdarcus@gmail.com>
+;; Maintainer: Bruce D'Arcus <bdarcus@gmail.com>
+;; Created: May 22, 2022
+;; Modified: May 22, 2022
+;; Version: 0.0.1
+;; Homepage: https://github.com/bdarcus/citar-org-roam
+;; Package-Requires: ((emacs "27.1") (org-roam "2.2") (citar "0.9.5"))
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; Commentary:
+;;
+;;  A tiny minor-mode to integrate 'citar' and 'org-roam'.
+;;
+;;  Provides custom functions for:
+;;  
+;;  1. updating the 'citar' UI from the 'org-roam' database
+;;  2. uses org-roam to open the notes
+;;
+;;; Code:
+
+(require 'org-roam)
+(require 'citar)
+
+(defvar 'citar-open-notes-functions)
+(defvar 'citar-create-note-function)
+(defvar 'citar-keys-with-notes-functions)
+
+;;; Functions
+
+(defun citar-org-roam--has-note-p (key &optional _entry)
+  "Return non-nil if a KEY has an associated org-roam ref note."
+  (let ((ref-node (org-roam-node-from-ref (concat "@" key))))
+    (when ref-node t)))
+
+(defun citar-org-roam--keys-with-notes ()
+  "Return a list of keys with associated note(s)."
+  (mapcar #'car (org-roam-db-query
+                 [:select ref :from refs :where (= type "cite")])))
+
+(defun citar-org-roam-cited (key-entry)
+  "Return a list of notes that cite the reference (KEY-ENTRY)."
+  (interactive (list (citar-select-ref
+                      :rebuild-cache current-prefix-arg)))
+  (let* ((ids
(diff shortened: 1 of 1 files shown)


bdarcus:
@publicimageltd as I dug into org-roam v2, I discovered a limitation in how it currently implements refs, described above, and in the linked issue at org-roam. Do you agree with my impulse that this limitation should be removed; that it should be possible to have multiple notes per ref/citekey?

Jethro says it would be a large change, but is open to it. I'm not sure how large, or what it exactly it would require. But I'd like to confirm my thinking is sound before going down the rabbit hole.

bdarcus:
Cases 1 and 3 should now be working correctly with `citar-open-notes`.

bdarcus:
I'm going to close this, in favor of a separate repo.

https://github.com/emacs-citar/citar-org-roam

Also, I decided to move both repos to a github org.

PR #611, opened 2022-05-26
Add back crossref support for notes to candidates

### LINKED ISSUE #610: Add back crossref support for notes to candidates
This was an oversight in #603 that needs to be fixed.

------

Not sure if this should be incorporated somehow, or where, but this adds the "N" prefix to candidates whose crossref has a note. 

```elisp
(defun citar-xref-notes ()
  (let ((all-keys (mapcar (lambda (x) (nth 1 x)) citar--candidates-cache))
        (note-keys (citar-file--keys-with-file-notes)))
    (remq nil      
          (mapcar (lambda (key)
                    (let* ((entry (citar--get-entry key))
                           (xref (citar--get-value "crossref" entry)))
                      (when (member xref note-keys)
                        key)))
                  all-keys))))
  
(setq citar-keys-with-notes-functions '(citar-xref-notes citar-file--keys-with-file-notes))
```

(As an aside, is there a better way to get a list of all keys?)

_Originally posted by @localauthor in https://github.com/bdarcus/citar/issues/603#issuecomment-1138528264_

localauthor:
Fix #610

### DIFF
--- a/citar-file.el
+++ b/citar-file.el
@@ -247,7 +247,15 @@ need to scan the contents of DIRS in this case."
 
 (defun citar-file--keys-with-file-notes ()
   "Return a list of keys with file notes."
-  (hash-table-keys (citar-file--has-file-notes-hash)))
+  (let ((all-keys (mapcar (lambda (x) (nth 1 x)) (citar--get-candidates)))
+        (note-keys (hash-table-keys (citar-file--has-file-notes-hash))))
+    (mapc (lambda (key)
+            (let* ((entry (citar--get-entry key))
+                   (xref (citar--get-value "crossref" entry)))
+              (when (member xref note-keys)
+                (push key note-keys))))
+          all-keys)
+    note-keys))
 
 (defun citar-file--keys-with-library-files ()
   "Return a list of keys with file notes."


citar-file.el
```diff
@@ -247,7 +247,15 @@ need to scan the contents of DIRS in this case."
 
 (defun citar-file--keys-with-file-notes ()
   "Return a list of keys with file notes."
-  (hash-table-keys (citar-file--has-file-notes-hash)))
+  (let ((all-keys (mapcar (lambda (x) (nth 1 x)) (citar--get-candidates)))
+        (note-keys (hash-table-keys (citar-file--has-file-notes-hash))))
```
bdarcus, line 251:
```suggestion
        (note-keys (citar--keys-with-file-notes)))
```
I believe this is the same, but simpler? Make sure to check my parens though.

bdarcus:
I guess a bigger question that just occurred to me, though it may or may not be an issue:

Perhaps this parsing shouldn't be on by default?

If that's reasonable question, perhaps it should be extracted into a separate function, per your original code?

localauthor:
> It seems there's some potential ambiguity about whether a note for an edited collection, for example, is also a note for a component of that collection.

Right, it’s not clear that files and notes ought  to be governed by the same logic here. Now that the note function variable is a list there’s no reason this can’t just be on the wiki.

bdarcus:
I'm agnostic about wiki vs include but have configurable; up to you.

I do often put code ideas I'm unsure of on the wiki first to think about a bit more, but to let people experiment if they want.

localauthor:
Now that the procedure for checking notes has been reversed, starting with files and identifying associated keys, there's no efficient way that I can think of to re-implement the xref notes feature.

The above function doesn't actually work --- it goes into an infinite loop for reasons that I can't get to the bottom of.

As you said, there isn't a clear logical association between a text and a note for its crossref anyway. So I'm thinking this might not be worth the effort it would take to implement.

Going to close it, if that's alright.

bdarcus:
I'll look into because, whether or not it's a good idea, it should be possible, or there might be a problem somewhere.

Do you have an idea about what causes that "inefficiency"?

Might it, for example, be extracting `all-keys` again?

bdarcus:
Another thought occurred to me:

I don't really use xrefs. Is there room to enhance their support in the UI somehow, in general?

Issue #613, opened 2022-05-27
Using `org-citar-edit-note` as note open function fails when Org Roam note template does not have `:immediate-finish t`


balajisivaraman:
**Describe the bug**

Setting `citar-open-note-functions` to `'(orb-citar-edit-note)` causes the `citar-open-notes` action to fail with the following error when the Org Roam capture template used does not have `:immediate-finish t`:

```
(wrong-number-of-arguments #<subr citar-org-format-note-default> 2)
```

**To Reproduce**
1. Set the value of `org-roam-capture-templates` as below (with other Org Roam configuration as needed):

```elisp
(setq 
org-roam-directory "~/notes"
org-roam-capture-templates
`(("n" "ref + noter" plain
      "%?"
      :target
      (file+head
       "bibliography/${citekey}.org"
       ":PROPERTIES:
:ID: %(shell-command-to-string \"uuidgen\"):ROAM_REFS: ${ref}
:END:
#+TITLE: ${citekey}: ${title}\n#+ROAM_KEY: ${ref}\n

- tags ::
- keywords :: ${keywords}

* ${title}
  :PROPERTIES:
  :Custom_ID: ${citekey}
  :URL: ${url}
  :AUTHOR: ${author}
  :NOTER_DOCUMENT: %(orb-process-file-field \"${citekey}\")
  :NOTER_PAGE:
  :END:\n\n")
  :unnarrowed t))
```

3. Set value of `citar-open-note-functions` to `'(orb-citar-edit-note)`.
4. Do `M-x` `citar-insert-reference`, choose an entry from the list, and then do `embark-act` on it.
5. From the Embark actions list, choose `n` for `Open note`.

**Expected behavior**
I should be taken to the newly created Org Roam note file for the entry where I can start typing my notes, and then do `C-c C-c` as per the normal Org Capture process to finalize the capture.

**Emacs version:** `GNU Emacs 28.1.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.33, cairo version 1.17.6) of 2022-05-22`

**Additional context**

I have performed initial debugging and figured out that this is due to the recent changes made as part of da5f0c5449675fd708df15aa635f0dbd7bb02a29.

Specifically, it is happening in the `citar--open-notes` function recaptured here:

```elisp
(defun citar--open-notes (key entry)
  "Open note(s) associated with KEY and ENTRY."
  (or (seq-some
       (lambda (opener)
[...]

PR #614, opened 2022-05-27
fix for issue 613

balajisivaraman:
As discussed in #613, this commit fixes the fallback in `citar--open-notes` and updates the README to add the note for needing `:immediate-finish t` in ORB capture notes.

Cheers.

### DIFF
--- a/README.org
+++ b/README.org
@@ -278,6 +278,9 @@ You can also use the ~citar-open-note-functions~ variable to replace or augment 
 
 Since ~citar-open-note-functions~ is a list, you can also include multiple functions, to handle different note scenarios.
 
+Please note that if you choose to use org-roam-bibtex using the above configuration, you will need to set ~:immediate-finish t~ in the template that you use for bibliography notes in ~org-roam-capture-templates~.
+Since ~citar-open-note-functions~ attempts multiple functions one after the other, this is needed to ensure the ~org-capture~ returns immediately without waiting for further user input.
+
 Citar also includes a ~citar-keys-with-notes-functions~ variable, which specifies a list of functions, each of which returns a list of keys that have associated notes.
 This function allows Citar to correctly format the completion UI candidates.
 The default function only supports one-file-per-key notes.


citar.el
```diff
@@ -1078,7 +1078,7 @@ With prefix, rebuild the cache before offering candidates."
   (or (seq-some
        (lambda (opener)
          (funcall opener key entry)) citar-open-note-functions)
-      (funcall citar-create-note-function key entry)))
+      (funcall citar-file--open-note key entry)))
```
bdarcus, line 1081: I'm a little confused ATM; it's possible I'm just distracted, but here goes ...

The logic of the code is:

1. try to open using the `citar-open-note-functions`
2. if that doesn't work, then create a new note using the `citar-create-note-function`

But here, unless I'm misunderstanding something, you're instead just falling back to `citar--open-note`, even though that function is already included by default in `citar-open-note-functions`.

Can you clarify?

If you want that behavior, it might be good to change the suggestion on the config approach in the README, and use `add-to-list` instead?

bdarcus: > EDIT: I see your note now on the issue about the 3 args. I need to think on that.

Seems the problem is that the default create-note function requires three args, but the `funcall` on it only includes two.

For whatever reason I can't figure out, I get no error!

balajisivaraman: @bdarcus, Understood what you're trying to say!

However, as you mention, the issue is that if we use `orb-citar-edit-note` as a possible override for the `citar-open-note-functions`, then they must accept two arguments. I haven't looked into either ORB or Citar code deep enough to be able to suggest an alternative.

For the time being, I agree that suggesting `add-to-list` might be better than what I have done, while the Roam integration evolves in Citar in your following commits.

bdarcus: What happens now if you specify the orb function for `citar-open-note-functions`, and then open a note for a reference that does not currently have a note?

My assumption/hope is that it calls the orb function, and creates the new note. If that's the case, that's a function call with two arguments, and it wouldn't get to the line I quote above?

balajisivaraman: If I have `:immediate-finish t` on my template, then things work as expected, since the ORB capture process returns `t`. Otherwise, it will return `nil`, and Citar will try to call create note with 2 arguments instead of 3.

I now realise that for the time being, it may be enough to add the note about `:immediate-finish t` to the Readme file, while the integration with Roam improves also.

bdarcus: That's good.

Do you want to drop the citar.el commit, then, I can merge the README change, and we can think a bit more about if the code needs a subtle refactor?

balajisivaraman: Done :+1:

bdarcus: One last little change:

I'm trying to keep the README one-sentence-per-line. Can you fix that please; add a line break between the two sentences?

bdarcus:
Thanks again!

balajisivaraman:
No problem! Happy to help and thank you for the great package.

citar.el
bdarcus: > I agree that suggesting `add-to-list` might be better than what I have done

Just a follow-up on this. In getting #609 working just now, I realize `add-to-list` wouldn't be right, as it won't stop at the first non-nil result. So if I'm opening a sub-file ref node, it will use org-roam to do that, and then after that use the default citar open function create a new file.

PR #616, opened 2022-05-28
Rework multiple selection to use completing read in a loop

aikrahguzar:
This more or less reproduces the usual ui without needing `completing-read-multiple` including `TAB` and `RET`. With a little convoluted logic vertico works out of the box without depending on it. We just rely on the fact that `exit-minibuffer` is remapped to `vertico-exit` on the `vertico-map` so any other ui that does this remapping will work out too.

Embark works for actions on single references but I haven't added the function which allows `embark-act-all` to work with the selected-candidates. I think it should work but I don't really know how to test it.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -77,8 +77,10 @@
 (defvar embark-transformer-alist)
 (defvar embark-multitarget-actions)
 (defvar embark-default-action-overrides)
+(defvar embark-candidate-collectors)
 
 (declare-function embark--target-buffer "ext:embark")
+(declare-function embark--metadata "ext:embark")
 
 ;;; Variables
 
@@ -94,6 +96,11 @@
 (defface citar-highlight
   '((t :weight bold))
   "Face used to highlight content in `citar' candidates."
+  :group 'citar)
+
+(defface citar-selection
+  '((t :inherit highlight :slant italic))
+  "Face used for the currently selected candidates."
   :group 'citar)
 
 (defcustom citar-bibliography nil
@@ -496,7 +503,7 @@ to filter them."
                           ('(nil nil) "Select Multiple")
                           ('(nil t)   "Selected")
                           ('(t   nil) cand)
-                          ('(t   t  ) (propertize cand 'face 'highlight)))))))
+                          ('(t   t  ) (add-face-text-property 0 (length cand) 'citar-selection nil cand) cand))))))
 
 (defvar citar--multiple-setup '("TAB" "RET" exit-minibuffer)
   "Variable whose value should be a list of three elements.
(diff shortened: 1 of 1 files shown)


bdarcus:
Awesome work @aikrahguzar!

It works really nicely; glad you got the TAB/RET behavior working, as it's a subtle thing that I think makes a big difference.

Two things:

First, is this our fault, or the Doom One theme?

image

Second, and this is really not important for you to resolve in the PR:

Given there's really no downside I can see with this, should we go back to making this available on all or most commands?

> Embark works for actions on single references but I haven't added the function which allows `embark-act-all` to work with the selected-candidates. I think it should work but I don't really know how to test it.

You mean, just how `embark-act-all` works from a UX POV?

I have to go from memory, because I didn't get around to figuring out what's wrong with my install, but try this:

1. filter candidate list to some subset
2. invoke `embark-act-all`
3. choose the action to use; say by hitting `RET`

That may not be 100% right, but it's some variation on that.

citar.el
```diff
@@ -470,6 +473,71 @@ documentation for the return value and the meaning of
...
+  (citar--completion-table
+   candidates filter
+   `(display-sort-function . (lambda (cands) (citar--sort-by-selection ,selected-hash cands)))
+   `(group-function . (lambda (cand transform)
+                        (pcase (list (not (not transform))
+                                     (gethash (substring-no-properties cand) ,selected-hash))
+                          ('(nil nil) "Select Multiple")
+                          ('(nil t)   "Selected")
+                          ('(t   nil) cand)
+                          ('(t   t  ) (propertize cand 'face 'highlight)))))))
```
bdarcus, line 499: OIC; here's the face. I guess this is an obvious one; are there are more subtle alternatives?

aikrahguzar: > OIC; here's the face. I guess this is an obvious one; are there are more subtle alternatives?
> 
> EDIT: see here for some possible ideas.
> 
> https://github.com/minad/vertico/blob/crm/vertico-crm.el
> 
> This is the vertico-crm prototype, which is actually superior to the consult-crm one. It does appear to use the same face display by default, but it defines a `vertico-crm-selected` face, which we maybe don't want to do?

I think we should define a face, I just didn't think about it after using one that worked ok for me. I will do that.

aikrahguzar:
> Second, and this is really not important for you to resolve in the PR:
> 
> Given there's really no downside I can see with this, should we go back to making this available on all or most commands?

I don't use other commands so I think it would be better to have input from people who do.

There is one problem with this implementation though: it is not guaranteed so work with every `completing-read` compliant api. The fix is not difficult though: just set the last element of `citar--multiple-setup` to the function that is used to select the current candidate for that ui (should that variable be public?). With that in mind it makes sense to use this for the commands where it makes sense but with `citar-select-multiple` still `nil` by default. 

> EDIT: I did rebuild my straight repo, but that didn't fix my embark keybinding issue. Turning off the WIP corfu module and back on company did fix it, though, but only temporarily. But in the time when it was working, I don't think `embark-act-all` was working correctly as expected.

`emabrk-act-all` isn't expected to work right now since I didn't add the collector for selected candidates.

citar.el
bdarcus: > EDIT: added a face but it inherits from highlight. If you suggest a different definition for it I will accept it.

No worry; that's bike-shedding we can worry about later. 

I merged this; thanks again!

PR #617, opened 2022-05-29
Fix calling fallback function

syohex:
`citar-file--open-note` is a function, not a variable.


bdarcus:
Thanks for this. But I'm confused.

That bug should never have made it into main.

This isn't the correct fix though. It's supposed to call `citar-create-note-function`.

Here's the weird thing, however: it's correct in #616 , even though it doesn't show in GitHub as a change!

bdarcus:
I merged #616. Can you update and confirm that works correctly?

PR #618, opened 2022-05-29
Don't compute `(citar--local-files-to-cache)` unless necessary.

aikrahguzar:
Mostly from my minor annoyance at seeing reftex flood my messages buffer. Moves the check to after the candidates have been computed. Should be a bit faster too.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -837,8 +837,6 @@ If the cache is unintialized, this will load the cache.
 If FORCE-REBUILD-CACHE is t, force reload the cache.
 
 If FILTER, use the function to filter the candidate list."
-  (unless (or citar-bibliography (citar--local-files-to-cache))
-    (error "Make sure to set citar-bibliography and related paths"))
   (when force-rebuild-cache
     (citar-refresh force-rebuild-cache))
   (when (eq 'uninitialized citar--candidates-cache)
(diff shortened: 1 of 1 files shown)


bdarcus:
Can you please reformat the commit message so first line is no more than 50 characters, and subsequent no more than 72?

See https://cbea.ms/git-commit/

aikrahguzar:
> Can you please edit the commit message so first line is no more than 50 characters, and subsequent no more than 72?
> 
> See https://cbea.ms/git-commit/
> 
> EDIT: oh, and a blank line between first and rest.

Done.

bdarcus:
There was no change ;-)

aikrahguzar:
> There was no change ;-)

This is what I see as commit message now,
```
Don't recompute local bibs unless necessary.

Mostly from my minor annoyance at seeing reftex flood my messages
buffer. Moves the check to after candidates have been computed.
Should be a bit faster too.
```
Which seems to be ok? Unless git is tripping me up somehow.

bdarcus:
I'm just looking at the GitHub UI.

aikrahguzar:
This is what I see when I go to commit,
image

So it looks alright to me there too except for the empty line between the first and other lines, bu a I don't know how to get github to show the raw commit message and it is there in emacs.

bdarcus:
I'll just assume some GitHub display oddity.

Thanks!

bdarcus:
> I'll just assume some GitHub display oddity.

Just to confirm, this is indeed the case. The actual git message looks perfect.

PR #622, opened 2022-06-01
Fix #620 Sort the candidates upfront + internal

### LINKED ISSUE #620: Broken history?
I updated Citar (but also million other things, including Emacs from Git), and `citar-history` stopped working (just for Citar).  For instance, I currently study a book on biology, citing from it (for my notes) every day.  Yet, the book does not come first in `citar-open` and comes last (???) in `org-cite-insert` when I type `biology`.  Citar does not have a test suite, so I decided to ask here first (because we should expect regressions).  Anyone sees the problem, or should I investigate my setup?  Thanks!

aikrahguzar:


bdarcus:
Could you please add `Fix #620` to the body of the commit message?

Github will then link commit to issue, etc.

aikrahguzar:
The first version I posted actually didn't work, it caused the same problem which was fixed by adding the sort function, if you deselcted the candidate, it would move to selected candidates to the bottom.

What I have done now is to remove a candidate from history if a candidate was deselected and removed the sorting altogether. This means that right now selected candidates are on the top if the sorting provided by UI sorts by history first which vertico has the option to do and does by default.

bdarcus:
So did I prematurely merge, or is that OK?

I should have tested, but didn't :-)

bdarcus:
In the future, you can just do a draft PR initially, so I won't even think about merging ;-)

See https://github.com/emacs-citar/citar/issues/620#issuecomment-1143742942; might need a supplemental PR?

aikrahguzar:
> In the future, you can just do a draft PR initially, so I won't even think about merging ;-)
> 
> See [#620 (comment)](https://github.com/emacs-citar/citar/issues/620#issuecomment-1143742942); might need a supplemental PR?

I don't think so. I think this should mostly fix the issue given a sort function that has the most recent items first, the edge case notwithstanding. I think that comment is with the old sort function.

PR #624, opened 2022-06-02
refactor(notes): generalize citar-has-note

bdarcus:
Revert the keys-with-notes defcustom and code, and instead generalize
'citar-has-note', which now iterates through 'citar-note-backends'.

------

## TODO

- [ ] a stupid thing, but settle on singular (note) or plural (notes) for symbol names

------

Can you review this please @roshanshariff?

Works on my end.

Aside: I'm not sure if there's something wrong with github, but I keep finding code I'm sure I fixed in main.


citar.el
```diff
@@ -703,7 +699,7 @@ key associated with each one."
          (raw-candidates
           (parsebib-parse bib-files :fields (citar--fields-to-parse)))
          (hasfilep (citar-has-file))
-         (hasnote (citar-keys-with-notes))
+         (hasnote (citar-has-note))
```
roshanshariff, line 702: Should probably call this `hasnotep` just like `hasfilep`.

Also, on line 710 below, should be `(funcall hasnotep citekey entry)` instead of `(member ...)`.

citar.el
```diff
@@ -239,8 +239,9 @@ If nil, single resources will open without prompting."
   :group 'citar
   :type '(function))
 
-(defcustom citar-keys-with-notes-functions
-  '(citar-file--keys-with-file-notes)
+(defcustom citar-note-backends
```
roshanshariff, line 242: I think I prefer your name for this, `citar-has-note-functions`. It immediately conveys that it's a list of functions, and that it's closely related to the `citar-has-note` function.

roshanshariff (changes requested):
LGTM, apart from the changes I've suggested.

citar.el
bdarcus: Is there a reason not to use hypens; `has-note-p`?

roshanshariff: Yeah, local variables [should not use hyphens in their names](https://with-emacs.com/posts/tutorials/almost-all-you-need-to-know-about-variables/#:~:text=One%20trick%20to%20prevent%20such%20conflicts%20is%20to%20avoid%20local%20variables%20which%20have%20dashes%20in%20their%20name.).

That's because in modern elisp `let` does lexical binding instead of dynamic binding, _unless_ the variable in question has a global definition. To make sure this can't happen, the convention is that all global variables have at least one hyphen (which is usually the case) whereas all local variables introduced in `let` bindings have no hyphens.

roshanshariff: Yeah, just keep it in mind and slowly change over if/when you touch some piece of code in the future. Since this only applies to local variables anyway, it can't really cause breaking changes outside the scope of a single function.

Issue #623, opened 2022-06-02
Optimizing `citar-keys-with-notes`


bdarcus:
@roshanshariff - I got the notes API stuff merged (via #603), and it works, but the performance discrepancy here is obvious.

```elisp
ELISP> (citar-select-ref :filter (lambda (key entry) (member key (citar-keys-with-notes))))
ELISP> (citar-select-ref :filter (citar-has-file))
```

After a bit of experimenting, I think the issue is really just list vs hash performance. If I simply convert `citar-keys-with-notes` from list to hash, the performance is much better again.

```elisp
ELISP> (citar-select-ref :filter (lambda (key _entry) (gethash key my/h2)))
```

So my impulse is just to do that.

Do you have any better suggestions?

For example, possible to generalize `citar-has-note` more?

TIA.

roshanshariff:
Just using a hash table would work, but perhaps the API of `citar-has-file` might be better? Instead of returning a hash table, it returns a predicate function that takes a `citekey` and `entry` and tells you whether there's an associated file:
```elisp
(defun citar-org-roam--has-note ()
  (let ((keys (make-hash-table :test #'equal)))
    ;; Store keys that have notes in hash table
    (dolist (record (org-roam-db-query
                     [:select ref :from refs :where (= type "cite")]))
      (puthash (car record) t keys))
    ;; Return predicate that queries hash table for given key
    (lambda (citekey _entry)
      (gethash citekey keys))))
```
The advantage is that the use of a hash table is just an internal implementation detail of this function, and is not part of its public API in any way. I hope that makes sense? I can clarify if needed.

bdarcus:
> Instead of returning a hash table, it returns a predicate function that takes a `citekey` and `entry` and tells you whether there's an associated file.

Just to clarify, more general than a file; an associated note, which may be a file, but could also be a sub-file node.

Your example makes sense to me, but how would I hook that up to citar?

Would `citar-has-note` become something more like a defcustom? So could do:

```elisp
(setq citar-has-note-functions '(citar-file--has-note citar-org-roam--has-note))
```

roshanshariff:
Yes, exactly, `citar-has-note` would be something like the code I had in https://github.com/emacs-citar/citar/issues/601#issuecomment-1133883990.

bdarcus:
Ok, I'll whip up a PR. Thanks.

I'm thinking I'll make those functions public, though, since we'd want to encourage their use for this sort of thing.

[...]

PR #625, opened 2022-06-04
refactor: cache the parsebib hashes

bdarcus:
Experimenting, per discussion on #623.

----------

Rather than caching the 'completing-read' candidates, instead cache the
hashes created by 'parsebib-parse'.

Change the API a bit to focus on using keys to access the data.

----------

## TODO

- [X] modify caches and `citar--get-candidates`, now renamed `citar--get-data`
- [X] adapt `citar--get-entry` and `citar--get-value`
- [ ] adapt `citar--format-candidates`
- [ ] add functions to update metadata in the cached hash

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -599,9 +599,10 @@ If no function is found, the DEFAULT function is called."
                   (citar-file--normalize-paths
                    citar-bibliography)))
 
-(defun citar--get-value (field entry)
-  "Return the FIELD value for ENTRY."
-  (cdr (assoc-string field entry 'case-fold)))
+(defun citar--get-value (citekey field)
+  "Return FIELD value for CITEKEY."
+  (let ((entry (citar--get-entry citekey)))
+    (cdr (assoc-string field entry 'case-fold))))
 
 (defun citar--field-with-value (fields entry)
   "Return the first field that has a value in ENTRY among FIELDS ."
@@ -778,13 +779,13 @@ key associated with each one."
                       "")
                 "")))
 
-(defvar citar--candidates-cache 'uninitialized
-  "Store the global candidates list.
+(defvar citar--data-cache 'uninitialized
+  "Store the global bibliographic data.
 
 Default value of 'uninitialized is used to indicate that cache
 has not yet been created")
 
-(defvar-local citar--local-candidates-cache 'uninitialized
+(defvar-local citar--local-data-cache 'uninitialized
   ;; We use defvar-local so can maintain per-buffer candidate caches.
   "Store the local (per-buffer) candidates list.")
 
@@ -802,13 +803,13 @@ are refreshed."
   (when force-rebuild-cache
     (run-hooks 'citar-force-refresh-hook))
   (unless (eq 'local scope)
-    (setq citar--candidates-cache
-      (citar--format-candidates
+    (setq citar--data-cache
+      (parsebib-parse
         (citar-file--normalize-paths citar-bibliography))))
   (unless (eq 'global scope)
-    (setq citar--local-candidates-cache
-          (citar--format-candidates
-           (citar--local-files-to-cache) "is:local"))))
+    (setq citar--local-data-cache
+          (parsebib-parse
+           (citar--local-files-to-cache)))))
 
 (defun citar--get-template (template-name)
   "Return template string for TEMPLATE-NAME."
@@ -818,7 +819,25 @@ are refreshed."
       (error "No template for \"%s\" - check variable 'citar-templates'" template-name))
     template))
 
-(defun citar--get-candidates (&optional force-rebuild-cache filter)
+(defun citar--update-hashes (table from-table)
+  "Update TABLE according to every key-value pair in FROM-TABLE."
+  ;; Adapted from 'ht-update!'
+  ;; REVIEW better to just require 'ht'?
+  (maphash
+   (lambda (key value) (puthash key value table))
+   from-table)
+  nil)
+
+(defun citar--merge-hashes (&rest tables)
+  "Create a new hash-table that includes all the key-value pairs from TABLES.
+If multiple have tables have the same key, the value in the last
+table is used."
+  ;; Adapted from 'ht-merge'
+  (let ((merged (make-hash-table :test 'equal)))
+    (mapc (lambda (table) (citar--update-hashes merged table)) tables)
+    merged))
+
+(defun citar-get-data (&optional force-rebuild-cache)
   "Get the cached candidates.
 
 If the cache is unintialized, this will load the cache.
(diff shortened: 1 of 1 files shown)


aikrahguzar:
The assumption in #623 that the bibtex-parsing is slow part of candidate generation is not accurate. In my experience the time is spent almost equally between bibtex parsing and formatting the candidates. Here is a time profile of `citar-refresh` for a file with about 1500 entries: 
citar-profile

So I this will hurt performance and we can't take advantage of the one advantage of not caching the formatted candidates: the ability to adapt to different frame widths. We specify the number of characters in each field absolutely and not relative to the frame widths so if the frame width changes the candidates are still going to look the same except for one field and they might get truncated.

That said the actual issue in #623  is pretty narrow: it basically only pertains to the `candidate-hidden` part of the entry and that doesn't seem to be the expensive part of the operation (though as my entries don't have notes or files or links, it is also mostly empty for me. Someone with lots of notes should check the actual impact, but it involves no expensive string manipulation so I think it will probably not contribute much). So to me the better way would be to pre-compute everything that can be deduced from the bibtex file including the formatted candidate string and then invisibly append `has-file` etc on the fly.

bdarcus:
OK, so I'll close this then :-)

> That said the actual issue in https://github.com/emacs-citar/citar/issues/623 is pretty narrow: it basically only pertains to the candidate-hidden part of the entry.

And the related affixations.

> So to me the better way would be to pre-compute everything that can be deduced from the bibtex file including the formatted candidate string and then invisibly append `has-file` etc on the fly.

Is that basically status quo? Or are you suggesting some change?

My example here was using a predicate function against the "pre-computed" data.

https://github.com/emacs-citar/citar/issues/623#issuecomment-1146492476

bdarcus:
To provide context to that, I was wanting to add a defcustom to allow pre-filtering of candidates for `citar-open-notes` and `citar-open-files`. So just wanted that to be snappy.

aikrahguzar:
> Is that basically status quo? Or are you suggesting some change?
> 
> My example here was using a predicate function against the "pre-computed" data.
> 
> [#623 (comment)](https://github.com/emacs-citar/citar/issues/623#issuecomment-1146492476)

I don't completely understand the issues in #623 but my impression is that motivating this change is the fact that the notes and associated files can change even without the bib files changing. What I am saying is that issue can be resolved by keeping the pre-computation of the candidate strings against the format but not pre-computing the invisible "has-file" "has-notes" etc. This part is recomputed every time `citar--get-candidates` is called. With this approach we longer need to invalidate the cache if a note or file is added/removed. But I am unsure how expensive doing this actually is since I don't have notes/files.

bdarcus:
> my impression is that motivating this change is the fact that the notes and associated files can change even without the bib files changing. 

And the UI may not align with what, for example, `citar-has-note` reports.

> What I am saying is that issue can be resolved by keeping the pre-computation of the candidate strings against the format but not pre-computing the invisible "has-file" "has-notes" etc. 

Here I'm not following. Can you restate?

Also, FYI, in #595 I added conses for notes and files to the cached entry data list. This was a simple change to the `citar-format-candidates` function.

In #623, I was just saying it made more sense to use that cached data, than to use the functions that create that data, to ensure the alignment I mentioned above.

> This part is recomputed every time `citar--get-candidates` is called. With this approach we longer need to invalidate the cache if a note or file is added/removed. But I am unsure how expensive doing this actually is since I don't have notes/files.

I _think_ maybe you're saying something similar here to https://github.com/emacs-citar/citar/issues/623#issuecomment-1146511688 ("Or perhaps include the string representation in the cache, but tack on the note and file indicators as-needed")?

In https://github.com/emacs-citar/citar/issues/623#issuecomment-1146589524, I was experimenting with basically that idea, and was planning to add it here. But I hadn't gotten that far on the details.

roshanshariff:
@aikrahguzar, thanks for the profile data! In this case, I agree that we can't recompute the string formatting every time and we need to cache it.

I did indeed mean that we could cache the parts of the candidate string that only depend on the bib file, i.e. everything but the invisible "has:note" and "has:string". There would still be some string concatenation to be done at the end, but should be much faster than actually doing the string template expansion from scratch.

That said, there's no urgency about any of this (and indeed I don't think there's likely to be a performance issue with the current implementation, because there shouldn't be a need to rescan the files/notes more than once per user operation). When I have some time I'll try implementing and benchmarking this proposal.

It occurred to me that we could also get the layout to be responsive to the window width without much more work: we'd format and cache two strings, one of unlimited length and another fixed-width string to be shown at the right margin. Then at display-time we could just truncate the first string to the required width and concatenate it with the second.

The truncation would be done with display properties (as by `truncate-string-to-width` with the `ELLIPSIS-TEXT-PROPERTY` argument) so that the hidden part of the text would remain searchable. I imagine that concatenating three strings instead of two is probably doable within the performance budget (i.e. the left column, right column, and invisible "has:note" and "has:file" tags).

bdarcus:
A couple of random thoughts. 

And to be clear, I don't believe these are critical; I'd just like to settle at the first before I tag 1.0 soon, since it's an API question. The other stuff is just internal details.

In working on this, it seemed to me that some of the "get" and formstting-related functions that have `entry` as arg could instead use `citekey`. It's a small thing, but would it make sense to change those?

The other thing is really just a high-level observation, which became more clear to me as I was playing with this branch:

Generally, we have two kinds of data access we need, and which we want to be performant.

1. Key lookup of reference data
2. Formatted string lookup of key

2 is purely for the completion candidates, while 1 is for the rest.

I was just thinking about the general fact that hash-tables are much faster in large collections (someone [here](https://www.reddit.com/r/emacs/comments/75h7k4/comment/do6h0xm/)  notes `delete-dups` uses list when less than  100, but hash-tables otherwise).

So notwithstanding caching and stuff (which does complicate this a lot), but just from that high level POV, you could imagine two hash-tables that reflect that above distinction; one to use for completing-read, and the other for everything else.

The caveats are important though, including that it's expensive to create the candidate strings regardless.

bdarcus:
PS - interesting performance detail of the formatted string:

In bibtex-completion, the candidate is a pre-forrnatted string; designed only for filtering, and NOT for display.

The formatting for display happens on-demand, in ivy and helm, which both allow that.

Completing-read doesn't, unless you count the new grouping support.

aikrahguzar:
> PS - interesting performance detail of the formatted string:
> 
> In bibtex-completion, the candidate is a pre-forrnatted string; designed only for filtering, and NOT for display.
>
> The formatting for display happens on-demand, in ivy and helm, which both allow that.

I was thinking about this and this is the correct way to do it. Only a few strings need to be displayed at one time and formatting them on the fly will be instantaneous.

> Completing-read doesn't allow that, unless you count the new grouping support in Emacs 28. We use it already for multi-selection, BTW, and in a couple of other places. See, for example:

Grouping support was my idea of how to enable that too and it is used for multiple selection but there it just prioritizes so if completing read doesn't understand group functions, the user wouldn't be able to which entries are selected. That is pretty bad but not disastrous. Showing unformatted  entries seems much worse to me.

But I wonder if the `compat` library helps with that? I don't know if it just provides the new symbols to older emacs version or actually backports improvements to exisiting functions too. 

I think performance without caching probably hurts older and less powerful machines most. On my laptop even file with 1500 entries refreshes in under a second which is noticeable and somewhat annoying but too bad.

bdarcus:
> But I wonder if the compat library helps with that? I don't know if it just provides the new symbols to older emacs version or actually backports improvements to exisiting functions too.

Somehow I doubt `group-function` would be something supported there. Completion seems not to get very much attention.

> I think performance without caching probably hurts older and less powerful machines most. On my laptop even file with 1500 entries refreshes in under a second which is noticeable and somewhat annoying but too bad.

To be clear, I'm not advocating removing caching; but it helps to evaluate the raw performance independent of it.

Bibtex-completion, for example, does caching as well, even with those optimizations.

bdarcus:
After playing more this morning in https://github.com/emacs-citar/citar/pull/625#issuecomment-1146777923, tentative conclusions:

1. not a huge difference in performance creating lists vs hashes
2. the hash is _much_ faster to access in my ~1000 entry bib file
3. formatting the candidates is indeed more expensive, but it may not be a huge practical deal

I've not considered some of the other nuances yet, but might provide some useful info.

bdarcus:
For some unknown reason (maybe because I haven't added new commit, but only force-pushed updates?), github won't let me reopen this.

But if either you want to play with this, the basic low-level stuff here works, at least to experiment with, or take a look at.

PR #626, opened 2022-06-05
refactor: cache the parsebib hashes

bdarcus:
Opening a new draft PR alternative to #625, not because I'll necessarily merge this, but because GitHub is being weird, and I think it worth discussing. I kind of went down a rabbit hole here.

This would be a major, breaking, change if we adopted it, but my impulse is this would make the code much easier to manage going forward. And now would be a good time to contemplate such a change, before I tag 1.0, or to reject it.

## Basic Idea

The `(candidate . (key . entry))` candidate structure is complex, and difficult to work with. There were good reasons for it at the time, but perhaps that was only necessary because we used the same cache for both candidates and entry data?

Aside from any performance considerations, am I missing something obvious, or would it not result in cleaner, easier to understand and debug code, if we got rid of the current candidate structure, and only use citekeys and hashes to access everything?

Here, schematically, is what I was thinking with this:

```elisp
(defun citar--get-data ()
  "Return hashes, with entry alists."
  (list
   citar--global-data-cache
   citar--local-data-cache))

(defun citar--get-ref-candidates ()
  "Return hash."
   citar--ref-candidates-cache)

(defun citar-select-ref ()
  "Select reference, return citekey."
  (let* ((candidates (citar--get--ref-candidates))
         (choice (completing-read "Ref: " candidates)))
    (gethash choice candidates)))

(defun citar-get-entry (citekey)
  "Return entry alist for CITEKEY."
  (seq-some
   (lambda (ht)
     (gethash key ht))
   (citar--get-data)))

(defun citar-get-value (citekey field)
  "Retern FIELD value for CITEKEY."
  (let ((entry (citar-get-entry citekey)))
    (cdr (assoc-string field entry))))
```

## The Problem

This impacts code throughout. A quick ripgrep shows 61 instances of "key-entry" in the repo, for example. And I haven't found it easy to adjust that code.

----------------

deleted previous content (benchmarks, etc., which showed hashes are much faster in general, but this may only have a marginal impact on performance in the end)

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -52,8 +52,6 @@
 
 ;; make all these private
 (make-obsolete 'citar-get-template 'citar--get-template "1.0")
-(make-obsolete 'citar-get-link 'citar--get-link "1.0")
-(make-obsolete 'citar-get-value 'citar--get-value "1.0")
 (make-obsolete 'citar-display-value 'citar--display-value "1.0")
 (make-obsolete 'citar-open-multi 'citar--open-multi "1.0")
 (make-obsolete 'citar-select-group-related-resources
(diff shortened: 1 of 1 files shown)


bdarcus:
@minad - is there any performance advantages for completion UIs in using hashes over lists?

Assume we want to support anywhere up to a few thousand candidates.

minad:
No.

bdarcus:
Correct me if I'm wrong: this is because a package like Vertico incrementally renders the candidates, so in effect is only processing a small number at a time?

minad:
It doesn't matter for any of the UIs. Hash tables and lists are handled by all-completions by iterating over them. There is no difference between the different supported data structures.

PR #628, opened 2022-06-07
Refactor cache architecture 

### LINKED ISSUE #623: Optimizing `citar-keys-with-notes`

### LINKED ISSUE #627: Split data and completion caches, change APIs to use citekeys
I went down a rabbit hole with some code experiments in #625 and #626, but maybe it's better to just talk about it generally, to decide if there's any value in the idea, or to reject it.

What I'm proposing here, I realized after opening this, is exactly the same implementation as in org-mode `oc-basic`. See below: https://github.com/emacs-citar/citar/issues/627#issuecomment-1148429228.

There's some demo code at the end of this.

## Basic Idea

The `(candidate . (key . entry))` candidate structure is complex, and difficult to work with. There were good reasons for it at the time, but perhaps that was only necessary because we used the same cache for both candidates and entry data?

Aside from any performance considerations (which I actually don't think are significant; see below), am I missing something obvious, or would it not result in cleaner, easier to understand and debug code, if we got rid of the current candidate structure, and only use citekeys and hash tables to access everything?

The other huge advantage of this, I learned subsequently, is customizable hash table equality testing; code like this to invalidate the cache:

```elisp
(unless (equal hash-1 hash-2))
```

A simple diagram of the hash structures:

citar-hashes

## The Problem

This would break code throughout. A quick ripgrep shows 61 instances of "key-entry" in the repo, for example. And I haven't found it easy to adjust that code; which maybe supports my point about it being too complex (mostly because of the candidate structure)?

So it might be possible to make the code simpler, cleaner and easier to maintain in the end, with a more intuitive API, but to get there may be a PITA.

## On Performance

I started out playing with this because I thought this could improve performance, and potentially solve a relatively minor issue in #623.

I no longer think this would improve performance overall, for two reasons:

1. The current candidate structure was actually mostly a performance optimization, since it means you don't need to look up the entry.
2. The parsebib hash values are, in fact, alists. So once you grab the entry, you're still working with small alists.

Rather, I think the improved and scalable entry access speed of the hashes makes it possible for us to simplify the code and API without _losing_ performance. 

Benchmark examples with ~1000 entries:

```elisp
ELISP> (benchmark-run-compiled 1000 (citar-get-entry "2005")) ; hash version
(0.001436192 0 0.0)

ELISP> (benchmark-run-compiled 1000 (citar--get-entry "2005")) ; list version
(0.26441816199999996 1 0.10594141199999996)

ELISP> (benchmark-run-compiled 1000 (citar-get-value "title" "2005")) ; hash version
(0.002393128 0 0.0)

ELISP> (benchmark-run-compiled 1000 (citar--get-value "title" (citar--get-entry "2005"))) ; list version
(0.939745279 2 0.17277752700000004)
```

Finally, to confirm @aikrahguzar's [profiling](https://github.com/emacs-citar/citar/pull/625#issuecomment-1146677182), parsebib is reasonably fast; what's slow is creating the candidate strings (and per my point above, there's no real difference between the hash and list versions there).

```elisp
ELISP> (benchmark-run-compiled 100 (parsebib-parse citar-bibliography))
(18.502350075 96 8.255232515000003)

ELISP> (benchmark-run-compiled 100 (citar--format-candidates)) ; hash version
(67.580158682 153 20.433643597)

ELISP> (benchmark-run-compiled 100 (citar--format-candidates citar-bibliography)) ; list version
(65.927820852 153 17.207595579)
```

## Demo Code

Here is demo code you can `eval-buffer` on:

<details>

```elisp


(require 'parsebib)
(require 'citar)

;; Internal variables

;; This design is adapted from org-mode 'oc-basic'.

(defvar citarn--bibliography-cache nil
  "Cache for parsed bibliography files.
This is an association list following the pattern:
  (FILE-ID . ENTRIES)
FILE-ID is a cons cell (FILE . HASH), with FILE being the absolute file name of
the bibliography file, and HASH a hash of its contents.
ENTRIES is a hash table with citation references as keys and fields alist as
values.")

(defvar citarn--completion-cache nil
  "Hash with key as completion string, value as citekey.")

;; Internal functions

(defun citarn--create-data-cache (files)
  "Create data cache for FILES."
  (seq-map
   (lambda (file)
     (let* ((path (file-truename file))
            (checksum
             (split-string
              (call-process "md5sum" nil t nil path) " "))) ; TODO doesn't work correctly
       (cons
        (cons file checksum) ; get the checksum here
        (parsebib-parse path))))
   files))

(defun citarn-ref-read-make-candidates (hashes)
  "Return a hash with completion candidates from HASHES."
  (let ((chash (make-hash-table :test #'equal)))
    (dolist (hash hashes)
      (maphash
       (lambda (citekey _entry)
         (puthash (citarn-get-value citekey "title") citekey chash))
       hash))
    ;; return new hash
    chash))

(defun citarn--get-candidates ()
  "Return completion hash."
  (citarn-refresh)
  citarn--completion-cache)

(defun citarn--get-data ()
  "Return hashes."
  (remq 'nil
        (list
         citarn--bibliography-cache)))

(defun citarn-refresh ()
  "Load caches."
  (interactive)
  (unless citarn--bibliography-cache
    (setq citarn--bibliography-cache ; load this first
          (parsebib-parse citar-bibliography)))
  (unless citarn--completion-cache
    (setq citarn--completion-cache
          (citarn-ref-read-make-candidates
           (citarn--get-data)))))

(defun citarn-select-ref ()
  "Select reference, return citekey."
  (let* ((candidates (citarn--get-candidates))
         (choice (completing-read "Ref: " candidates)))
    (gethash choice candidates)))

(defun citarn-get-entry (key)
  "Return entry alist for KEY."
   (seq-some
   (lambda (ht)
     (gethash key ht))
   (citarn--get-data)))

(defun citarn-get-value (key-or-entry field)
  "Retern FIELD value for KEY-OR-ENTRY."
  (let ((entry (if (stringp key-or-entry)
                   (citarn-get-entry key-or-entry)
                 key-or-entry)))
    (cdr (assoc-string field entry))))

;; Interactive commands

(defun citarn-example ()
  "Return title as message."
  (interactive)
  (let* ((choice (citarn-select-ref))
         (title (citarn-get-value choice "title")))
  (message title)))

(citarn-example)

</details>

### PR DESCRIPTION
**Pull request by bdarcus**
Convert the cache from a single list, to a list of structured objects, each of which contains two hash tables, each keyed by citekey:

- `:entries` the raw bib data
- `:completions` the formatted completion strings

The completion candidates themselves are now no longer cached, but quickly assembled on-the-fly.

The bibliography objects also includes the following additional properties, to selecting update bibliographic data:

- `:hash` the checksum of the file
- `:buffers` the associated buffers

Close #627, fix #623 #610

## Status

Roshan is working on a PR against this that I think will work better for us. 

----------------------------

Caching and completion all work, as do at least some of the interactive commands (I haven't tested them all; some may break currently because completion now always returns a list of keys).

image

I've added keyword comments in places.

## Tasks

- [X] refactor `citar--format-candidates` to create the appropriate completion hash
- [X] refactor cache code, and it's broader integration
- [X] update `citar-select-ref` and `citar-select-refs` to use citekeys
- [X] hook up enhanced formatting for completion candidates
- [X] re-implement basic `has-*` support
- [X] re-add or fix filtering (I'm actually not sure how to do this ATM, and it may be the same as the above)
- [X] adapt `citar-has-file` and `citar-has-note` to this architecture (f there's a reason to do so; IDK), or remove them
- [X] review `citar--completion-table` predicate-function code
- [ ] make sure local bibs correctly supported
- [ ] check filenotify code (per below, we should remove it perhaps)
- [ ] update embark handling code to use citekeys (it may be fine now actually; see https://github.com/emacs-citar/citar/pull/628/files#r894474531)
- [ ] update everything else, including interactive commands
- [ ] change `citar-has-file` to `citar-has-library-file`
- [ ] benchmarks, to evaluate this in comparison to main? Performance alone is not the only metric here (code clarity is more important to me actually), but it's one.

## Questions and notes 

1. Can we avoid reintroducing a separate local cache? I think we can. See `citar--parse-bibliography` (and also `org-cite-list-bibliography-files`).
2. Can we remove some of the inotify code, since this should obsolete the part that deals with bib files? If yes, then we can also remove `citar-refresh` etc, which I have done for now (though adding it back would basically be a wrapper for `(clrhash citar--completion-cache)`.
3. Can we resolve #623 along with this? (**I think I did.**)
4. Do we need "rebuild cache" now (see [comments](https://github.com/emacs-citar/citar/pull/122#issuecomment-1150749793) from its original author)? I assumed no, so removed it. What about `ensure-entries`?
5. Should we wrap the completion cache in `citar--get-candidates`, or deprecate that function (as now)?
6. Where else can we simplify, and remove code?
7. `citar-select-ref` now always returns a list. Is that a mistake? My key concern is avoiding bugs in Embark and such.
8. `citar--ref-completion-table` is conceptually the same as `citar--get-candidates`; should we rename it `citar--get-ref-completions` to be consistent?

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -179,6 +179,25 @@ All functions that match a particular field are run in order."
   :type '(alist :key-type   (choice (const t) (repeat string))
                 :value-type function))
 
+(defcustom citar-prefilter-entries '(nil . t)
+  "When non-nil pre-filter note and library files commands.
+For commands like 'citar-open-notes', this will only show
+completion candidates that have such notes.
+
+The downside is that, if using Embark and you want to use a different
+command for the action, you will not be able to remove the
+filter.
+
+The value should be a cons of the form:
+
+(FILTER . TOGGLE)
+
+FILTER turns this on by default
+
+TOGGLE use prefix arg to toggle behavior"
+  :group 'citar
+  :type 'cons)
+
 (defcustom citar-symbols
   `((file  .  ("F" . " "))
     (note .   ("N" . " "))
--- a/citar.el
+++ b/citar.el
@@ -458,10 +458,9 @@ FILTER: if non-nil, should be a predicate function taking
   function returns non-nil will be offered for completion.  For
   example:
 
-  (citar-select-ref :filter 'citar-has-library-file-p)
+  (citar-select-ref :filter (citar-has-note))
 
-  (citar-select-ref :filter 'citar-has-note-p)"
-  ;; TODO readd an example filter or two above?
+  (citar-select-ref :filter (citar-has-file))"
   (let* ((candidates (or (citar--ref-completion-table)
                          (user-error "No bibliography set")))
          (chosen (if (and multiple citar-select-multiple)
@@ -573,6 +572,10 @@ HISTORY is the 'completing-read' history argument."
          ((string-match "http" resource 0) "Links")
          (t "Library Files")))))
 
+(defun citar--bibliography-files ()
+  "The list of global and local bibliography files."
+  (seq-concatenate 'list citar-bibliography (citar--local-files-to-cache)))
+
 (defun citar--ref-completion-table ()
   "Return completion table for cite keys, as a hash table.
 In this hash table, keys are a strings with author, date, and
@@ -637,9 +640,7 @@ strings or nil."
   (unless (hash-table-p citar--file-id-cache)
     (setq citar--file-id-cache (make-hash-table :test #'equal)))
   (let ((results nil))
-    ;; FIX the files to parse needs to be a function that returns the right
-    ;; local and/or global bibliography files for the current buffer.
-    (dolist (file citar-bibliography)
+    (dolist (file (citar--bibliography-files))
       (when (file-readable-p file)
         (with-temp-buffer
           (when (or (file-has-changed-p file)
--- a/citar.el
+++ b/citar.el
@@ -236,16 +236,17 @@ If nil, single resources will open without prompting."
 (defcustom citar-open-note-functions
   '(citar-file--open-note)
   "List of functions to open a note."
+  ;; REVIEW change to key only arg?
   :group 'citar
   :type '(function))
 
 (defcustom citar-has-note-functions
-  '(citar-file-has-notes)
-  "Functions to return a predicate testing whether a reference has
-associated notes.
+  '(citar--has-file-notes)
+  "Functions to test whether a reference has associated notes.
 
-Such functions must take arguments KEY and ENTRY and return
-non-nil when the reference has associated notes."
+Such functions must take KEY and return non-nil when the
+reference has associated notes."
+  ;; REVIEW change to key only arg?
   :group 'citar
   :type '(function))
 
@@ -257,6 +258,7 @@ A note function must take two arguments:
 
 KEY: a string to represent the citekey
 ENTRY: an alist with the structured data (title, author, etc.)"
+  ;; REVIEW change to key only arg?
   :group 'citar
   :type 'function)
 
@@ -269,6 +271,7 @@ A note function must take three arguments:
 KEY: a string to represent the citekey
 ENTRY: an alist with the structured data (title, author, etc.)
 FILEPATH: the file name."
+  ;; REVIEW change to key only arg?
   :group 'citar
   :type 'function)
 
@@ -349,6 +352,12 @@ of all citations in the current buffer."
   :group 'citar
   :type '(repeat string))
 
+(defcustom citar-select-multiple t
+  "Use `completing-read-multiple' for selecting citation keys.
+When nil, all citar commands will use `completing-read`."
+  :type 'boolean
+  :group 'citar)
+
 ;;; Keymaps
 
 (defvar citar-map
@@ -380,24 +389,34 @@ of all citations in the current buffer."
     map)
   "Keymap for Embark citation-key actions.")
 
-;;; Completion functions
+;; Internal variables
 
-(defcustom citar-select-multiple t
-  "Use `completing-read-multiple' for selecting citation keys.
-When nil, all citar commands will use `completing-read`."
-  :type 'boolean
-  :group 'citar)
+;; Most of this design is adapted from org-mode 'oc-basic',
+;; written by Nicolas Goaziou.
+
+(defvar citar--bibliography-cache nil
+  "Cache for parsed bibliography files.
+This is an association list following the pattern:
+  (FILE-ID . ENTRIES)
+FILE-ID is a cons cell (FILE . HASH), with FILE being the absolute file name of
+the bibliography file, and HASH a hash of its contents.
+ENTRIES is a hash table with citation references as keys and fields alist as
+values.")
+
+(defvar citar--completion-cache (make-hash-table :test #'equal)
+  "Hash with key as completion string, value as citekey.")
+
+;;; Completion functions
 
 (defun citar--completion-table (candidates &optional filter &rest metadata)
   "Return a completion table for CANDIDATES.
 
-CANDIDATES is an alist with entries (CAND KEY . ENTRY), where
-  CAND is a display string for the bibliography item given
-  by (KEY . ENTRY).
+CANDIDATES is a hash with references CAND as key and CITEKEY as value,
+  where CAND is a display string for the bibliography item.
 
 FILTER, if non-nil, should be a predicate function taking
-  arguments KEY and ENTRY.  Only candidates for which this
-  function returns non-nil will be offered for completion.
+  argument KEY. Only candidates for which this function returns
+  non-nil will be offered for completion.
 
 By default the metadata of the table contains the category and
 affixation function. METADATA are extra entries for metadata of
@@ -406,20 +425,22 @@ the form (KEY . VAL).
 The returned completion table can be used with `completing-read`
 and other completion functions."
   (let ((metadata `(metadata . ((category . citar-reference)
-                                . ((affixation-function . ,#'citar--affixation)
+                                . ((affixation-function . ,#'citar--ref-affix)
                                    . ,metadata)))))
     (lambda (string predicate action)
       (if (eq action 'metadata)
           metadata
+        ;; REVIEW this now works, but probably needs refinement
         (let ((predicate
                (when (or filter predicate)
-                 (lambda (cand-key-entry)
-                   (pcase-let ((`(,cand ,key . ,entry) cand-key-entry))
+                 (lambda (cand _)
+                   (let* ((key (gethash cand candidates))
+                          (entry (citar--get-entry key)))
                      (and (or (null filter) (funcall filter key entry))
-                          (or (null predicate) (funcall predicate cand))))))))
+                          (or (null predicate) (funcall predicate string))))))))
           (complete-with-action action candidates string predicate))))))
 
-(cl-defun citar-select-ref (&optional &key rebuild-cache multiple filter)
+(cl-defun citar-select-ref (&optional &key multiple filter)
   "Select bibliographic references.
 
 A wrapper around 'completing-read' that returns (KEY . ENTRY),
@@ -429,9 +450,6 @@ data.
 
 Takes the following optional keyword arguments:
 
-REBUILD-CACHE: if t, forces rebuilding the cache before offering
-  the selection candidates.
-
 MULTIPLE: if t, calls `completing-read-multiple` and returns an
   alist of (KEY . ENTRY) pairs.
 
@@ -440,49 +458,32 @@ FILTER: if non-nil, should be a predicate function taking
   function returns non-nil will be offered for completion.  For
   example:
 
-  (citar-select-ref :filter (citar-has-file))
+  (citar-select-ref :filter 'citar-has-library-file-p)
 
-  (citar-select-ref :filter (citar-has-note))
-
-  (citar-select-ref
-   :filter (lambda (_key entry)
-             (when-let ((keywords (assoc-default \"keywords\" entry)))
-               (string-match-p \"foo\" keywords))))"
-  (let* ((candidates (citar--get-candidates rebuild-cache))
+  (citar-select-ref :filter 'citar-has-note-p)"
+  ;; TODO readd an example filter or two above?
+  (let* ((candidates (or (citar--ref-completion-table)
+                         (user-error "No bibliography set")))
          (chosen (if (and multiple citar-select-multiple)
                      (citar--select-multiple "References: " candidates
                                              filter 'citar-history citar-presets)
                    (completing-read "Reference: " (citar--completion-table candidates filter)
-                                    nil nil nil 'citar-history citar-presets nil)))
-         (notfound nil)
-         (keyentries
-          (seq-mapcat
-           ;; Find citation key-entry of selected candidate.
-           ;; CHOICE is either the formatted candidate string, or the citation
-           ;; key when called through `embark-act`.  To handle both cases, test
-           ;; CHOICE against the first two elements of the entries of
-           ;; CANDIDATES.  See
-           ;; https://github.com/bdarcus/citar/issues/233#issuecomment-901536901
-           (lambda (choice)
-             (if-let ((cand (seq-find
-                             (lambda (cand) (member choice (seq-take cand 2)))
-                             candidates)))
-                 (list (cdr cand))
-               ;; If not found, add CHOICE to NOTFOUND and return nil
-               (push choice notfound)
-               nil))
-           (if (listp chosen) chosen (list chosen)))))
-    (when notfound
-      (message "Keys not found: %s" (mapconcat #'identity notfound "; ")))
-    (if multiple keyentries (car keyentries))))
+                                    nil nil nil 'citar-history citar-presets nil))))
+    ;; Return a list of keys regardless of 1 or many
+    (if (stringp chosen)
+        (list (gethash chosen candidates))
+      (seq-map
+       (lambda (choice)
+         (gethash choice candidates))
+       chosen))))
 
-(cl-defun citar-select-refs (&optional &key rebuild-cache filter)
+(cl-defun citar-select-refs (&optional &key filter)
   "Select bibliographic references.
 
 Call 'citar-select-ref' with argument :multiple; see its
 documentation for the return value and the meaning of
 REBUILD-CACHE and FILTER."
-  (citar-select-ref :rebuild-cache rebuild-cache :multiple t :filter filter))
+  (citar-select-ref :multiple t :filter filter))
 
 (defun citar--multiple-completion-table (selected-hash candidates filter)
   "Return a completion table for multiple selection.
@@ -526,7 +527,7 @@ HISTORY is the 'completing-read' history argument."
                          (completing-read
                           (format "%s (%s/%s): " prompt
                                   (hash-table-count selected-hash)
-                                  (length candidates))
+                                  (hash-table-count candidates))
                           (citar--multiple-completion-table selected-hash candidates filter)
                           nil t nil history `("" . ,def)))))
              (unless (equal item "")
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by bdarcus** [comment]
@roshanshariff - FYI, I have a short status section at the top of the PR, where I will briefly update where things stand as I make changes.

I moved everything from `citarn.el` to `citar.el`, but have left the `citarn.el` file there in case it's useful for to use experimentally.

**Comment by bdarcus** [comment]
Hmm .... for reasons I can't understand, `citar--ref-completion-table` is currently MUCH faster than `citar--format-candidates`.

**Comment by bdarcus** [comment]
@roshanshariff 

> I ...  have left the `citarn.el` file there in case it's useful for to use experimentally.

See one of these "experiments" I just added for `has-note` support.

As always with that file, you can just do `eval-buffer`.

The nice thing is it makes use of your code here without modification.

image

**Comment by roshanshariff** [comment]
Just a quick high-level question based on a cursory skim of the code: I see you've added a long-term cache for the has-notes information. And it looks like the cache is updated whenever the completion table is generated? Is there a reason to have this cache in that case? Wouldn't it just be enough to re-scan the notes directory/org-roam database/etc. whenever needed, as with the current code?

Second, it seems a little awkward to have a hash table mapping completion strings to cite keys. Couldn't we avoid this just by prepending the cite key followed by a space to the completion string with a display property to hide it? Then, when we get back the completion string, we just extract the first word and we have the associated cite key. That should clean up the code a little. I was wondering if there was some reason why you didn't do it that way originally.

**Comment by bdarcus** [comment]
> Couldn't we avoid this just by prepending the cite key followed by a space to the completion string with a display property to hide it?

Yeah, as I suspected: no. Unfortunately, `completely-read` strips the text properties at the end.

```elisp
ELISP> (setq my/c2 (list (propertize "one" 'citar-key "1") (propertize "two" 'citar-key "2")))
(#("one" 0 3
   (citar-key "1"))
  #("two" 0 3
    (citar-key "2")))

ELISP> (get-text-property 0 'citar-key (completing-read "test:" my/c2))
nil
```

roshanshariff:
Yeah, I figured string properties wouldn't survive the round-trip through the completion system, but we already use hidden text for other things so that should work.

My thought behind the design of the has-note/has-file functions was this: the truly essential operation is to test if a given key+entry has an associated note/file. So `has-file` and `has-note` return a function that lets you answer just that question. I explicitly didn't want any users of those functions to rely on a full hash table listing all the possible keys with notes. As far as I can tell, we still don't need that capability. These functions also don't access the cache and are completely self-contained, which makes them easier to reason about and test (for when we have a more complete test suite).

bdarcus:
> Couldn't we avoid this just by prepending the cite key followed by a space to the completion string with a display property to hide it? Then, when we get back the completion string, we just extract the first word and we have the associated cite key. 

Done.

It has two other benefits, BTW.

1. this becomes a non-issue, because no need to do the lookup: https://github.com/emacs-citar/citar/pull/631#discussion_r894043612
2. `consult--multi` only accepts lists of strings for candidates (though I'm not sure that's critical, always nice to have options for us and for others).

Just need to look closely at `citar--ref-completion-table` to make sure it's correctly handling cache invalidation, etc.

> My thought behind the design of the has-note/has-file functions was this: the truly essential operation is to test if a given key+entry has an associated note/file. So `has-file` and `has-note` return a function that lets you answer just that question. I explicitly didn't want any users of those functions to rely on a full hash table listing all the possible keys with notes. As far as I can tell, we still don't need that capability. These functions also don't access the cache and are completely self-contained, which makes them easier to reason about ...

But they're hard, at least for me, to understand to begin with. I can figure it out, if the value proposition is sound though; am just not sure ATM.

Really the only other issue that was motivating me in thinking about that is integrating, say via third-party code, with other note packages. `citar-org-roam` is the immediate test of this, but there will likely be others; see, for example:

https://github.com/Vidianos-Giannitsis/zetteldesk.el/issues/5

> ... and test (for when we have a more complete test suite).

A test suite tied into GitHub Actions CI would be pretty useful going forward.

bdarcus:
> I've partially reverted back to the `citar-has-note`and `citar-has-file` approach in main here in the latest commit, which maybe suggests getting rid of these caches and functions. 

Done.

I also added a convenience function that builds on your approach, so you can do this:

```elisp
ELISP> (citar-get-data-entries :filter (citar-has-note))
#<hash-table equal 32/65 0x193e43f>
```

bdarcus:
I don't fully understand this, @roshanshariff, but it's clearly key to the caching strategy.

https://github.com/emacs-citar/citar/blob/2c03ede3889e6f2c57f1453ec7cddd3b3b79299f/citar.el#L694-L698

And I'm worried it requires using a hash table for both.

Any ideas?

```elisp
ELISP> org-cite-basic--bibliography-cache
((("~/org/bib/academic.bib" . "0f1fa900095a0a6b581197510122cafdfc14a8a4")
  . #<hash-table equal 757/1095 0x2976199>))

ELISP> org-cite-basic--completion-cache
#<hash-table equal 756/1095 0x14f8ae5>

ELISP> (setq my/entries (org-cite-basic--parse-bibliography))
(("~/org/bib/academic.bib" . #<hash-table equal 757/1095 0x2976199>))

ELISP> (gethash my/entries org-cite-basic--completion-cache)
t
```

bdarcus:
> And I'm worried (like at this point am 99% sure) it requires using a hash table for both, and so I will need to revert those changes.

So I've done this.

Caching now works correctly, as does everything else (filtering, for example).

And the notes and files caches are gone.

citar.el
```diff
@@ -1017,6 +969,7 @@ FORMAT-STRING."
   "Look up key for a citar-reference TYPE and TARGET."
   (cons type (or (cadr (assoc target
                               (with-current-buffer (embark--target-buffer)
+                                ;; FIX how?
```
bdarcus, line 972: @aikrahguzar I think you wrote this, but I don't understand it, and how to adapt.

bdarcus:
@aikrahguzar @roshanshariff 

On item 1 in "questions".

I haven't tested this, but here's what I think conceptually:

I added a new `citar--bibliography-files` function that concatenates the local and global bib files.

https://github.com/emacs-citar/citar/commit/a7031aa07daa02db1d9ab746746ec220dd244934

So what _should_ happen is that upon switching to a new buffer, with a different list of these files, `citar--parse-bibliography` will recognize that, and invalidate the caches.

Compared to main branch, there won't be distinct local and global caches, so the downside is that if you're switching buffers a lot, each with different lists of bib files, there will be a pause on the first call to `citar-select-ref`.

I think this is a good balance, and will be a feature enhancement for users, who no longer have to configure filenotify.

WDYT?

bdarcus:
> The PR I'm working on ...

OK, cool.

Is it feasible to rebase off this branch now? I changed a LOT this morning ;-)

I can then leave it alone for awhile.

Unless your's is drastically different of course, in which case better to branch off main.

roshanshariff:
I haven't yet modified any code, so should be okay to rebase. I'm currently working on the lowest level of the cache, which involves some bookkeeping to keep track of which buffers use which bibliographies (so we can remove a bibliography from the cache when a buffer stops using it).

Once that's done, I'll work my way up the stack modifying code as needed to new the new cache.

bdarcus:
I should add I'm not 100% certain that my summary of the org cache design I borrowed is correct.

It may be more sophisticated than I'm characterizing, and I just missed it!

roshanshariff:
I'm changing that part of the code, so I'll see if that bug is still there afterwards. I don't entirely see why `citar--bibliography-cache` needs to be keyed by both the filename and its hash value? You would only need that if you wanted to cache multiple versions of the same file, with possibly different contents.

In the new code, the cache is a hash table keyed by filename, and the values are records containing the hash value, parsed entries, etc. Whenever the cache is refreshed, it checks to see if the hash value has changed and only re-parses the file if it has. This seems more in line with what we need, I think? (Unless you can think of a reason we'd need to keep multiple versions of the same file in cache?)

bdarcus:
> I don't entirely see why `citar--bibliography-cache` needs to be keyed by both the filename and its hash value? You would only need that if you wanted to cache multiple versions of the same file, with possibly different contents.

No, that would be overkill in my view.

Maybe it has something to do with the hash table equality testing I was saying yesterday I didn't understand?

> In the new code, the cache is a hash table keyed by filename, and the values are records containing the hash value, parsed entries, etc. Whenever the cache is refreshed, it checks to see if the hash value has changed and only re-parses the file if it has. This seems more in line with what we need, I think? 

So what's different in the end?

They both only reparse if a file has changed?

In any case, I think we agree on what we need, so that's good :-)

roshanshariff:
After reading `oc-basic.el` more carefully, this is my understanding of what `org-cite-basic--parse-bibliography` does. This is the only function that references `org-cite-basic--bibliography-cache`.

1. For each bibliography file, if the file has never been seen before or it might have changed (based on file size and modification time), compute the hash of the file contents and store it in `org-cite-basic--file-id-cache`.
2. Look up each `(filename . hash)` pair in the `org-cite-basic--bibliography-cache` alist. If found, the file's current contents have already been parsed before, so return them. If the file has changed (i.e. the hash was updated in step 1) then this lookup will fail.
3. If not found in step 2, parse the file's new contents and push them onto `org-cite-basic--bibliography-cache`, keyed by the filename and hash value.

Note that no filename+hash+contents item is ever removed from `org-cite-basic--bibliography-cache`. The effect is that if you have a large, constantly changing bibliography file, every version of it is parsed and stored in the cache, which keeps growing. In return, if the contents of a bibliography file ever change back to a previous value, it won't be re-parsed.

I suspect `oc-basic` is geared towards small bibliography files that are associated with individual Org files. That's not how I use citar though; I have a large centralized bibliography file that changes all the time, along with possibly smaller local files for individual projects. I believe this is probably a more common usage of Citar.

In the new PR, the bibliography cache is keyed only by filename, not the hash. If the file changes, we don't bother to save the old version. Then, instead of having two different hash tables keyed by filename (the file-id cache and the bibliography cache), we have one hash table and store both the caches in fields in the value of that hash table. We then also use this hash table to store some additional metadata, like which buffers are using that bibliography. `oc-basic` doesn't keep track of this because it never removes anything from the cache.

bdarcus:
Awesome analysis; makes total sense!

Yeah, Nicolas isn't a domain expert, and never got feedback on these
low-level details, so not surprising there may be an assumption or two that
doesn't hold generally.

bdarcus:
This is helpful in clarifying my questions about equality testing:

https://www.emacswiki.org/emacs/HashMap#h5o-1

citar.el
```diff
@@ -179,6 +179,25 @@ All functions that match a particular field are run in order."
...
+The value should be a cons of the form:
+
+(FILTER . TOGGLE)
+
+FILTER turns this on by default
+
+TOGGLE use prefix arg to toggle behavior"
+  :group 'citar
+  :type 'cons)
+
```
bdarcus, line 200: @salutis - you've previous asked about this. WDYT about this in-progress idea?

citar.el
```diff
@@ -599,9 +702,36 @@ If no function is found, the DEFAULT function is called."
                   (citar-file--normalize-paths
                    citar-bibliography)))
 
-(defun citar--get-value (field entry)
-  "Return the FIELD value for ENTRY."
-  (cdr (assoc-string field entry 'case-fold)))
+;; Data access functions
+
+(cl-defun citar-get-data-entries (&optional &key filter)
```
bdarcus, line 707: I started this as a more general function, with additional variables. It probably makes sense to require the one argument.
```suggestion
(cl-defun citar-get-data-entries (filter)
```

citar.el
```diff
@@ -572,6 +591,90 @@ HISTORY is the 'completing-read' history argument."
          ((string-match "http" resource 0) "Links")
          (t "Library Files")))))
 
+(defun citar--bibliography-files ()
+  "The list of global and local bibliography files."
+  (seq-concatenate 'list citar-bibliography (citar--local-files-to-cache)))
+
+(defun citar--ref-completion-table ()
```
bdarcus, line 598: I don't find this an intuitive name in context.
```suggestion
(defun citar--get-ref-completions ()
```

citar.el
```diff
@@ -440,49 +477,31 @@ FILTER: if non-nil, should be a predicate function taking
...
-               (push choice notfound)
-               nil))
-           (if (listp chosen) chosen (list chosen)))))
-    (when notfound
-      (message "Keys not found: %s" (mapconcat #'identity notfound "; ")))
-    (if multiple keyentries (car keyentries))))
-
-(cl-defun citar-select-refs (&optional &key rebuild-cache filter)
+                                    nil nil nil 'citar-history citar-presets nil))))
+    ;; Return a list of keys regardless of 1 or many
```
bdarcus, line 490: To me, this is a key (no pun intended!) question.

bdarcus:
A thought: right now, this code handles updating for bib files. And it also elegantly supports plugging difference notes systems, like org-roam.

But it doesn't do both. If using org-roam, which automatically updates it's database, I see no way currently to update the citar completions.

I think it's why I had add a notes cache, but even that alone probably wouldn't solve that.

citar.el
aikrahguzar: I am pretty sure I didn't write this code but it is looking up the key given the candidate string. So I think the adapted version would something like
```emacs-lisp
(defun citar--reference-transformer (type candidate)
  "Look up key for a citar-reference TYPE and CANDIDATE."
  (with-current-buffer (embark--target-buffer) (cons type (citar-lookup-key-for-candidate candidate))))
```
I don't know enough about the changes here to know exactly what should be `citar-lookup-key-for-candidate` probably a lookup in one of the hashes.

bdarcus: > I am pretty sure I didn't write this code  ...

OK; sorry for the distraction :-)

 @roshanshariff  may know then.

It would help to clarify the docstring here. Is this for the minibuffer, or the buffer?

I was assuming the latter, but I guess it doesn/t matter. In either case, the code should just use the key, which we already have.

bdarcus: Actually, the problem is specific to `citar-open-entry`, when accessed from `embark-act`.

aikrahguzar: > With both this and your example, I get a "key 'nil' not found" error, which hints to me at a `gethash` not returning a result. 

Is there error in `citar--reference-transformer` or some other function? What stack trace do you get if you try after `toggle-debug-on-error`? My guess would be that this is some other function causing the problem.

bdarcus: I have debug on error turned on, but it doesn't return a stack trace unfortunately.

aikrahguzar: > Actually, the problem is specific to `citar-open-entry`, when accessed from `embark-act`.

This is the problem then
```elsip
(key (car key-entry))
```
This is a binding in the `while-let` inside `citar-open-entry` which expects the key-entry and not just the key.

bdarcus: That's what I was thinking but that also doesn't fix it!

Time for a reboot!

citar.el
```diff
@@ -572,6 +591,90 @@ HISTORY is the 'completing-read' history argument."
          ((string-match "http" resource 0) "Links")
          (t "Library Files")))))
 
+(defun citar--bibliography-files ()
+  "The list of global and local bibliography files."
+  (seq-concatenate 'list citar-bibliography (citar--local-files-to-cache)))
```
bdarcus, line 596:
```suggestion
(defun citar--bibliography-files ()
  "The list of global and local bibliography files."
  (mapcar
   (lambda (file)
     (file-truename file))
   (seq-concatenate 'list citar-bibliography (citar--local-files-to-cache))))
```

roshanshariff:
I actually think we can add the notes/files data fast enough without a cache, which would simplify matters greatly. I think that should be our first preference unless there's a compelling feature we can't get without a cache.

I think we really only need a cache per bibliography file, shared among all the buffers. Ideally, all the processing needed to generate completion candidates (including adding the has-note and has-file metadata) can be done very quickly, and we shouldn't need a cache for that.

I will update with more details after some experimentation.

bdarcus:
> I actually think we can add the notes/files data fast enough without a cache, which would simplify matters greatly. I think that should be our first preference unless there's a compelling feature we can't get without a cache.

If we're just talking about https://github.com/emacs-citar/citar/pull/628#issuecomment-1153110077, I guess that would solve the problem I raised there. 

But creating the completion candidates is still the slowest operation (turning off the completion cache is noticeable, even with my not very large library), and the completion cache is brain-dead simple, so I'll be curious what you find in your experiments.

roshanshariff:
Even if we can invalidate a cache from outside, it's still inefficient to regenerate the cache repeatedly for every buffer every time any metadata changes. A cache only goes so far, and if it's invalidated too often you'll lose the benefits of caching.

Currently, generating the completion candidates is doing a lot of unnecessary work in the template expansion. I think it can be boiled down to just concatenating a few strings, which can in fact be done as-needed. The trick is to do most of the string formatting ahead of time per-bibliography, and only redo it whenever the bibliography changes.

Then the only work that needs to be done to generate the completion candidates is to gather the list of bibliographies, get their associated formatted strings, concatenate them (possibly adjusting for the frame width), and add the has-note/has-file metadata. This last step (I believe) can be done very quickly.

bdarcus:
OIC.

I guess this also goes back to the earlier discussion of as-needed display formatting, which completing-read per doesn't support yet, but its `group-function` does.

roshanshariff:
I don't think it's a matter of just `completing-read` supporting as-needed formatting. You pretty much need all the strings in memory if you're going to search them. You may be displaying only a few strings at a time, but as soon as you enter a single character/tab-complete you need to search all the strings in the completion table.

The only thing you can do at display-time is to add non-searchable but visible metadata (like Marginalia does). This doesn't help Citar much because we pretty much want the whole completion string to be searchable, so we must necessarily generate all the completion strings every time we call `completing-read`. The only thing that is "ornamental" is the column-width truncation that hides text when it's too long to fit on one line, and we can just do that with the completion strings when we generate them (they won't reformat if you resize the Emacs frame while completion is running, but that's a small tradeoff).

> In the interim, I guess you're thinking, we can just stash the primary completion string in an entry field while parsing the bibliography?

I was thinking of another hash table in the cache with the display strings, just to avoid messing with the entry too much. Otherwise, it gets hard to keep track of all the non-standard bibliography fields we're adding and they become a de facto part of our public API, which means we can't change implementation details without breaking external code.

bdarcus:
OIC; so basically moving the hash that is now the completion cache into the bibliography cache?

roshanshariff:
Yeah, exactly. The structure of the cache is actually not that complex, and even that complexity is localized in a few functions.

And yes, it is optimized for that use case, but there's no disadvantage to having a large number of small bibliographies either. It's just that all those bibliographies will be stored in a central cache, but fetching things from that cache is extremely efficient so it doesn't really matter.

bdarcus:
> The structure of the cache is actually not that complex ...

I was realizing after I wrote that something like this would probably work to represent a file, and it's indeed simple.

```elisp
("/home/test/bib/test.bib" 
   :data #<hash-table equal 1/65 0x1b5b1e9> 
   :completions #<hash-table equal 0/65 0x208978f> 
   :checksm "1234" 
   :buffers nil)
```

roshanshariff:
Yeah, that's pretty much what I have now, but defined using a `cl-defstruct` so you get accessor functions, documentation, etc.

```elisp
(cl-defstruct (citar--bibliography
               (:constructor citar--make-bibliography (filename))
               (:copier nil))
  "Cached bibliography file."
  (filename
   nil
   :read-only t
   :documentation
   "True filename of a bibliography, as returned by `file-truename`.")
  (hash
   nil
   :documentation
   "Hash of the file's contents, as returned by `buffer-hash`.")
  (buffers
   nil
   :documentation
   "List of buffers that require this bibliography.")
  (entries
   (make-hash-table :test 'equal)
   :documentation
   "Hash table mapping citation keys to bibliography entries,
   as returned by `parsebib-parse`.")
  (formatted
   (make-hash-table :test 'equal)
   :documentation
   "Formatted strings used to display bibliography entries."))
```

bdarcus:
> but defined using a `cl-defstruct` so you get accessor functions, documentation, etc.

Oh, good; hadn't thought of that, but I like it a lot!

I had just run into `cl-defstruct` for the first time recently when looking through the org-roam codebase.

bdarcus:
This is a totally minor thing, but do you know if there's a way to get it so that marginalia actually shows the documentation strings?

image

Is it worth asking Daniel and Omar about? I don't see anything about it in that issue tracker.

roshanshariff:
It looks like `cl-defstruct` always makes the first line of the documentation be a generic string like what you see in that screenshot. The `:documentation` string you supply ends up on the next line, and you can see it if you use `describe-function` etc. It's slightly annoying, but I think you'd need to modify `cl-defstruct` itself to fix it. The first line of the documentation string is treated specially throughout Emacs, so I doubt it would be easy/desirable to hack around that in Marginalia.

bdarcus:
Ah yeah; that makes sense.

BTW, another minor thing we can control ...

As I'm reading the `cl-defstruct` docs, we may as well include a `:type` property on each definition?

roshanshariff:
I was looking at that, and it looks like that's mainly for compatibility with Common Lisp? It's ignored in Elisp, and in fact there isn't really a good vocabulary for naming types in Emacs.

bdarcus:
This has some good tips on `cl-defstruct` etc:

https://nullprogram.com/blog/2018/02/14/

bdarcus:
FWIW, I just pushed a change here, but it was only adding the new citar-capf.el file I just merged to main.

Vidianos-Giannitsis:
From my perspective, this idea seems pretty good. If I understand this correctly, this makes bibliography into a `cl-defstruct` which should make it much easier to then access each separate object of it. The `org-roam-node` which is defined this way has a lot of supporting functions (I think they are called methods in CL terminology) which allow you to access each component separately and I think it works very well and is extensible by the user. For example I have defined a method that accesses the buffer of an org-roam-node for use in zetteldesk. Making the citar bibliography like this would then make it much easier to access parts of it I feel.

bdarcus:
Thanks for taking a look!

> If I understand this correctly, this makes bibliography into a `cl-defstruct` which should make it much easier to then access each separate object of it.

It makes the cache a list of such objects, so we have finer-grained control of UI updating, and also, as a consequence the "has-note" indicators in the UI are always up-to-date with the bibliography files.

The higher level functions (like `citar--get-entry`) will stay the same.

Make sense?

Vidianos-Giannitsis:
Yeah I get it. This definitely sounds like a step in the right direction. And by not changing the higher level functions it also doesn't break anybody's workflow which is a good thing (unless they use the low level functions at which point they can figure out what to change).

bdarcus:
@roshanshariff any ETA on when you can open up at least a draft PR?

Even if not all functions are working ATM, I can still weigh on what they are, etc.

roshanshariff:
> ETA on when you can open up at least a draft PR?

Sure, I can do that in a couple of hours. Pretty much all the code is ready and tested, and the performance looks promising. Just need to do a bit of work to hook it up to the existing citar functions so you can use it.

bdarcus:
Closing in favor of #634

PR #629, opened 2022-06-08
Add citar-capf

mclearc:
Lemme know if this is what you had in mind. I didn't make any changes to the README -- I figured you would want to do that.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -922,7 +922,7 @@ FORMAT-STRING."
    (lambda (str _status)
      ;; take completion str and replace with key
      (delete-char (- (length str)))
-     (insert (cadr (assoc str candidates))))
+     (insert (gethash str candidates)))
    :exclusive 'no)
   "Completion extra properties for `citar-capf'.")
 
--- a/citar.el
+++ b/citar.el
@@ -914,6 +914,71 @@ FORMAT-STRING."
      (let ((field-names (split-string raw-field "[ ]+")))
        (citar--display-value field-names entry)))))
 
+;;; Completion at Point Function
+
+(defvar citar-capf--extra-properties
+  (list
+   :exit-function
+   (lambda (str _status)
+     ;; take completion str and replace with key
+     (delete-char (- (length str)))
+     (insert (cadr (assoc str candidates))))
+   :exclusive 'no)
+  "Completion extra properties for `citar-capf'.")
+
+;;;###autoload
+(defun citar-capf ()
+  "Complete citation key at point for org, markdown, or latex.
+Parsing by citar is of bibtex, biblatex and csl json files, and
+supports multiple different files. Please see citar's
+documentation for further info."
+  (let ((element (org-element-at-point))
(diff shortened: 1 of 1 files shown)


citar.el
```diff
@@ -914,6 +914,71 @@ FORMAT-STRING."
...
 
+;;; Completion at Point Function
+
+(defvar citar-capf--extra-properties
+  (list
+   :exit-function
+   (lambda (str _status)
+     ;; take completion str and replace with key
+     (delete-char (- (length str)))
+     (insert (cadr (assoc str candidates))))
```
bdarcus, line 925: I'm looking at this one my phone, but is this right?

Not `gethash` here?

bdarcus:
Merged; thanks again!

minad:
> The main one from a design POV is that the integrated function relies on different packages for different major modes.
> But citar clearly separate those into separate files, that are only auto-loaded as needed.
> WDYT about breaking this apart, and using the "major mode adapter" system?

Indeed, there are issues. But I think going via an extra adapter system is a bit overengineered, which I would avoid. The function just needs a few tweaks here and there. I'll add comments.

bdarcus:
> I think going via an extra adapter system is a bit overengineered.

It wouldn't be an extra "system"; it would make use of the existing system.

Like, there's already a "key-at-point" major mode function.

I don't have in mind, to be clear, this is a good idea. It's just an obvious question to me, given that the system was aimed at addressing this sort of issue.

mclearc:
I think the function is such that `org-element-at-point` will only be called when in org mode, and that is the only dependency. So the suggested changes seem the simplest. But totally up to you obviously!

bdarcus:
> think the function is such that `org-element-at-point` will only be called when in org mode, and that is the only dependency. So the suggested changes seem the simplest. But totally up to you obviously!

In that, seems we're fine just declaring the function, per @minad's point.

mclearc:
> > think the function is such that `org-element-at-point` will only be called when in org mode, and that is the only dependency. So the suggested changes seem the simplest. But totally up to you obviously!
> 
> In that case, seems we're fine just declaring the function, per @minad's point.

Yeah I had this in the original package -- just forgot to include it in the pull request....whoops. :)

bdarcus:
I force-pushed what should mostly fix the issues, but I think the extra-properties part might be wrong.

At least, it won't work (though gives no errors) for me. But my doom/corfu install is messed up, as you know, so ...

citar.el
```diff
@@ -914,6 +914,71 @@ FORMAT-STRING."
...
+               (or (eq ?@ (char-before))
+                   (looking-back capf-citar-markdown-regexp
+                                 (line-beginning-position))))))
+
+      ;; Get and insert candidate
+      (let* ((candidates (citar--ref-completion-table))
+             (begin (save-excursion (backward-word) (point)))
+             (end (point)))
+        (list begin end candidates
+              citar-capf--extra-properties
```
minad, line 978: The properties are not specified correctly. Take a look at cape!

bdarcus:
See #631 for a followup PR to finalize this.

PR #631, opened 2022-06-09
citar-capf follow-up

bdarcus:
Followup to #629.

I decided to base this on main, as it's trivial to adjust it for the new branch, and I don't want to run into git history problems.

So I think just need to settle the exit function on the defvar. Maybe need to wrap it in a lambda?

### DIFF
--- a/citar-capf.el
+++ b/citar-capf.el
@@ -1,0 +1,91 @@
+;;; citar-capf.el --- Description -*- lexical-binding: t; -*-
+;;
+;; Copyright (C) 2022 Bruce D'Arcus
+;;
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; Commentary:
+;;
+;;  Description
+;;
+;;  Citation key 'completion-at-point' for org, markdown, or latex.
+;;
+;;  Supports org-cite in org, and pandoc in markdown.
+;;
+;;; Code:
+(require 'citar)
+
+;;;; Utility Vars & Functions
+;; Declare org-element related vars & functions
+(defvar org-element-citation-key-re)
+(defvar org-element-citation-prefix-re)
+(declare-function org-element-at-point "ext:org-element")
+(declare-function org-eement-at-point "ext:org-element")
+(declare-function org-element-property "ext:org-element")
+;; Declare function from citar
+;; (declare-function citar--ref-completion-table "citar") ;; pending cache revisions
+
+;; Define vars for capf
+(defvar citar-capf--candidates (or (citar--get-candidates)
+                                   (user-error "No bibliography set"))
+  "Completion candidates for `citar-capf'.")
+
+(defvar citar-capf--properties
+  (list :exit-function #'citar-capf--exit
+        :exclusive 'no)
+  "Completion extra properties for `citar-capf'.")
+
+;; Define exit function
+(defun citar-capf--exit (str _status)
+  "Return key for STR from CANDIDATES hash."
+  (delete-char (- (length str)))
+  (insert (cadr (assoc str citar-capf--candidates))))
+
+;;;; Citar-Capf
+;;;###autoload
+(defun citar-capf ()
+  "Citation key 'completion-at-point' for org, markdown, or latex."
+  (let ((citar-capf-latex-regexp
+         "\\(?:cite\\(?:\\(?:[pt]\\*\\|[pt]\\)?{\\)\\)\\([[:alnum:]_-]*,\\)*\\([[:alnum:]_-]*\\)")
+        (citar-capf-markdown-regexp
+         (concat "-?@"                         ; @ preceded by optional -
+                 "\\(?:"
+                 "{\\(?1:.*?\\)}"              ; brace-delimited key
+                 "\\|"
+                 "\\(?1:[[:alnum:]_][[:alnum:]]*\\(?:[:.#$%&+?<>~/-][[:alnum:]]+\\)*\\)"
+                 "\\)")))
+    ;; only list candidates in certain contexts
+    (when
+        ;; conditional recognition of citation key by mode
+        (cond
+         ;; latex-mode
+         ((derived-mode-p 'latex-mode)
+          (looking-back citar-capf-latex-regexp 2))
+         ;; org-mode
+         ((and (derived-mode-p 'org-mode)
+               (let ((element (org-element-at-point)))
+                 (or (eq 'citation (org-element-type (org-element-context element)))
+                     (and (or (eq ?@ (char-before))
+                              (looking-back org-element-citation-key-re
+                                            (line-beginning-position)))
+                          (let ((origin (point)))
+                            (save-excursion
+                              (and (re-search-backward org-element-citation-prefix-re
+                                                       (org-element-property
+                                                        :begin element)
+                                                       t)
+                                   (not (search-forward "]" origin t))))))))))
+         ;; markdown-mode
+         ((and (derived-mode-p 'markdown-mode)
+               (or (eq ?@ (char-before))
+                   (looking-back citar-capf-markdown-regexp
+                                 (line-beginning-position))))))
+      ;; Get and insert candidate
+      (let ((begin (save-excursion (backward-word) (point)))
+            (end (point)))
+        (append (list begin end citar-capf--candidates)
+                citar-capf--properties)))))
+
+(provide 'citar-capf)
+;;; citar-capf.el ends here


mclearc:
Yeah - sorry have been super busy. Will try it tonight if I can. 

--
Colin McLear
Associate Professor
Department of Philosophy 
University of Nebraska–Lincoln
colinmclear.net

bdarcus:
No huge rush. I put it in a separate file so no need to stress about when
to merge it to main..

citar-capf.el
```diff
@@ -0,0 +1,82 @@
...
+(declare-function org-element-at-point "ext:org-element")
+(declare-function org-element-property "ext:org-element")
+(declare-function citar--ref-completion-table "citar")
+
+(defvar citar-capf--properties
+  (list :exit-function #'citar-capf--exit
+        :exclusive 'no)
+  "Completion extra properties for `citar-capf'.")
+
+(defun citar-capf--exit (str _status candidates)
```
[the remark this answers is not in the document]
bdarcus: Actually, @mclearc, I just made a change on the development branch that will solve this.

https://github.com/emacs-citar/citar/blob/b884e1666f3f5b48514546fc3a4fb3c9578b00f7/citar.el#L572

https://github.com/emacs-citar/citar/blob/b884e1666f3f5b48514546fc3a4fb3c9578b00f7/citar.el#L575-L577

mclearc:
FYI -- this works against main for me, though I had to change `citar--get-candidates` to `citar--ref-completion-table`.

bdarcus:
> Is there a good drop in replacement for `citar--get-candidates`?

In effect, the new `citar--parse-bibliography` (adapted from `-org-cite-basic--parse-bibliography`) does the same thing I think.

I need to think about that a little more though, and also whether we should keep the same symbol name, when the underlying implementation changes.

E.g. a lot of citar will work the same, but the underlying change is drastic, and there are places like this that it will be exposed (when someone has a function that expects a list and they get a hash table instead; also, data is stored in a separate cache).

bdarcus:
> FYI -- this works against main for me, though I had to change `citar--get-candidates` to `citar--ref-completion-table`.

Actually, here you should use `citar--completion-cache`.

mclearc:
`citar--completion-cache` displays fine but still exits giving the whole table line rather than just the key.

bdarcus:
> `citar--completion-cache` displays fine but still exits giving the whole table line rather than just the key.

Even if you change that key line to use `gethash`?

mclearc:
Yeah -- with `      (let* ((candidates citar--completion-cache)` and `(defun citar-capf--exit (str _status candidates)
  "Return key for STR from CANDIDATES hash."
  (insert (gethash str candidates)))` I'm still not getting just the key on exit. 

I have to run to meetings. Will try this more later!

bdarcus:
Sounds good. As I said, there's no rush, and taking our time, we can get both #628 and this right.

bdarcus:
> > FYI -- this works against main for me, though I had to change `citar--get-candidates` to `citar--ref-completion-table`.
> 
> Actually, here I _think_ you should use `citar--completion-cache`.
> 
> I may reintroduce `citar--get-candidates` as a wrapper for this, in which case I can add error handling.

Thinking about this some more, and qualifying this may be subject to change ...

The key place to look into the branch code is here:

https://github.com/emacs-citar/citar/blob/a1d00859d9079d723c090daa55679f9c535ab751/citar.el#L483-L489

And my current thinking is to deprecate `citar--get-candidates`.

The key change in that branch is splitting what was a single data/completion cache into two. So it makes sense to make that clear in the function names.

But should be settled in the next couple of days in any case.

mclearc:
I'm not sure I entirely understand the above. But the current problem is that the capf presents each line in the table as an entire string (so the entry isn't simply the citekey, but the whole line including author, title, etc.). Currently the exit function is set to insert this entire line as a string. I don't see how to get it to insert only the citekey (which one can do, for example, with `(citar--all-keys)` as part of the exit-function. There is likely a simply solution here, I just don't understand how gethash could recover just part of the string.

bdarcus:
> ...  the current problem is that the capf presents each line in the table as an entire string (so the entry isn't simply the citekey, but the whole line including author, title, etc.). Currently the exit function is set to insert this entire line as a string. 

This I don't understand. This line is the same as you had earlier, and it says to that that string, and lookup its match in the `candidates`.

https://github.com/emacs-citar/citar/blob/7492b8e178d255d8324050489819643a708fa6aa/citar-capf.el#L34

That works on main; right?

The hash version would be the same idea:

```elisp
 (insert (gethash str candidates)))
```

> I don't see how to get it to insert only the citekey (which one can do, for example, with `(citar--all-keys)` as part of the exit-function. There is likely a simply solution here, I just don't understand how gethash could recover just part of the string.

My question was really about the argument passing.

mclearc:
Ok -- I think I've found the problem (at least the one I was having). The insertion part of the capf needs to put the various lists together with `append` like so:

```emacs-lisp
      (let* ((candidates (or (citar--ref-completion-table)
                             (user-error "No bibliography set")))
             (begin (save-excursion (backward-word) (point)))
             (end (point)))
        (append (list begin end candidates)
                citar-capf--properties))))
```
And the exit function needs to look like so:

```emacs-lisp
(defun citar-capf--exit (str _status)
  "Return key for STR from CANDIDATES hash."
  (let* ((candidates (or (citar--ref-completion-table)
                         (user-error "No bibliography set"))))
    (delete-char (- (length str)))
    (insert (gethash str candidates))))
```

What I don't like about this is `candidates` is defined twice in the two let forms. It would be nice to eliminate the redundancy. But I think this otherwise works against the api-refactor branch. I'll be able to do more testing later today. Any thoughts?

bdarcus:
> What I don't like about this is `candidates` is defined twice in the two let forms. It would be nice to eliminate the redundancy. But I think this otherwise works against the api-refactor branch. I'll be able to do more testing later today. Any thoughts?

I agree it's a little awkward.

OTOH, it's referencing a variable; the cache.

@minad is there an elegant way to pass the candidates to the exit function here, or is what @mclearc suggesting here reasonable?

https://github.com/emacs-citar/citar/pull/631#issuecomment-1153262593

minad:
Looks okay. You could also use a closure.

citar-capf.el
```diff
@@ -0,0 +1,82 @@
...
+          ((and (derived-mode-p 'markdown-mode)
+                (or (eq ?@ (char-before))
+                    (looking-back citar-capf-markdown-regexp
+                                  (line-beginning-position))))))
+      ;; Get and insert candidate
+      (let* ((candidates (citar--get-candidates))
+             (begin (save-excursion (backward-word) (point)))
+             (end (point)))
+        (list begin end candidates
+              citar-capf--properties)))))
```
bdarcus, line 79:
```suggestion
        (append (list begin end candidates)
                citar-capf--properties))))
```

@mclearc line comment suggestions like this are convenient with PRs.

citar-capf.el
```diff
@@ -0,0 +1,82 @@
...
+(declare-function citar--ref-completion-table "citar")
+
+(defvar citar-capf--properties
+  (list :exit-function #'citar-capf--exit
+        :exclusive 'no)
+  "Completion extra properties for `citar-capf'.")
+
+(defun citar-capf--exit (str _status candidates)
+  "Return key for STR from CANDIDATES hash."
+  (insert (cadr (assoc str candidates))))
```
bdarcus, line 34:
```suggestion
  (let* ((candidates (or (citar--ref-completion-table)
                         (user-error "No bibliography set"))))
  (insert (cadr (assoc str candidates))))
```

bdarcus:
> Looks okay. You could also use a closure.

I guess in the end, @mclearc, it's an aesthetic question, and I don't have a strong opinion.

We can think on it a few days, while we finish #628.

mclearc:
> > Looks okay. You could also use a closure.
> 
> I guess in the end, @mclearc, it's an aesthetic question, and I don't have a strong opinion.
> 
> We can think on it a few days, while we finish #628.

I'll look at it a bit more then I'll just open a pull request with the relevant changes (if that sounds good)

mclearc:
@bdarcus  -- is the easiest thing to open a whole new pull request? Or is there a way for you to update this pull request with the commit referenced above? Sorry for the question, but I'm just not that familiar with the github workflow.

bdarcus:
> @bdarcus -- is the easiest thing to open a whole new pull request? Or is there a way for you to update this pull request with the commit referenced above? Sorry for the question, but I'm just not that familiar with the github workflow.

the line comment suggestions are great for little changes, because they're easy for you, and also for me. I just clicked to commit them, so the branch now reflects those changes.

I would use that for small changes.

citar-capf.el
```diff
@@ -0,0 +1,84 @@
...
+(declare-function org-element-at-point "ext:org-element")
+(declare-function org-element-property "ext:org-element")
+(declare-function citar--ref-completion-table "citar")
+
+(defvar citar-capf--properties
+  (list :exit-function #'citar-capf--exit
+        :exclusive 'no)
+  "Completion extra properties for `citar-capf'.")
+
+(defun citar-capf--exit (str _status candidates)
```
mclearc, line 32: https://github.com/mclearc/citar/commit/e0f7bf379e28855256b7c0ac6481a75758132d54#diff-25e85872b3080fbaf13121056eb76bfae3c5bcafabd34fd28463bc0e7f4e08bbR38-R42

bdarcus:
Oh, I missed your commit.

I can take a look in the morning, and maybe just cherry pick it.

PR #633, opened 2022-06-13
Revise citar-capf; followup to #631

mclearc:
- Make sure exit function works
- Add some notes & vars


mclearc:
Sorry -- I'm sure doing line comments is easier for someone who knows what they are doing on github, but I can't really spend any more time on this so do with it what you will.

Issue #449, opened 2021-11-29
Prefill e.g. "has: notes" for "citar-open-notes"


salutis:
More often than not, when I run `citar-open-notes`, I type `has:notes`. Could this be set as the default input? From what I understand, `citar-presets` _almost_ does what I want but not really. I would like the default input prefilled for all _specialized_ `citar-open` functions (except `citar-open-entry`, obviously). Or perhaps I have a bad idea?

bdarcus:
In the past, we had that, using basically a hack.

So I removed it.

Do note that you can just type `:n`.

bdarcus:
The other possibility is we could add a defcustom to pre-filter candidates.

But the downside would be that one would not be able to back out of that and access the full bib list. So it would need to be nil be default.

salutis:
> [...], using basically a hack.
> 
> But the hack itself was a problem, and there aren't any great, pain free, alternatives.

Does this mean that there exists no API to prefill that input field at the bottom of the screen?

bdarcus:
https://www.gnu.org/software/emacs/manual/html_node/elisp/Initial-Input.html

salutis:
That makes sense. How about we have a variable that says "show only relevant stuff" for these specialized functions?  For example, when I call `citar-open-notes`, I want to see stuff with notes, because I want to "open notes".

salutis:
Maybe I should start using `citar-open` instead. I remember that you wanted to remove the specialized functions. At this point, what purpose do they serve?

salutis:
Done.

`(global-set-key (kbd "C-c b") 'citar-open)`

In nobody uses the specialized functions (that do not even filter the items), perhaps we should remove them, as you suggested in the past.

bdarcus:
> In nobody uses the specialized functions (that do not even filter the items), perhaps we should remove them, as you suggested in the past.

Do you use embark with citar?

As oantolin has made the point, it's best not to think of this too rigidly, given all these commands are contextually available. All the command does in that context is define the default behavior (e.g. opening a note).

So, for example, you can just choose whatever is your preferred entry point, and use that, and use `embark-act` to access different actions.

You can also give that command an alias if you prefer, like `salutis/cite`.

If that strategy, it would likely make sense that it be a command that always needs the full candidate list; like `citar-insert-citation`.

Really, this UI is designed to work best with Emacs 28 and embark.

[...]

Issue #283, opened 2021-09-13
Source files only opening for some bibtex entries.


anaxonda:
**Describe the bug**
When running either `bibtex-actions-open` or `bibtex-actions-open-library-files`, the source files only open for some entries.  For bibtex entry types described below I get the error `No file(s) found for ("name of key")`, or a "Related resources" minibuffer  that has no entries.
**Expected behavior**
Either source file should open, or selected library file should open based off of the bibtex `file` field. Opening files works fine in helm-bibtex, so I guess its not a bibtex-completion issue.

**Emacs version:**
28.0.5
**Additional context**
I have two bibliography files, one generated by Zotero and one by Calibre. 

Zotero bibtex entries with multiple files listed in the `file` field do not open. Entries with single files listed work fine. Here is an example of a bibtex entry with multiple files listed that does not work:
```
@article{kaneReconsideringFireEcology2020,
  title = {Reconsidering the Fire Ecology of the Iconic {{American}} Chestnut},
  author = {Kane, Jeffrey M. and Varner, J. Morgan and Stambaugh, Michael C. and Saunders, Michael R.},
  date = {2020},
  journaltitle = {Ecosphere},
  volume = {11},
  number = {10},
  pages = {e03267},
  issn = {2150-8925},
  doi = {10.1002/ecs2.3267},
  url = {https://esajournals.onlinelibrary.wiley.com/doi/abs/10.1002/ecs2.3267},
  urldate = {2021-02-20},
  langid = {english},
  keywords = {American chestnut,Castanea dentata,fire ecology,fire history,fire-adaptive traits,foundation species,restoration},
  annotation = {\_eprint: https://esajournals.onlinelibrary.wiley.com/doi/pdf/10.1002/ecs2.3267},
  file = {../../Zotero/storage/IXT6SZDI/Kane et al. - 2020 - Reconsidering the fire ecology of the iconic Ameri.pdf;../../Zotero/storage/M4T63FTP/ecs2.html}
```
The other case is bibtex entries generated by calibre (of which none work):
```
@book{ ThompsonDArcyWentworth567,
    author_sort = "Wentworth, Thompson D'Arcy",
    author = "Thompson D'Arcy Wentworth",
    cover = "/home/Documents/Library/Thompson D'Arcy Wentworth/On Growth and Form (567)/cover.jpg",
    formats = "pdf",
[...]

PR #634, opened 2022-06-13
New caching and formatting functionality

roshanshariff:
This pull request continues the work in #628.

### Formatting

This is in `citar-format.el`. Formatting references is now done in three stages:

1. In the first stage, all required entry fields are looked up and formatted into strings.
2. In the second stage, these strings are joined together after accounting for the current screen width.
3. (In `citar--ref-completion-table`) Metadata like the "has:notes" and "has:files" tags are looked up and added.

The first step is done when the bibliography is parsed, while the second and third are fast enough to do for every completion session.

Behind the scenes, there is new code to parse format strings in the existing format into a new, efficient, list-based format. All the functions now operate on this parsed representation internally. When formatting or pre-formatting entire bibliographies, this parsing only needs to be done once.

### Caching

This functionality is in `citar-cache.el`. Instead of buffer-local and global caches, we now maintain a centralized cache which holds records for each bibliography file in use. This cache also keeps track of which buffers are using each bibliography, so that the cached files can be discarded when they are no longer used by any buffer.

The cache also holds "pre-formatted" candidate strings (i.e. the result of step 1 above). These strings do not depend on any transient information like whether each entry has files/notes or the screen width.

## Remaining work

- [ ] Allow the cache to be refreshed manually using `citar-refresh`. Currently the cache is never updated after the first time each bibliography is parsed.
- [ ] Integrate the file-notify functionality so the cache can be updated whenever the underlying bibliography file changes.
- [ ] Clean up doc strings.


bdarcus:
I was going to compile it first before proceeding, but I get these errors:

```
Compiling file /home/bruce/Code/citar/citar.el at Mon Jun 13 18:03:01 2022

In citar-get-data-entries:
citar.el:673:27: Warning: reference to free variable
    ‘citar--bibliography-cache’

In end of data:
citar.el:841:16: Warning: the function ‘citar--parse-bibliography’ is not
    known to be defined.
```

roshanshariff:
Right, as it stands the cache never gets updated at all, even when calling `citar-select-ref`. I should change this so it re-parses the bibliography on calling `citar-select-ref` if the file has changed. Then we'd hook it up to filenotify so that the update happens automatically and there isn't a big wait on the next completion session.

bdarcus:
First glance looks good.

When you think it's ready to merge to the other branch, mark it ready for review?

I'll test more a bit later.

bdarcus:
Or if easier, I suppose you could rebase on main, and I can close the other.

Up to you.

roshanshariff:
We don't need file notify, but it's nice to have. File-notify can also report spurious change events in some cases, so the checksum guards against unnecessary updates. And yes, we can turn on file-notify by default now. But we should probably have a fallback for people who can't use file-notify (maybe they're on a network file system or something?)

We'll have to fine-tune the details on this and add appropriate customization options. For example, I can imagine somebody might have a bibliography on a tramp connection and they don't want emacs to try to update it every time they start completion. Even just checking the file size and time could involve a long round-trip they'd rather not have.

bdarcus:
Can you also take a look at the new capf and make sure it works?

I have a problem testing it.

Should just require a slight tweak.

bdarcus:
I'm going to edit your TODO directly, to keep it all in one place.

bdarcus:
Seems that `citar--ref-completion-tables` now is a bit more than 5.5x faster than `citar--format-candidates`:

```elisp
ELISP> (benchmark-run-compiled 100 (citar--ref-completion-table))
(12.98203636 43 6.759782593000001)
```

citar.el
```diff
@@ -1160,26 +1005,23 @@ With prefix, rebuild the cache before offering candidates."
...
-                'list
-                citar-bibliography
-                (citar--local-files-to-cache))))
-    (bibtex-search-entry key t nil t)))
+  (interactive (list (citar-select-ref)))
+  (when-let ((bibtex-files
+              (seq-concatenate
+               'list
+               citar-bibliography
+               (citar--local-files-to-cache))))
```
bdarcus, line 1016:
```suggestion
  (when-let ((bibtex-files (citar--bibliography-files))
```

citar.el
```diff
@@ -572,6 +615,35 @@ HISTORY is the 'completing-read' history argument."
...
+In this hash table, keys are a strings with author, date, and
+title of the reference.  Values are the cite keys.
+Return nil if there are no bibliography files or no entries."
+  ;; Populate bibliography cache.
+  (when-let ((bibs (citar--bibliographies)))
+    (let* ((entries (citar-cache--entries bibs))
+           (preformatted (citar-cache--preformatted bibs))
+           (fmtstr (citar--get-template 'completion))
+           (hasnotep (citar-has-note))
+           (hasfilep (citar-has-file))
```
bdarcus, line 629: I mentioned on #628 that I think we should rename this to `citar-has-library-file`; for clarity.

citar.el
```diff
@@ -179,6 +193,25 @@ All functions that match a particular field are run in order."
   :type '(alist :key-type   (choice (const t) (repeat string))
                 :value-type function))
 
+(defcustom citar-prefilter-entries '(nil . t)
```
bdarcus, line 196: @roshanshariff WDYT of this idea of mine?

This would address #449. Earlier I had closed it because I couldn't figure out a flexible way to do this, but this seems a good approach?

bdarcus:
@roshanshariff one of the items I put on the todo is testing with very large libraries, including a link to an online data with a lot of them.

Try this 6000+ reference/3.2 MB file.

http://www-alg.ist.hokudai.ac.jp/~thomas/COLTBIB/colt.bib.gz

There is a noticeable pause, akin to having no cache on my ~1000 item bib.

I guess we'll need to decide what kind of performance user should expect, with what size libraries?

I never did test the previous implementation with this size of file though.

bdarcus:
> We don't need file notify, but it's nice to have?

Why?

citar-cache.el
```diff
@@ -0,0 +1,239 @@
...
+
+Unless FORCE is non-nil, the file is re-read only if it has been
+modified since the last time BIB was updated."
+  (let* ((filename (citar-cache--bibliography-filename bib))
+         (entries (citar-cache--bibliography-entries bib))
+         (newhash (with-temp-buffer
+                    (insert-file-contents filename)
+                    (buffer-hash))))
+    ;; TODO Also check file size and modification time before hashing?
+    ;; See `file-has-changed-p` in emacs 29, or `org-file-has-changed-p`
```
bdarcus, line 191: I didn't realize this was a new function, but below is the source code. I think the org version wraps this, with some backward compatibility stuff.

So it doesn't look at a checksum, but at file size and modification time.

```elisp
(defun file-has-changed-p (file &optional tag)
  "Return non-nil if FILE has changed.
The size and modification time of FILE are compared to the size
and modification time of the same FILE during a previous
invocation of `file-has-changed-p'.  Thus, the first invocation
of `file-has-changed-p' always returns non-nil when FILE exists.
The optional argument TAG, which must be a symbol, can be used to
limit the comparison to invocations with identical tags; it can be
the symbol of the calling function, for example."
  (let* ((file (directory-file-name (expand-file-name file)))
         (remote-file-name-inhibit-cache t)
         (fileattr (file-attributes file 'integer))
	 (attr (and fileattr
                    (cons (file-attribute-size fileattr)
		          (file-attribute-modification-time fileattr))))
	 (sym (concat (symbol-name tag) "@" file))
	 (cachedattr (gethash sym file-has-changed-p--hash-table)))
     (when (not (equal attr cachedattr))
       (puthash sym attr file-has-changed-p--hash-table))))
```

roshanshariff:
> But citar-select-ref is slower, and it is noticeable with large libraries.

Thanks for that big file; I will test with it on top of own ~1000 key bibliography. I'm implementing some low-hanging fruit in terms of improving performance that should help.

However, you're right that no matter how hard we try, for large enough libraries performance will be a concern. I'd prefer not to add another cache, however, with all the accompanying headaches. Instead, I think we should expose a few options that gradually make the results less "live" but improve performance. The idea would be to maintain the current architecture, but have the "preformat" stage do more and more work.

So, for example, we currently adapt the completions to the current frame width. We could have a defcustom to specify a fixed width for completion candidates. Then the preformat can produce a single string of that width instead of a list of strings, and there's little work to be done at display-time.

An even more extreme option would make the preformat include the has:note and has:file markers, and they would only be updated when the bibliography changes or you run `citar-refresh`.

We'd have these options disabled by default, but document them in the Readme as "options for large bibliography files". What do you think? If you enabled all these options, I think you'd get pretty much the performance of having a candidates cache (but only a few users will need to go that far).

> Why?

Without file-notify, if the bibliography changes, you will have to wait for it to be parsed the next time you call completion. People might prefer to have Emacs re-parse it automatically (perhaps when it's idle) so there's no wait.

> Can you also take a look at the new capf and make sure it works?

I modified `citar-capf.el` to account for the new API, and when I run `(citar-capf)` manually in a Markdown or Org buffer it seems to return reasonable results. But I still can't get the completion to work for some reason. I've added `citar-capf` to `completion-at-point-functions`, and completion does work normally in other buffers. I'll investigate this further when I have some time. In the meantime I've pushed the small changes needed.

bdarcus:
> I think we should expose a few options that gradually make the results less "live" but improve performance. The idea would be to maintain the current architecture, but have the "preformat" stage do more and more work.

In any case, I do like the idea of incremental iteration (certainly not doing everything in this PR), in response to real user needs, but to have a clean foundation.

Do you agree this foundation is, indeed, cleaner, and easier to maintain going forward, then the current code?

> So, for example, we currently adapt the completions to the current frame width. We could have a defcustom to specify a fixed width for completion candidates. Then the preformat can produce a single string of that width instead of a list of strings, and there's little work to be done at display-time.
> 
> An even more extreme option would make the preformat include the has:note and has:file markers, and they would only be updated when the bibliography changes or you run `citar-refresh`.
> 
> We'd have these options disabled by default, but document them in the Readme as "options for large bibliography files". What do you think? If you enabled all these options, I think you'd get pretty much the performance of having a candidates cache (but only a few users will need to go that far).

Do you imagine two boolean defcustoms? Or one with choices?

> > Why?
> 
> Without file-notify, if the bibliography changes, you will have to wait for it to be parsed the next time you call completion. People might prefer to have Emacs re-parse it automatically (perhaps when it's idle) so there's no wait.

Right. But then would want to be when idle, or using the async library.
 
> > Can you also take a look at the new capf and make sure it works?
> 
> I modified `citar-capf.el` to account for the new API, and when I run `(citar-capf)` manually in a Markdown or Org buffer it seems to return reasonable results. But I still can't get the completion to work for some reason. I've added `citar-capf` to `completion-at-point-functions`, and completion does work normally in other buffers. I'll investigate this further when I have some time. In the meantime I've pushed the small changes needed.

+1

roshanshariff:
> Do you agree this foundation is, indeed, cleaner, and easier to maintain going forward, than the current code?

Yes, I think all the future improvements I can think of would be easier to build upon this new architecture than what we have now.

As you say, we should respond to real user needs. If we decide that we need better performance on very large files, I'm not opposed to adding new caches if needed. But this should be a last resort.

Sometimes it's easy to avoid improving performance just by adding a cache on top, but the complexity of the system increases rapidly when you do this. And if we do end up adding another cache in the end, it's good if it can be invalidated and reconstructed quickly anyway :)

> Do you imagine two boolean defcustoms? Or one with choices?

I'll have to think about this more, and I think the correct design will become apparent as we go. I need to do some profiling to figure out where the bottlenecks currently are so we're not optimizing the wrong thing.

citar.el
roshanshariff: Hmm, I agree with renaming it, but to me "library file" also seems very vague and generic. How about something like "attachment"? And would the "has:files" tag in the completion string also change? It is currently easy to type at least.

citar.el
roshanshariff: I think it's better to implement a Consult-like narrowing system. I looked at its code, and behind the scenes it maintains a list of predicates that are combined into `minibuffer-completion-predicate`. We'd bind some keys to allow predicates to be enabled and disabled on-the-fly, just like Consult does.

Then we could make `citar-select-ref` activate a certain predicate by default, but you'd always be able to press a key to remove the filter and/or add other filters.

bdarcus: I agree that would be valuable, without us requiring consult.

But this (really the car) is orthogonal, and is just saying whether to pre-filter with `citar-open-notes` and such, as per #449.

In any case, feel free to remove or adapt this at will.

citar.el
bdarcus: > Hmm, I agree with renaming it, but to me "library file" also seems very vague and generic. How about something like "attachment"? 

"attachment" has too much association with emails, I think?

The "library" word would just be to be consistent with the defcustom etc.

But it's not a big deal, particularly since ...

> And would the "has:files" tag in the completion string also change? It is currently easy to type at least.

No; I would keep it as is, for the reason you note.

citar.el
```diff
@@ -834,12 +821,6 @@ repeatedly."
              (propertize (citar--get-template 'suffix) 'face 'citar)))
    (error "No template for \"%s\" - check variable 'citar-templates'" template-name)))
 
-(defun citar--all-keys ()
```
bdarcus: On your question, FYI, this was just a function used in the oc-basic completion table function.

And you added an equivalent to the first one, though without the filter arg.

roshanshariff: Yeah, it occurred to me that in general, to do filtering properly, you have to construct the entire hash table and then remove the entries you don't want. Otherwise there can be complications with multiple bibliographies, where a key in an earlier bibliography overrides a later one. If you naively filter the bibliographies and then merge them (or equivalent) then you can end up filtering out the first instance of that key and not the second, which then incorrectly gets "unshadowed". So you pretty much have to first merge all the bibliographies and _then_ filter.

That being the case, it's probably easier to just have the function that returns all the keys/items as a hash table, and use `map-filter` to filter them:
```
(map-filter (citar-has-note) (citar--get-entries)) ; Returns an alist
```

bdarcus: That particular code won't work though, since `citar--get-entries` returns a hash table.

roshanshariff: The functions in `map.el` do indeed work on hash tables (and other kinds of associative data structures). It's the counterpart of `seq.el`, whose functions work generically on sequences.

bdarcus: BTW, WDYT of making the get functions (at least for the data) public again?

We want people to feel comfortable using them; right?

bdarcus: > The functions in map.el do indeed work on hash tables (and other kinds of associative data structures.

Hmm ... I got an error.

roshanshariff: I just ran it and it works fine, so the error must be something else.

> WDYT of making the get functions (at least for the data) public again?

Yeah, I think these functions are probably going to pretty stable (and should be easy to keep updated if the implementation changes).

bdarcus:
> I'll have to think about this more, and I think the correct design will become apparent as we go.

It might be the defcustom values are user-oriented.

Like the current behavior results in constantly updated related resource indicators, which sacrifices some ultimate performance and responsiveness, while the other optimizations you note sacrifice the former to optimize the latter.

roshanshariff:
Actually, all this might be moot. With some optimization (a little hacky for now), I'm able to get it down to about 800 ms to format 7000 entries, with live has:notes and has:files indicators and adaptive column widths. And I think there are some more avenues for improvement.

Turns out ~70% of the time was being spent in garbage collection, not in string handling...

andersjohansson:
This sounds great! Just to add a user perspective, I currently always use a file exported from Zotero with my full library (ca 6500 entries), so keeping that fast is important for me. The initial parsing always takes 5-10 seconds or something (with json) which is of course unavoidable, but in case significant and slow work with all entries is done at other stages, I think the suggestion of configuring more "static" display could be good. I’ll gladly help testing when you think it is ready enough.

bdarcus:
> The initial parsing always takes 5-10 seconds or something (with json) ...

Did you ever test JSON vs BibTeX?

My hypothesis is JSON would be faster, but I've never tested.

If it is, we might not that in the docs.

> I’ll gladly help testing when you think it is ready enough.

I think once Roshan pushes that commit he's working on it, it should be ready to test this piece.

bdarcus:
> In reality, I could probably clean up my Zotero library, but the time required for that is better spent on research (or Emacs hacking). Another workflow is of course to organize collections in zotero to export bib-files for different projects, but I generally want to be able to cite whatever whenever (also using org-roam for notes and relating various notes) so having the full library available is best for my workflow.

For sure.

FWIW, I use Zotero too. I had in mind a setup where you auto-export a very large file, that contains all references added before X days ago, and then a newer one that contains all references since.

So the latter might change a lot (for example, while you're writing), but the latter not.

Does that make sense?

> I did compare bibtex and json a while back and came to the conclusion that json was faster for the initial parsebib parsing (especially with native json support, from Emacs 27). Don’t have any saved measurements though.

That's good enough. We can recommend people try it.

bdarcus:
> But splitting auto export of into older+newer would require changes to BBT or a new Zotero plugin, wouldn’t it?  If I remember correctly auto export doesn’t work well with saved searches?

If that's the case, it's a bug we should report to BBT.

Because it _should_ work.

image

OTOH, now that I think about it, not sure that would impact this PR, other than to make the export process from BBT faster.

das-s:
If I might chime in on the topic of saved searches: the author of BBT states that it is impossible to export saved searches with the current Zotero versions (has been possible with hacks in v4) [1]. One Zotero developer claims its very expensive to do [2].

- [1] https://github.com/retorquere/zotero-better-bibtex/issues/904
- [2] https://forums.zotero.org/discussion/70534/auto-exporting-saved-searches

citar-cache.el
```diff
@@ -0,0 +1,239 @@
...
+
+
+(defun citar-cache--preformat-bibliography (bib)
+  "Updated pre-formatted strings in BIB."
+  (let* ((entries (citar-cache--bibliography-entries bib))
+         (fmtstr (citar-cache--bibliography-format-string bib))
+         (preformat (citar-format--preformat fmtstr :hide-elided t))
+         (preformatted (citar-cache--bibliography-preformatted bib)))
+    (clrhash preformatted)
+    (maphash (lambda (citekey entry)
```
bdarcus, line 208: Again, just a minor suggestion:
```suggestion
    (map-do (lambda (citekey entry)
```

citar-cache.el
```diff
@@ -0,0 +1,239 @@
...
+
+(defun citar-cache--merge-hash-tables (hash-tables)
+  "Merge hash tables in HASH-TABLES."
+  (when-let ((hash-tables (reverse hash-tables))
+             (first (pop hash-tables)))
+    (if (null hash-tables)
+        first
+      (let ((result (copy-hash-table first)))
+        (dolist (table hash-tables result)
+          (maphash (lambda (key entry) (puthash key entry result)) table))))))
```
bdarcus, line 235:
```suggestion
          (map-do (lambda (key entry) (puthash key entry result)) table))))))
```

citar-cache.el
```diff
@@ -0,0 +1,239 @@
...
+  "Dissociate BUFFER from cached bibliographies.
+If BUFFER is nil, use the current buffer.  Otherwise, BUFFER
+should be a buffer object, buffer name, or a symbol like 'global.
+KEEP-FILENAMES is a list of file names that are not dissociated
+from BUFFER.
+
+Remove any bibliographies from the cache that are no longer
+needed by any other buffer."
+  (let ((buffer (citar-cache--canonicalize-buffer buffer)))
+    (maphash
```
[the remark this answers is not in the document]
roshanshariff: I used to do this, using the more generic `seq` and `map` functions wherever possible. But lately, since I've been focusing on performance, I've come to realize that unnecessary genericity is not always good.

Using `maphash` tells the compiler that you're definitely operating on a hash table, which is the case because this code is accessing our own internal variables (it would be a huge bug if we got anything but a hash table). It's implemented in C and is as close to a direct loop over hash tables as you can get. Potentially, it could even get optimized away with the new native compilation work, and the lambda function inlined (which would be a huge win).

The `seq` or `map` functions, on the other hand, must check the type of the object at runtime (necessarily, because they're not macros). The check itself is quite fast, but this means that the compiler has no visibility into what code is going to run. It also has to package up the lambda and call it as a function on every iteration.

My current approach is to directly use the built-in C functions, like `mapcar` and `maphash`, for basic access when possible. I only use the generic functions for higher-level algorithms that wouldn't make sense to implement separately for each type, or for handling data of unknown types.

bdarcus:
> Can we just assume that if an entry has a `file` field, then it must have files? If the files don't actually exist, then we'd give an error when trying to open them.

IDK myself, but probably that will work.

Can you just do that, and if we get complaints, can revert?

roshanshariff:
Arguably, the UX is actually better if we show the file indicator even if we can't successfully parse the file field. Then, when you try to open the file, we can detect the case that there's a file field but the parser didn't work and direct users to the appropriate documentation. Rather than the "has:file" indicator just not showing up (the status quo), which looks more mysterious (people might not even know it's supposed to be there).

citar-capf.el
```diff
@@ -24,10 +24,10 @@
 (declare-function org-eement-at-point "ext:org-element")
```
bdarcus, line 24: Just saw this bug when testing an eldev linter.

```suggestion
(declare-function org-element-at-point "ext:org-element")
```

bdarcus:
I don't know if this is relevant here, @roshanshariff, but since you're performance optimizing ...

Someone on the doom discord mentioned compiler macros; specifically `cl-define-compiler-macro`. The example they gave was was removing `cf-defun` function keywords at the compiler level, so that `(foo :bar t)` looks like `(foo t)` internally.

```elisp
(defun foo--no-keyword (bar)
  bar)

(cl-define-compiler-macro foo (&rest args)
  `(funcall #'foo--no-keyword ,(pcase-exhaustive args
                                 ('() nil)
                                 (`(:bar ,bar) bar))))

(cl-defun foo (&key bar)
  (funcall #'foo--no-keyword bar))

(cl-compiler-macroexpand '(foo :bar 1))
=> (funcall #'foo--no-keyword 1)
```

The documentation shows this, which changes `cl-member` "to a call to the faster `memq` in the common case where a is a non-floating-point constant".

```elisp
(cl-define-compiler-macro cl-member (&whole form a list &rest keys)
     (if (and (null keys)
              (eq (car-safe a) 'quote)
              (not (floatp (cadr a))))
         (list 'memq a list)
       form))
```

roshanshariff:
Thanks @bdarcus. I'd already found one place where keyword arguments were causing a problem inside a tight loop, because they require arguments to be passed as lists and that requires memory allocations. I ended up using `cl-defsubst` which expands out keyword arguments at compile time:

```elisp
(cl-defsubst citar-format--string (string
                                &key width text-properties hide-elided ellipsis)
  "Truncate STRING to WIDTH and apply TEXT-PROPERTIES.
If HIDE-ELIDED is non-nil, the truncated part of STRING is
covered by a display property that makes it invisible, instead of
being deleted.  ELLIPSIS, when non-nil, specifies a string to
display instead of the truncated part of the text."
  (when text-properties
    (setq string (apply #'propertize string text-properties)))
  (when (numberp width)
    (setq string (truncate-string-to-width string width 0 ?\s ellipsis hide-elided)))
  string)
```
Then running `(macroexpand-all '(citar-format--string "foo" :width 80))` produces
```elisp
(progn
  (if nil
      (progn
        (setq "foo"
              (apply #'propertize "foo" nil))))
  (if
      (numberp 80)
      (progn
        (setq "foo"
              (truncate-string-to-width "foo" 80 0 32 nil nil))))
  "foo")
```
Since we're not applying text properties (there's no `:text-properties` argument), it turns into an `(if nil ...)` in the output which gets removed by the byte compiler.

Thankfully there was only one place I needed to do this. In fact, the latest commit in this PR is reasonably good performance-wise. It's taking between 0.07 seconds and 0.2 seconds to format 7500 candidates (the first when ~1500 references have notes/files, the second when all of them do).

The time goes up by about 0.2 seconds for every `*` that appears in the format string. Previously I'd proposed a `fixed-width` option to avoid this, but now I think there might be a way to not incur that slowdown either. I'll leave the implementation of that idea to a future PR, but because of that I won't put in the fixed-width option yet. In the interim, anyone impacted can just not use any `*` fields in their format strings (and the defaults don't have any now anyway).

As I said, I'll leave further performance work to future PRs now that it's acceptably fast for ~10,000 item bibliographies. I've turned my attention to just cleaning up the API and getting all the Citar functions to work with the new API.

I've also implemented the proposal to include the citation key as an invisible prefix of each candidate, which simplifies a few things.

bdarcus:
Sounds like a plan.

My thinking is that once I merge this, we'll give it a week or so, and then tag v1.0.

So key thing is the API stability, and to make sure there aren't any bugs.

aikrahguzar:
> Thankfully there was only one place I needed to do this. In fact, the latest commit in this PR is reasonably good performance-wise. It's taking between 0.07 seconds and 0.2 seconds to format 7500 candidates (the first when ~1500 references have notes/files, the second when all of them do).

I haven't followed these changes but this is extremely impressive performance for on the fly formatting!

> The time goes up by about 0.2 seconds for every `*` that appears in the format string. Previously I'd proposed a `fixed-width` option to avoid this, but now I think there might be a way to not incur that slowdown either. I'll leave the implementation of that idea to a future PR, but because of that I won't put in the fixed-width option yet. In the interim, anyone impacted can just not use any `*` fields in their format strings (and the defaults don't have any now anyway).

This is surprising. The `*` just means that a width needs to be calculated and there is at most one star. So at the very beginning of the formatting one can calculate the width and then pass along this fixed width format. Calculating the width one time shouldn't take 0.2 seconds so what is causing the slow down? 

> As I said, I'll leave further performance work to future PRs now that it's acceptably fast for ~10,000 item bibliographies. I've turned my attention to just cleaning up the API and getting all the Citar functions to work with the new API.

Regarding API I think one area that can use some improvement especially in view of this is how the widths are specified in the format. With an absolute number of characters, the formatting can't handle different frame sizes well so I think it will be good idea to allow fractional widths. This can be done even in a backward compatible manner where integer widths are interpreted as number of characters while fractional widths are treated as fractions of frame width with the format converted into a fixed width one once for each call to `citar-select-ref` etc.

roshanshariff:
> there is at most one star

I've removed this restriction and now you can have any number of stars. The available width (i.e. the frame width minus all the fixed-width text) is distributed evenly amongst all the stars.

> I think it will be good idea to allow fractional widths

I was thinking this too, and my plan was just to allow a specification like `*2`, with just `*` being equivalent to `*1`. Then if you had two fields, `*2` and `*3`, the first field would get 2/5 of the space and the second would get 3/5 (for example). This is similar to how such things are specified in HTML, iirc. This should be a fairly easy extension to add, but I thought I'd defer it to after this PR.

> what is causing the slow down

The calculation is very fast, as you say. The slowdown is because of using `truncate-string-to-width`. Internally, the preformatting phase produces a list of strings, which alternate between fixed-width and `*`-width. See the function `citar-format--star-widths` in this PR, which takes such a list of strings, truncates alternate ones and concatenates them. So the number of strings that need to be concatenated is 2x the number of stars (and there's no concatenation if there are no stars).

Instead, it would be better to produce the entire concatenated string at the preformat phase, and just apply text properties as needed at display time. I believe this will reduce the number of memory allocations and hence GC time. This is the idea I was referring to above about how to reduce the performance impact of having multiple `*`-width columns. But I figured it's better not to get too ambitious with this PR and leave that for the next step.

bdarcus:
Another interesting article on performance tuning; found when searching for `cl-defsubst`.

https://nullprogram.com/blog/2019/12/10/

The `cl-first` section is particularly interesting.

bdarcus:
> the latest commit in this PR is reasonably good performance-wise.

Just in case you don't know, I can no longer compile this branch. There are a lot of compiler errors on `citar.el`.

I can just wait to test again when you mark this "ready" though. I'm sure it's more than good enough.

roshanshariff:
Yeah, all the compile errors are coming from functions that haven't yet been updated to use the new API. I'm slowly going through and fixing them. I'll mark this PR as ready for review when that's done.

roshanshariff:
Right, so you're saying that `citar-select-refs` should always return a list (even if called with `:multiple nil`) and `citar-select-ref` should always return a string? (Internally one is a wrapper around the other, but that doesn't matter for the API.) I think that makes sense, since my instinct is that a function should always return the same type of thing, regardless of its arguments.

The status quo in the main branch is that `citar-select-refs` always returns a list, whereas `citar-select-ref` only returns a list when called with `:multiple t`. So we would just remove that latter behaviour. Since I doubt that many users are explicitly calling `citar-select-ref` with `:multiple t`, there's probably no need for a breakage warning.

bdarcus:
> Right, so you're saying that `citar-select-refs` should always return a list (even if called with `:multiple nil`) and `citar-select-ref` should always return a string?

Yes, only from the standpoint of a clear and consistent API.

And so for embark acting at point within an org `citation-reference`, it's also returning a key string; not a list.

It's a subtle important point; just wanted to make sure we explicitly discussed it.

We now no longer use crm, but we do have a good alternative.

roshanshariff:
We do have Embark actions for citations as well, where the targets are of the form "key1 & key2 & key3". As far as I can tell, this is currently broken.

However, there's an easy fix; make a new set of functions in `citar-embark.el` (which I've created and will push soon) like this:
```
(defun citar-embark-open-citation (keystring)
  (citar-open (split-string-and-unquote keystring " & ")))
```

We make a new keymap that is just like `citar-citation-map`, but instead of the the standard functions like `citar-open`, it will have `citar-embark-open-citation`. Since these are not interactive functions, they shouldn't confuse users too much. And Embark will call them just passing in the "key1 & key2 & key3" string as an argument, instead of messing around with the minibuffer.

What do you think? I can't think of another simple way to unbreak `embark-act` for multi-key citations.

bdarcus:
> We do have Embark actions for citations as well, where the targets are of the form "key1 & key2 & key3". 

Those details come out of the crm implementation. But ...

> What do you think? I can't think of another simple way to unbreak `embark-act` for multi-key citations.

Thinking out loud ... I once wondered whether we really need at point embark actions on citations. Do we?

Certainly the citation references (the single keys) are most important.

WDYT?

bdarcus:
> Especially since it's so easy to add and not much of a maintenance burden, we might as well?

That's kind of my thinking behind the commands taking keys, and using multiple (now that there's no real downside) by default.

roshanshariff:
Yeah, those little wrapper functions are conceptually transformers that take an Embark target of type `citar-citation`, split the string target into a list, and call the real action with that list. But these can't be implemented as actual Embark transformers because those can only turn a target type+string into another type+string; they can't produce lists; and in any case, Embark doesn't know what to do with lists except in the limited case of multi-actions.

bdarcus:
> ... these can't be implemented as actual Embark transformers because those can only turn a target type+string into another type+string; they can't produce lists ...

Somewhere on here, Omar explains how this worked with CRM. Embark just passed the target string as is, and CRM would read it and convert it to the list.

I just don't think that would work with our CR-based alternative, which does work with real lists.

If, in fact, we need to add new shadow functions, it would make sense to decouple them from assumptions about embark, so they could also be used with hydra or transient menus, or even just at point keymaps?

bdarcus:
Sounds good; go for it then ;-)

It adds another reason to release a separate little `citar-embark` package!

bdarcus:
> In one of the discussions we had about Embark ages ago, I believe Daniel Mendler suggested creating such non-interactive functions if we wanted to get Embark to not use the minibuffer-text-insertion convention.

I think it may have been in this, long, thread?

https://github.com/emacs-citar/citar/issues/486

... in which we were also discussing changing the API to focus on citekeys!

bdarcus:
I think, once we merge this (and the move back to a key-based API), I'm going to experiment again with a transient and/or hydra menu, as a possible optional extension.

It might be that it ends up working better for us, or least for some people, at point.

bdarcus:
I'm not sure about keeping this as a separate repo, but for now, here's what the (untested) `citar-embark` looks like:

https://github.com/emacs-citar/citar-embark/blob/main/citar-embark.el

So explicit package include/require for embark, and no `with-eval-after-load` statements.

I also implemented the embark part of the idea I suggested in #642.

bdarcus:
I pushed a commit that now allows it to build, but there are still some CI errors to do with missing functions.

Will see if I can fix this.

roshanshariff:
Ah, I was just about to push my edits lately that also have a new `citar-embark` package. Do you I think I should merge your version?

bdarcus:
Sorry about that!

And of course whatever you decided was fine.

roshanshariff:
No worries, it's all sorted now! I'm updating the original PR description with the API changes made here. I will finish with the interactive commands tomorrow. Eventually, we can include that in the developer documentation. In the latest commit, I've renamed a couple of those commands and added obsolete aliases for the old names.

I think it would be easier to develop the Embark integration in sync with Citar if they were in the same repository. You could open PRs that changed both, for example. Maybe it's best to just ignore package-lint on this? I can go back to making it not fail the CI checks.

There's also a mysterious failure to compile on emacs 29 snapshot. I'll try to obtain a working copy of that and reproduce.

bdarcus:
I updated the test/manual script, both because it was out-of-date, and to help test this.

There, Embark still isn't working right for me. I assume it's just some little bug somewhere.

citar-embark.el
```diff
@@ -0,0 +1,130 @@
...
+;;  Description
+;;
+;;; Code:
+
+(require 'citar)
+(require 'embark)
+
+;;;; Keymaps
+
+(defvar citar-embark-map (make-composed-keymap citar-map embark-general-map)
```
bdarcus, line 27: Why not just move the maps here? They aren't useful outside Embark; are they?

roshanshariff: I bind `citar-map` to the prefix key `C-c b`, so I can open a bibliography file with `C-c b o`, insert a citation with `C-c b c`, etc. And it's nice to not have this prefix keymap cluttered with unnecessary keybinds from `embark-general-map`, so I figured it was better not to modify the original map in-place by setting its parent (which also makes `citar-embark-mode` more self-contained.

README.org
```diff
@@ -73,9 +73,17 @@ This is the minimal configuration, and will work with any completing-read compli
...
-Citar will automatically integrate with Embark if it is installed, offering contextual access to actions in the minibuffer and at-point.
+The =citar-embark= package adds contextual access actions in the minibuffer and at-point.
 When using Embark, the Citar actions are generic, and work the same across org, markdown, and latex modes.
 
+#+BEGIN_SRC emacs-lisp
+(use-package citar-embark
+  :bind (("M-;" . embark-act)
+         ("M-," . embark-dwim))
+  :custom
+  (citar-embark 1))
```
roshanshariff, line 84: I think we need to decide on the autoload strategy. As I see it, there are two options:

1. Load `citar-embark` after both `citar` and `embark` are loaded. Then `embark-act` will load Embark, but not detect in-buffer citations unless Citar also gets loaded somehow (either through a mode hook, or using a Citar command).
2. Load `citar-embark` as soon as `embark` is loaded, which will also pull in `citar`. This will "just work" because running any Embark command will also load and activate Citar, but may annoy people who just want to use Embark and aren't doing anything with Citar yet.

In either case, I think it's better to have separate `use-package` stanzas for `citar`, `embark`, and `citar-embark`. This would be the `citar-embark` configuration for option 1:
```elisp
(use-package citar-embark
  :after citar embark                   ; drop "citar" for option 2
  :no-require
  :config (citar-embark-mode))
```
The equivalent without use-package is just
```elisp
(with-eval-after-load 'embark
  (with-eval-after-load 'citar
    (citar-embark-mode)))
```

bdarcus: > I think we need to decide on the autoload strategy.

Do you mean just WRT to config? Or does this impact the code base?

If the former, and this is just for the README, we could include both options?

Or add an explanation?

citar-embark.el
bdarcus: I like the idea of the buffer actions being usable from a keybinding, or if possible also via a transient or hydra menu, without needing Embark.

citar.el
```diff
@@ -41,49 +41,50 @@
   (require 'subr-x))
 (require 'seq)
 (require 'browse-url)
+(require 'citar-cache)
+(require 'citar-format)
 (require 'citar-file)
-(require 'parsebib)
 (require 'crm)
```
bdarcus, line 47:
```suggestion
```

We can remove this now.

bdarcus:
I pushed a small commit to simplify the open note part.

https://github.com/emacs-citar/citar/pull/634/commits/171b2c90d78a59b547ca8e2eda1327438161a294

roshanshariff:
Re: the create/open note API, I think it makes sense to remove the `entry` argument for the `citar-open-notes-functions` because it's hard to imagine any note opener needing that information. But for the `citar-create-note-function`, pretty much every implementation is going to want the `entry` as well. Instead of forcing those functions to turn around and call `citar-get-entry`, I think we should pass the `entry` argument as well. That way, in many cases, the implementation of the function might be able to avoid a dependency on Citar.

As for the notes directories, I think the status quo is actually pretty good? We search in a list of directories but create in the first directory in the list? If we disallowed lists, then that would make it impossible to search in multiple directories, which seems like an unnecessary limitation for no good reason. If we had a separate variable for the creation directory, then you'd have to also add that directory to the search list or else Citar wouldn't find the notes it creates. Of course, we could have two different customization variables and implicitly include the creation directory in the search path, but I can't say I see the need...

Finally, currently the `citar-open-notes` is misnamed because it only lets you open a single note. Should we turn this into a multi-action? I think it's useful to be able to select multiple references and open all their notes at once. There's a UX design consideration about what to do when multiple keys don't already have notes: create all of them, possibly confirming first? Or ask them to select which notes to create?

bdarcus:
> Re: the create/open note API, I think it makes sense to remove the `entry` argument for the `citar-open-notes-functions` because it's hard to imagine any note opener needing that information. But for the `citar-create-note-function`, pretty much every implementation is going to want the `entry` as well. Instead of forcing those functions to turn around and call `citar-get-entry`, I think we should pass the `entry` argument as well. That way, in many cases, the implementation of the function might be able to avoid a dependency on Citar.

OK.

> As for the notes directories, I think the status quo is actually pretty good? We search in a list of directories but create in the first directory in the list. If we disallowed lists, then that would make it impossible to search in multiple directories, which seems like an unnecessary limitation for no good reason. 

My thinking is we have a default function that implements the file-per-note approach, and a flexible API for other options. So `citar-org-roam` will enable support for things like #518, for example, and we won't add it to core citar.

In the latter case, it doesn't really make sense to use the same variable.

So in the end, it comes down to whether we want to allow the default note function to search across multiple directories, AND to have this somewhat odd convention of only creating notes in the first directory, which if someone wants the first, they may well complain about the second.

That was my thinking anyway. I'm not 100% wedded to it.

> Finally, currently the `citar-open-notes` is misnamed because it only lets you open a single note. Should we turn this into a multi-action? I think it's useful to be able to select multiple references and open all their notes at once. There's a UX design consideration about what to do when multiple keys don't already have notes: create all of them, possibly confirming first? Or ask them to select which notes to create?

I think we can tease apart a few things:

- the main (`citar-open-notes`) command should probably, indeed, allow multi selection/keys
- the function it references, which calls the notes function (`citar--open-notes`) only accepts a single, which I think makes sense, since can run it in a `dolist` from `citar-open-notes`
- a wrinkle is that right now, for example, `org-roam` only returns a single note per reference; but that's not a constraint at the db level, and I've discussed with them removing that constraint, which in case selecting one reference in citar may open more than one from `org-roam`

Does that all make sense?

WDYT now?

bdarcus:
Except, passing `entry` means, in the case of ORB, forcing them to define an argument they don't actually use (they use `bibtex-completion` for that); they just throw it out.

roshanshariff:
> In the latter case, it doesn't really make sense to use the same variable.

Yes, that was my thinking as well. Something like `citar-org-roam` would extend all of `has-notes`, `open-notes`, and `create-note` for consistency, so it would be completely independent of these Citar configuration variables.

As far as Citar's built-in note functionality goes, you're right that the convention is a little odd. Maybe we can just have a `citar-note-path` (where notes are created), and the built-in has-note/get-note functions would implicitly search that in addition to `citar-open-notes-paths` (names TBD)? So if you only had the one directory, you could set the former and leave the latter nil. 

> the function it references, which calls the notes function (citar--open-notes) only accepts a single

I think this makes sense, but perhaps then this function should only open existing notes, not create new ones. If, for example, the user wants to be prompted before creating notes, then it makes sense to amalgamate all the prompts until after the existing notes have been opened. This can only be done if the higher-level `citar-open-notes` function is responsible for creating notes, not `citar--open-notes` which is called in a loop.

But the main thing is to design the UX for this. It's not clear to me how interactive the command should be. Some questions:
* If you ask to open notes for a single key, but there are multiple notes, should we open all of them or prompt to choose one or more?
* Ditto, but with multiple keys: if some keys have only a single note, we should just open them; otherwise why would the user have given us those keys in the first place? But if some keys have multiple notes, do we ask which ones to open? And once for each key or all together?
* If one or more keys don't have any notes, do we create all of them (potentially after confirmation) or ask which ones to create? And again, once per key or together for all of them?

Depending on the answers to these questions, we may have to change the `citar-open-notes-functions`+`citar-create-note-function` API. Currently, because we just have a list of open-note functions, it's hard for them to be interactive or prompt the user (otherwise you end up with an annoying series of prompts, one for each function).

Anyway, sorry for the wall of text :) The design question is somewhat involved and perhaps we can defer some of this work for later. It's hard to balance extensibility with a good UI.

bdarcus:
> The design question is somewhat involved and perhaps we can defer some of this work for later. It's hard to balance extensibility with a good UI.

Perhaps this is the key point. You ask the right questions, and I agree the answers are not obvious.

In that case, let's keep status quo for now, and consider changing later? We could turn your post into a new issue, for later resolution?

roshanshariff:
> Except, passing `entry` means, in the case of ORB, forcing them to define an argument they don't actually use (they use `bibtex-completion` for that); they just throw it out.

Ah, yes, that's not ideal. Sure, we can leave out the second argument for compatibility.

Though I don't suppose we could get ORB to just use the information in the `entry` argument? That way, people wouldn't have to setup `bibtex-completion` in addition to Citar. It's a little painful to do that, because then there are two caches, every bibliography file has to be parsed twice, etc...

Maybe if the necessary features were in `citar-org-roam`, there wouldn't be a need to use both? I haven't kept up with ORB's feature set lately. Is it the capture template preformatting feature that's missing?

bdarcus:
To backtrack a little, I don't think we should worry too much about ORB. He made the argument optional, so things won't break regardless.

> Though I don't suppose we could get ORB to just use the information in the entry argument? 

I have a feeling there are other, deeper, dependencies on `bibtex-completion`.

The maintainer is a `citar` user, so maybe he'll get around to it.

OTOH ....

> Maybe if the necessary features were in `citar-org-roam`, there wouldn't be a need to use both? 

I'd say that's correct, from what I understand.

> I haven't kept up with ORB's feature set lately. Is it the capture template preformatting feature that's missing?

I do currently have a capture template based function in `citar-org-roam`, but it doesn't work correctly ATM. However, you can see how it would work here, and I don't see what value it provides over the default create note function.

https://jethrokuan.github.io/org-roam-guide/#orgc48eb0d

So my impulse is to remove that function.

bdarcus:
@roshanshariff - you're aware that there's something with "has-files"?

As in, in many cases, the indicator will show a reference, but when selecting, it says no.

I can't immediately see what the problem is.

roshanshariff:
> As in, in many cases, the indicator will show a reference, but when selecting, it says no.

This is probably because the indicator is just checking if the entry has a `file` field, not parsing it (and therefore not checking file extensions). But when you actually go to open the file, it does filter out non-existent files or those with extensions not in `citar-library-file-extensions`. Perhaps we can improve the feedback on this somehow.

> There's also a little bug in the default parser (or more likely the function it calls, but I can't figure out how that's supposed to work):

`citar-file--split-escaped-string` takes a single character separator, not a string. Try this:

```elisp
(citar-file--split-escaped-string "one.txt;two.txt" ?\;)
```

The result of `citar-file--parser-default` makes sense when you consider that it's trying both `;` and `:` as separators. When it tries `;`, it correctly splits `"one.txt;two.txt"` into `'("one.txt" "two.txt")`. But it also tries `:` as the separator, in which case it just has the original string. Since it doesn't know what the actual separator character is, it keeps all the options. But the non-existent ones should get discarded.

Incidentally, have you actually come across bib files in the wild using `:` as the file field separator? Or were they actually triplet-format and looked like they were `:`-separated?

> I guess they're just not hooked up yet

If you run `eldev test`, it runs all the test files in `test/*.el`. They don't actually need to be hooked up to anything.

bdarcus:
> > As in, in many cases, the indicator will show a reference, but when selecting, it says no.
> 
> This is probably because the indicator is just checking if the entry has a `file` field, not parsing it (and therefore not checking file extensions). But when you actually go to open the file, it does filter out non-existent files or those with extensions not in `citar-library-file-extensions`. Perhaps we can improve the feedback on this somehow.

I need to double check, but I'm pretty sure it's not just this.

IIRC, when I reported this, I was looking at at file for a reference, so it was present. And it was indicated in the UI. But when going to open it, it wasn't found.

Again IIRC, this is with directory files; not file fields.

> `citar-file--split-escaped-string` takes a single character separator, not a string. Try this:
> 
> ```emacs-lisp
> (citar-file--split-escaped-string "one.txt;two.txt" ?\;)
> ```

Got it.
 
> The result of `citar-file--parser-default` makes sense when you consider that it's trying both `;` and `:` as separators. When it tries `;`, it correctly splits `"one.txt;two.txt"` into `'("one.txt" "two.txt")`. But it also tries `:` as the separator, in which case it just has the original string. Since it doesn't know what the actual separator character is, it keeps all the options. But the non-existent ones should get discarded.

Right.

> Incidentally, have you actually come across bib files in the wild using `:` as the file field separator? Or were they actually triplet-format and looked like they were `:`-separated?

I don't actually remember. There may be something in the closed issues; I'll take a look later.

> > I guess they're just not hooked up yet
> 
> If you run `eldev test`, it runs all the tests defined in `test/*.el`. They don't actually need to be hooked up to anything.

```elisp
❯ eldev test
Running 2 tests (2022-06-27 14:22:30-0400, selector ‘t’)
   passed  1/2  citar-format-test--parsing (0.000635 sec)
   passed  2/2  citar-format-test--star-widths (0.017110 sec)
```

bdarcus:
BTW, there's another little wrinkle in "has-file":

I have, it turns out, a lot of duplicate files; where they are listed both in the file field (from Zotero), and exist independently in the library directory.

I don't know what, if any, general conclusion to draw from that. Certainly I can clean up my data, or modify `citar-file-field-functions`, which is maybe as it should be.

roshanshariff:
There was a bug in `citar-get-files`: if `citar-library-file-extensions` was nil, it would just filter out all files instead of keeping all of them. Since `citar-open-file` let-binds that variable to nil, it wouldn't show any files. I've pushed a new commit that fixes this issue.

The new commit also improves a number of other things (see the commit message). There's now much better feedback to the user if file parsing didn't work, explaining what went wrong. Whenever `has-files` is true for some key but `get-files` fails, there will always be a message explaining the discrepancy.

Also, the Calibre example in #283 is actually a triplet-style file field, not `:`-separated. Calibre uses `,` separated fields, each element being a triplet of the form `:/some/file.pdf:PDF` (the first component of the triplet happens to be empty). These are handled by `citar-file--parser-triplet`, so I've made `citar-file--parser-default` now only handle `;`-separated fields (not `:`).

I've also added tests for all the above functionality. When you ran `eldev lint`, it was actually running all the tests but `citar-format-test--parsing` was misnamed; it should have been `citar-file-test--parsing`. In any case, I've split that test up into separate ones for each parser and the overall `citar-file--parse-file-field` function.

> Maybe instead of naming the functions for the applications, we name them for the format?

I'm not sure I understand, but we don't have any functions named for applications. There's only `citar-file--parser-{default,triplet}`.

> I have, it turns out, a lot of duplicate files

`citar-get-files` removes duplicates; now that it's fixed, you should only see each file once.

bdarcus:
> I'm not sure I understand, but we don't have any functions named for applications.

I was just meaning if we add a new one for Calibre, we shouldn't include that in the name.

citar.el
```diff
@@ -1107,89 +1092,78 @@ For use with `embark-act-all'."
...
-  (interactive (list (citar-select-ref
-                      :rebuild-cache current-prefix-arg)))
-  (let* ((embark-default-action-overrides '((file . find-file)))
-         (key (car key-entry))
-         (entry (cdr key-entry)))
+(defun citar-open-notes (key)
+  "Open notes associated with the KEY."
+  ;; REVIEW KEY, or KEYS
+  (interactive (list (citar-select-ref)))
+  (let ((embark-default-action-overrides '((file . find-file))))
```
bdarcus, line 1135: @roshanshariff WDYT about this line?

Would it not be the case that (again, just for practical demonstration) `citar-org-roam` candidates would not be files, but org-roam "nodes"?

So would to deal with that?

roshanshariff:
> So on citar-open, you removed "notes" from that?

This was supposed to be temporary, until we got a `citar-get-notes` function. But that got sidetracked and I forgot to add it back in.

If we're going to stabilize the API at 1.0, maybe we need to bite the bullet and figure out how to do notes properly? Something like a list of `get-notes` functions that return a display string, internal identifier, and opener function for each note? Then the `org-roam` function would maybe use the title of the note as the display string, the node ID as the identifier, and a wrapper around `org-roam-node-visit` as the opener function.

bdarcus:
> If we're going to stabilize the API at 1.0, maybe we need to bite the bullet and figure out how to do notes properly? Something like a list of `get-notes` functions that return a display string, internal identifier, and opener function for each note? Then the `org-roam` function would maybe use the title of the note as the display string, the node ID as the identifier, and a wrapper around `org-roam-node-visit` as the opener function.

I'm not exactly sure what the details should be. But this is what I currently have in citar-org-roam:

```elisp
(defun citar-org-roam-open-note (key &optional _entry)
  "Open org-roam node for KEY."
  (let ((ref-node (org-roam-node-from-ref (concat "@" key))))
    (when ref-node
      (org-roam-node-open ref-node))))
```

So the identifier can also be the citekey there, since org-roam has a dedicated node type called a ref, which is basically a zettelkasten reference note; aka bibliographic note.

I think the issue I was a bit stuck on was the type is no longer a "file", but a more generic "node".

We use the candidate type, of course, also for marginalia annotations.

Maybe in the end it's actually a simple problem to solve?

bdarcus:
Actually, I guess the issue is the related resource candidates.

`org-roam-ref-find` provides candidates for ref nodes., though they aren't designed well. I need to think about this a bit more, but certainly the display string should be the node title, as you suggested.

roshanshariff:
Yeah, I will also think about it more. I thought one of the considerations (like you mention in org-roam/org-roam#2207) was that a ref could be associated with more than one node (even if `org-roam` doesn't quite handle this correctly right now)? And similarly, for the file-based notes we have built-in support for, there could be multiple notes for the same key.

So perhaps it makes sense to use a different note ID than just the citation key? For file-based notes it would be the note filename, and for org-roam notes it could be the node ID.

bdarcus:
> I thought one of the considerations (like you mention in https://github.com/org-roam/org-roam/issues/2207) was that a ref could be associated with more than one note (even if org-roam doesn't quite handle this correctly right now)? And similarly, for the file-based notes we have built-in support for, there could be multiple notes for the same key.

This isn't really a problem though, since we already support more than one file per key in the open UI. The paths distinguish them.

> So perhaps it makes sense to use a different note ID than just the citation key? For file-based notes it would be the note filename, and for org-roam notes it could be the node ID.

Probably.

Yes, there was that issue in org-roam, still not entirely resolved.

But I fixed a narrow related issue with the completions by adding the node id as a hidden property to the ref (citekey) candidate string.

So here's what that looks like with two nodes with for the same ref:

image

OTOH, you can also do stuff like this in org-roam:

```elisp
(org-roam-db-query [:select * :from nodes :where (= title "Protest")])
```

Could even use the trick you suggested earlier of putting the ID as a hidden string at the beginning, and use the title, and maybe citekey, for the visible display.

bdarcus:
The nice thing about `citar-get-files` is it just returns a simple list of files, which is also compatible with the expectations of `consult--multi`. Since we already know the type, that's all we need. And Marginalia can handle any annotations.

With notes, the issue is that we don't know what the target is: it could be a file (as it always has been thusfar), or it could be some sub-file section. For purposes of this discussion, the type can be `file` or `org-roam-node`.

If the type is `file`, then the candidates need be no different than `citar-get-files`; it's only with the `org-roam-node` case that it gets complicated. And even more specifically, it's only if we insist on allowing different types within the same group, which I'm skeptical about.

You were suggesting:

>  Something like a list of get-notes functions that return a display string, internal identifier, and opener function for each note? 
 ... [P]erhaps it makes sense to use a different note ID than just the citation key? For file-based notes it would be the note filename, and for org-roam notes it could be the node ID.

A display string may or may not be part of the annotation. If it is, it could be a list of id, annotation function, opener function.

The id does indeed, per my example, make sense to be the node-id.

But if we go this more complex way, I would in general be much more complex than the other parts of `citar-open`, and it wouldn't be compatible with `consult--multi`.

OTOH, with that, you could just do:

```elisp
(defvar citar-org-roam--source-notes
  `(:name "Notes"
    :narrow ?n
    :category org-roam-node
    :annotate ,#'citar-org-roam--annotate ; title and maybe citekey annotations
    :items ,#'citar-org-roam--notes)) ; a list of node-ids
```

That's of course much simpler, with the sole constraint that all notes must be the same type.

bdarcus:
Here's the status of `citar-open`:

image

bdarcus:
This isn't truly done, but I've marked it ready for review, as I want to merge this soon.

If you have time to test it, @andersjohansson or @aikrahguzar, now would be a good time.

I still have a tiny little issue to fix with resource grouping, but will do that soon.

Other than that:

- I can't get the new defcustom-based alist to work per https://github.com/emacs-citar/citar/issues/646#issuecomment-1172982936
- the big one: filenotify. I don't know how to adjust this code at all, but my suspicion is it's easier to just rewrite into a much simpler form. If either @aikrahguzar or @roshanshariff have time for that, that'd be great. If not, I'll remove that file for now.

andersjohansson:
Great work! I’m away from my computer and won’t be able to test until next weekend, unfortunately.

bdarcus:
>
> Thanks!

Am wondering about removing citar-open-files and open-links.

Thoughts?

roshanshariff:
I would be inclined to remove the file-notify functionality for now. The code is rather complex, and I can't think of a good way to have it work without locking out the UI. Without it, you get the parsing delay the first time you use a Citar command after the bibliography has changed, rather than every time it changes.

In the future I'd like to look into a way of doing the parsing in the background, but that is going to be challenging. A background emacs process would need to send the parsed data back as text, parsing which is probably just as slow as `parse-bib`. And apparently the Emacs multi-threading support is very immature and not suitable for serious use.

roshanshariff:
I've pushed a new commit which changes the caching behaviour. Now calling any citar command will refresh any files if and only if needed; see the commit message. I feel this is the right default for almost all users.

I've removed the `citar-cache-refresh` command since it shouldn't be needed any more. For development purposes (while working on the parsing code for example) it's probably more useful to just clear the cache with `(clrhash citar-cache--bibliographies)`.

If there is demand, we can add a mode where Citar never updates files after first reading them, unless you manually request it. This could be useful for people who have large, frequently changing bibliographies and want Citar to intentionally use outdated information.

Finally, as a small help for benchmarking, Citar now prints a message whenever it's updating a bibliography, and prints the elapsed time. If you want to benchmark the candidate formatting, which is the other performance-sensitive feature, use
```elisp
(citar-get-bibliographies) ;; make sure bibliographies are in cache
(benchmark-run-compiled 10 (citar--format-candidates)) ;; Remember to divide the elapsed time by 10
```

README.org
```diff
@@ -216,65 +223,17 @@ You can save this history across sessions by adding =citar-history= to =savehist
...
-  (run-with-idle-timer 0.5 nil #'citar-refresh))
-
-(add-hook 'LaTeX-mode-hook #'gen-bib-cache-idle)
-(add-hook 'org-mode-hook #'gen-bib-cache-idle)
-#+END_SRC
+Citar uses a cache to speed up library display.
+If a bib file changes, the cache will automatically update the next time you run a Citar command.
 
-For additional configuration options on this, see [[https://github.com/bdarcus/citar/wiki/Configuration#automating-path-watches][the wiki]].
+The =citar-cache-refresh= command will also reload the caches manually.
```
andersjohansson, line 229: citar-cache-refresh was removed now as well wasn’t it? (got some time to start checking this out before leaving my computer)

citar-file.el
```diff
@@ -351,6 +321,21 @@ of files found in two ways:
...
+  (citar-file--directory-files
+   citar-notes-paths keys citar-file-note-extensions
+   citar-file-additional-files-separator))
+
+(defun citar-file-has-notes (&optional _entries)
+  "Return predicate testing whether cite key has associated notes."
+  ;; REVIEW why this optional arg when not needed?
+  (let ((files (citar-file--get-notes-hash)))
+    (lambda (key)
+      (gethash key files))))
```
bdarcus, line 337: @roshanshariff can you review this little thing on `citar-has-note`?

See my comment here.

I didn't need entries here, and I don't need them in the org-roam one.

roshanshariff: I'll do a more full code review later this week when I have some time, but the main reason is for commonality with the file-related functions. There it's useful to have the `entries` argument to check if an entry has the `file` field.

On the other hand, with our current notes backends, it is true that you only need the cite key. But I can imagine adding support in the future for the BibTeX [note field](https://www.bibtex.com/f/note-field/), so maybe it's a good idea to leave the API as-is and just ignore the `entries` argument if your backend doesn't need it,

bdarcus: Ok, we can think on it a bit.

But if a function needs it, it can pull the entries; no?

E.g. if I understand right, doesn't seem justified to design for a hypothetical need when we have two concrete backends neither of which need it.

roshanshariff: Maybe this is a matter of taste to some extent. The code that's calling `has-notes` has a set of entries that it is operating on. To me, it seems cleaner to pass that on to the backend, which can either use it or ignore it. The alternative is for the backend to call back into Citar to get the entries, but this is not trivial. It involves asking the major-mode function for the list of bibliographies, getting them from the cache, merging their entries into another big hash table (which can use a lot of memory for no reason), etc. This is the reason some other API functions accept an optional `:entries` keyword argument, so they can avoid calling `citar-get-entries` if the `entries` hash table is already available.

And finally, it's easier to unit test a function just by passing it a hash table, knowing that its behaviour only depends on the arguments you pass in (rather than having to set up an environment where `citar-get-entries` returns the right thing).

In general, I prefer functions that take arguments for the information they need and return results that depend only on those arguments rather than global state. Of course, global state is unavoidable, but I think it's better for testability, reasoning, etc. if access to it is not scattered throughout the code base but carefully controlled to a few places.

bdarcus: Ok. In any case, I'll defer to you; just thought I'd raise it.

roshanshariff: Sure, and of course I might come around to a different view later. But also see my edit above about race conditions.

In general, accessing the cache is not "free", especially now that it can unpredictably decide to re-parse files. In general, the correct way to use it is to get everything you need from the cache at the beginning of some operation (i.e. call `citar-get-entries` in this case), use that information throughout, and then throw it away only when you're done.

bdarcus: In other news, I fixed at least one obvious bug (that was causing a compilation error noted in the MELPA review) in the capf. Perhaps it works now?

I can't easily test it.

bdarcus: > I'll do a more full code review later this week when I have some time ...

I'm not sure if it's useful, but I just added [this branch](https://github.com/emacs-citar/citar/tree/pre-1.0-cleanup).

You should have write access, if you need/want it.

andersjohansson:
Some feedback on real-world performance:

On my 6500 items library, with native json support:
```
Updating bibliography ~/bibliotek/bibliotek.json...done (3.700 seconds)
Updating bibliography ~/bibliotek/bibliotek.bib...done (10.567 seconds)
```

Running:
```emacs-lisp
(benchmark-run-compiled 10 (citar--format-candidates))
```
Gives: `(7.243678976 39 6.092884648)`. .7 seconds doesn’t seem too bad.

Thanks to the updates of org-cite-csl-activate, @andras-simonyi, one parsing step is cut out from my workflow as well.

I guess people on older computers may perhaps need to take action.

Nevertheless, my switch from citing in org-mode with the help of [org-zotxt](https://gitlab.com/egh/zotxt) which calls the really slow Zotero server, to citing with citar feels like a great update. Citing once the library is parsed is really quick!

bdarcus:
@andersjohansson - we have identified some performance issues in #659, which should be addressed in parsebib and in citar "soon". But it only impact initially parsing, or when the file changes.

PR #638, opened 2022-06-17
Add start of eldev linting, etc.

bdarcus:
Would something like this be useful, @roshanshariff, that in time we could hook up to the CI, including unit tests, etc?

Right now, all that works is `make lint`, where it will flag a ton of things, mostly code formatting/indenting.

I had an earlier github actions setup (code still here), but I found it too interventionist.

This would reject PRs (in the sense mark them needing fixes) with these warning though.

But my thought is in the interim, could add it and ask contributors to use it? I know it would be useful for me!

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,0 +1,11 @@
+
+((emacs-lisp-mode
+  (fill-column . 110)
+  (indent-tabs-mode . nil)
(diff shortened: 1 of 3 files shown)


bdarcus:
Merged.

Easy enough to remove or ignore.

But give it a go.

PR #639, opened 2022-06-18
Add GitHub CI action and fix compilation and linting errors

roshanshariff:
I've modified the Eldev configuration so that it correctly compiles, tests, and lints all of Citar. I've also gone through the code base and fixed everything that was causing errors or warnings:

- Byte-compilation warnings
- Errors in function and variable declarations
- Whitespace and indentation issues
- Docstring inconsistencies found by checkdoc
- Regexp lint warnings found by relint
- Docstring quote issues found by the emacs 29 byte-compiler

The GitHub CI action runs byte compilation, tests, and all linters: [elisp-lint](https://github.com/gonewest818/elisp-lint), checkdoc, [relint](https://github.com/mattiase/relint), and [package-lint](https://github.com/purcell/package-lint). It runs these on emacs 27, 28, and 29, triggered by pull requests and pushes to the main branch.

Currently, package-lint warns on our use of `with-eval-after-load`, so its failure does not mark the check as failed. But its output is still available in the action logs.

With the fixes in this PR, all these checks pass on emacs 27.2, 28.1, and current 29 snapshot.


bdarcus:
Awesome; thanks!

> Currently, package-lint warns on our use of with-eval-after-load ...

This has always annoyed me, so good it doesn't fail the check.

roshanshariff:
From [this discussion](https://github.com/emacs-citar/citar/pull/640#issuecomment-1159462604)

> Now that's on, it doesn't run, or isn't an action under "checks".That may just because of the sequencing here, but FYI.

You should be able to trigger the action manually from the [Actions page](https://github.com/emacs-citar/citar/actions/workflows/check.yml). However, I [did this](https://github.com/emacs-citar/citar/actions/runs/2520896502) and got the error

```
purcell/setup-emacs@master is not allowed to be used in emacs-citar/citar. Actions in this workflow must be: within a repository owned by emacs-citar, created by GitHub, or verified in the GitHub Marketplace.
```

I was able to fix it on my forked repository by adding the `purcell/setup-emacs@master` to the [Actions permissions](https://github.com/emacs-citar/citar/settings/actions) page:

image

I'm not sure if you can manually trigger the action on a pull request; perhaps just running it on the origin branch of the pull request will work? In any case, it should also run again if you push a new/modified commit.

bdarcus:
> I'm not sure if you can manually trigger the action on a pull request; perhaps just running it on the origin branch of the pull request will work? 

I don't think you can. I had tried earlier setting the "all actions" option, and that also didn't give me the option to manually run it.

But I did add that line to the more restrictive option. So should work next PR (or this one changed, as you say).

PR #640, opened 2022-06-18
Autoload citar--embark-selected

bdarcus:
See https://github.com/doomemacs/doomemacs/issues/6464

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -1023,6 +1023,7 @@ FORMAT-STRING."
                               (with-current-buffer (embark--target-buffer)
                                 (citar--get-candidates)))))))
 
+;;;###autoload
 (defun citar--embark-selected ()
   "Return selected candidates from `citar--select-multiple' for embark."
   (when-let (((eq minibuffer-history-variable 'citar-history))


bdarcus:
Two things @roshanshariff:

1. do you think this is the right solution for the linked issue at doom emacs?
2. I forgot to turn back on github actions before opening this. Now that's on, it doesn't run, or isn't an action under "checks".That may just because of the sequencing here, but FYI.

roshanshariff:
The fix looks fine as far as fixing that bug, but I wonder if we should be trying to avoiding Citar being autoloaded in that case; just running `embark-export` from grep results is completely unrelated to Citar after all?

Looking at [the code](https://github.com/emacs-citar/citar/blob/b6cd49ffb56824b8d1793b0c4268237b3d89fb45/citar.el#L1039-L1045), we autoload Citar whenever you run
* `embark-act`, due to `citar-citation-finder` and `citar-key-finder`, or
* `embark-collect` (as in this bug), due to `citar--embark-selected`.

But presumably if you don't have Citar loaded, you won't have any Citar completion candidates to collect, so we shouldn't yet have added ourselves to `embark-candidate-collectors`. Maybe we should move that out of the autoloaded `with-eval-after-load`? (Similar reasoning applies to `citar--reference-transformer`.)

```elisp
;;;###autoload
(with-eval-after-load 'embark
  (add-to-list 'embark-target-finders 'citar-citation-finder)
  (add-to-list 'embark-target-finders 'citar-key-finder))

(with-eval-after-load 'embark
  (add-to-list 'embark-transformer-alist '(citar-reference . citar--reference-transformer))
  (add-to-list 'embark-candidate-collectors #'citar--embark-selected))
```

Maybe we shouldn't even be autoload-adding Citar functions to `embark-target-finders`. We'd wait until we were in a major mode where Citar is usable. I'd have to think about how to accomplish that reliably.

bdarcus:
> he fix looks fine as far as fixing that bug, but I wonder if we should be trying to avoiding Citar being autoloaded in that case ...

That's what I was wondering too.

bdarcus:
Is there any value in moving all the embark code to `citar-embark.el` and having it a minor mode?

roshanshariff:
I think that would be a very good idea. So the minor mode would add Embark target finders when enabled, and remove them when disabled? And you'd have to do something like
```
(add-hook 'org-mode-hook #'citar-embark-minor-mode)
```
for all the modes where you wanted Citar to work in-buffer.

bdarcus:
Yes. I would do it myself, but given #634, it might be better to add it there, or wait to merge it and then do it?

Issue #507, opened 2021-12-27
Make file-open more configurable


bdarcus:
We should make opening files more configurable: use an alist as minad suggests below (though we'd maybe need to think through more what the defaults should be; particularly for pdfs):

```elisp
(defcustom citar-file-open-functions-alist
  '(("html" . #'citar-open-external) ; what about pdf?
    ("org" . #'citar-open-note)  ; do we need this here?
    (t . #'find-file))
  "Function to use to open files.

Each cons should be of the form (EXTENSION . FUNCTION)."
  :group 'citar
  :type '(alist))
```

I think this also allows us to remove `citar-file-note-extensions`.

Also, see this [related query about mailcap](https://www.mail-archive.com/emacs-orgmode@gnu.org/msg144468.html), though I didn't really understand.

I guess, beyond borrowing code, in that case they associate mime-types with open functions; so it's a different approach.

minad:
Or make it an alist with configurable open functions for each extension? The `t` key could specify the default open function. The `t`  key as default is a common pattern in Emacs.

PR #651, opened 2022-07-05
pre-1.0-cleanup branch

bdarcus:
I was realizing I was making too many small commits, so setup this branch to collect various commits to get this repo ready for tagging 1.0 in the next week or so.

The first commit emerged out of a suggestion from the MELPA review for citar-embark, and that is to use these for the file headers:

https://spdx.dev/ids/

So I've done that, and also removed the human-readable license text. I hope that's OK; I guess it should be if MELPA folks recommended it.

@localauthor and @aikrahguzar - not sure if you wanted your names on the copyright for some of these files?

-------------

### Questions/Issues

1. I'm not sure on this, but on the note API, I'm leaning towards only defining the `:action` function (or we could all it `:open`?), and leaving out the `:create` one. My reasoning is that creating new notes is pretty closely-tied to opening them, and so better to just have the API reflect that. Yes, if someone wants something like the default source for markdown, it will mean having to copy and adjust the default function(s). But that seems a small price to pay. And at least with `citar-org-roam`, there's no value in a separate configurable function for new notes.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -66,6 +66,27 @@
 
 ;;; Variables
 
+(defvar-local citar--entries nil
+  "Override currently active citar entries.
+
+When non-nil, should be a hash table mapping citation keys to
+entries, as returned by `citar-get-entries'. Then all citar
+functions will use that hash table as the source of bibliography
+data instead of accessing the cache.
+
+This variable should only be let-bound locally for the duration
+of individual functions or operations. This is useful when using
+multiple Citar functions in quick succession, to guarantee that
+all potential cache accesses and updates are performed up-front.
+In such cases, use a pattern like this:
+
+  (let ((citar--entries (citar-get-entries)))
+    ...)
+
+Note that this variable is buffer-local, since Citar has a
+different list of bibliographies (and hence entries) for each
+buffer.")
+
 ;;;; Faces
 
 (defgroup citar nil
@@ -258,7 +279,8 @@ If nil, single resources will open without prompting."
                        :hasnote #'citar-file-has-notes
                        :open #'citar-file--open-note
                        :create #'citar-org-format-note-default ; TODO remove?
-                       :items #'citar-file--get-note-files)))
+                       :items #'citar-file--get-note-files
+                       :transform #'file-name-nondirectory)))
   "The alist of notes backends available for configuration.
 
 The format of the cons should be (NAME . PLIST), where the
@@ -274,7 +296,9 @@ plist has the following properties:
 
   :items function to return candidate strings for keys
 
-  :annotate annotation function (optional)"
+  :annotate annotation function (optional)
+
+  :transform transformation function (optional)"
   :group 'citar
   :type '(alist :key-type symbol :value-type plist))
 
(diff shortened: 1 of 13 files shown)


aikrahguzar:
> @localauthor and @aikrahguzar - not sure if you wanted your names on the copyright for some of these files?

I don't think my name is needed on on the copyright.

I haven't tested the new interface yet but I will over this weekend.

localauthor:
> > @localauthor and @aikrahguzar - not sure if you wanted your names on the copyright for some of these files?
> 
> I don't think my name is needed on on the copyright.

Agreed. No need to worry about it for me. :)

I'm testing and reconfiguring some of my custom functions now. All seems good so far, and its easy to see the API is much more sensible/consistent. I'll be sure to flag any issues when/if I come across them.

bdarcus:
@localauthor - OK cool. 

Do you use Embark, such that you're testing the new minor mode?

localauthor:
Ok, so I've made just added a note source tailored to my `zk` package, here: [zk-citar](https://github.com/localauthor/zk/blob/main/zk-citar.el)

Re: question 1 above, I think `:open` makes more sense than `:action`. No opinion on leaving or taking out `:create`, except that leaving it in makes things a bit more transparent.

bdarcus:
> And no opinion on leaving or taking out `:create`, except that leaving it in makes things a bit more transparent.

The one wrinkle here is the distinction between format and create.

With the former, we could use same open function, but different format functions, for markdown and org.

roshanshariff:
This is one place where the API is a little non-orthogonal. If your note-taking package is agnostic about Org/Markdown/etc., it still has to be able to format the note appropriately because that is the responsibility of the note source. Even though perhaps `citar-org` or `citar-markdown` are in a better position to handle that. I'm not sure if/how to improve this.

bdarcus:
> This is one place where the API is a little non-orthogonal. If your note-taking package is agnostic about Org/Markdown/etc., it still has to be able to format the note appropriately because that is the responsibility of the note source. Even though perhaps `citar-org` or `citar-markdown` are in a better position to handle that. I'm not sure if/how to improve this.

We should rename `:action` to `:open` regardless.

But beyond that, it seems the options are:

1. keep a single function, and leave it developers to determine how to handle creating new notes.
2. also have a single `:create` or `:format` property
3. add both, and clearly distinguish them

Right?

roshanshariff:
I agree there should only be one function, just as emacs has `find-file` for both opening or creating files. The question is what that function should do after it has created a note:
1. Format the note as it sees fit. Perhaps `citar-org-roam` will do this, because the notes are necessarily in Org format.
2. Just call `find-file` and do nothing else. This will give a blank note by default.
3. Call `find-file`, then use some Citar function that we provide for formatting notes based on the major mode.

This is of course a question for the developer of the note package, but if they want Citar to handle the note formatting (option 3) then we might able to provide a function to help them.

bdarcus:
> This is of course a question for the developer of the note package, but if they want Citar to handle the note formatting (option 3) then we might able to provide a function to help them.

Does that not suggest an (optional) `:format` function?

So then the default (`citar-file`) would include that, and `citar-org-roam` would probably not.

That reminds me; we should rename `citar-file` to `citar-file-org` or some such.

roshanshariff:
> Does that not suggest an (optional) :format function?

Would Citar call this after the `:open` function has created a note? In that case, I don't think we need it; the `:open` function could just call the formatting function itself, couldn't it? Instead of telling Citar what to run via the `:format` function?

> That reminds me; we should rename citar-file to citar-file-org or some such.

The `citar-file` note source has very little to do with Org except for the formatting when it creates a new note. This suggests that the formatting should be handled by Citar rather than the note source.

The simplest way to do this would be to add a defcustom `citar-format-note-function`, which the note `:open` function calls when it has created new note. By default, its value could be `citar-org-format-note-default`, but the user can change it as they please.

Another option would be to add a command like `citar-format-note`, which is modeled on the emacs `auto-insert` function. It would look at the major mode of the current file, look up the ref to which the note belongs, and dispatch to a major-mode specific function that actually inserts the note format into the current buffer.

We'd need a new note API function (`:noteref`?) which returns the citekey for the currently open note (or nil if the current file isn't a note). This seems useful to add anyway, because we could use it to offer actions that operate on the current ref.

bdarcus:
> > Does that not suggest an (optional) :format function?
> 
> Would Citar call this after the `:open` function has created a note? In that case, I don't think we need it; the `:open` function could just call the formatting function itself, couldn't it? Instead of telling Citar what to run via the `:format` function?

Yes. I was just doing the thought experiment of asking what it would look like to add a markdown function that otherwise works the same as the current org one.

But I do think perhaps if in doubt, we should stay with just the one.

> > That reminds me; we should rename citar-file to citar-file-org or some such.
> 
> The `citar-file` note source has very little to do with Org except for the formatting when it creates a new note. This suggests that the formatting should be handled by Citar rather than the note source.

Except, would that generalize?

What if someone wanted a note source where there was a single file, and markdown?
 
> The simplest way to do this would be to add a defcustom `citar-format-note-function`, which the note `:open` function calls when it has created new note. By default, its value could be `citar-org-format-note-default`, but the user can change it as they please.
> 
> Another option would be to add a command like `citar-format-note`, which is modeled on the emacs `auto-insert` function. It would look at the major mode of the current file, look up the ref to which the note belongs, and dispatch to a major-mode specific function that actually inserts the note format into the current buffer.
> 
> We'd need a new note API function (`:noteref`?) which returns the citekey for the currently open note (or nil if the current file isn't a note). This seems useful to add anyway, because we could use it to offer actions that operate on the current ref.

Do you think it's something you're confident we need to add now, or should we just add it later?

My impulse is the latter. I'm less concerned with adding note functions, that changing how they operate.

bdarcus:
For now, I just pushed https://github.com/emacs-citar/citar/pull/651/commits/7a04d3e2e2d2f71a1597d56c3dba44954697c6ff

localauthor:
Lest this get lost in the shuffle:

Should `citar-open-notes` should be added to `citar-embark--multitarget-actions`?

bdarcus:
> Lest this get lost in the shuffle:
> 
> Should `citar-open-notes` should be added to `citar-embark--multitarget-actions`?

Sorry; meant to add that also today. Done.

roshanshariff:
I've pushed a new commit that makes some API changes. As @bdarcus hoped in #634, the API functions no longer take an optional `entries` argument to prevent them from using the cache. Instead, there is a buffer-local `citar--entries` variable which can be let-bound when needed to achieve the same effect.

This change also affects notes sources since the `:hasnote` and `:items` functions no longer get the `entries` argument. Instead, you can just `citar-get-entry` or `citar-get-entries` if needed.

I've also simplified some of the code around `citar-open`. See the commit message for details.

citar.el
```diff
@@ -1089,21 +1097,23 @@ are potentially cross-referenced from elements of KEYS."
...
+  (let ((embark-default-action-overrides `((file . ,#'citar-file-open)
+                                           (url . ,#'browse-url)
+                                           ,@(when-let ((notecat (citar--get-notes-config :category))
+                                                        (open (citar--get-notes-config :open)))
+                                               (list (cons notecat open)))
+                                           . ,embark-default-action-overrides)))
+      (if-let ((resource (citar--select-resource keys
+                                                 :files t :links t :notes t
+                                                 :always-prompt citar-open-prompt)))
+          ;; TODO Does this work for non-file notes?
```
bdarcus: @roshanshariff - do you mean, for example, citar-org-roam?

If yes, the answer is yes; it works.

bdarcus:
> This change also affects notes sources since the `:hasnote` and `:items` functions no longer get the `entries` argument. Instead, you can just use `citar-get-entry`, `citar-get-entries`, or `citar-get-value`, etc. if needed.

So you're no longer concerned about this?

bdarcus:
A probably minor thing I just realized, and maybe not worth trying to address (unless there's an easy way to do it).

I just added a message to the `citar-org-roam` open function. But, of course, it will only report the number opened per citekey.

```elisp
(defun citar-org-roam-open-note (id-key entry)
  "Open or creat org-roam node for ID-KEY and ENTRY."
  (let* ((key (car (split-string id-key " ")))
         (ref-node-ids
          (flatten-list
           (org-roam-db-query
            [:select node-id :from refs
             :where (= ref $s1)] key))))
    (if ref-node-ids
        (let ((number (length ref-node-ids)))
          (dolist (id ref-node-ids)
            (let ((ref-node (org-roam-node-from-id id)))
              (org-roam-node-open ref-node)))
          (message (format "%s notes opened" number)))
      (citar-org-roam--create-capture-note key entry))))
```

citar.el
```diff
@@ -560,124 +584,136 @@ HISTORY is the `completing-read' history argument."
...
-          ('url "Links"))))))
-
-(cl-defun citar--format-candidates (&key (bibs (citar--bibliographies))
-                                         (entries (citar-cache--entries bibs)))
-  "Format completion candidates for ENTRIES.
-
-BIBS should be a list of `citar-cache--bibliography' objects that
-are the source of ENTRIES. Use the pre-formatted strings in BIBS
-to format candidates.
+  (if (get-text-property 0 'citar--note resource)
```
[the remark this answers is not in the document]
roshanshariff: The configurable transform functions are just an side benefit. To make `citar-open-files` and `citar-open-links` wrappers around `citar--select-resource`, I changed `citar--get-resource-candidates` to return the completion category for the list it returns; it only uses `multi-category` if there is more than category of item.

The old group function would only assign groups to the candidates if they had the `multi-category` property. But Vertico (and probably other completion UIs) ignore the transformation if there is no group assigned. This created a bug, where you would see transformed candidates only for `citar-open` (a `multi-category` completion) but not for `citar-open-files`.

While fixing that, I decided to restructure the group function so there are three separate branches, for notes, files, and links. Since each branch explicitly both transforms the resource and assigns it to a group, it's much harder to accidentally do one without the other. I thought the small increase in lines of code was worth it for the clarity.

This naturally raised the question of how to transform notes. In the old code, all candidates that satisfied `file-regular-p` were transformed regardless of whether they were notes or library files. Instead of conflating files and notes, I figured it was cleaner for note sources to explicitly specify the transformation via a `:transform` function.

roshanshariff:
> So you're no longer concerned about this?

My concern was accessing the cache multiple times, which could potentially cause it to be updated in the middle of an operation and thus inconsistent data.

To solve this, I've added a new buffer-local variable called `citar--entries` which acts as a very short-lived cache. It is usually nil, but when it isn't, `citar-get-entries` returns it instead of accessing the main cache and `citar-get-entry` looks up keys in it. So none of the Citar API functions will access the cache when `citar--entries` is non-nil.

Along with that, all functions that perform an extended operation that might otherwise access the cache multiple times (like formatting candidates) now let-bind `citar--entries`:
```elisp
(let ((citar--entries (citar-get-entries)))
  ...)
```
The effect is that at the beginning of the operation, all the entries are extracted from the cache and saved in `citar--entries`; this will trigger whatever updates are necessary. For the duration of the let-binding, all functions that would otherwise access the cache will instead use those saved entries. Afterwards, `citar--entries` reverts to nil, so there is no problem with it getting out of date.

The benefit is that all the code (including the `hasfile`/`hasnote` predicates) can just use the API normally to get data, but things are set up so they don't access the cache when they do so.

citar.el
roshanshariff: I just tested it and `citar-open` doesn't actually work when opening notes with `citar-org-roam`:
1. Set `citar-notes-source` to `'citar-org-roam`.
2. Run `citar-open` and select a ref with notes.
3. Select a note and press RET.

It just opens a new file named like `"citekey nodeid"`. This makes sense because `citar-open` calls `citar--open-multi`, which has no idea what to do with a string returned by a note source's `:items` function. It just treats it like a filename and calls `find-file` on it.

bdarcus: Ok.

I don't know what happened. Could have sworn it worked on main.

But I'm kind of distracted ATM.

aikrahguzar:
I have tried out the new interface and it works very nicely in the limited experience I have had with it. 

And trying it I have started thinking about one improvement it can potentially bring to my workflow. I only ever insert entries to my bibtex files using an emacs interface to https://inspirehep.net/ while allows me search and insert bibtex for high energy physics papers using a keybind. With the new design it seems like it should be straightforward for this same keybind to update the cache too. I am thinking of a function `citar-insert-entries-and-update-cache` which inserts some entries into a bibtex file but if the files is cached it also parses just those entries and inserts them into the cache along with updating the hash for that file and I think mtime so that there is no cache reloading when I need to cite something the next time.

I will probably write such a function, but I am not sure if it belongs to citar as there might be edge cases. There are some I can think of which can be resolved by first checking that the cache is current by comparing the hash but I don't know if I am missing others. Any opinions on it?

roshanshariff:
> Any opinions on it?

I understand the current practical need for such a solution, but at the same time I would be worried about losing the conceptual clarity that comes from knowing that the cache is only ever updated by calling `parsebib-parse` on a file. I would worry about subtle errors that make the cache diverge from the bibliography on disk.

Is your bibliography file too big to just reparse whenever you insert a new entry? Another simpler option would be to have a separate bib file that you add new entries to. Whenever that file grows big enough that parsing it is too slow, you'd just append its contents to your main bibliography.

It only takes about half a second to parse a bibliography with ~1300 entries, so you wouldn't have to do this too often; I understand if that doesn't appeal to you though.

bdarcus:
> Another simpler option would be to have a separate bib file that you add new entries to. Whenever that file grows big enough that parsing it is too slow, you'd just append its contents to your main bibliography.

I added that as an explicit recommendation to the README, in part so people understand the cache design better.

bdarcus:
Seems like a shell script that automates that would be a good addition to the wiki?

aikrahguzar:
> I understand the current practical need for such a solution, but at the same time I would be worried about losing the conceptual clarity that comes from knowing that the cache is only ever updated by calling `parsebib-parse` on a file. I would worry about subtle errors that make the cache diverge from the bibliography on disk.

I agree with this, those divergences would be hard to track down if they happen. And speed is definitely not a concern for me. The new cache generation is very fast and most of the my bib files have less than 200 entries so even the first run feels instantaneous.

citar.el
bdarcus: OK, I see now. I'm not sure how to fix it ATM though.

bdarcus: I see the problem, though not yet a solution:

`citar--select-resource` returns a plain string.

bdarcus: We might need an additional note property; a function to say what to do with the string?

But we still have no way of knowing it's a note.

roshanshariff:
I think the best way to fix `citar-open` is:
1. `citar--get-resource-candidates` annotates all candidates with an extra `citar--resource` text property, which is either `file`, `url`, or `note` (instead of just annotating notes). We want to use a new property with no meaning for other packages, not reuse `multi-category`; we only use that if there are actually multiple categories.
2. `citar--select-resource`, after it gets the selected string from `completing-read`, looks up that string in the candidates list to get the propertized version, retrieves the `citar--resource` property, and returns that along with the string.
3. `citar-open` acts appropriately, calling `citar-file-open`, `browse-url`, or the note `:open` function as appropriate.

I think the `:open` function should just open notes, not create them; its argument should be a string returned by the `:items` function. What is now the `:open` function should be called `:create`, since it takes the citation key and entry and creates new notes.

Personally, I think this makes for a more logical notes API: `:items` returns a list of strings, any of which can be passed to `:open`. They can also be annotated with `:annotate` or transformed with `:transform`, but otherwise are not interpreted or modified in any way. And if you want to create a new note, you call `:create`.

As an aside, doing (1) also simplifies the grouping function a great deal, since it doesn't have to guess what type of resource it's been given.

I can implement this plan if it sounds acceptable.

bdarcus:
Yes, that's acceptable; it's basically what I thinking (the lookup part).

Thanks!

bdarcus:
Not sure if my second message showed up on embark implications?

bdarcus:
I see @roshanshariff pushed the note fixes. Thank you!

So far, we haven't gotten any reported bugs or suggestions, except our own. And I believe we've resolved all.

Anything else that needs done before I merge, and tag 1.0?

bdarcus:
I merged this. Will hold off on tagging 1.0.

roshanshariff:
Would it be okay if we held off on tagging 1.0 for a week or two? There are some parts of the code base I would like to review more thoroughly. It's possible that a Citar 1.0 release will get some publicity, and it would be good to minimize the chances of bugs that might put off new users if they encounter them.

There are also a number of TODOS we should probably resolve as well. Maybe we could just do another point release in the meantime?

bdarcus:
Yes; sure thing.

Should I go the point release now, or wait?

roshanshariff:
> Should I go the point release now, or wait?

I don't _think_ we've introduced any serious new bugs, so I'd be okay with releasing now :sweat_smile:

bdarcus:
> Would it be okay if we held off on tagging 1.0 for a week or two? There are some parts of the code base I would like to review more thoroughly.

Where do we stand ATM @roshanshariff? 

You obviously were able to address at least some of this with the notes functions, but I wasn't sure if there were other pieces of the code you felt needed review before we tag 1.0?

I know your time is limited these days, and mine will likely be for a couple/few months, so it's just a question of when the right time to pull the trigger is, without delaying too long.

From my POV, it certainly makes sense to close #507, but I _think_ other issues can wait, though I did assign a few others to this milestone.

https://github.com/emacs-citar/citar/milestone/3

aikrahguzar:
One thing I noticed while testing my recent changes on an old file: the new caching seems to be case-sensitive on the field names of the entry i.e. "Title" field is not equivalent to "title" now while it used to be. Maybe some uses of `assoc-string` string got replaced with something or no-longer have optional case-fold argument non-nil. 

I don't know if this was intended and I don't think it is crucial since this can be resolved by adding both cased versions in the template.  But seems like something that should be settled before 1.0 whatever the answer here is.

bdarcus:
That sounds like an oversight.

It could be changes in parsebib maybe?

I do vaguely remember we needed to use `assoc-string` because of inconsistencies between bibtex and JSON as parsebib parses it (the former as keys as strings, and the latter as symbols).

It should remain the case that it works for both formats, as should ignoring case.

@roshanshariff?

roshanshariff:
@aikrahguzar, thanks for spotting the case-sensitivity bug. In 64c332a, I've added back the case-fold option to the `assoc-string` in `citar-get-value` which got lost in the reshuffle.

@bdarcus, I think the main pending issue for 1.0 is the performance regression in #659. I did some preliminary investigation there and on my (fairly old) computer I don't see the problem right away (loads that bibliography in ~5 sec). I'll have to investigate a little more to see where the slowdown is.

Other than that, since we did the notes refactoring, I don't foresee any needed changes to external APIs in the immediate future. There's just the matter of mopping up the small bugs you've already tagged under the v1.0 milestone (and any others that might be uncovered in the meantime).

bdarcus:
Only question I have is if not using multi-category has implications for
embark?

roshanshariff:
It shouldn't make a difference to Embark, because from what I remember of its code, it doesn't distinguish between single-category completion and multi-category with all items having the same category.

It might make a difference to other hypothetical packages that interact with the completion system, since specifying a single category is part of the standard Emacs completion API, whereas the multi-category mechanism is a non-standard extension that is followed by Embark, Marginalia, etc.

bdarcus:
Seems some sort of GitHub hiccup here. The comment says I posted that today. But I don't even remember doing that, which means it was probably months ago!

Shrug.

PR #652, opened 2022-07-09
add citar-open-gscholar command

teeann:
During my research workflow, I often find myself querying Google Scholar for academic papers' info.
This PR aims to make it more convenient to do so in `citar`.
Any comment is welcome!

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -1189,6 +1189,17 @@ For use with `embark-act-all'."
     (bibtex-search-entry key t nil t)))
 
 ;;;###autoload
+(defun citar-open-gscholar (key-entry)
+  "Open Google Scholar query result associated with the KEY-ENTRY.
+With prefix, rebuild the cache before offering candidates."
+  (interactive (list (citar-select-ref
+                      :rebuild-cache current-prefix-arg)))
+  (let* ((entry (cdr key-entry))
+         (title (alist-get "title" entry nil nil #'equal))
+         (url (url-encode-url (format "https://scholar.google.com/scholar?q=%s" title) )))
+    (browse-url-xdg-open url)))
+
+;;;###autoload
 (defun citar-insert-bibtex (keys)
   "Insert bibliographic entry associated with the KEYS."
   (interactive (list (citar-select-refs)))


bdarcus:
Thanks for this!

I'm not sure this is the right approach, given how diverse these lookup services are. We wouldn't want to end up with separate commands for all of them.

And I'm not sure how best to do it.

But, have you seen this in the wiki (where we often put preliminary code ideas)?

https://github.com/emacs-citar/citar/wiki/Embark#embark-become

A gacholar option isn't included there, but could easily be added.

Finally, you should update your repo; I merged a major breaking change last week, so your code won't work ;-)

But adjusting it will be easy enough.

citar.el
```diff
@@ -1188,6 +1188,17 @@ For use with `embark-act-all'."
   (when-let ((bibtex-files (citar--bibliography-files)))
     (bibtex-search-entry key t nil t)))
 
+;;;###autoload
+(defun citar-open-gscholar (key-entry)
```
bdarcus, line 1192:
```suggestion
(defun citar-open-gscholar (key)
```

citar.el
```diff
@@ -1188,6 +1188,17 @@ For use with `embark-act-all'."
...
+  "Open Google Scholar query result associated with the KEY-ENTRY.
+With prefix, rebuild the cache before offering candidates."
+  (interactive (list (citar-select-ref
+                      :rebuild-cache current-prefix-arg)))
+  (let* ((entry (cdr key-entry))
```
bdarcus, line 1197:
```suggestion
  (let* ((entry (citar-get-entry key))
```

citar.el
bdarcus: Examples of what would need to be changed for the new API.

teeann:
Thanks for the review, I've updated my code.
Regarding the overhead in dealing with different lookup services, I think adding this single query command for **Google Scholar** to `citar` is sufficient.
**Google Scholar** is already a popular and comprehensive lookup service for papers.
Alternatively, we could wrap `citar-open-gscholar (key)` inside `citar-web-lookup-entry (key, url-query)`. Supporting other lookup services is then trivial.

roshanshariff:
I think in the long term it would be nice if we had a way to handle multiple of these kinds of "open" shortcuts without needing a separate command for each. Maybe the UI could be another section in `citar-open` which would offer things like "Open in Google Scholar".

In the meantime, maybe the Wiki is a good home for these functions, which we could mention in the Readme. Or perhaps a `citar-extras` package. After all, the API exists to make these sorts of integrations possible.

bdarcus:
> Thanks for the comment! I'll now close this PR

And thanks again for submitting.

Please do put it on the wiki if you can, and we can maybe come back to the idea later.

PR #656, opened 2022-07-15
Use citar-always-prompt as default value when selecting resource

furkanusta:
Otherwise citar-open-files prompts even for a single resource

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -640,7 +640,7 @@ CANDIDATES."
              (annotate (citar--get-notes-config :annotate)))
     (funcall annotate (substring-no-properties candidate))))
 
-(cl-defun citar--select-resource (keys &key files notes links (always-prompt t))
+(cl-defun citar--select-resource (keys &key files notes links (always-prompt citar-always-prompt))
   ;; FIX the arg list above is not smart
   "Select related FILES, NOTES, or LINKS resource for KEYS.
 


furkanusta:
I get what you mean. I had a problem with the existing behavior and added this fix to solve it, didn't think about other use cases.

In my case I have one big Notes.org file with each paper as a heading and I have ROAM_REFS property with citation key.
I use the function below to open the corresponding PDF, and I wanted to not ask since I (almost) always have a single file associated with a key.

```emacs-lisp
(defun my-citar-open-current-pdf ()
    "Open REFs of the node at point."
    (interactive)
    (let ((keys (when-let* ((prop (org-entry-get (point) "ROAM_REFS" t))
                            (refs (when prop (split-string-and-unquote prop)))
                            (oc-cites
                             (seq-map (lambda (ref) (substring ref 7 (- (length ref) 1))) refs)))
                  oc-cites)))
      (if keys
          (let ((citar-file-open-function (lambda (file)
                                            (other-window 1)
                                            (find-file file))))
            (citar-open-files (car keys)))
        (user-error "No ROAM_REFS found"))))
```

I thought about changing `citar--select-resource` to pass `:always-prompt` key but then realized there is already an existing variable, so why not use it?

For me both solutions work. If you do not want to change the existing behavior I am open for a workaround on my side as well (if possible).

bdarcus:
I'll think on it a bit.

Can you post your function on the wiki, regardless?

Also, seen this?

https://github.com/emacs-citar/citar-org-roam

roshanshariff:
Maybe it makes sense for `citar-open-files`, `citar-attach-files`, and `citar-open-links` to never prompt for single resources, because of what @bdarcus said. The other option would be to use the `citar-always-prompt` variable for all open commands, as in this PR, but then we should adjust its docstring.

furkanusta:
I've added the function to wiki and I didn't know citar-org-roam that looks quite useful. Thanks!

roshanshariff:
This issue should be addressed by the changes to `citar-open-prompt` in #651. The default value will make Citar always prompt for `citar-open`, `citar-attach-files`, and `citar-open-note`, but not for any other commands when there's only a single item to act on. You can customize the value of the list and add any other commands you don't want to be prompted for.

PR #658, opened 2022-07-16
Use citar--select-resource for citar-open-notes

localauthor:
Is there a reason why `citar-open-notes` doesn't use `citar--select-resource`?

I've got a few cases of multiple notes for a key and I wanted a chance to select one, rather than have all notes opened at once. Not sure this is the right answer, but maybe in the right direction?

Also, the message in the last line of `citar-open-notes`  tells the user to "Select one key to create note". Was that meant to be interactive? I've thrown in a possible way of doing that here.

Thoughts?

### DIFF
--- a/citar-file.el
+++ b/citar-file.el
@@ -325,12 +325,12 @@ extensions in `citar-file-note-extensions'."
 (defun citar-file--create-note (key)
   "Create a note file from KEY and ENTRY."
   (let ((entry (citar-get-entry key)))
-  (if-let ((filename (citar-file--get-note-filename key)))
-      (prog1 (find-file filename)
-        (unless (file-exists-p filename)
-          (citar--check-configuration 'citar-note-format-function)
-          (funcall citar-note-format-function key entry)))
-    (user-error "Make sure `citar-notes-paths' and `citar-file-note-extensions' are non-nil"))))
+    (if-let ((filename (citar-file--get-note-filename key)))
+        (prog1 (find-file filename)
+          (unless (file-exists-p filename)
+            (citar--check-configuration 'citar-note-format-function)
+            (funcall citar-note-format-function key entry)))
+      (user-error "Make sure `citar-notes-paths' and `citar-file-note-extensions' are non-nil"))))
 
 (defun citar-file--get-notes (keys)
   "Return list of notes associated with KEYS."
--- a/citar-file.el
+++ b/citar-file.el
@@ -322,14 +322,15 @@ extensions in `citar-file-note-extensions'."
   (let ((files (citar-file--note-directory-files)))
     (lambda (key) (gethash key files))))
 
-(defun citar-file--create-note (key entry)
+(defun citar-file--create-note (key)
   "Create a note file from KEY and ENTRY."
+  (let ((entry (citar-get-entry key)))
   (if-let ((filename (citar-file--get-note-filename key)))
       (prog1 (find-file filename)
         (unless (file-exists-p filename)
           (citar--check-configuration 'citar-note-format-function)
           (funcall citar-note-format-function key entry)))
-    (user-error "Make sure `citar-notes-paths' and `citar-file-note-extensions' are non-nil")))
+    (user-error "Make sure `citar-notes-paths' and `citar-file-note-extensions' are non-nil"))))
 
 (defun citar-file--get-notes (keys)
   "Return list of notes associated with KEYS."
--- a/citar.el
+++ b/citar.el
@@ -622,7 +622,7 @@ CANDIDATES."
            (sources (nconc (when files (list (cons 'file (withtype 'file files))))
                            (when links (list (cons 'url (withtype 'url links))))
                            (when notes (list (cons notecat (withtype 'note notes))))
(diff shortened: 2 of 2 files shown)


roshanshariff:
I think the current implementation of `citar-open-notes` is a bit of a placeholder, pending some [design questions](https://github.com/emacs-citar/citar/pull/634#issuecomment-1166624864). It has some problems, even with this PR:
* It only lets you create notes if _none_ of the selected keys already have notes. You can't open some existing notes, but also create others.
* You can only create one note, even if you selected multiple keys.

After thinking about it some more, I'm proposing the following design for `citar-open-notes`:
1. As we do now, prompt for a list of keys (or get them from `embark-act` or `embark-act-all`). Compile a list of all notes associated with the chosen keys, as well as a list of the keys that don't have any notes.
2. Use `completing-read` with a group function that creates an "Open Notes" group for the notes that already exist, and "Create Notes" for the keys that don't yet have notes.
3. This should a be a multi-select, so you can choose which notes to open and/or create.
4. (Customizable) Pre-select the existing notes? If enabled, just pressing RET will open all existing notes, but create none. Otherwise, you can de-select any notes you don't want to open and then press RET. You'd disable this option if you have lots of notes for the same keys and usually just open one.
5. (Optional) Instead of just listing the keys that don't have notes, display them using a new template in `citar-templates`.

For consistency, we can make `citar-open` also show a "Create Notes" section as above, so that `citar-open-notes` would be just like `citar-open` except omitting all the non-note resources. We would also make `citar-open` a multi-select so you can open several resources at once.

Some benefits of this design:
* It should work well whether you usually have multiple notes per key or not, since you can choose which ones to open (or not to open, by enabling 4 above).
* You explicitly opt-in to creating notes, so you don't accidentally create a bunch of them and have to track down their buffers. This was the reason for the current restriction of creating a single note.
* There's a minimum of sequential menus/interactions. Just one completing-read for selecting keys and then another for selecting notes to open/create, just like `citar-open`.
* `citar-open` just shows a superset of `citar-open-notes`, combining it with `citar-open-files` and `citar-open-links`.
* The `citar-create-note` command is no longer essential; we can still keep it for creating just a single note, perhaps for use in a bigger workflow. You'd use it to create another note for a key that already has one, which is otherwise impossible with the above design (or create a note when a cross-referenced key already one, e.g. for a book chapter when there's already a note for the book).

Any thoughts on this basic idea, or refinements? It'll take a bit of work to implement, but I don't think it'll be too hard once we settle on the UX.

bdarcus:
I like it in general, and don't have any specific suggestions.

We do keep relying on grouping (and of course affixation); I wonder at what point we require Emacs 28?

roshanshariff:
> I wonder at what point we require Emacs 28?

Vertico/Selectrum/etc. support grouping and affixations well enough for our purposes even on 27, so I'm not sure we should have a hard dependency on emacs 28. That would just make life harder for people who can't upgrade yet (like me, at least for the moment). I think the experience wouldn't be too bad even without grouping/affixation, but installing a modern completing-read is strongly preferred for older emacs.

> maybe we somehow split interactive and non-interactive functions?

That's a good point. I'll keep in mind that we should provide simple non-interactive functions for all the basic create/open operations if the interactive commands don't have appropriate semantics.

bdarcus:
> Vertico/Selectrum/etc. support grouping and affixations well enough for our purposes even on 27

I didn't know that; good to know though!

localauthor:
Thanks for the review, @roshanshariff! I just pushed a commit that addresses some of your suggestions. I just wasn't sure how to make multiple-selection work...

roshanshariff:
@localauthor, do you mind if I implement this differently in a new PR?

A lot of the complexity this adds to `citar-open-notes` should be in `citar--select-resource`, I think. It already does very similar things for other resource types (files/links) so adding a "create note" resource there would be much simpler. There's also some work to be done to set up Embark. And I think the note-source API also needs some small changes so we don't have to call the "get-items" function in a loop.

localauthor:
No, go for it! I’ll close this one with a link to your PR.

localauthor:
I think the latest commit might approach something like what you've suggested in your last comment.

I'm just playing around at this point though, so feel free to ignore and carry on as you intend. :)

bdarcus:
@roshanshariff 

> Compile a list of all notes associated with the chosen keys, as well as a list of the keys that don't have any notes.

Would a user be able to both open some notes and create others in the same action?

The one question I could think of ATM.

localauthor:
> Would a user be able to both open some notes and create others in the same action?

With the latest commit in this branch, this works with `embark-act-all`. Not sure how feasible it would be to implement without embark, esp. without using `completing-read-multiple`.

bdarcus:
> > Would a user be able to both open some notes and create others in the same action?
> 
> With the latest commit in this branch, this works with `embark-act-all`. Not sure how feasible it would be to implement without embark, esp. without using `completing-read-multiple`.

If we thought this a valuable requirement (IDK), we could use our own homegrown multiple select (as in `citar-select-refs`)?

I haven't looked; perhaps that code would need to be generalized?

localauthor:
Closed in favor of #661

PR #664, opened 2022-07-20
Make `citar-file-open` use customizable functions to open files

### LINKED ISSUE #507: Make file-open more configurable

ShuguangSun:
The new customization variable `citar-file-open-functions` is an alist from file extensions to functions used to open them. `citar-file-open` will now use this variable to decide how to open library files.

Fixes #507

### DIFF
--- a/citar-file.el
+++ b/citar-file.el
@@ -82,7 +82,7 @@ separator that does not otherwise occur in citation keys."
                  (const :tag "Find files with space after key" "[[:space:]]")
                  (regexp :tag "Filename separator")))
 
-(defcustom citar-file-open-external-extension '("html")
+(defcustom citar-file-open-external-extensions '("html")
   "List of file extensions to be opened with external application.
 
 These are the extensions the `citar-file-open-external'
--- a/citar-file.el
+++ b/citar-file.el
@@ -82,6 +82,14 @@ separator that does not otherwise occur in citation keys."
                  (const :tag "Find files with space after key" "[[:space:]]")
                  (regexp :tag "Filename separator")))
 
+(defcustom citar-file-open-external-extension '("html")
+  "List of file extensions to be opened with external application.
+
+These are the extensions the `citar-file-open-external'
+will open, via `citar-file-open'."
+  :group 'citar
+  :type '(repeat string))
+
 (defvar citar-notes-paths)
 (defvar citar-library-paths)
 (defvar citar-library-file-extensions)
@@ -288,7 +296,9 @@ need to scan the contents of DIRS in this case."
 
 (defun citar-file-open (file)
   "Open FILE."
-  (if (equal (file-name-extension file) "html")
+  (if (cl-member (downcase (file-name-extension file))
+                 citar-file-open-external-extension
+                 :test #'equal)
       (citar-file-open-external (expand-file-name file))
     (funcall citar-file-open-function (expand-file-name file))))
 
--- a/citar-file.el
+++ b/citar-file.el
@@ -296,9 +296,7 @@ need to scan the contents of DIRS in this case."
 
 (defun citar-file-open (file)
   "Open FILE."
-  (if (cl-member (downcase (file-name-extension file))
-                 citar-file-open-external-extension
-                 :test #'equal)
+  (if (member-ignore-case (file-name-extension file) citar-file-open-external-extensions)
       (citar-file-open-external (expand-file-name file))
     (funcall citar-file-open-function (expand-file-name file))))
 
--- a/citar-file.el
+++ b/citar-file.el
@@ -82,13 +82,21 @@ separator that does not otherwise occur in citation keys."
                  (const :tag "Find files with space after key" "[[:space:]]")
                  (regexp :tag "Filename separator")))
 
-(defcustom citar-file-open-external-extensions '("html")
+(defcustom citar-file-open-external-extension-alist '(("html" . system))
   "List of file extensions to be opened with external application.
 
 These are the extensions that `citar-file-open' will open via
 `citar-file-open-external'."
   :group 'citar
-  :type '(repeat string))
+  :type '(repeat (cons string
+                       (choice
+                        (function :tag "Open with function")
+                        (symbol :tag "Open with sytem default application"
+                                system)
+                        (symbol :tag "Open with Emacs"
+                                default)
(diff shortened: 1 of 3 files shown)


citar-file.el
```diff
@@ -82,6 +82,14 @@ separator that does not otherwise occur in citation keys."
                  (const :tag "Find files with space after key" "[[:space:]]")
                  (regexp :tag "Filename separator")))
 
+(defcustom citar-file-open-external-extension '("html")
+  "List of file extensions to be opened with external application.
+
+These are the extensions the `citar-file-open-external'
+will open, via `citar-file-open'."
```
roshanshariff, line 89:
```suggestion
These are the extensions that `citar-file-open' will open via
`citar-file-open-external'."
```

citar-file.el
```diff
@@ -288,7 +296,9 @@ need to scan the contents of DIRS in this case."
 
 (defun citar-file-open (file)
   "Open FILE."
-  (if (equal (file-name-extension file) "html")
+  (if (cl-member (downcase (file-name-extension file))
+                 citar-file-open-external-extension
+                 :test #'equal)
```
roshanshariff, line 301:
```suggestion
  (if (member-ignore-case (file-name-extension file) citar-file-open-external-extensions)
```

citar-file.el
```diff
@@ -82,6 +82,14 @@ separator that does not otherwise occur in citation keys."
                  (const :tag "Find files with space after key" "[[:space:]]")
                  (regexp :tag "Filename separator")))
 
+(defcustom citar-file-open-external-extension '("html")
```
roshanshariff, line 85:
```suggestion
(defcustom citar-file-open-external-extensions '("html")
```

roshanshariff (changes requested):
Thanks, looks good to me! I just suggested some small changes.

ShuguangSun (commented):
updaded accordingly.

roshanshariff (approved):

citar-file.el
```diff
@@ -82,6 +82,14 @@ separator that does not otherwise occur in citation keys."
                  (const :tag "Find files with space after key" "[[:space:]]")
                  (regexp :tag "Filename separator")))
 
+(defcustom citar-file-open-external-extensions '("html")
+  "List of file extensions to be opened with external application.
+
+These are the extensions that `citar-file-open' will open via
+`citar-file-open-external'."
+  :group 'citar
+  :type '(repeat string))
```
bdarcus, line 91: The other approach is suggested in #507, where this might instead be:

```
(defcustom citar-file-open-extensions (list ("html" . 'citar-open-external)
                                            (t . 'find-file)))
```

roshanshariff: Yeah, I agree that it's more future-proof to make this an alist as you suggest. @ShuguangSun, are you willing to implement this? Otherwise I can take a look, but probably not right away.

bdarcus: Just checking back, @ShuguangSun; can you answer @roshanshariff's question?

> ... are you willing to implement this?

E.g. make it match the plans in #507?

Otherwise, we may need to address this post-1.0 (which, hopefully, we can tag soon, pending a couple of minor things).

ShuguangSun:
I'll work on it.

citar-file.el
ShuguangSun: I'll work on it.

citar-file.el
```diff
@@ -82,6 +82,22 @@ separator that does not otherwise occur in citation keys."
                  (const :tag "Find files with space after key" "[[:space:]]")
                  (regexp :tag "Filename separator")))
 
+(defcustom citar-file-open-external-extension-alist '(("html" . system))
```
bdarcus, line 85: Note: we don't need a new defcustom; we need to refactor the current one.

See my minimal example elsewhere on this PR, and on #507 ;-)

ShuguangSun: I think there are four:
- open with system default  which will be easy handled by current `citar-open-external`
- open with a customized application which need a new function for each command or make `citar-open-external` accept some args (and it will make the `citar-file-open-functions-alist` a little complicated, for example, `("pdf" . #'citar-open-external "sumatrapdf")`
- open with a funciton as `citar-file-open-functions-alist` does

Are you happy with somethink like `("pdf" . #'citar-open-external "sumatrapdf")`. If so, I just need to refactor my current code.

bdarcus: Let's see what input @roshanshariff might have. He's been working on related code more recently than I, I believe, and I've been distracted with some IRL things.

I do wonder, however (it's a genuine question) if your `("pdf" . #'citar-open-external "sumatrapdf")` example could just be a lambda instead for the `cdr`?

ShuguangSun: It could be a lambda, but I don't expect user might be familar with lambda or elisp funciton.

bdarcus: Yes, that's true.

bdarcus: @ShuguangSun - your example of `sumatrapdf` reminded me of a discussion on the org list awhile back, where someone was wanting to use `sioyek` (which I'd never heard of). This was his solution; just another real example to check to make sure we have this right:

```elisp
(setq citar-file-open-function '(lambda (file)
 (call-process "sioyek" nil 0 nil
 file)))
```

bdarcus (approved):
Looks good to me; thanks!

@ShuguangSun - you want to give a quick check/test before I merge?

ShuguangSun (commented):
fine update

bdarcus:
Merged; thank you!

Issue #668, opened 2022-07-29
Do not replace characters in paths


autrimpo:
**Describe the bug**
`file` fields are transformed using `parsebib-TeX-markup-replace-alist`, which results in mangled paths.

**To Reproduce**
1. Add a `file` field to an entry containing `--`.
2. Call `citar-open-files`.
3. Get an error about missing files, as `--` has been replaced with `–` (en dash).

**Expected behavior**
The file is opened correctly.

**Screenshots**
Unnecessary?

**Emacs version:**
`GNU Emacs 29.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.34, cairo version 1.16.0)`

**Additional context**
`--` is included in the default alist. It is also used in the Denote naming scheme, creating conflicts.

bdarcus:
Do you have a sense of whether this might be better fixed in parsebib, or if there's an obvious oversight in this codebase?

autrimpo:
Unfortunately, I don't know. I was just trying to figure out why citar couldn't find my files :)

joostkremers:
It's a parsebib issue, because parsebib doesn't allow one to specify which fields to untexify. That needs to be fixed.

It is possible to untexify a single string (see `parsebib-clean-tex-markup`), but that's cumbersome to do from within citar. Better to make parsebib mor flexible.

bdarcus:
Right; so you'd make that change, and we'd want to pass a list of fields to `parsebib-parse`, which would exclude `file` and such (the actual field for this we define in a defcustom)?

roshanshariff:
Depending on how @joostkremers implements this in`parsebib`, do we need any change to `citar--fields-to-parse`? Presumably, `parsebib` would have a list of fields not to untexify, which would include `file`. We would continue to include `file` as one of the fields to parse. Removing `file` from that list would just mean that it's not included in the entry alist, which isn't what we want.

bdarcus:
That's true; it does depend on whether this list of fields here is an exclusion list (more convenient in this case), or an inclusion list. I was assuming the latter.

For sure; let's wait. I was just curious also if we had suggestions for @joostkremers.

joostkremers:
> For sure; let's wait. I was just curious also if we had suggestions for @joostkremers.

[...]

PR #669, opened 2022-07-31
Remove citar--clean-string

bdarcus:
Parsebib now handles removing TeX markup, so we remove it.

See also #668


roshanshariff (approved):
Looks good to me! :+1:

Issue #662, opened 2022-07-19
can not generate local-bib after escape symbols


hy-zhao:
**Describe the bug**
when I cite after an escaping symbol, the following citation item can not be exported to local-bib file.

**To Reproduce**
1. `\%\cite{gaoEfficientKnowledgeGraph2019}`
2. `citar-export-local-bib-file`

**Expected behavior**
Referece item related to `gaoEfficientKnowledgeGraph2019` should be exported. But it doesn't. If i make `\cite{gaoEfficientKnowledgeGraph2019}` to a new line, the export will work. Is it impossible to solve this bug?

**Emacs version: 28**

bdarcus:
Any thoughts @aikrahguzar?

aikrahguzar:
As far as I can tell, for latex  `citar-export-local-bib-file` gets keys from `reftex-all-used-citation-keys`. So my guess is reftex ignores comment but isn't smart about determining what is a comment. I don't think we should try to fix this in `citar` itself since it will be non-trivial and probably introduce more corner cases.

Since reftex is a builtin package @hy-zhao can try to verify if it is indeed a problem with `reftex-all-used-citation-keys` and send a bug report via `M-x report-emacs-bug` if that is the case.

hy-zhao:
Oh yes, I did see the function about reftex you mentioned. And it shows as a void function. Seems so weird.

aikrahguzar:
`symbol's function definition is void: reftex-all-used-citation-keys` means that `reftex-cite` is not loaded. That is actually a problem with `citar`.  `citar-latex` requires `reftex-cite` but if you start emacs open a latex file and the first thing you do is `citar-export-local-bib-file` then you will get that error.

Try this, first insert a citation using `citar-insert-citation` and then try to export evaluate `(reftex-all-used-citation-keys)`, this should be a list of keys. See `gaoEfficientKnowledgeGraph2019` is in that list. If it is not send a bug report using `report-emacs-bug`.

bdarcus:
> I can do a pr but probably not before weekend.

I'll ping you if I can manage time, but it's pretty limited ATM.

I'll tag this with a 1.0 milestone, so we make sure to include the small fix before we tag 1.0.

cc @roshanshariff

bdarcus:
Any chance you have time to fix this @aikrahguzar?

bdarcus:
Partially fixed in #670.

The rest is not our problem :-)

hy-zhao:
[...]

PR #670, opened 2022-07-31
Fix autoload issue + multiple selection corner case

aikrahguzar:
Fixes part of #662 related to auto-loading. 

Also fixes an edge case for the multiple selection. Now on deselection we just restore the history to what it was before instead of trying to remove the topmost element from the history.

### DIFF
--- a/citar-latex.el
+++ b/citar-latex.el
@@ -207,5 +207,8 @@ With ARG non-nil, rebuild the cache before offering candidates."
   (seq-find (lambda (x) (member command (car x)))
             citar-latex-cite-commands))
 
+;;;###autoload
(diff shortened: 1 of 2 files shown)


bdarcus:
Thank you!

aikrahguzar:
I just noticed that, in helpful buffer the first two lines are,
```
citar-latex-list-keys is a function alias for
reftex-all-used-citation-keys, defined in reftex-cite.el.gz.
```
I don't know if the problem is with `helpful` or if I don't understand `defalias` (there is something in its documentation about current load file) but the `defined in reftex-cite.el.gz` above makes me wonder if it would have been better to just define the function.

I does works as expected so this is pretty minor.

Issue #659, opened 2022-07-17
First load is too slow


bdarcus:
On the latest commit (really I think since the change to the new cache system) I've noticed that citar takes 25-30 seconds to load on first being called (and same for any subsequent change in the bibfile). I have a somewhat large .bib file (~7200) but it did not take this long to load it previously. Is there a way to avoid this without splitting my bib file into multiple smaller ones (something I am not intending to do)? Thanks!

_Originally posted by @mclearc in https://github.com/emacs-citar/citar/issues/653#issuecomment-1186601992_

roshanshariff:
@mclerc, I'll have to look into what's causing this, since in my testing a bibliography file of that size took much less time. Would you be willing to share your bibliography, privately if you prefer?

mclearc:
Sure -- here it is.
[bibfile.bib.zip](https://github.com/emacs-citar/citar/files/9128671/bibfile.bib.zip)

mclearc:
Oh and if it is of use here's my config:

```emacs-lisp
(use-package citar
  :straight (:host github :repo "emacs-citar/citar")
  :commands (citar-open-beref
             citar-open-notes
             citar-insert-citation)
  :bind (:map citar-map
         ("b" .  #'citar-open-beref))
  :custom
  ;; Use with org citation
  (org-cite-global-bibliography `(,lem-bibliography))
  (citar-bibliography `(,lem-bibliography))
  (org-cite-insert-processor 'citar)
  (org-cite-follow-processor 'citar)
  (org-cite-activate-processor 'citar)
  :config
  ;; use embark with at-point
  (setq citar-at-point-function 'embark-act)
  (setq citar-default-action 'citar-open-beref)
  ;; add beref entry for bookends
  (setq citar-additional-fields '("doi" "url" "beref"))
  (setq citar-templates
        `((main . " ${=key= id:15} ${title:48}")
          (suffix . "${author editor:30}  ${=type=:12}  ${=beref=:12} ${tags keywords:*}")
          (preview . "${author editor} (${year issued date}) ${title}, ${journal journaltitle publisher container-title collection-title}.\n")
          (note . ,lem-citar-note)))
  (when (display-graphic-p)
    (setq citar-symbols
          `((file ,(all-the-icons-octicon "file-pdf"      :face 'error) . " ")
            (note ,(all-the-icons-octicon "file-text"     :face 'warning) . " ")
            (link ,(all-the-icons-octicon "link-external" :face 'org-link) . " "))))
[...]

PR #674, opened 2022-08-03
v1.0 last call

bdarcus:
I'm planning to merge this and then tag 1.0 "very soon"; ideally once Joost has a chance to improve performance of #659 a bit, and I can bump parsebib to 4.3 instead.

If any last minute changes, @roshanshariff, you can add them here, though I think we're good.

Edit: seeing your latest message, can also hold off until the weekend if you prefer. Just want to get this off my TODO list in the next few days.

If you do want to add a more substantial commit, you can also just do that, and I can then rebase on main.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -8,7 +8,7 @@
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;; Version: 0.9.7
 ;; Homepage: https://github.com/emacs-citar/citar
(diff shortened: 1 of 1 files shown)


roshanshariff:
@bdarcus, if we're okay with the performance, I don't think we have to hold up 1.0. We can always make another point release later that'll improve matters for people with large bibliographies. I don't think there's going to be any API change anyway.

bdarcus:
Sounds good. I'll still wait until tomorrow to see if Joost pushes any commits to address that.

Issue #680, opened 2022-08-25
Generalize parsing for cache


bdarcus:
Just pulling this out, so easier to track.

I think this should be fairly straightforward, though I got a bit stuck trying to implement it.

----------

I would think a clean way to do this would be to allow the cache to use different data sources to retrieve bibliographic data (which is to say, a hash table of entries). Then we would refactor the current `parsebib` code to be one source among others, and there could be another package supporting `org-bibtex`.

There should only be a few integration points:

* For each bibliography filename, you'd need a way to figure out which data source to use.
* We need a function that tells us if a data source has been updated, and if it has, another function that returns a hash table of the entries,

Then the `citar-org-bibtex` implementation would maintain a buffer-local hash table of the keys/entries and update them as text is edited. It could just return this hash table whenever Citar needs it.

The cache optimization I was thinking of was for one layer down: speeding up the display transformation of strings after they've been parsed. That may or may not be needed anymore, and in any case, should be fairly independent of these changes.

_Originally posted by @roshanshariff in https://github.com/emacs-citar/citar/issues/397#issuecomment-1222735452_

Issue #397, opened 2021-11-11
Support org-bibtex as a bibliographic backend


JonathanReeve:
**Is your feature request related to a problem? Please describe.**
I've been really enjoying Citar lately, and I'd love to get it to work with my favorite bibliographic format, `org-bibtex`. 

In org-bibtex, instead of maintaining separate files for your notes and your bibliographic metadata, and therefore requiring functions for jumping between the two, you just have everything written in your org file. The heading which contains your notes has a properties drawer that contains elements like author, title, year, and so on. You don't need `parsebib` or another library to parse it, since it's already in a structured data format:

```org
* Active documents with org-mode
  :PROPERTIES:
  :bib-type: article
  :author: Schulte, Eric and Davison, Dan
  :journal: Computing in Science & Engineering
  :volume: 13
  :number: 3
  :pages: 66--73
  :year: 2011
  :publisher: Institute of Electrical and Electronic Engineers, Inc. 
  :END:

  Here are my notes on this document. This article describes org-mode documents. 
```

Instead of jumping between two files all the time, you can edit your notes and your source's metadata all from the same place, and hide it when you don't need it by closing the property drawer. 

Best of all, org-bibtex is already included in org by default, and it's already supported by [citeproc-el](https://github.com/andras-simonyi/citeproc-el). 

**Describe the solution you'd like**
Add a backend for bibliographic metadata that reads the properties of note files, and, if those properties include the bibliographic type, use that as the metadata source.

bdarcus:
I really know nothing about org-bibtex, and what would be involved in supporting it. 

Any input on this @andras-simonyi?

bdarcus:
I wonder if this is something parsebib can support?

andras-simonyi:
I think it would be rather trivial to add some support for that format utilizing the functions provided by `ol-bibtex.el` in Org. I'm probably not familiar enough with Citar to come up with a PR but could post an outline of the required functions/changes here. OTOH, adding support to `parsebib` seems to be a better (more general) solution. @bdarcus, @joostkremers, WDYT?

andras-simonyi:
[...]

PR #683, opened 2022-08-29
Make bib parser configurable

bdarcus:
Implement #680.

This should be enough to enable #397?

Please review and comment. 

--------------

## On the org-bibtex parser, etc

I don't think I have time to do the org-bibtex parser and updatep functions. Would be great if someone can test this with that.

The first should look something like the below, and return a hash-table exactly like `parsebib-parse` does:

```elisp
(cl-defun citar-cache--org-bibtex-parser (files &optional &key entries fields)
  "Return a parsebib-compatible hash-table for a list of org-bibtex FILES."
  ;; FILES is a list of file paths, though in citar, will only be a single item list
  (map-into                                ; convert alist into hash-table
    (org-element-cache-map ...
```

EDIT - I tried to figure out how to add the parser to parsebib on this branch:

https://github.com/bdarcus/parsebib/tree/org-bibtex

But I ran into a UX and technical issue: to take advantage of `org-element-cache-map`, you can't use temp-buffers, and need to be in the org major-mode. So unlike the other formats, it would seem to be more sensible to open all listed org files. But that's a departure from how parsebib and citar work. And what happens if someone does something crazy like do file-per-reference using org?

This may suggest that parser is better off here, and we just clearly document the different workflows.

### DIFF
--- a/citar-cache.el
+++ b/citar-cache.el
@@ -15,6 +15,7 @@
 (require 'citar-format)
 (require 'seq)
 (require 'map)
+(require 'files)
 
 (declare-function citar--get-template "citar" (template-name))
 (declare-function citar--fields-to-parse "citar" ())
@@ -142,13 +143,14 @@ need them."
          (cachedfmtstr (and cached (citar-cache--bibliography-format-string cached)))
          (props (citar-cache--get-bibliography-props filename cachedprops))
          (fmtstr (citar--get-template 'completion))
+         (ext (file-name-extension filename))
          (bib (or cached (citar-cache--make-bibliography filename))))
     (prog1 bib
       ;; Set the format string so it's correct when updating bibliography
       (setf (citar-cache--bibliography-format-string bib) fmtstr)
       ;; Update bibliography if needed or forced
       (if (or force-update
-              (citar-cache--update-bibliography-p cachedprops props))
+              (funcall (citar-cache--get-update-p ext) cachedprops props))
           (citar-cache--update-bibliography bib props)
         ;; Otherwise, update props anyway in case mtime has changed:
         (setf (citar-cache--bibliography-props bib) props)
@@ -209,7 +211,7 @@ re-hashing the file contents."
                    (buffer-hash)))))
     `(:size ,size :mtime ,mtime :hash ,hash :fields ,fields)))
 
-(defun citar-cache--update-bibliography-p (oldprops newprops)
+(defun citar-cache--update-default-bibliography-p (oldprops newprops)
   "Return whether bibliography needs to be updated.
 Compare NEWPROPS with OLDPROPS and decide whether the file
 contents have changed or the list of bibliography fields to be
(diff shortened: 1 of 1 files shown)


citar-cache.el
```diff
@@ -218,6 +219,19 @@ parsed is different."
             (equal (plist-get oldprops :hash) (plist-get newprops :hash))
             (equal (plist-get oldprops :fields) (plist-get newprops :fields)))))
 
+;; Parsers
+
+(defvar citar-cache--parser
+  (list (cons t #'parsebib-parse)))
```
bdarcus, line 225: This maybe should be a triplet, with the last arg an updatep function.

citar-cache.el
```diff
@@ -209,7 +211,7 @@ re-hashing the file contents."
                    (buffer-hash)))))
     `(:size ,size :mtime ,mtime :hash ,hash :fields ,fields)))
 
-(defun citar-cache--update-bibliography-p (oldprops newprops)
+(defun citar-cache--update-parsebib-bibliography-p (oldprops newprops)
```
bdarcus, line 214: This probably needs a more general name. Maybe "parsebib" -> "default"?

citar-cache.el
```diff
@@ -209,7 +211,7 @@ re-hashing the file contents."
                    (buffer-hash)))))
     `(:size ,size :mtime ,mtime :hash ,hash :fields ,fields)))
 
-(defun citar-cache--update-bibliography-p (oldprops newprops)
+(defun citar-cache--update-default-bibliography-p (oldprops newprops)
```
bdarcus, line 214: One question on this; what's the implications of the very last line of this docstring for this PR?

citar-cache.el
```diff
@@ -218,6 +220,38 @@ parsed is different."
             (equal (plist-get oldprops :hash) (plist-get newprops :hash))
             (equal (plist-get oldprops :fields) (plist-get newprops :fields)))))
 
+;; Parsers
+
+(defconst citar-cache--parser
+  (list
+   (cons t (cons #'parsebib-parse #'citar-cache--update-default-bibliography-p)))
```
bdarcus, line 227: This is maybe a bit ugly; does that matter?

aikrahguzar:
I am not sure how much this will help. The configuration it introduces is seems unlikely to be used outside of citar since a new parser is likely a lot of work. If the intention is to allow formats not supported by parsebib think a simpler way would be to essentially define an analog of `parsebib-parse` that does the calls to `parsebib-parse-bib-buffer` etc as `parsebib-parse` and includes the additional calls too. If a user really wants to support some format outside of citar, this function can be advised.

bdarcus:
That's a good point.

I went this way because I wasn't sure if we want the org-bibtex parser in the citar package, or in a separate package. I'm still not sure on that actually.

WDYT?

aikrahguzar:
> I went this way because I wasn't sure if we want the org-bibtex parser in the citar package, or in a separate package. I'm still not sure on that actually, largely because there are outstanding UX and performance issues to sort out there.
> 
> WDYT?

So the problematic scenario is when the parser is in a separate package and we don't want to depend on it? I think the soft dependency can be handled via `featurep` check inside the citar version of  `parsebib-parse` and in that case we can warn users that they need the package installed if we are called to parse an org-bibtex file.

bdarcus:
@aikrahguzar 

> So the problematic scenario is when the parser is in a separate package and we don't want to depend on it?

That's the immediate issue I was thinking of. But perhaps there could be other in the future?

PR #709, opened 2022-10-30
Call predicate on candidate display string (fix #708)

### LINKED ISSUE #708: Citar with vanilla Emacs (no Vertico)
**Describe the bug**

Citar does not work well with Emacs' default completions.

**To Reproduce**

1. Run `citar-open`
2. Type `TAB`

**Expected behavior**

The completions buffer appears, like in `M-x` or `C-x b`.

**Emacs version:**

29

**Additional context**

- When I type something at the "Reference:" prompt and then press `TAB`, nothing happens either.  The newly typed text disappears and Citar asks again (and again).

- Everything works well with Vertico.

oantolin:
citar--completion-table mistakenly called the completion system's predicate on the minibuffer input typed by the user instead of calling it on the completion candidate, a display string for a reference.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -513,9 +513,9 @@ and other completion functions."
         ;; REVIEW this now works, but probably needs refinement
         (let ((predicate
                (when (or filter predicate)
-                 (lambda (_ key)
+                 (lambda (cand key)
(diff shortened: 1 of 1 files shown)


bdarcus:
Thank you!

PR #710, opened 2022-11-01
Create limited itemgetter for citeproc-formatting

### LINKED ISSUE #702: `citar-copy-reference` slow?
Thanks for this great package.

I noticed that it takes Emacs 5–7 seconds after I invoke `citar-copy-reference` to actually copy the formatted reference to the kill ring. Emacs runs generally pretty fast on my current machine and I very rarely notice these sorts of delays, so I was wondering if this behavior is normal or if there is something that could be done to improve performance (this wouldn't normally bother me, but I use `citar-copy-reference` dozens of times per day on average). I attach a profiler report, in case it helps, and the relevant parts of my config follow.

[profiler.txt](https://github.com/emacs-citar/citar/files/9804546/profiler.txt)

``` emacs-lisp
(use-package oc
  :demand t
  :custom
  (org-cite-global-bibliography `(,ps/file-bibliography))
  (org-cite-csl-styles-dir ps/dir-csl-styles)
  (org-cite-csl-locales-dir ps/dir-csl-locales)
  (org-cite-follow-processor 'citar) ; `org-open-at-point` integration

(use-package citar
  :straight (citar :type git :host github :repo "emacs-citar/citar" :includes (citar-org))
  :demand t
  :after (oc vertico embark marginalia citeproc)
  :custom
  (citar-open-note-functions '(orb-citar-edit-note))
  (citar-bibliography org-cite-global-bibliography)
  (citar-notes-paths `(,ps/dir-bibliographic-notes))
  (citar-at-point-function 'embark-act)
  (citar-symbols
     (link ,(all-the-icons-octicon "link" :face 'all-the-icons-orange :v-adjust 0.01) . " ")))
  (citar-symbol-separator "  ")
  (citar-format-reference-function 'citar-citeproc-format-reference))

(use-package citar-citeproc
  :after citar
  :demand t
  :custom
  (citar-citeproc-csl-styles-dir org-cite-csl-styles-dir)
  (citar-citeproc-csl-locales-dir org-cite-csl-locales-dir))

```

andersjohansson:
This avoids parsing the full file.

This an adaptation from `citeproc-hash-itemgetter-from-any`, but I have only implemented the json-parsing yet (what is needed for me). Seems to work fine and much faster already.

Partially fixes #702

### DIFF
--- a/citar-citeproc.el
+++ b/citar-citeproc.el
@@ -98,13 +98,44 @@ With prefix-argument, select CSL style."
                   (expand-file-name citar-citeproc-csl-style citar-citeproc-csl-styles-dir)))
          (bibs (citar--bibliography-files))
          (proc (citeproc-create style
-                                (citeproc-hash-itemgetter-from-any bibs)
+                                (citar-citeproc--limited-itemgetter-by-keys bibs keys)
                                 (citeproc-locale-getter-from-dir citar-citeproc-csl-locales-dir)
                                 "en-US"))
          (references (car (progn
                             (citeproc-add-uncited keys proc)
                             (citeproc-render-bib proc 'plain)))))
     references))
+
+(defun citar-citeproc--limited-itemgetter-by-keys (file-or-files keys)
+  "Return a getter for FILE-OR-FILES in any supported format, limited to KEYS."
+  (let ((files (if (listp file-or-files)
+                   file-or-files
+                 (list file-or-files)))
+        (cache (make-hash-table :test #'equal)))
+    (dolist (file files)
+      (pcase (file-name-extension file)
+        ("json"
+         (let ((json-array-type 'list)
+               (json-key-type 'symbol))
+           (with-temp-buffer
+             (insert-file-contents file)
+             (dolist (key keys)
+               (goto-char (point-min))
+               (when (search-forward key nil t)
+                 (beginning-of-line)
+                 (let ((item
+                        (if (json-available-p)
+                            (json-parse-buffer :array-type 'list :object-type 'alist)
+                          (json-read))))
+                   (puthash (cdr (assq 'id item)) item cache)))))))))
+    (lambda (x)
+      (pcase x
+        ('itemids
+         (hash-table-keys cache))
+        ((pred listp) (mapcar (lambda (id)
+                                (cons id (gethash id cache)))
+                              x))
+        (_ (error "Unsupported citeproc itemgetter retrieval method"))))))
 
 (provide 'citar-citeproc)
 ;;; citar-citeproc.el ends here


bdarcus:
I've pinged Roshan to take a look. In the meantime, the CI is reporting some linting errors you could fix?

andersjohansson:
Yes, a stray parenthesis crept in to what I committed somehow. Fixed.

roshanshariff:
I think this is a good proof of concept, and as you say it works for your use case. There are a few shortcuts that probably work fine with Zotero-produced JSON files, but should be cleaned up for more general usage.
* The code would get confused if a citation key appears anywhere in the JSON file, not just as a key. You might be able to get around this with a more targeted regexp.
* It assumes that the JSON object starts at the beginning of the line containing the key.

If you or somebody else does end up making this a more robust alternative to what we have now, I think it would be a useful addition to Citar (or perhaps it would fit better in citeproc-el, since it has no other dependency on citar). Ideally, it would also support cross-references and bibtex files.

andersjohansson:
>     * The code would get confused if a citation key appears anywhere in the JSON file, not just as a key. You might be able to get around this with a more targeted regexp.
> 
>     * It assumes that the JSON object starts at the beginning of the line containing the key.

Yes! Some test cases should be added with different JSON possibilities. Does anyone know of other common sources for CSL-JSON except from Zotero and BBT?

bdarcus:
> * The code would get confused if a citation key appears anywhere in the JSON file, not just as a key. You might be able to get around this with a more targeted regexp.
> * It assumes that the JSON object starts at the beginning of the line containing the key.

Wouldn't it be better to use the json library to actually parse the file into native data structures?

Not sure if that's possible, as I recall it may have some limitations there.

andersjohansson:
> Wouldn't it be better to use the json library to actually parse the file into native data structures?

Parsing the entire file is what citeproc-el does, and what we want to avoid doing repeatedly, isn’t it? Re-using the citar cache would be preferable, but it doesn’t (and shouldn’t?) contain all the data needed for citeproc.

bdarcus:
> Re-using the citar cache would be preferable, but it doesn’t (and shouldn’t?) contain all the data needed for citeproc.

I think this is the key question, and I may have asked it elsewhere here.

Any thoughts on this @andras-simonyi?

Is there a reasonable path for us to enhance the cache a bit to allow functionality that relies on citeproc to work better?

Without digging into details, it seems likely that names and dates are the key issue?

bdarcus:
> Does anyone know of other common sources for CSL-JSON except from Zotero and BBT?

I did realize yesterday that pandoc is also a good bib file converter. It supports CSL JSON.

andras-simonyi:
> > Re-using the citar cache would be preferable, but it doesn’t (and shouldn’t?) contain all the data needed for citeproc.
> 
> I think this is the key question, and I may have asked it elsewhere here.
> 
> Any thoughts on this @andras-simonyi?
> 
> Is there a reasonable path for us to enhance the cache a bit to allow functionality that relies on citeproc to work better?
> 
> Without digging into details, it seems likely that names and dates are the key issue?

My gut feeling is that adding the missing fields that could be needed for proper rendering shouldn't be a huge problem in terms of performance, but doing it methodically would require a list of the fields that are made use of by  (the most popular?) CSL styles. Has anybody assembled such a list? (Of course, this might be too complicated and we could just solve the problem using some simple rules of thumb...)

bdarcus:
Closing this, in favor of #742.

Thanks for working on this @andersjohansson; I hope the solution I merged works for you!

PR #728, opened 2023-02-16
capf: Refactor to complete on key, use cache

### LINKED ISSUE #660: Refactor `citar-capf`
I think `citar-capf` is going to need some more extensive changes. Currently it just saves the active bibliographies whenever `citar-capf` is loaded and then never updates them. But we also can't update it every time `capf` is called. We'll need to do something a little more involved so it's reasonably fast but also updates when the bibliiographies change (and respects the buffer's local bibliographies).

_Originally posted by @roshanshariff in https://github.com/emacs-citar/citar/issues/653#issuecomment-1186612775_

bdarcus:
Change the capf so completion is on the citekey, and add an annotation.

Close #660

------

image

IUC (and it's possible I don't), this addresses the limitations that @roshanshariff noted [here](https://github.com/emacs-citar/citar/issues/660#issuecomment-1419377554), and that @mclearc referenced.

The candidates are drawn from `citar-get-entries`.

I'm not 100% sure on the UX (completing on citekeys), though my hunch is this makes more sense.

### DIFF
--- a/citar-capf.el
+++ b/citar-capf.el
@@ -25,27 +25,10 @@
 ;; Declare function from citar
 (declare-function citar--completion-table "citar") ;; pending cache revisions
 
-;; Define vars for capf
-(defvar citar-capf--candidates
-  (or (citar--format-candidates)
-      (user-error "No bibliography set"))
-  "Completion candidates for `citar-capf'.")
-
-(defvar citar-capf--properties
-  (list :exit-function #'citar-capf--exit
-        :exclusive 'no)
-  "Completion extra properties for `citar-capf'.")
-
-;; Define exit function
-(defun citar-capf--exit (str _status)
-  "Return key for STR from CANDIDATES hash."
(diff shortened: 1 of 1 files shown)


bdarcus:
> IUC (and it's possible I don't), this addresses the limitations that @roshanshariff noted [here](https://github.com/emacs-citar/citar/issues/660#issuecomment-1419377554) ...

@roshanshariff when you get a chance, can you confirm my assumption is correct here?

To summarize, `citar-capf` now simply completes keys drawn from `citar-get-entries`.

roshanshariff:
@bdarcus I'm traveling and don't have access to my computer, but I agree that only considering citation keys makes sense for completion-at-point. Maybe more bibliographic information as annotations?

bdarcus:
> Maybe more bibliographic information as annotations?

Yep; that's there now too.

PR #730, opened 2023-02-17
Don't use setf with buffer-substring

### LINKED ISSUE #729: setf and buffer-substring
With Emacs 29.1, use of `setf` with `buffer-substring` is effectively obsolete. Hopefully they reverse this, but am guessing they won't.

```console
citar-org.el:407:12: Warning: ‘buffer-substring’ is an obsolete generalized
    variable.
citar-org.el:445:12: Warning: ‘buffer-substring’ is an obsolete generalized
    variable.
```

https://github.com/emacs-citar/citar/blob/3d84879b1793f327c68cabaddd6bedb35a030274/citar-org.el#L407-L408

See linked PR for an ugly fix.

See also:

https://github.com/minad/consult/issues/636

bdarcus:
Fix #729

--------

I don't understand this code well, and not sure this is the best solution, but it seems to work.

Here's the definition of `cl--set-buffer-substring`:

```elisp
(defun cl--set-buffer-substring (start end val)
  (save-excursion (delete-region start end)
          (goto-char start)
          (insert val)
          val))
```

Note: I uncovered a bug in the shift function that preceded this. If you try to move right, for example, when point is on the last reference, you will get a type error.

I think either it should do nothing but not trigger an obscure error, or move it to be the first reference.

I have implemented the former, but added a more useful error message.

Finally, I also added `citar-capf-setup` here.

### DIFF
--- a/README.org
+++ b/README.org
@@ -64,11 +64,21 @@ This is the minimal configuration, and will work with any completing-read compli
 
 #+begin_src emacs-lisp
 (use-package citar
-  :bind (("C-c b" . citar-insert-citation)
-         :map minibuffer-local-map
-         ("M-b" . citar-insert-preset))
   :custom
   (citar-bibliography '("~/bib/references.bib")))
+#+end_src
+
+*** =citar-capf=
+
+This package includes a ~completion-at-point~ function to complete citation keys in the buffer, which you can configure like so:
+
+#+begin_src emacs-lisp
+(use-package citar
+  :custom
+  (citar-bibliography '("~/bib/references.bib"))
+  :hook
+  (LaTeX-mode . citar-capf-setup)
+  (org-mode . citar-capf-setup))
 #+end_src
 
 *** Embark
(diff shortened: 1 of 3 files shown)


bdarcus:
@yantar92 - this recent change in Emacs that obsoletes using `setf` with, in this case, `buffer-substring`: does this have broader impact in org?

If yes, what's the alternative (this PR works, but I don't like it, in part because I'm using a private cl function)?

If there's none, is their room for you to argue to emacs-devel to revert this?

yantar92:
bdarcus ***@***.***> writes:

> @yantar92 - this recent change in Emacs that obsoletes using `setf` with, in this case, `buffer-substring`: does this have broader impact in org?

Nothing other than just minor annoyance refactoring the code that used
the obsolete generalized variables.

> If yes, what's the alternative (this PR works, but I don't like it, in part because I'm using a private cl function)?

Using private function is certainly not a good idea.
Here is what Emacs devs answered:
***@***.***/

> If there's none, is their room for you to argue to emacs-devel to revert this?

I tried ;)

-- 
Ihor Radchenko // yantar92,
Org mode contributor,
Learn more about Org mode at <https://orgmode.org/>.
Support Org development at <https://liberapay.com/org-mode>,
or support my work at <https://liberapay.com/yantar92>

PR #733, opened 2023-03-08
Fixes to citar--shorten-names

pprevos:
When I use `truncate` as 1 or 2, it does  not show "et al.", which should always show when `namelength` > `truncate`. 

Changing `tnamelength` to `namelength` in the conditional fixes the issue. 

Minor point, "et al" needs to be "et al.".

I used a citation with 7 authors for testing: https://doi.org/10.1038/s41545-022-00208-8

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -1157,9 +1157,9 @@ replace last comma."
          (namelength (length namelist))
          (tnamelist (seq-take namelist (or truncate namelength)))
          (tnamelength (length tnamelist)))
-    (cond ((equal tnamelength 1)
+    (cond ((equal namelength 1)
            (citar--shorten-name (car tnamelist)))
-          ((equal tnamelength 2)
+          ((equal namelength 2)
            (concat (citar--shorten-name (car tnamelist))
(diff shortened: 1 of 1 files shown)


pprevos:
Still needs some work. I get an issue with a paper with two authors and using `truncate`  = 1

bdarcus:
> Still needs some work. I get an issue with a paper with two authors and using `truncate` = 1

Let me know when you want me to merge it to my branch?

bdarcus:
Yeah, in retrospect, that first `cond` shouldn't be needed.

bdarcus:
> Yeah, in retrospect, I have a hunch that first `cond` shouldn't be needed, and probably confuses things. We should try removing that.

Confirmed.

With that, I now get this:

```elisp
ELISP> (citar--shorten-names "Doe, J and Smith, N and Jones, K and Lamb, H" 1)
"Doe et al."
```

Should I just merge that then and close this?

pprevos:
Thanks for all the hard work while I was asleep :)

PR #734, opened 2023-03-09
Enhance citar-display-transform-functions

### LINKED ISSUE #694: Long author name lists and `citar--shorten-names`
Very long author lists is not something I deal with in my field, and hence not surprising to see it's an oversight here, but:

- `citar--shorten-names` outputs the entire author list
- when rendering for the minibuffer, this isn't a huge problem, and is actually a benefit, since you can truncate the field width for display, but the full content is still searchable
- but it IS a bit awkward when creating new notes, since there's no way to truncate the list for the title

I'm not sure what the right solution here is, but I do NOT think it is to enable shortening in the main UI.

Perhaps `citar--shorten-names` could take an optional argument to truncate that is only used for notes and the formatted reference, something like this?

```elisp
(defun citar--shorten-names (names &optional truncate and)
  "Return a list of family names from a list of full NAMES.

To better accommodate corporate names, this will only shorten
personal names of the form \"family, given\".

With TRUNCATE, shorten the list of names.

AND is the string to precede the last entry in the list."
  (when (stringp names)
    (let* ((namelist (split-string names " and "))
           (namelen (length namelist))
           (snamelist (when truncate (seq-take namelist truncate)))
           (snamelen (length snamelist))
           (rendered
            (mapconcat
             (lambda (name)
               (if (eq 1 (length name))
                   (cdr (split-string name " "))
                 (car (split-string name ", "))))
             (if truncate snamelist namelist) ", ")))
        (concat rendered
                (when
                    (and truncate (> namelen snamelen)) " et al")))))
```

bdarcus:
Modify the defcustom so that transform functions can be specified by template, and adjust `citar-get-display-value` accordingly by adding an optional "template" argument.

So default works the same, but you can now also do this:

```elisp
ELISP> (citar-get-display-value '("author" "editor") "francis2012" 'notes)
"Francis, Giles-Corti & Wood et al."
```

Close #694

Also, disable linting on everything before Emacs snapshot, since there are recent changes in indenting rules that cannot be accounted for cross-version.

-----

Aside: I'm confused on how to clear the candidate cache, so I can force regeneration. This does not work, for example:

```elisp
(clrhash (citar--format-candidates))
```

If you have time over the next few days, can you take a look at this @aikrahguzar @roshanshariff?

EDIT: I think I'm missing a piece here ATM, which is the template connection. But I'm tired, so will need to look at it again later. I mainly want confirmation this is a good idea though.

EDIT 2: took another look this morning. I think `citar-format--entry` is where this would be, but it currently takes a `format` rather the symbol for it. So will need to think about that.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -1,6 +1,6 @@
 ;;; citar.el --- Citation-related commands for org, latex, markdown -*- lexical-binding: t; -*-
 
-;; Copyright (C) 2021-2022 Bruce D'Arcus
+;; Copyright (C) 2021-2023 Bruce D'Arcus
 
 ;; Author: Bruce D'Arcus <https://github.com/bdarcus>
 ;; Maintainer: Bruce D'Arcus <https://github.com/bdarcus>
@@ -147,7 +147,7 @@ for the title field for new notes."
 (defcustom citar-ellipsis nil
   "Ellipsis string to mark ending of truncated display fields.
 
-If t, use the value of `truncate-string-ellipsis'.  If nil, no
+If t, use the value of variable `truncate-string-ellipsis'.  If nil, no
 ellipsis will be used.  Otherwise, this should be a non-empty
 string specifying the ellipsis."
   :group 'citar
@@ -176,18 +176,19 @@ references as a string."
 
 (defcustom citar-display-transform-functions
   ;; TODO change this name, as it might be confusing?
-  '((("author" "editor") . citar--shorten-names))
+  '((t . ((("author" "editor") . (citar--shorten-names))))
+    (notes . ((("author" "editor") . (citar--shorten-names 3 "&")))))
   "Configure transformation of field display values from raw values.
 
 All functions that match a particular field are run in order."
   :group 'citar
-  :type '(alist :key-type   (choice (const t) (repeat string))
-          :value-type function))
+  :type '(alist :key-type symbol
+          :value-type (cons list list)))
 
 (defcustom citar-symbols
-  `((file  .  ("F" . " "))
-    (note .   ("N" . " "))
-    (link .   ("L" . " ")))
+  `((file . ("F" . " "))
+    (note . ("N" . " "))
+    (link . ("L" . " ")))
   "Configuration alist specifying which symbol or icon to pick for a bib entry.
 This leaves room for configurations where the absense of an item
 may be indicated with the same icon but a different face.
@@ -649,8 +650,8 @@ CATEGORY is one of:
  * `multi-category' when CANDIDATES has resources of multiple
    types. The `multi-category' text property is applied to each
    element of CANDIDATES."
-  (cl-flet ((getresources (table) (when table
-                                    (delete-dups (apply #'append (hash-table-values table)))))
+  (cl-flet ((getresources (table)
+              (when table (delete-dups (apply #'append (hash-table-values table)))))
             (keycands (type citekeys)
               (let ((format (citar-format--parse (citar--get-template 'completion)))
                     (width (- (frame-width) 2)))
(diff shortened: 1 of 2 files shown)


citar.el
```diff
@@ -176,18 +176,19 @@ references as a string."
 
 (defcustom citar-display-transform-functions
   ;; TODO change this name, as it might be confusing?
-  '((("author" "editor") . citar--shorten-names))
+  '((t . ((("author" "editor") . (citar--shorten-names))))
+    (notes . ((("author" "editor") . (citar--shorten-names 3 "&")))))
```
aikrahguzar, line 180: The doc string here needs to be changed. In particular we need to specify how the arguments will be used. Note that  for value `(citar--shorten-names 3 "&")`,  `citar--shorten-names` will be called as `(citar--shorten-names STR 3 "&")` and this needs to be documented.

citar.el
```diff
@@ -901,17 +902,18 @@ was found to have a value, and VALUE is its value."
...
 
-(defun citar-get-display-value (fields citekey-or-entry)
+(defun citar-get-display-value (fields citekey-or-entry &optional template)
   "Return the first non nil value for CITEKEY-OR-ENTRY among FIELDS .
 
-The value is transformed using `citar-display-transform-functions'"
+The value is transformed using `citar-display-transform-functions' for TEMPLATE."
   (let ((fieldvalue (citar-get-field-with-value fields citekey-or-entry)))
     (seq-reduce (lambda (string fun)
+                  ;; FIX: can't get this to work, and don't understand this line.
```
aikrahguzar, line 911: What is the problem here? `fun` is badly named since it will actually be an element of `citar-display-transform-functions` but I probably couldn't (still can't) think of a name.

citar.el
```diff
@@ -901,17 +902,18 @@ was found to have a value, and VALUE is its value."
...
                   (if (or (eq t (car fun))
                           (seq-contains-p (car fun) (car fieldvalue) #'string=))
-                      (funcall (cdr fun) string)
+                      (apply (cadr fun) (append (list string) (cddr fun)))
```
aikrahguzar, line 914: I think this `(apply (cadr fun) string (cddr fun))`

aikrahguzar:
I am a bit hazy on what is the purpose but the code looks alright and I left some minor comments.

aikrahguzar:
Actually since `citar-display-transform` is getting so deeply nested it might be better to use a plist with template as keys.

bdarcus:
> Actually since `citar-display-transform` is getting so deeply nested it might be better to use a plist with template as keys.

I was wondering this also, and whether the transform details should stay with the template?

The purpose is to have different transforms for different contexts, most immediately for minibuffer vs note titles. See #694.

citar.el
bdarcus: I need to remove the comment, as it does not work.

But I still don't understand the line immediately below it.

aikrahguzar: > I need to remove the comment, as it does not work.

I assume the not is a typo?

> But I still don't understand the line immediately below it.

`fun` is an element of `citar-display-transform-functions` so `(car fun)` is either a list of field names or it is `t` where the `t` means the function which is in the `(cdr fun)` should be applied to all fields.

bdarcus: Yeah; typo, now fixed.

Thanks for explanation!

aikrahguzar:
> Aside: I'm confused on how to clear the candidate cache, so I can force regeneration. This does not work, for example:
> 
> ```emacs-lisp
> (clrhash (citar--format-candidates))
> ```

I think `(clrhash citar-cache--bibliographies)` should do it.

aikrahguzar:
> > Actually since `citar-display-transform` is getting so deeply nested it might be better to use a plist with template as keys.
> 
> I was wondering this also, and whether the transform details should stay with the template? Maybe something like:
> 
> ```emacs-lisp
> (note
>  (:template "${author editor} - ${title}" :transforms
>             ((("author" "editor")
>               citar--shorten-names))))
> ```
> 
> ... or even:
> 
> ```emacs-lisp
> (note
>  (:template "${'author' 'editor'}" :transforms
>             (:fields
>               ("author" "editor")
>               :funsymbol citar--shorten-names :args
>               (3 "&")))
>  main
>  (:template "${'title'}")
>  suffix
>  (:template "${'=key=}"))
> ```

I like the idea of unifying templates and transforms since they are very related and them being close together makes them easier to grasp. There is some chance of breakage because of that and unintended consequences.

bdarcus:
> I like the idea of unifying templates and transforms since they are very related and them being close together makes them easier to grasp. There is some chance of breakage because of that and unintended consequences.

On the second sentence, you mean if we keep them split, or make the change?

bdarcus:
> Maybe I need to think again about variable modifiers ...

I've reverted this to draft, and opened up an alternate branch with this idea in #735.

bdarcus:
I'm going to close this, at least for now.

PR #735, opened 2023-03-11
Allow transform modifiers in templates

### LINKED ISSUE #694: Long author name lists and `citar--shorten-names`

bdarcus:
Modify:

* `citar-display-transform-functions` so it becomes a list of shortcuts to include in template specs (like `"${author:20%etal}"` where `etal` is the key that then points to a list of a function symbol and optional arguments).

* `citar-get-display-value` so it allows an optional transform function

* `citar-format--parse` to include transform list

* formatting functions to properly integrate the above

Close #694

-----

This is an alternative to #734; I think easier to implement and configure, and more flexible.

I think it all works, and **unless I hear otherwise, I'll likely merge it tomorrow, March 13**.

EDIT: I got sidetracked by a stupid variable declaration bug; now fixed.

@andersjohansson @pprevos - any feedback on this? I think you were, or may be, interested in such a feature.

PS - It has occurred to me that perhaps when strings are transformed, the original value should still be included as hidden (making use of display vs hidden properties)?

```elisp
(propertize "Doe, J and Jones, K and Smith, L" 'hidden t 'display "Doe et al.")
```

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -32,6 +32,10 @@ jobs:
         exclude:
           - emacs_version: snapshot
             action: compile
+          - emacs_version: 28.2
+            action: lint
+          - emacs_version: 27.2
+            action: lint
     steps:
     - name: Set up Emacs ${{matrix.emacs_version}}
       uses: purcell/setup-emacs@master
--- a/README.org
+++ b/README.org
@@ -180,10 +180,10 @@ Here's the default value:
 
 #+begin_src emacs-lisp
 (setq citar-templates
-      '((main . "${author editor:30}     ${date year issued:4}     ${title:48}")
+      '((main . "${author editor:30%sn}     ${date year issued:4}     ${title:48}")
         (suffix . "          ${=key= id:15}    ${=type=:12}    ${tags keywords:*}")
-        (preview . "${author editor} (${year issued date}) ${title}, ${journal journaltitle publisher container-title collection-title}.\n")
-        (note . "Notes on ${author editor}, ${title}")))
+        (preview . "${author editor%etal} (${year issued date}) ${title}, ${journal journaltitle publisher container-title collection-title}.\n")
+        (note . "Notes on ${author editor:%etal}, ${title}")))
 #+end_src
 
 Note:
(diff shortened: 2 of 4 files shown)


pprevos:
Hi Brian, good stuff.

There is a package for [org-ref that prettifies citations](https://github.com/emacsmirror/org-ref-prettify). Is that what you meant by hidden?

bdarcus:
> There is a package for [org-ref that prettifies citations](https://github.com/emacsmirror/org-ref-prettify). Is that what you meant by hidden?

No, that's what org-cite-csl-activate does.

More what we already do a bit of in the minibuffer candidates. But it's an optimization to consider later.

andersjohansson:
Seems like a great idea! 
I dug into the parsing and realized this: #743

Issue #645, opened 2022-06-26
Allow "has" affixation to be configurable


bdarcus:
I don't know if this is a good idea, but with #634, it occurred to me:

It might be good to allow configurable indicators?

That would allow to edit the current ones (for example, remove "links"), or to add new ones.

roshanshariff:
I was also thinking this, and was going to implement a more configurable system, but it looked like it would increase the complexity more than necessary. I think it's a good idea, but we should wait until we have a concrete use case and make sure whatever we implement works for that purpose.

The main constraint is that any indicators have to be very fast at display time. This means that if you want to do any non-trivial work, you have to do it earlier, when the bibliography is parsed (at least as much as possible). So you'd want to have a customizable list of "data augmentation" functions that run when the bibliography is parsed. They would pre-compute extra information that gets attached to each bibliography entry. Then, ideally, the indicators would just be able to look at this augmented information.

I have some thoughts on how to implement all this but, as I said, I think we should wait for a concrete use case.

Just allowing the built-in indicators to be removed is very easy. We don't do this now, but we can just disable the file and note indicators when `has-files-functions` or `has-notes-functions` are nil. Currently we just won't show the indicators, but still reserve the space for them. We could easily add a similar option for links.

bdarcus:
To be clear, I do not advocate adding this now, or even necessarily soon.

I just thought it worth considering.

For that reason, I marked the milestone "future".

bdarcus:
FYI, I just linked an issue with a "concrete use case": in this case, for citar-org-roam, indicating when a note has cited resources.

But TBH, that needs some more thought in any case.

bdarcus:
I think we have enough concrete use cases at this point to figure this out.

I'll see if I can find the time to look at the code and at least figure out what be involved in implementing it, if not actually code it. I'm not sure ATM how big a PITA this would be.

bdarcus:
This was added in #753.

PR #744, opened 2023-03-15
Add optional affixation symbol for currently cited references

andersjohansson:
It's helpful to see what you’ve already cited in the current document.This is something I have missed from my earlier system that extended org-zotxt (before org-cite).

I'm not sure if you want to include something like this, and it is not finished for anything more than org.

An implementation for fetching cite keys in org-mode is included, with configuration for fetching in narrowed buffer or not, and from several org buffers.

Optional grouping by currently cited is also included, however, this seems to interfere with the multi-selection display based on grouping.

(Took some time to go through my citar config and this branch today)

### DIFF
--- a/citar-org.el
+++ b/citar-org.el
@@ -69,6 +69,15 @@ Each function takes one argument, a citation."
   :group 'citar-org
   :type '(repeat function))
 
+(defcustom citar-org-current-refs-in-widened-buffer t
+  "If non-nil, gathers current references from the widened buffer."
+  :group 'citar-org
+  :type 'boolean)
+
+(defvar-local citar-org-current-refs-in-several-org-buffers nil
+  "When non-nil, gathers current references from given list of files, or from all org-buffers (if t).")
+
+
 ;;; Keymaps
 
 (defvar citar-org-citation-map
@@ -139,6 +148,31 @@ With PROC list, limit to specific processor(s)."
               (push fstyle styles))))))
     styles))
 
+(defun citar-org--current-refs-fetch ()
+  "Fetch refs in current org buffer and store in ‘citar--cited-references’."
+  (when (memq citar-indicator-cited citar-indicators)
+    (clrhash citar--cited-references)
+    (cl-loop for buf in (pcase citar-org-current-refs-in-several-org-buffers
+                          ('t (org-buffer-list 'files t))
+                          ('nil (list (current-buffer)))
+                          ((and (pred listp) files)
+                           (mapcar #'find-file-noselect files)))
+             do
+             (with-current-buffer buf
+               ;; This is slow, would be good to use cache, but ‘org-element-cache-map’ can’t
+               ;; be used with object granularity (for citation-reference).
+               (org-element-map
+                   (if citar-org-current-refs-in-widened-buffer
+                       (save-restriction
+                         (widen)
+                         (org-element-parse-buffer))
+                     (org-element-parse-buffer))
+                   'citation-reference
+                 (lambda (ref) (puthash
+                                (org-element-property :key ref)
+                                t
+                                citar--cited-references)))))))
+
 ;;; Org-cite processors
 
 ;; NOTE I may move some or all of these to a separate project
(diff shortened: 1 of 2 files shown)


bdarcus:
Cool; related: #645.

andersjohansson:
Screenshot:
image

This is with: 
```emacs-lisp
(setq citar-symbols
     (link ,(all-the-icons-octicon "link" :face 'all-the-icons-orange :v-adjust 0.01) . " ")
     (cited ,(all-the-icons-faicon "quote-right" :face 'org-cite) . " ")))

```

bdarcus:
So on #645, If it's too much of a hassle, we can save it for later., but as you work on this, can you think about how to do both together?

It could avoid us having to add additional defcustoms and such.

andersjohansson:
By the way, it does a lot of parsing of org-files (on every invocation of `citar-org-select-key` or `citar-org-insert-citation`, but I have so far (have used this for a few months) not experienced this as slowing down citing.

aikrahguzar:
> An implementation for fetching cite keys in org-mode is included, with configuration for fetching in narrowed buffer or not, and from several org buffers.


This can be done for latex by parsing the aux file produced after compilation. `reftex` probably has something like that too.
Parsing aux file should be simpler and faster and will include citations from all the files that are used to produce the document. It has the disadvantage that it will be reflect the document at the time of last compilation but I don't think it possible to write latex without compiling very often.

bdarcus:
Thinking about it from the user-facing end, would not something like this work for addressing #645?

```elisp
(defvar citar-indicators ; a defcustom really
  `(("F" . #'citar-has-files)
    ("L" . #'citar-has-links)
    ("N" . #citar-has-notes)
    ("C" . #'citar-has-cited)))
```

andersjohansson:
Cool! I re-implemented it with following #753. This actually makes it reasonable to do this as a plugin to citar. The only point where I would need to use advice in that case would be for the points where I do the buffer parsing (the not very carefully chosen ` citar-org-select-key` and `citar-org-insert-citation`).

I'm not sure if you are interested to merge this eventually?

The configuration of pretty symbols would look like this then. A bit more tedious to edit the `:symbol` slot of the indicator structs.

```emacs-lisp
(setf (citar-indicator-symbol citar-indicator-files)
      (all-the-icons-faicon "file-o" :face 'all-the-icons-green :v-adjust -0.1)
      (citar-indicator-symbol citar-indicator-links)
      (all-the-icons-octicon "link" :face 'all-the-icons-orange :v-adjust 0.01)
      (citar-indicator-symbol citar-indicator-notes)
      (all-the-icons-material "speaker_notes" :face 'all-the-icons-blue :v-adjust -0.3)
      (citar-indicator-symbol citar-indicator-cited)
      (all-the-icons-faicon "quote-right" :face 'org-cite))

(setq citar-indicators (list citar-indicator-links
                             citar-indicator-files
                             citar-indicator-notes
                             citar-indicator-cited))
```

andersjohansson:
Oh, I saw you added indicator configuration here as well: https://github.com/emacs-citar/citar/wiki/Indicators.

bdarcus:
> I'm not sure if you are interested to merge this eventually?

I'm not opposed at all; with this all configurable, users can use or not. And I do like the idea.

From my POV whether to include is really about the general fit, and this one seems to fit, as it's really about writing in the major modes, and isn't about any specific other packages.

I think the design question would just be whether and how to integrate into the major-mode adapter system. 

I suppose the "list-keys" key is all that's really required, though it's currently defined as only include the citekeys in the "current buffer."

>  This actually makes it reasonable to do this as a plugin to citar. 

Yes; I'll leave it up to you, though just note that the `citar-capf` was originally spun out, but we merged it back in because it was such a small amount of code.

> The only point where I would need to use advice in that case would be for the points where I do the buffer parsing (the not very carefully chosen  citar-org-select-key and citar-org-insert-citation).

Not following here, but if you need to advice, perhaps there are tweaks to that code that would avoid that?

I developed `citar-org-roam` independently of `citar` in part to test and refine the API, and I think it's better because of it.

I'll also try to test this (been busy with other things). Is it ready for that?

PR #753, opened 2023-03-20
feat!: make indicator affixes configurable

bdarcus:
The current indicators are hard-coded, with the only configurable details being the symbols in `citar-symbols`, and the space between them in `citar-symbol-separator`.

This makes all of it configurable by adding:

* `citar-indicators` variable; a list of plist indicator specs
* `citar--make-indicator-processors` function (creates a list of plists with the needed properties, including the compiled predicate functions)
* `citar--make-indicators-tags` and  `citar--make-indicator-symbols` functions
* add indicator variables, and a [new wiki page](https://github.com/emacs-citar/citar/wiki/Indicators) with others

BREAKING CHANGE: `citar-symbols` will be deprecated in favor of `citar-indicators`

Close: #601 

-------

## Status

From what I can tell, everything works correctly; this example removes the links indicator, and uses default for files, and icons for notes:

image

And while I haven't benchmarked, I don't believe this should be any slower.

Aside: I had an idea to extend this later so people can separately configure left (what we do now) and right sides of the affixation.

I think I may turn `citar-indicators` into a defcustom, but later.

In the second commit I switched to using a `cl-defstruct` for the `citar-indicator` type. Maybe be a tad overkill, but it mirrors what we do with `citar-cache`. I've made this public to encourage people to use it in their configs.

@andersjohansson with some help, I figured out the code for this. When I merge this, you could use it for #744.


citar.el
```diff
@@ -781,32 +851,22 @@ Return nil if `citar-bibliographies' returns nil."
...
            (preformatted (citar-cache--preformatted bibs))
-           (hasfilesp (citar-has-files))
-           (hasnotesp (citar-has-notes))
-           (haslinksp (citar-has-links))
-           (hasfilestag (propertize " has:files" 'invisible t))
-           (hasnotestag (propertize " has:notes" 'invisible t))
-           (haslinkstag (propertize " has:links" 'invisible t))
-           (symbolswidth (string-width (citar--symbols-string t t t)))
+           (indicatorprocs (citar--make-indicator-processors citar-indicators))
+           ;; TODO remove the below function call and function, but replace with what?
```
bdarcus, line 855: Remove comment; already fixed.

bdarcus:
OK, merged. I think it's pretty solid, though let me know, if send a PR, if any little bugs.

Does occur to me I didn't really include error handling in the key functions, so that's a good next step.

PR #761, opened 2023-03-23
Indicator for cited references, re-using major-mode `list-keys` function

andersjohansson:
This is a simpler variant of #744. 

I am not sure about whether it is smart to transform a list to a hash table in this way, compared to generating a hash-table from the beginning. A hash table should be most suitable for lookup (`citar-is-cited`) though.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -253,6 +253,12 @@ in some cases when using icons.")
    :function #'citar-has-notes
    :tag "has:notes"))
 
+(defvar citar-indicator-cited
+  (citar-indicator-create
+   :symbol "C"
+   :function #'citar-is-cited
+   :tag "is:cited"))
+
 ;; Indicator config
 
 (defvar citar-indicators
@@ -620,6 +626,8 @@ FILTER: if non-nil, should be a predicate function taking a
   (citar-select-ref :filter (citar-has-file))"
   (let* ((candidates (or (citar--format-candidates)
                          (user-error "No bibliography set")))
+         (_cited (when (memq citar-indicator-cited citar-indicators)
+                   (citar--populate-cited-hash)))
          (chosen (if (and multiple citar-select-multiple)
                      (citar--select-multiple "References: " candidates
                                              filter 'citar-history citar-presets)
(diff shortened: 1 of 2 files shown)


citar.el
```diff
@@ -620,6 +626,8 @@ FILTER: if non-nil, should be a predicate function taking a
   (citar-select-ref :filter (citar-has-file))"
   (let* ((candidates (or (citar--format-candidates)
                          (user-error "No bibliography set")))
+         (_cited (when (memq citar-indicator-cited citar-indicators)
```
andersjohansson, line 629: This was the most general place I found to add this? Is it reasonable?

citar.el
```diff
@@ -620,6 +626,8 @@ FILTER: if non-nil, should be a predicate function taking a
...
+                   (citar--populate-cited-hash)))
```
andersjohansson, line 630: This was the most general place I found to add this. Is it reasonable?

citar.el
```diff
@@ -1311,6 +1319,23 @@ replace last comma."
...
+  "Return a function for checking whether CITEKEY is cited.
+Just looks for CITEKEY in ‘citar--cited-references’."
+  (lambda (citekey)
+    (gethash citekey
+             citar--cited-references)))
+
+(defun citar--populate-cited-hash ()
+  "Parse buffer to populate ‘citar--cited-references’."
+  (clrhash citar--cited-references)
+  (cl-loop for key in (citar--major-mode-function 'list-keys #'ignore)
```
andersjohansson, line 1336: Could as well use `dolist` for this of course. I’m so used to using `cl-loop`

andersjohansson:
This seems to work ok at least for smaller org, markdown and latex files. Haven’t tested extensively for lager files.

bdarcus:
> I am not sure about whether it is smart to transform a list to a hash table in this way, compared to generating a hash-table from the beginning. 

My question as well, particularly since we now rely on hash tables so pervasively.

I haven't looked; do you have a handle on how easy that change would be?

If easy and straightforward, perhaps we should just do it?

bdarcus:
This is weird. It says I closed this, but I didn't; at least not intentionally!

andersjohansson:
> Why do you it here? My intention was all the indicators could be decoupled from this sort of code.
> 
That would be good. But wouldn’t the other alternative be to do it in the indicator-function, which is run once for every (visible ?) key that is up for completion (if my understanding of affixation functions is correct). We only want to parse the buffer once, and then do lookups in the hash-table (is my idea). Or am I missing something?

bdarcus:
> Why do you it here? My intention was all the indicators could be decoupled from this sort of code.
> That would be good. But wouldn’t the other alternative be to do it in the indicator-function, which is run once for every (visible ?) key that is up for completion (if my understanding of affixation functions is correct). 

Yes. That's how the indicators work, to ensure they're always up-to-date.

> We only want to parse the buffer once, and then do lookups in the hash-table (is my idea). Or am I missing something?

The buffer changes, and every time you add or remove a citation you'd want the UI to update?

Same with bib files, notes, files, etc.

Do you find the performance problematic without the hash table?

bdarcus:
Maybe try with this as the function and see what happens? It's the simplest approach, but I think it may work because the list shouldn't be longer, but the values are citekeys, so the same as the keys in the larger hash tables.

```elisp
(defun citar-is-cited ()
  "Return function to check if reference is cited in buffer."
  (let ((iscited
         (citar--major-mode-function 'list-keys #'ignore)))
    (lambda (citekey)
      (member citekey iscited))))
```

Issue #759, opened 2023-03-23
Indicator spacing wrong with icons


bdarcus:
_converting from a discussion_

----------------------------

After updating, and replacing my citar config with the example out of the wiki, I noticed that citar does look different. For example, the spacing between the icons is gone. How do I get it back?

Thank you!

<img width="77" alt="Bildschirm­foto 2023-03-22 um 20 07 40" src="https://user-images.githubusercontent.com/39911339/227012121-c511c65e-1b0b-4ebe-ae36-c76ed0d57715.png">

_Originally posted by @gwbrck in https://github.com/emacs-citar/citar/discussions/754#discussioncomment-5397164_

bdarcus:
[This](https://github.com/emacs-citar/citar/commit/dcd8c3f0dd4bb5ca75082eb2005e9604a4957c37#diff-edf00f713d343a2437767c9ff28665dda570370c7c5e02e76127171b7d685bddL1239-L1259) is the previous code that I adapted.

There's something subtle I'm not yet seeing that seems to incorrectly calculate the alignments here.

I checked, for example, and `string-width` returns `1` on the icons, so don't think that's the problem.

robmoss:
I'm also experiencing this issue (I didn't realise there were meant to be spaces between the icons) and `(string-width (citar--make-indicator-symbols "abc has:files has:links has:notes"))` evaluates to 6 — is that correct, or too narrow? I've tried `(setq citar-symbol-separator " a ")`, which suggests it's too narrow. 

I have no idea how to fix this. The following hack inserts a space between each pair of tags, but isn't a good solution: `:align-to (+ 1 (string-width str))`

bdarcus:
> The following hack inserts a space between each pair of tags, but isn't a good solution: `:align-to (+ 1 (string-width str))`

What I had in mind is adding a property to the indicator spec, so one could do this:

```elisp
     :symbolwidth 2
```

Make sense?

I think that's the right place to put it; I was just struggling how to integrate that integer into the output, but your comment points me in the right direction, so thank you!

bdarcus:
I can't test it ATM, but I've opened a PR with a fix that _should_ work in #762,

gwbrck:
I tried #762 and can confirm, the following shows the spacing correct:

```elisp
(setq citar-indicator-notes-icons
    (citar-indicator-create
     :symbol (all-the-icons-material
              "speaker_notes"
              :face 'all-the-icons-blue
              :v-adjust -0.3)
[...]

PR #762, opened 2023-03-23
fix(indicators): add `padding` slot

bdarcus:
* add `padding` to `citar-indicator`, set to default value of 1
* incorporate into `citar--make-indicator-symbols`
* deprecate `citar-symbol-separator`
* update README

Close: #759

---------------

I think I fixed it.

image

So basically, there's no longer a global variable for the separator; instead it's defined in the `citar-indicator` specs, with a default of 1.

If you two can either test, or at least check the basic logic, that'd be great.

@gwbrck @robmoss

Hmm .. as I look at documentation, this seems a bit odd:

```elisp
   :symbolwidth 2
   :emptysymbol "  "
```

The latter is required, because one might want to use an icon also.

But maybe they should be consistent? Probably not, since they serve different purposes?

### DIFF
--- a/README.org
+++ b/README.org
@@ -227,7 +227,22 @@ Here's a screenshot using this configuration, which removes the links indicator,
 #+name: fig-indicators
 [[images/indicators.png]]
 
-You can create your own indicators, of course, though keep in mind the included predicate functions must be performance-optimized, since the completion UI runs them on your entire library every time you open it.
+You can create your own indicators, of course.
+Here's an example indicator definition incorporating icons:
+
+#+begin_src emacs-lisp
+(defvar citar-indicator-notes-icons
+  (citar-indicator-create
+   :symbol (all-the-icons-material
+            "speaker_notes"
+            :face 'all-the-icons-blue
+            :v-adjust -0.3)
+   :function #'citar-has-notes
+   :padding " "
+   :tag "has:notes"))
+#+end_src
+
+Keep in mind, however, the included predicate functions must be performance-optimized, since the completion UI runs them on your entire library every time you open it.
 
 ** Test Script
     :PROPERTIES:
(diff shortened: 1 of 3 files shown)


bdarcus:
Maybe I should change `symbolwidth` to just `padding`, since that's all it's doing now (I think)?

README.org
```diff
@@ -227,7 +227,23 @@ Here's a screenshot using this configuration, which removes the links indicator,
...
+#+begin_src emacs-lisp
+(defvar citar-indicator-notes-icons
+  (citar-indicator-create
+   :symbol (all-the-icons-material
+            "speaker_notes"
+            :face 'all-the-icons-blue
+            :v-adjust -0.3)
+   :function #'citar-has-notes
+   :symbolwidth 2
+   :emptysymbol "  "
```
gwbrck, line 242: For me, `emptysymbol` needs to be set to just one space to make it look right.

bdarcus:
> Maybe I should change `symbolwidth` to just `padding`, since that's all it's doing now (I think)?

In the latest, separate, commit, I made this change. So now works like:

```elisp
  :padding "  "
```

It works the same (I had problems before because I was using two spaces for the width).

I think this is more clear.

Any feedback before I merge?

README.org
bdarcus: I remove the line in the README.

bdarcus: Given that it makes no difference for icons or plain text, maybe we can remove that entirely?

gwbrck: Yes, I can remove it. 
But in general, I wouldn't remove this option. Having a dim placeholder icon for whenever there is no note, pdf or link seems like a nice option to have.

bdarcus: Right; I meant in this case, not generally.

So I removed them from the examples in the docs and wiki page.

PR #763, opened 2023-03-24
feat(indicators): add cited

bdarcus:
@andersjohansson - can you give this a try? 

It's just a few lines (no cache or hash table), but works for me.

```elisp
(defun citar-is-cited ()
  "Return function to check if reference is cited in buffer."
  (let ((iscited
         (citar--major-mode-function 'list-keys #'ignore)))
    (lambda (citekey)
      (member citekey iscited))))
```

image

My test file was small though.

If this is good to go, I can merge it.

I gave co-author credit on the commit :-)

We can then later think of other enhancements as and if needed (like possibly extending `'list-keys` beyond the immediate buffer, or optimizing performance).

PS - `'list-keys` was originally added to allow exporting of local bib files.

### DIFF
--- a/README.org
+++ b/README.org
@@ -212,6 +212,7 @@ By default, it includes plain text indicators for, each of which indicates the p
 - notes
 - library files
 - links
+- cited (for references cited in the current buffer)
 
 For other indicators, see the [[https://github.com/emacs-citar/citar/wiki/Indicators][wiki]].
 
--- a/citar.el
+++ b/citar.el
@@ -887,6 +887,15 @@ visible in the completion UI."
                    (if matchtagp sym emptysym)
                    padding))
              (pos (length str)))
+        ;; See https://github.com/emacs-citar/citar/issues/764 for explanation of what this code is doing.
+        ;;
+        ;; We say that the last character of the padding should be replaced by a space which stretches to the
+        ;; position we want to.
+        ;;
+        ;; So for example if the icon has string width 1 but occupies 0.9 the space will stretch to occupy 1.1
+        ;; and if the icon occupies 1.5 it will shrink to occupy 0.5.
+        ;;
+        ;; Emacs 29 has `vtable', which we might use here in the future as well.
         (put-text-property (- pos 1) pos 'display
                            (cons 'space
                                  (list :align-to (string-width str)))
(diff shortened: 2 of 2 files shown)


bdarcus:
Update: I think I'm going to merge this, as it's the necessary first step regardless of whether it needs further optimization.

Just trying to set the CI to pass; seems there's currently some internet-related problems not related to this code.

andersjohansson:
This works well! And the buffer parsing is only run once per selection session just as it should be. My other attempts was based on a misunderstanding that this would not be the case (I didn't realize that this would be taken care of only once and stored in the compiled function when running `citar--indicator-processors`).

And totally reasonable to just check against a list, I suppose people have few documents with several hundred (thousands?) different citations where `member` versus `gethash` would make any reasonable difference.

I tried in a small document and in one with 40 000 words and 330 references. In the large document there is a bit of a more noticeable delay (half a second)  when launching selection (mainly, not unsurprisingly  `org-element-parse-buffer`).

But this is totally ok for me! Trying to introduce good caching would be quite complicated. Too bad the `org-element--cache` can’t be re-used with `org-element-cache-map` (as I understand it, it doesn’t store objects, like citations).

bdarcus:
> I tried in a small document and in one with 40 000 words and 330 references. In the large document there is a bit of a more noticeable delay (half a second)... But this is totally ok for me! 

But not ideal.

Ooh, you just reminded of the long #397 thread.

What if we used `org-element-cache-map` instead? It's new, but wouldn't that be perfect here?

andersjohansson:
﻿
> 
> What if we used org-element-cache-map instead? It's new, but wouldn't that be perfect here?
> 
Yes, but as @yantar92 explained in the other #766, that doesn’t work for objects (as citations are in org). At least not yet. 🙂

By the way, I haven’t looked deeply into how reftex does its listing of citations, is that more performant perhaps?

bdarcus:
Yes, I missed what you were saying on that earlier!

We can always add it when that functionality exists.

I did look at reftex yesterday to consider ideas for org. I don't remember details, but it's pretty low-level code that uses `re-search-forward`, that I assume is pretty performant.

But not the code I want to include when we can wait on Ihor for something easier to code and maintain. :-)

PR #766, opened 2023-03-25
perf(org): use cached org data for list-keys

bdarcus:
@andersjohansson - doesn't quite work ATM (returns `nil`), but I'm sure it's just some little thing.

### DIFF
--- a/citar-org.el
+++ b/citar-org.el
@@ -352,11 +352,23 @@ or citation-reference."
 
 (defun citar-org-list-keys ()
   "List citation keys in the org buffer."
-  (let ((org-tree (org-element-parse-buffer)))
-    (delete-dups
-     (org-element-map org-tree 'citation-reference
-       (lambda (r) (org-element-property :key r))
-       org-tree))))
+  (let* ((org-tree (org-element-parse-buffer))
+         (citekeys
+          (if (fboundp 'org-element-cache-map)
+              (org-element-cache-map
+               (lambda (ref)
+                 (citar-org--get-ref-citekey ref))
+               :restrict-elements '(citation-reference)
+               :granularity 'element)
+            (org-element-map org-tree 'citation-reference
+              (lambda (ref)
+                (citar-org--get-ref-citekey ref))
+              org-tree))))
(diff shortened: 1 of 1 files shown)


citar-org.el
```diff
@@ -352,11 +352,23 @@ or citation-reference."
 
 (defun citar-org-list-keys ()
   "List citation keys in the org buffer."
-  (let ((org-tree (org-element-parse-buffer)))
-    (delete-dups
-     (org-element-map org-tree 'citation-reference
-       (lambda (r) (org-element-property :key r))
-       org-tree))))
+  (let* ((org-tree (org-element-parse-buffer))
```
bdarcus, line 355: @yantar92 - could you be so kind as to tell me what I'm doing wrong here with `org-element-cache-map`?

I'm trying to get the list of all the `citation-reference` keys, just as in the then statement below it.

yantar92: `citation-reference` is not an element, it is an object.

bdarcus: Got it; thanks!

bdarcus: As I'm looking, is there any reason `org-cite-list-keys` shouldn't be available outside of export?

That's basically what we're recreating here.

yantar92: Full parsing is expensive.

yantar92: Basically, `org-cite-list-keys` is not going to be performant on large files. And it is not written to be working outside export scope.

bdarcus: As I'm attempting to do here; is that a reasonable use of `org-element-cache-map`, that's likely to be performant?

bdarcus: Oh wait, `citation` is also an object.

yantar92: `org-element-cache-map` does not currently support objects. Objects are not even cached. The best I can suggest for now is `re-search-forward` + `org-element-context`. 

Object caching and fast mapping over arbitrary buffer AST elements are in the works. They are not trivial though.

bdarcus:
OK, closing this. Not convenient to do this now.

Issue #685, opened 2022-09-05
Open PDF in Zotero


dlukes:
Zotero has recently acquired a very nice [PDF reader and annotator](https://www.zotero.org/support/pdf_reader), which can be [invoked via the command line](https://forums.zotero.org/discussion/98234/is-there-anyway-i-can-open-pdfs-in-zotero-pdf-viewer-via-the-command-line). So I started investigating whether I can make citar use it.

Turns out -- yes, see below. I'm still posting this here because this was the first place I looked to see whether anyone had already figured it out, so I'm hoping it might help other people who want to achieve something similar. Additionally, it might be a source of inspiration to rework some of citar's internals to make this type of setup a bit easier (basically, all that's needed is a some special-casing for URIs starting with `zotero://...`). But if that's something none of the maintainers consider worth their while at this point, that's perfectly fine too, feel free to close the issue right away :)

### Big picture

The Zotero PDF reader is activated via URLs of the general form `zotero://open-pdf/library/items/<ZOTERO_KEY>`, which you shoud be able to just `(xdg-)open` on the command line. So we need to:

1. generate such URLs during export from Zotero
2. sheperd them safely through citar's attachment files machinery

### Zotero side

I'm using [Better BibTeX](https://retorque.re/zotero-better-bibtex) for the exports, because it allows you to [customize the export process](https://retorque.re/zotero-better-bibtex/exporting/scripting/). You'll need to add a custom script to the export process via the /postscript/ setting:

<img width="805" alt="image" src="https://user-images.githubusercontent.com/2734517/188439594-d44e62a3-e3d9-4b39-ba99-5504081cdc2a.png">

Here's the full script for copy-paste, with explanatory comments:

```javascript
// I'm using this with BetterCSLJSON, but BetterBib(La)TeX is also possible, see
// https://retorque.re/zotero-better-bibtex/exporting/scripting/.
if (Translator.BetterCSLJSON) {
  entry.file = item.attachments.map(
    a => (
      // If this is a PDF attachment...
      /.pdf$/i.test(a.localPath) ?
        // ... generate a link to open the PDF in Zotero, abusing the query string to
        // provide a human readable file name like so:
        // zotero://open-pdf/library/items/RANDOM_ZOTERO_KEY?human-readable-name.pdf
        `zotero://open-pdf/library/items/${a.key}?${a.localPath.split(a.key)[1].substring(1)}`
        // ... otherwise, just return the normal file path.
        : a.localPath
[...]

PR #775, opened 2023-04-07
feat(file): support non-path files

bdarcus:
* add citar-file-scheme-skip to define URI file schemes to pass on without validating or normalizing
* citar-file--find-files-in-dirs: use citar-file-scheme-skip

Close: #685

----------------

@dlukes - this turned out to be trickier than I thought, because the existing code was all assuming regular file paths.

EDIT: need to look now at `citar-has-files`, as the indicators aren't correctly matching returned file paths in `citar-open`, and opening files in general aren't working correctly.

### DIFF
--- a/citar-file.el
+++ b/citar-file.el
@@ -395,9 +395,9 @@ SEPCHAR."
       (push (buffer-string) strings))
     (nreverse strings)))
 
-(defun citar-file--scheme-skip-rx ()
-  "Return regexp sting from `citar-file-scheme-skip'."
-  (mapconcat (lambda (s) (concat s "://")) citar-file-scheme-skip "\\|"))
+(defvar citar-file--scheme-skip-rx
+  (mapconcat (lambda (s) (concat s "://")) citar-file-scheme-skip "\\|")
+  "Return regexp sting from `citar-file-scheme-skip'.")
 
 (defun citar-file--normalize-path (file)
   "Return FILE as full path, or if non-file URI schema.
@@ -405,7 +405,7 @@ SEPCHAR."
 Expand non-absolute file paths, but include as is URI schemes in
 `citar-file-scheme-skip'."
   (cond
-   ((string-match (citar-file--scheme-skip-rx) file 0) file)
+   ((string-match citar-file--scheme-skip-rx file 0) file)
    ((and (file-name-absolute-p file)(file-exists-p file))
     (expand-file-name file))))
 
@@ -420,7 +420,7 @@ Expand non-absolute file paths, but include as is URI schemes in
                      (lambda (dir)
                        (let ((filepath (expand-file-name file dir)))
                          (when (or (file-exists-p filepath)
-                                   (string-match (citar-file--scheme-skip-rx) filepath 0))
+                                   (string-match citar-file--scheme-skip-rx filepath 0))
                            filepath)))
                      dirs)))
           (push filepath foundfiles))))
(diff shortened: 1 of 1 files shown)


citar-file.el
```diff
@@ -388,14 +395,24 @@ SEPCHAR."
...
+                    (seq-some
+                     (lambda (dir)
+                       (let ((filepath (expand-file-name file dir)))
+                         (when (or (file-exists-p filepath)
+                                   (string-match schemerx filepath 0))
+                           filepath)))
+                     dirs)))
           (push filepath foundfiles))))
-    (nreverse foundfiles)))
+    (nreverse (remove 'nil foundfiles))))
```
bdarcus, line 415: Don't like this.

citar-file.el
```diff
@@ -388,14 +395,24 @@ SEPCHAR."
...
+               ((string-match schemerx file 0) file)
+               ((and (file-name-absolute-p file)(file-exists-p file))
+                (expand-file-name file)))))
+        (push npath foundfiles)
+        (when-let ((filepath
+                    (seq-some
+                     (lambda (dir)
+                       (let ((filepath (expand-file-name file dir)))
+                         (when (or (file-exists-p filepath)
+                                   (string-match schemerx filepath 0))
```
bdarcus, line 411: Feels odd to have to check this twice.

citar-file.el
```diff
@@ -384,18 +395,36 @@ SEPCHAR."
...
+   ((and (file-name-absolute-p file)(file-exists-p file))
+    (expand-file-name file))))
+
 (defun citar-file--find-files-in-dirs (files dirs)
   "Expand file names in FILES in DIRS and keep the ones that exist."
   (let (foundfiles)
     (dolist (file files)
-      (if (file-name-absolute-p file)
-          (when (file-exists-p file) (push (expand-file-name file) foundfiles))
-        (when-let ((filepath (seq-some (lambda (dir)
```
bdarcus, line 393: I don't think I should remove this.

bdarcus:
@dlukes: I think I'm going to put this aside, as once I make progress in one area, I end up breaking other things.

Any suggestions on easiest/best to support #775 @roshanshariff?

Basic we need to refactor the file code so it doesn't assume only ever standard file paths, with the initial case zotero links.

roshanshariff:
I haven't been following this development very much, but I wonder if it's better to just bypass the `citar-file` functionality completely. In other words, make a separate `citar-zotero` module that handles everything to do with Zotero attachments, including showing the indicators and handling the "open" action. That way `citar-file` can stay focused on actual files, since most of its functionality is not relevant to Zotero attachments anyway.

bdarcus:
> I wonder if it's better to just bypass the citar-file functionality completely. In other words, make a separate citar-zotero module that handles everything to do with Zotero attachments, including showing the indicators and handling the "open" action.

So zotero resources then would be a separate type of related resource, in a separate group from "files"?

And therefore also  a separate command from `citar-open-files`?

They are still files, though, they just use a different mechanism to find them, and are opened in a different viewer, effectively.

I did a few weeks ago merge a change that allow one to set `citar-open-entry` to open in Zotero, rather than a bib or json file. That felt right to me.

But this has gotten more complicated, as I mentioned.

And a separate module opens up other questions, which I'm loathe to tackle, at least anytime soon.

bdarcus:
I think I'm going to close this for now.

dlukes:
Apologies, the whole family's been ill on and off for the past week or so, and I just can't afford to invest much into this right now, especially as the Citar codebase is mostly unfamiliar to me. Again, sorry!

bdarcus:
No problem!

PR #787, opened 2023-07-14
Modify if statements in citar-markdown-insert-citation

### LINKED ISSUE #786: prefixes and suffixes in `citar-markdown-insert-citation`
When I run `citar-insert-citation` in a Markdown buffer and select the work I want to cite, the identifier is inserted preceded by a space and followed by a comma and a space, e.g. `[ @Ord2020PrecipiceExistentialRisk, ]`. The function responsible for this behavior is `citar-markdown-insert-citation` (specifically, the lines `(prenote (if (string= "" prenote)  "" (concat prenote " ")))` and `(postnote (if (string= "" postnote) "" (concat ", " postnote)))`. ) I would prefer not to have these extra characters inserted when no prefix or suffix is provided, which would also be consistent with how citations are inserted in org-mode buffers. Would it make sense to add a user option to control this behavior?

benthamite:
Closes #786


roshanshariff:
I think a better way to write this is `(and prenote (not (string-empty-p prenote)))`, explicitly checking for prenote being non-nil and non-empty. Using `length` only works because `(length nil) => 0`, but it's a little non-obvious.

benthamite:
@roshanshariff, thanks. I think it should be `(or (not prenote) (string-empty-p prenote))`, i.e. the negation of your condition. I made the changes. (I tried to squash the new commit, which I believe is the recommendation, but had an issue with Forge/Magit, so I ended up force-pushing it. I'm not a programmer; apologies if this wasn't the correct process.)

bdarcus:
> I tried to squash the new commit, which I believe is the recommendation, but had an issue with Forge/Magit, so I ended up force-pushing it ...

No worries; I can just squash merge it.

Thanks!

Issue #606, opened 2022-05-23
Suffix appended without whitespace


denismaier:
1. Enter a citation
2. Run `citar-org-update-pre-suffix`
3. Enter a suffix.

=> the suffix is inserted without any separating whitespace. Is this the intended behaviour?

bdarcus:
I vaguely recall this is something about org-cite. I'll see if I can find the email.

bdarcus:
Here's my email to the list, and the subsequent thread.

https://lists.gnu.org/archive/html/emacs-orgmode/2021-10/msg00472.html

bdarcus:
OIC; yes, that's an obvious bug. I never use that function, so never came across it!

bdarcus:
So what's should the behavior be, do you think?

Add a space after the prefix and before the suffix?

denismaier:
> Add a space after the prefix and before the suffix?

I'd think so, yes.

bdarcus:
Only thing is, if one is updating an existing affix, one wouldn't want to add the spaces.

denismaier:
Ok, looks like this is really caused by `org-element-interpret-data`

```
(org-element-interpret-data `(citation-reference (:key "asdf" :prefix "prefix" :suffix "suffix")))

; vs:

(org-element-interpret-data `(citation-reference (:key "asdf" :prefix "prefix " :suffix " suffix")))
```

bdarcus:
Yeah; I kind of think that's a bug in org. WDYT?

denismaier:
Probably yes. I can't claim to totally understand what is happening there, though.

PR #788, opened 2023-07-15
Upgrade citar-org-update-pre-suffix to iterate through all references in citation

### LINKED ISSUE #607: Query affixes after adding key
I think we've discussed that before, but would it be possible to prompt for affixes after selecting a key? Maybe in a loop so users can select multiple keys, add affixes for each until they are finished?

krisbalintona:
To resolve #607 and #606. The following pull request does the following:
1. Renames `citar-org-update-pre-suffix` to `citar-org-update-prefix-suffix`. I found the initial naming potentially misleading. This rename may increase the discoverability of the command.
2. Adds a helper function, `citar-org--update-prefix-suffix`.
3. (Unchanged) When on a reference, will prompt for a new prefix and suffix, with default values being their current values.
4. (New) When on a citation (but not a reference), will prompt for a new prefix and suffix **for every reference in the citation at point** from left to right.
5. (New) When on a reference and called with prefix arg, does the same as (4).
6. (New) There will always be *only one space* between citekey and citation suffix, unless there is no non-whitespace character in the suffix, in which case the suffix will be empty.

This new functionality uses purely the org-element API, meaning this should be a robust solution.

Additionally, I have manually tested the pull request through several edge cases, and have detected no bugs.

Finally, if this pull request goes through, I think it would be a good idea to (i) have `citar-org-update-prefix-suffix` be run after inserting a citation and (ii) have a customizable option to have toggle this behavior, potentially having it enabled by default.

### DIFF
--- a/citar-org.el
+++ b/citar-org.el
@@ -442,14 +442,19 @@ or citation-reference."
 
 (defun citar-org--update-prefix-suffix ()
   "Change the prefix and suffix text of the reference at point."
-  (let* ((datum (org-element-context))
-         (datum-type (org-element-type datum))
-         (ref (if (eq datum-type 'citation-reference) datum
-                (error "Not on a citation reference")))
+  (let* ((ref (org-element-context))
          (key (org-element-property :key ref))
-         ;; TODO handle space delimiter elegantly.
-         (pre (read-string "Prefix text: " (org-element-property :prefix ref)))
-         (post (read-string "Suffix text: " (org-element-property :suffix ref)))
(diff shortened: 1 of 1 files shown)


bdarcus:
> Finally, if this pull request goes through, I think it would be a good idea to (i) have citar-org-update-prefix-suffix be run after inserting a citation and (ii) have a customizable option to have toggle this behavior, potentially having it enabled by default.

Yeah; I was wondering about that after testing it. ;-)

Perhaps on that you could also update the documentation?

Thanks!

PR #790, opened 2023-07-17
Fix edge case for `citar-org-shift-reference-{left,right}`

krisbalintona:
The current behavior for `citar-org-shift-reference-left` and `citar-org-shift-reference-right` fails when, in the same citation, there are multiple references with the same citekey (regardless of prefix and suffix differences). The reason for this lies in `citar-org--get-ref-index` using citekeys to retrieve the index of a reference within a citation.

This pull request has that function use the point at the beginning of a reference instead. Doing so required a slight change to how `citar-org--shift-reference` leaves the point on the correct reference.

Finally, the final commit enhances how the point is left on the reference: rather than left at the beginning, the point retains its position relative to where the command was called in the reference. For instance, if the command was called on the fourth character of the reference, it will remain there.

### DIFF
--- a/citar-org.el
+++ b/citar-org.el
@@ -396,6 +396,8 @@ or citation-reference."
   (let*  ((current-citation (if (eq 'citation (org-element-type datum)) datum
                               (org-element-property :parent datum)))
           (current-ref (when (eq 'citation-reference (org-element-type datum)) datum))
+          (point-offset
+           (- (point) (org-element-property :begin current-ref)))
           (refs (org-cite-get-references current-citation))
           (index
            (citar-org--get-ref-index refs current-ref)))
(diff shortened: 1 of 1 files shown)


bdarcus:
Looks good.

Minor thing, for future reference: we're using conventional commit messages.

https://www.conventionalcommits.org

krisbalintona:
> Looks good.
> 
> Minor thing, for future reference: we're using conventional commit messages.
> 
> https://www.conventionalcommits.org

Thanks for letting me know. I haven't contributed to packages before, so I wasn't aware of the convention. I'll use it moving forward (for `citar` and other packages, if ever).

bdarcus:
It does depend on the project. A lot are using it, however, because the conventions make it suitable for processing.

I can just squash merge this when you're done, however, and rewrite the commit message.

krisbalintona:
> It does depend on the project. A lot are using it, however, because the conventions make it suitable for processing.
> 
> I can just squash merge this when you're done, however, and rewrite the commit message.

I just remade the commit messages to be more helpful and (hopefully) follow the conventions. I haven't discovered any lingering bugs, so I believe this should be ready to go?

bdarcus:
> I just remade the commit messages to be more helpful and (hopefully) follow the conventions. I haven't discovered any lingering bugs, so I believe this should be ready to go?

You have it marked as a draft. I'll unmark it.

Issue #695, opened 2022-10-03
Add database/file annotation to completions


jsilve24:
**Is your feature request related to a problem? Please describe.**
Yes. I have my own bibtex database which is my main / global / default database. I also frequently work with others on local databases that are in a given folder with a document we are working on. Others may add citations to the local bibtex file that are duplicates of something in my database but has a different key. This can cause problems when I am trying to figure out which of two seemingly identical (save different keys) completion candidates to use. In those cases I would prefer to just use the one that is in the local bibtex file. 

**Describe the solution you'd like**
I think its too hard to automate the choice between two candidates but a simple option / annotation in the completions buffer would be enough. Would it be possible to show whether something is from a global file or the local file? This could simply be a ("global"/"local") indicator or simply just given the filename for where the candidate is comming from. 

**Describe alternatives you've considered**
none. automating seems like more of a pain, indicating which file seems easiest.

bdarcus:
It's an interesting problem.

> This could simply be a ("global"/"local") indicator or simply just given the filename for where the candidate is coming from.

Keep in mind, however, one can in theory have multiple global and/or local bibliographies associated with a buffer.

In that situation, I would think the first option likely wouldn't be very useful, and the second could get unwieldy.

But this is also tied to broader issues, like how citar merges multiple bibliographies. In theory, there should really be duplicate entries with different keys, but we currently don't check for that.

I could imagine a feature where we did, and we have some customization option on what to do with duplicates.

cc #645

jsilve24:
While I appreciate the long-term goal / vision of having some complete method of handling duplicates. I think a very simple solution would be to simply check whether the database is included in the variable `org-cite-global-bibliography` or whatnot and just have a binary label (yes/no)/(global/local). Yes you could have multiple global bibliographies with duplicates but how common is that? Or perhaps its just a placeholder annotation that has access to some key variables allowing user to write their own function with that info.

bdarcus:
Right, but we have to think about the bigger picture before considering hard coding something like this.

I don't use local files, for example, so I don't really want such an indicator taking up real estate.

Which is why this:

> Or perhaps its just a placeholder annotation that has access to some key variables allowing user to write their own function with that info.

...may well be a better, more flexible and general, solution? It's why I mentioned #645.

WDYT?

jsilve24:
[...]

PR #800, opened 2023-08-22
Make citar-{bibliography,library-paths,notes-paths} safe local vars

ashton314:
It is incredibly handy to customize `citar-bibliography` as a directory-local variable on a per-project basis. This worked so fabulously with the paper I just finished that I can't see myself not using this workflow again. I think a lot of people would like to use Citar in this way too.

I've added autoload cookies to put the safe local variable predicate on the symbol; the predicate is simply a list of strings. I followed how Denote by Prot was setting the safe local variable predicate on some denote variables.

### DIFF
--- a/README.org
+++ b/README.org
@@ -171,6 +171,17 @@ By default, this uses ~citar-open-entry-in-file~, which will open the relevant b
 The other included option is ~citar-open-entry-in-zotero~, which will select the item in Zotero.
 Note that functionality depends on [[https://retorque.re/zotero-better-bibtex/][Better BibTeX]] (which you should be using anyway!).
 
+*** Per-project customization
+
+When collaborating with others, it might be convenient to have a project-specific bibliography file that can be version controlled and shared among collaborators.
+You can set ~citar-bibliography~ with =add-file-local-variable= or =add-dir-local-variable= to add a custom value to ~citar-bibliography~ for just that file or directory respectively.
+
+For example, if you are sharing a =custom_refs.bib= file adjacent to the =cool_new_paper.tex= that you're working on, using the aforementioned functions or adding the following to your =.dir-locals.el= will make citar /only/ search your shared bibliography:
+
+#+begin_src emacs-lisp
+((latex-mode . ((citar-bibliography . '("custom_refs.bib")))))
+#+end_src
+
 ** Rich UI
 :PROPERTIES:
 :CUSTOM_ID: rich-ui
--- a/README.org
+++ b/README.org
@@ -174,13 +174,10 @@ Note that functionality depends on [[https://retorque.re/zotero-better-bibtex/][
 *** Per-project customization
 
 When collaborating with others, it might be convenient to have a project-specific bibliography file that can be version controlled and shared among collaborators.
-You can set ~citar-bibliography~ with =add-file-local-variable= or =add-dir-local-variable= to add a custom value to ~citar-bibliography~ for just that file or directory respectively.
-
-For example, if you are sharing a =custom_refs.bib= file adjacent to the =cool_new_paper.tex= that you're working on, using the aforementioned functions or adding the following to your =.dir-locals.el= will make citar /only/ search your shared bibliography:
-
-#+begin_src emacs-lisp
-((latex-mode . ((citar-bibliography . '("custom_refs.bib")))))
-#+end_src
+Citar automatically uses bibliographies specified with ~\bibliography~ or ~\addbibresource~ in LaTeX and the ~#+bibliography~ keyword in org-mode.
+However, Citar will only use these if ~citar-bibliography~ is not set.
+Thus, if you have a global bibliography you've set in your ~init.el~ file, you can unset ~citar-bibliography~ on a per-project basis by setting it to ~nil~ as a file-local or a directory-local variable.
+For example, to set it to ~nil~ in your file-local variables, do ~M-x add-file-local-variable RET citar-bibliography RET nil RET~.
 
 ** Rich UI
 :PROPERTIES:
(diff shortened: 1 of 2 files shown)


bdarcus:
Thanks for this!

Can you just clarify, including in the README, how and why one would use this, rather than just local bibliographies?

ashton314:
Conventional Commits are new to me—I've seen them but not used them. Would you consider this a `feat`?

ashton314:
OK @bdarcus I've updated the commit message and added a little blurb in the README—happy to expand if you'd like.

> If you're comfortable rewriting history…

Magit makes this dangerously easy. :) I can't imagine using Git now without it. In fact, I opened this fork and PR entirely with Magit+Forge!

roshanshariff:
I'd also like to hear about your workflow @ashton314. For the bibliographies, perhaps it makes more sense to have a separate buffer-local variable, perhaps `citar-local-bibliography`, which is added to the list of buffer-local bibliographies parsed from Latex, Org, Markdown, etc.

And while I can also see the utility in project-specific notes, I'm curious to learn how you're using per-project library paths.

I'm not opposed to marking these variables safe-local, but first I'd like to understand if there are better ways of accomplishing the purpose you had in mind. For instance, just overriding these variables locally makes it hard to do things like configurably combining local and global paths.

bdarcus:
> Conventional Commits are new to me—I've seen them but not used them. Would you consider this a `feat`?

Yes.

You'd want the commit message to be 50 characters max, and lowercase.

Thanks for the README addition, but I'm still not totally clear, nor it seems is @roshanshariff. Could you please respond to him?

ashton314:
Hey—sorry for the long wait. Semester started and I got inundated. 😅 Surfacing now!

I used a local value for `citar-bibliography` for a shared paper to great effect. This is what I had at the bottom of my `.tex` file:

```latex
% Local Variables:
% jinx-local-words: "..."
% citar-bibliography: ("./ref.bib")
% End:
```

I was collaborating with several other authors—this made it nice for me to scope citar's UI to just the papers that we had in our bibliography. If a ref I wanted was missing from the list, it was a good clue for me to go back to my personal `.bib` file and add it.

I'll confess I haven't actually used the other two variables: they seemed like good variables to change in tandem with the `citar-bibliography` file, as I (perhaps mis-)understood it, these paths should be somewhat related to one another.

If going the route of a buffer-local variable, I think it might be nice to have a switch or some kind of narrowing key to restrict citar results to only what is available locally. That way a missing ref in the project-specific `.bib` file can get caught faster before trying to render the paper.

ashton314:
If you'd prefer; I'll narrow the scope of this PR to just updating the `citar-bibliography` variable and we can discuss the other two separately. Lmk what you'd like me to do.

bdarcus:
> If going the route of a buffer-local variable, I think it might be nice to have a switch or some kind of narrowing key to restrict citar results to only what is available locally.

I think this is related to #695.

> If you'd prefer; I'll narrow the scope of this PR to just updating the citar-bibliography variable and we can discuss the other two separately. Lmk what you'd like me to do.

I'm unsure, since safe local variables are new to me.

Let's see if @roshanshariff has further input, though I know he, like us, is busy.

roshanshariff:
Maybe we can think of the two requested subfeatures separately?

1. Specifying local bibliographies: we currently support mode-specific ways of doing this, like the `\bibliography` or `\addbibresource` macros in Latex (handled by Reftex) or the `#+bibliography` keyword in Org mode. If these aren't sufficient (are they?) we could add a safe local variable that lets you specify more files, maybe `citar-local-bibliography`.
2. Restricting the citar interface to local bibliographies only. A less intrusive alternative would be to add a custom indicator as in #695. If that's not sufficient, we could add a local minor mode (`citar-local-only-mode`?) that makes citar only show entries from local bibliographies. I'm inclined not to make this a safe local variable because I don't want files to be able to silently alter the behaviour of Citar, which I rely on to access my bibliography throughout Emacs. But, of course, you could always add the variable to `safe-local-variable-values` if you want to allow this, and Emacs will prompt you to do so if you open a file that tries to set it.

bdarcus:
> 1. If these aren't sufficient (are they?) ...

And if they aren't, why not?

The advantage of how we do things now what local files tex, org etc see is the same as what citar sees.

aikrahguzar:
> The advantage of how we do things now is what local files tex, org etc see is the same as what citar sees.

I think this is a big advantage and reading the discussion, it seems like the same use case can be met by just unsetting the `citar-bibliography`. Citar will use just the local bibliographies in that case. So, in my mind the only local value for `citar-bibliography` ever needed should be `nil`. Is there any use case not covered by that?

ashton314:
(Hey, sorry again for the super long delay. I've been slammed this semester.)

I didn't realize that citar would read what tex and org see—that's perfect.

In that case, I am in favor of making `nil` the only baked-in safe buffer-local value for `citar-bibliography`.

ashton314:
@bdarcus I think this is ready now! Thanks all for your patience.

roshanshariff:
@ashton314,

Citar uses local bibliographies _in addition_ to those set in `citar-bibliography`. It combines the entries from all local and global bibliographies. In particular, you don't need to set the global bibliography to `nil` to use local bibliographies.

So if your collaborators don't use Citar themselves, they will have no global bibliographies and only local ones will be used. If they do have global bibliographies set up, I'm not sure they would expect Citar to automatically, without warning, ignore them when inside a particular project.

I see that you prefer the global bibliographies to be suppressed for certain projects, but since allowing this is a personal preference, it naturally belongs in your own private configuration?

bdarcus:
Per Roshan's explanation, and in https://github.com/emacs-citar/citar/pull/831#issuecomment-2105980044, I am closing this.

Thanks for the PR, @ashton314, but we'll need to find another way (which Roshan suggests in the above linked comment).

Issue #804, opened 2023-10-08
Customizable `citar-add-file-to-library`


mpedramfar:
**Is your feature request related to a problem? Please describe.**
Currently, the behavior of the function `citar-add-file-to-library` is not that customizable.
For example:
- What if I want to confirm the filename before it's being saved?
- What if I want the path of the saved file to be added to the `file` field of the bibtex entry?

**Describe the solution you'd like**
One way to implement it would be to have something like this:
```emacs-lisp
(defun citar-save-file (fun citekey name)
  "FUN is a function that takes one argument, DESTFILE, and saves the file to DESTFILE.
NAME is the initial name of the source file."
  (let* ((directory (if (cdr citar-library-paths)
                        (completing-read "Directory: " citar-library-paths)
                      (car citar-library-paths)))
         (extension (or (and (stringp name) (file-name-extension name))
                        (read-string "File extension: ")))
         (destfile (expand-file-name citekey directory))
         (destfile (if (string-empty-p extension)
                       destfile
                     (concat destfile "." extension))))
    (funcall fun destfile)))


(setq citar-save-file-function #'citar-save-file) ;; this can be a customizable variable


(defun citar-add-file-to-library (citekey)
  "Add a file to the library for CITEKEY.
The FILE can be added from an open buffer, a file path, or a
URL."
  (interactive (list (citar-select-ref)))
  (citar--check-configuration 'citar-library-paths)
  (unless citar-library-paths
    (user-error "Make sure `citar-library-paths' is non-nil"))
  (let (fun name)
    (pcase (read-char-choice "Add file from [b]uffer, [f]ile, or [u]rl? " '(?b ?f ?u))
      (?b
       (let ((buf (read-buffer "Add file buffer: " (current-buffer))))
         (setq name (and buffer-file-name (file-name-nondirectory buffer-file-name)))
         (setq fun (lambda (destfile)
                     (with-current-buffer buf
                       (write-file destfile 'confirm))))))
      (?f
       (let* ((file (read-file-name "Add file: " nil nil t)))
         (setq name (file-name-nondirectory file))
         (setq fun (lambda (destfile)
                     ;; last arg integer means to confirm before overwriting
                     (copy-file file destfile 1)))))
      (?u
       (let* ((url (read-string "Add file URL: ")))
         ;; TODO: Use Content-Type HTTP response header to guess file extension
         (setq name (url-file-nondirectory url))
         (setq fun (lambda (destfile)
[...]

PR #806, opened 2023-10-15
Add options to customize citar-add-file-to-library

mpedramfar:
PR for issue #804

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -419,8 +419,8 @@ the user, this function should decide the appropriate
 destination DESTFILE to save the file.
 It should also decide what to do if DESTFILE already
 exists by choosing OK-IF-ALREADY-EXISTS.
-(See the documentation of `copy-file' for the behavior of
-this variable)
+\(See the documentation of `copy-file' for the behavior of
+this variable\)
 
 Then it should call FUN with DESTFILE and
 OK-IF-ALREADY-EXISTS, which will save the file."
@@ -1331,7 +1331,13 @@ getters for file, note, and link resources."
             (putresult citekey (citar-get-entry citekey))))))))
 
 (defun citar-save-file (citekey fun info)
-  "Default save function for `citar-save-file-function'."
+  "Default save function for `citar-save-file-function'.
+
+If there is more than one directory in `citar-library-paths',
+query the use for the directory to save.
+If no extension is provided in INFO, query the user for extension.
+Call FUN to save the file as CITEKEY.EXTENSION and
+if file already exists, confirm before overwriting it."
   (let* ((directory (if (cdr citar-library-paths)
                         (completing-read "Directory: " citar-library-paths)
                       (car citar-library-paths)))
@@ -1345,7 +1351,10 @@ getters for file, note, and link resources."
          (ok-if-already-exists 1))
     (funcall fun destfile ok-if-already-exists)))
 
-(defun citar-add-file--from-buffer (citekey)
+(defun citar-add-file--from-buffer (_citekey)
+  "Add file to library from buffer.
+
+See the documentation for `cita-add-file-sources' for more details."
   (let* ((buf (get-buffer (read-buffer "Add file buffer: " (current-buffer))))
          (fun `(lambda (destfile ok-if-already-exists)
                  (if (and (not ok-if-already-exists)
@@ -1359,14 +1368,20 @@ getters for file, note, and link resources."
                        (file-name-extension (buffer-file-name buf))))))
     (cons fun info)))
 
-(defun citar-add-file--from-file (citekey)
+(defun citar-add-file--from-file (_citekey)
+  "Add an existing file to library by copying it.
+
+See the documentation for `cita-add-file-sources' for more details."
   (let* ((file (read-file-name "Add file: " nil nil t))
          (fun `(lambda (destfile ok-if-already-exists)
                  (copy-file ,file destfile ok-if-already-exists)))
          (info (list :extension (file-name-extension file))))
     (cons fun info)))
 
-(defun citar-add-file--from-url (citekey)
+(defun citar-add-file--from-url (_citekey)
+  "Add file to library from url.
+
+See the documentation for `cita-add-file-sources' for more details."
   (let* ((url (read-string "Add file URL: "))
          (fun `(lambda (destfile ok-if-already-exists)
                  (url-copy-file ,url destfile ok-if-already-exists)))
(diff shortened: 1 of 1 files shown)


bdarcus:
Thanks!

The CI has flagged a few things you'll want to fix; seems mostly documentation related.

mpedramfar:
Thanks for the tips!
It's probably easiest if you squash merge them when done.

citar.el
```diff
@@ -1638,30 +1746,29 @@ URL."
...
-              (destfile (funcall withext (file-name-extension file))))
-         (copy-file file destfile 1)))  ; last arg integer means to confirm before overwriting
-      (?u
-       (let* ((url (read-string "Add file URL: "))
-              ;; TODO: Use Content-Type HTTP response header to guess file extension
-              (ext (url-file-extension url))
-              (destfile (funcall withext (when (> (length ext) 1) (substring ext 1)))))
-         (url-copy-file url destfile 1))))))
+  (unless citar-add-file-sources
+    (user-error "Make sure `citar-add-file-sources' is non-nil"))
```
roshanshariff, line 1750: Instead of checking `citar-add-file-sources` here, add a clause to the `citar--check-configuration` function and just call `(citar--check-configuration 'citar-add-file-source)`. We'd like to centralize all such validation in one place.

bdarcus:
This is really weird, and I guess a bug in GH. But @roshanshariff left a comment that is not now showing in the web UI.

I'll reproduce it below just to be sure:

> Instead of checking citar-add-file-sources here, add a clause to the citar--check-configuration function and just call (citar--check-configuration 'citar-add-file-source). We'd like to centralize all such validation in one place.

If it shows up, I'll delete this.

roshanshariff:
@bdarcus Sorry, I was doing a code review and accidentally submitted a single comment instead of submitting as a batch.

I like the general direction of this change. I just thought the UI for selecting a file source could be simplified using the `read-multiple-choice` function instead of reimplementing similar functionality. For example, suppose `citar-add-file-sources` was defined as follows:
```elisp
(setq citar-add-file-sources
  '((?b "buffer" "current buffer" citar-add-file--from-buffer)
    (?f "file" "existing file" citar-add-file--from-file)
    (?u "url" "download from URL" citar-add-file--from-url)))
```
Then the code to select a file source would just be
```elisp
(nth 3 (read-multiple-choice "" citar-add-file-sources))
```
This handles highlighting the shortcut character, shows extra help if requested, etc.

mpedramfar:
I added the check for `citar-save-file-function` to `citar--check-configuration`.
We can add the check for `citar-add-file-sources` after we're certain about its type.

When it comes to `citar--check-configuration`, there are some things in the code seem inconsistent.
1. `error` vs `user-error`: Currently `citar--check-configuration` checks if the type of `citar-library-paths` is correct (a list of strings) but it doesn't check if it's non-nil. If the type is incorrect, it raises an `error`, with backtrace and everythin. Inside `citar-add-file-to-library` it's checked if this variable is non-nil and if it is nil, it raises a `user-error`.  

It's not clear to me why `error` is being used at all. Wouldn't `user-error` be better? And wouldn't it be better if we check if `citar-library-paths` is non-nil inside `citar--check-configuration`?

2. There seems to be a lot of customization options that do not appear in `citar--check-configuration`. Here are the the source/resource variables:
- `citar-open-resources`: not checked
- `citar-file-sources`: not checked
- `citar-notes-sources`: not checked in `citar--check-configuration`. Instead, there is a separate function that checks a single source (i.e. `citar--check-notes-source`) and this is only called from `citar-register-notes-source`. If the user changes `citar-notes-sources` using `setq` or the customize interface, it won't be checked at all.  

Wouldn't it be better to move all the checks to `citar--check-configuration`?

roshanshariff:
Another small suggestion: instead of having the file sources return a `(function . plist)`, just have them return a plist with the function included under a `:write-file` key. In the future, we could have other kinds of file sources that don't fit that abstraction (e.g. open and return a new Emacs buffer) and expose actions other than `:write-file`.

roshanshariff:
@mpedramfar, yeah the usage of `citar--check-configuration` is somewhat inconsistent at the moment. Ideally, it would check the types of customization variables and raise `user-error`. Any functions that have additional requirements (like `citar-library-paths` being non-empty) would do that check on their own, after the call to `citar--check-configuration`.

New code should follow this pattern, and over time the old code can be changed as well.

mpedramfar:
I see, there is a use of `citar-library-paths` that doesn't need it being non-nil, so it makes sense to check it only when we need it to be non-nil.
There is no use case of `citar-add-file-sources` when it is nil. Is there a way to specify this in defcustom `:type`?
Should we check for `citar-add-file-sources` being non-nil in `citar--check-configuration` or `citar-add-file-to-library`?

roshanshariff:
Now that I look at it, it doesn't seem like there's a good built-in type for non-nil lists. I think it's fine to go back to what you had, just checking the list is non-nil before using it.

bdarcus:
So what's the status here, given the back-and-forth?

mpedramfar:
I applied the comments. Let me know if there are other issues to fix.

roshanshariff:
I pushed a couple of commits with some changes. I also tested the code a little, but @bdarcus if you have time to give it a whirl, that couldn't hurt :)

If we don't discover any issues, I'm okay with merging now!

* I copy-edited the docstrings and refined the type declarations to present a better customization UI.
* @mpedramfar, `lambda` functions don't need to be quoted because they are [self-quoting forms](https://www.gnu.org/software/emacs/manual/html_node/elisp/Anonymous-Functions.html#index-lambda). And since we're using lexical bindings, lambda functions can capture local variables from the surrounding context. I made the needed changes.
* The "add file from buffer" function was using `write-file`. I changed it to use `write-region` instead. The former makes the buffer visit the destination file and marks it as unmodified, which struck me as rather dangerous. Instead, we should probably just write the contents of the buffer to the destination file but not touch the buffer in any other way, which is what `write-region` does. What do you think?

Thanks @mpedramfar for suggesting this change! I'm looking forward to writing more functions to do things like saving to Zotero libraries (which is what I use). Then we can probably rename `citar-add-file-to-library` to just `citar-add-file`. When I have time, I'd like to collect some of the integration code that's floating around into a `citar-zotero` package.

mpedramfar:
Thanks for the tips!

Using `write-region` makes sense. I think the only exception is the one you mentioned, when the buffer is visiting the destfile that we want to save into. In that case, calling `write-region` saves the buffer to its file, but calling `buffer-modified-p` returns `t`. So it makes sense to simply call `save-buffer` instead. I implemented that in the last commit.

As a side-note, you might find the [Zotra](https://github.com/mpedramfar/zotra) package useful. It basically gives Emacs one of the main features of Zotero, i.e. capturing bib entries and attachments from the web. After this PR is merged, I'll add the integration code to Zotra, so that it will can be added as one of the sources in `citar-add-file-sources`.

bdarcus:
CI is complaining :-)

mpedramfar:
Should be ok now :)

bdarcus:
Merged; thanks!

I did test the basics, and didn't see any issues. But I don't tend to use this command anyway!

PR #807, opened 2023-10-23
Generalize rendering of names for completion

torfjelde:
Generalizes `citar--shorten-names` + implements rendering of full names as discussed in #805.

Issue #805, opened 2023-10-10
Display transform function for full namees


torfjelde:
**Is your feature request related to a problem? Please describe.**
I often find myself wanting to search through my bibliographies using the first name. This might just be because I remember the first name but not the last name, or there are too many authors with the same last name.

**Describe the solution you'd like**
I'm thinking something like (basically just a copy-paste of `citar--shorten-names`):

```emacs-lisp
;; Similar to `citar--shorten-name' and `citar--shorten-names'
;; but using the full name instead of just the last name.
(defun citar--full-name (name)
  "Return full NAME, i.e. `given family`, from `family, given' string.

Otherwise, return as is."
  (let (split-names (split-string name ", "))
    (if (> (length split-names) 1)
        (concat (cadr split-names) " " (car split-names))
      name)))

(defun citar--full-names (namestr &optional truncate andstr)
  "Return a list of full names from a list of full NAMESTR.

With an integer TRUNCATE, will shorten the list, and ANDSTR will
replace last comma."
  (let* ((namelist (split-string (or namestr "") " and "))
         (namelength (length namelist))
         (tnamelist (seq-take namelist (or truncate namelength)))
         (tnamelength (length tnamelist)))
    (mapconcat
     (lambda (n)
       (let* ((shortname (citar--full-name n))
              (pos (citar--shorten-name-position tnamelist n))
              (suffix
               (cond
                ;; if last name in the list and we're truncating add et al.; otherwise, no suffix
                ((equal pos tnamelength)
                 (if (< tnamelength namelength) " et al." ""))
                ;; if second to last in the list, and ANDSTR, use that
                ((and andstr (equal pos (- tnamelength 1)))
                 (concat " " andstr " "))
                ;; otherwise, use a comma
                (t ", "))))
         (concat shortname suffix)))
     tnamelist "")))
```

in combination with
```emacs-lisp
(setq citar-display-transform-functions
   `((sn . (citar--shorten-names))
     (etal . (citar--shorten-names 3 "&"))
     (fn . (citar--full-names))))
```

so I can use this in templates with the transform key `fn`.

**Describe alternatives you've considered**
One alternative is to "maintain" the usage of "and" in the author list, but IMO this becomes quite cluttered when searching, and so I prefer the "given1 family1, given2 family2, ..." instead of `given1 family1 and given2 family2 and ...".

bdarcus:
Given this behavior is already configurable (just create your own function and use that in your templates), is there really any point in changing the default function there?

[...]

Issue #780, opened 2023-05-09
Expand test coverage, integrate into CI


bdarcus:
Self-explanatory; we haven't done much with this.

PR #808, opened 2023-10-23
Generalize `citar--shorten-names` + add display transform function for full names

torfjelde:
Generalizes citar--shorten-names + implements rendering of full names as discussed in #805.

Replaces #807 as I did not follow the contribution guidelines in #807.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -191,6 +191,7 @@ references as a string."
 (defcustom citar-display-transform-functions
   ;; TODO change this name, as it might be confusing?
   `((sn . (citar--shorten-names))
+    (fn . (citar--full-names))
     (etal . (citar--shorten-names 3 "&")))
   "Configure transformation of field display values from raw values.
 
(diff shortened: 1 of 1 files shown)


torfjelde:
@bdarcus are there any tests for this stuff?

bdarcus:
> are there any tests for this stuff?

Do you mean in this project, or more generally?

If the former, this is an example (see #780) where we should probably add some.

torfjelde:
For the project, and specifically of the display transform functions:)

But aight; should I just make a new `ert-deftest` for the short- and full-name methods? Are there an existing list of author names that I could maybe draw from as test-cases, or should I add some?

bdarcus:
> Are there an existing list of author names that I could maybe draw from as test-cases, or should I add some?

There isn't, but I can give you an example of a key type that I included when developing the existing function; a corporate (organizational) name:

```
World Bank
```

That shouldn't be shortened.

Though you shouldn't worry about it, you should be aware of the discussion here:

https://github.com/emacs-citar/citar/issues/532

bdarcus:
So what do we want to do with this, given @aikrahguzar's alternative?

https://github.com/emacs-citar/citar/issues/805#issuecomment-1783892531

bdarcus:
> So what do we want to do with this, given @aikrahguzar's alternative?
> 
> [#805 (comment)](https://github.com/emacs-citar/citar/issues/805#issuecomment-1783892531)

I'm going to close this, and we can consider @aikrahguzar's alernative.

PR #815, opened 2023-12-26
Fix bug in `citar-citeproc-select-csl-style`

benthamite:
Since `list` is a list of cons cells, to get the name of the csl style when `list` has one element `caar` needs to be used; `car` returns the cons cell itself.

### DIFF
--- a/citar-citeproc.el
+++ b/citar-citeproc.el
@@ -83,7 +83,7 @@ accepted.")
                 files))
          (style
           (if (= (length list) 1)
-            (car list)
+              (caar list)
             (completing-read "Select CSL style file: " list nil t)))
          (file (cdr (assoc style list))))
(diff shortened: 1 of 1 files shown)


benthamite:
Separately (as I just noticed), I think the prompt on line 87 should be "Select CSL style: ", since the user is prompted to select a style, not a file.

bdarcus:
> Separately (as I just noticed), I think the prompt on line 87 should be "Select CSL style: ", since the user is prompted to select a style, not a file.

Could you please make that change while you're at it?

benthamite:
Yep, done.

bdarcus:
Thanks!

I did squash the commits so I could easily rewrite the commit message;

https://www.conventionalcommits.org

PR #840, opened 2024-09-05
feat: enable recursive library paths

### LINKED ISSUE #839: Allow recursive library paths
**Is your feature request related to a problem? Please describe.**

I like to keep my library documents in subdirectories, one for each library entry. In order to make this work, I need to add each subdirectory to `citar-library-paths`. Additionally, whenever I add a new library file, I have to go and change `citar-library-paths`. It would be nice to instead have the possibility of letting `citar-library-paths` be interpreted recursively. 

**Describe the solution you'd like**

Add a new variable, e.g., `citar-library-paths-recursive`, that, when true, makes Citar look in all subdirectories of each entry in `citar-library-paths`. This new variable would be set to `nil` by default for backwards compatibility.

**Describe alternatives you've considered**

The obvious alternative is to transpose the directory layout; i.e., have a separate directory for each file type. This won't work for my use-case, which is to have Xournal++ files linked to my PDFs.

The other solution is the null solution; i.e., just use a flat directory structure.

heraplem:
Fixes  #839.

### DIFF
--- a/citar-file.el
+++ b/citar-file.el
@@ -80,6 +80,7 @@ separator that does not otherwise occur in citation keys."
 
 (defvar citar-notes-paths)
 (defvar citar-library-paths)
+(defvar citar-library-paths-recursive)
 (defvar citar-library-file-extensions)
 (defvar citar-note-format-function)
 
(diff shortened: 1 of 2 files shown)


bdarcus:
Thanks for this!

The CI is complaining about some things. Beyond that, is this still in progress? The feature doesn't appear to actually be implemented here.

Issue #842, opened 2024-09-16
More performant rendering/activation of citations


krisbalintona:
**Is your feature request related to a problem? Please describe.**
I notice that the rendering/activation of org-mode citations with citar (i.e. when `org-cite-activate-processor` is `'citar`) is *very* slow. When editing a line with one or more citations, Emacs freezes for multiple seconds on every character inserted or deleted. This is even the case for small .bib files.

**Describe the solution you'd like**
When editing (inserting or deleting characters) a line with citation(s), render citations as quickly as possible so that Emacs does not freeze.

**Describe alternatives you've considered**
1. Other packages such as `org-cite-activate-processor` that activate citations differently. Different packages have different pros and cons to how they activate citations --- and having to install another package is not much of a solution with respect to citar. I did not like this solution.
2. I defined by own activate function and replaced the `org-cite-basic-activate` in `citar-org-activation-functions` with it. My version, below, is identical to `org-cite-basic-activate` but with faster rendering by using citar's functions which also leverage its cache (I believe). **I currently use this approach.**
``` emacs-lisp
   (defun kb/citar-basic-activate (citation)
    "Set various text properties on CITATION object.

Fontify whole citation with `org-cite' face.  Fontify key with `error' face
when it does not belong to known keys.  Otherwise, use `org-cite-key' face.

Moreover, when mouse is on a known key, display the corresponding bibliography.
On a wrong key, suggest a list of possible keys, and offer to substitute one of
them with a mouse click.

My version of this function uses the speed of `citar' (and its cache) to
eliminate the extraordinary slow down of the default function's
rendering. I believe the bottlenecks are `org-cite-basic--all-keys' and
`org-cite-basic--get-entry' so I have replaced them with equivalent
functions from `citar'."
    (pcase-let ((`(,beg . ,end) (org-cite-boundaries citation))
                ;; NOTE 2024-09-05: Use `citar' (and its cache) to get all keys
                (keys (let (keys)
                        (maphash (lambda (key value) (push key keys))
                                 (citar-get-entries))
                        keys)))
      (put-text-property beg end 'font-lock-multiline t)
      (add-face-text-property beg end 'org-cite)
      (dolist (reference (org-cite-get-references citation))
        (pcase-let* ((`(,beg . ,end) (org-cite-key-boundaries reference))
                     (key (org-element-property :key reference)))
          ;; Highlight key on mouse over.
          (put-text-property beg end
                             'mouse-face
[...]

PR #845, opened 2024-10-22
Use a more performant org-cite activate processor by default

krisbalintona:
As described in #842, citar uses the built-in `org-cite-basic-activate` which can cause significant slowdowns. This PR defines our own activate function, `citar-org-cite-basic-activate`, and uses it by default (replaces `org-cite-basic-active` in `citar-org-activation-functions`). This new activate function is identical to the built-in basic activate function but uses citar functions to speed up the expensive operations of the built-in activate function.

### DIFF
--- a/README.org
+++ b/README.org
@@ -361,7 +361,7 @@ For example, if point:
 - is on an existing citation-reference, you will be prompted to replace it
 - follows or precedes a citation-reference, you will be prompted to add a new citation-reference
 
-The "activate processor" runs the list of functions in ~citar-org-activation-functions~, which by default is the ~basic~ processor from ~oc-basic~ to provide fontification, and also a little function that adds a keymap (~citar-org-citation-map~) for editing citations at point.
+The "activate processor" runs the list of functions in ~citar-org-activation-functions~, which by default uses ~citar-org-cite-basic-activate~, a version of the ~basic~ processor from ~oc-basic~ to provide fontification that leverages citar's performant caching, as well as a little function that adds a keymap (~citar-org-citation-map~) for editing citations at point.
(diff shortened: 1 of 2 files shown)


bdarcus:
Thanks!

The CI has picked up a linting error. Can you see about fixing that?

Come to think about it, it may be related to an earlier issue. I'll check back later today though.

krisbalintona:
> Thanks!
> 
> The CI has picked up a linting error. Can you see about fixing that?
> 
> Come to think about it, it may be related to an earlier issue. I'll check back later today though.
> 
> EDIT: looks like recently org devs moved `org-element-property` and `org-element-type` to a new `org-element-ast.el`, that's what generating the error. Not sure how to address that ATM.
> 
> Seems like maybe:
> 
> ```emacs-lisp
> (if (require 'org-element-ast nil 'noerror)
>   (require 'org-element)
>   (message "Loaded org-element"))
> ```
> 
> ... and:
> 
> ```emacs-lisp
> (if (featurep 'org-element-ast)
>     (declare-function org-element-property "org-element-ast")
>   (declare-function org-element-property "org-element"))
> ```

So, to be clear, should I be waiting for a resolution on main?

krisbalintona:
> > So, to be clear, should I be waiting for a resolution on main?
> 
> I don't think it matters. I was partly just updating to includes notes for myself.
> 
> If I update it on main before merging, do you think it will be a problem for you?

Nope, no problem for me 👍

bdarcus:
I ended up merging this first, as I haven't yet figured out how to fix the issue with Emacs Master.

krisbalintona:
> I ended up merging this first, as I haven't yet figured out how to fix the issue with Emacs Master.

Sounds good. Thank you!

PR #847, opened 2024-10-26
fix(org): update load, declare-function

bdarcus:
In Emacs 31, org-element-property and org-element-type has been moved to a new file: org-element-ast.

Update code to handle this properly, depending on version.


bdarcus:
@yantar92 - this little PR is me trying to update citar to accommodate you moving a couple of functions to `org-element-ast`, but my CI is complaining about the code I have here. Any suggestions for fixing it?

yantar92:
Update https://github.com/emacs-citar/citar/blob/main/citar-org.el#L31

bdarcus:
Right, but how do I do it conditionally? I still have to use `org-element` for earlier versions.

yantar92:
I suspect that you cannot do it conditionally. You may consider using ext:, which will suppress CI failures.

PR #848, opened 2024-10-27
fix(org): update load, declare-function

bdarcus:
In Emacs 31, org-element-property and org-element-type have moved to a new file: org-element-ast.

Update code to handle this properly.

PS - closed #847 in favor of this because I made a git mistake and didn't want to fix it.

### DIFF
--- a/citar-org.el
+++ b/citar-org.el
@@ -21,15 +21,20 @@
 
 (require 'citar)
 (require 'org)
-(require 'org-element)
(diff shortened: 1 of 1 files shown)


bdarcus:
So I now have it so that the CI doesn't complain, and it seems to work as expected on my Emacs version (29).

But I don't have time to install and test master. So it's possible this will require some additional tweaks at some point.

PR #850, opened 2024-11-03
Add indicators nerd icons

larrasket:


bdarcus:
Hi - thanks for this!

But the way that part of the README is designed, it points (and in this version continues to point) to the wiki for custom configurations like this.

I could go either way on whether to include this, but care to explain why this makes more sense to include in the README proper?

And if I find it convincing, you'll need to revise the text preceding your change (maybe move it that down and add "other"?) so it's better integrated.

larrasket:
bdarcus ***@***.***> writes:

> I could go either way on whether to include this, but care to explain
> why this makes more sense to include in the README proper?
>
> And if I find it convincing, you'll need to revise the text preceding
> your change (maybe move it that down and add "other"?) so it's better
> integrated.

I just thought it will be useful for people who want to get started
quickly, especially that many people moved to using nerd-icons instead
of all-the-icons, made sense to have an example for that.

--
Regards,
Salih

PR #855, opened 2025-01-13
Don't make reftex do a rescan everytime

### LINKED ISSUE #854: Performance of `citar-latex-local-bib-files`
After looking at `citar-latex-local-bib-files` for #852, I am a bit worried about its performance. When it was originally written it was supposed to be called rarely. We would call it and use the files returned to generate the cache for entries and then continue happily until the cache was somehow invalidated. As a result it made sense to ask `reftex` to parse the document to make sure we find the file. However, with the redesign of cache, it is called almost every time user needs to select an entry for any reason and reftex ends up reparsing the document to find the local bib file every time. I tend to have fairly small latex files (less than 100 pages) and I am not too sensitive to latency for `LaTeX` at least. For people with bigger documents or older computers the time this takes might be too much. Reftex already caches the path of the bib file and with the current design it makes sense to not ask it to reparse every time.

With the current definition, (on a 244 kb LaTeX document)
```emacs-lisp
(benchmark-run-compiled 100 (citar--bibliographies))
```
give me
```
(3.8437003109999996 4 0.39145653800005675)
```
while with,
```emacs-lisp
(defun citar-latex-local-bib-files ()
  "Local bibliographic for latex retrieved using reftex."
  (ignore-errors (copy-sequence (reftex-get-bibfile-list))))
```
it reduces to,
```
(0.010558950000000001 0 0.0)
```
i.e. almost all the time is taken by the parsing of document.

So, I want to change the definition to the one above. However, there should be way to reparse the document. For this reftex has two interactive functions: `reftex-parse-all` and `reftex-parse-one`. I think merely suggesting these in the documentation if `citar` doesn't pick up the local bib file is enough. This should happen only when a bib file is added or changed. For me that is one time per document so very rare.

@roshanshariff any ideas/suggestions?

aikrahguzar:
fix #854

### DIFF
--- a/README.org
+++ b/README.org
@@ -333,6 +333,14 @@ If a bib file changes, the cache will automatically update the next time you run
 Note that cached data preformatted completion candidates are independently tracked by file.
 So, for example, if you have one very large bibliography file that changes a lot, you might consider splitting into one large file that is more stable, and one-or-more smaller ones that change more frequently.
 
+*** Reftex and local bibliographic files in LaTeX
+
+Citar relies on the builtin library reftex to find the bibliographic files included in a LaTeX document.
+Reftex scan the LaTeX document to find these files and caches them.
+If you find that Citar isn't picking up bibliographic entries in the included files it is most likely due to the fact that reftex needs to do a rescan.
+You can use ~reftex-parse-all~ or ~reftex-parse-one~ to do such a scan interactively.
+This should be only rarely needed.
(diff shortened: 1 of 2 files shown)


bdarcus:
Thanks! 

I just changed the commit message to conform to [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/)s.

PR #860, opened 2025-02-19
say local bibliographies override global ones and consequence on file

rdiaz02:


bdarcus:
Sorry; missed or forgot about this earlier.

Issue #784, opened 2023-06-14
Discrepancies in citation format when using `citar-insert-reference`


benthamite:
Apologies if I'm missing something obvious, but I notice that when I insert a formatted reference with `citar-insert-reference`, there are various discrepancies between the inserted reference and the same reference as it appears when exported with one of the `org-mode` export commands, such as `org-md-export-to-markdown`. Most notably, the titles are not capitalized correctly (e.g. the braces surrounding a word are not respected).

As an example, consider the following bibtex entry:

``` bibtex
@online{Hanson2023CanHumansBe,
  abstract =	 {It is one of the most fundamental questions in the
                  social and human sciences: how culturally plastic are
                  people? Many anthropologists have long championed the
                  view that humans are very plastic; with matching
                  upbringing people can be made to behave a very wide
                  range of ways, and to want a very wide range of
                  things. Others say human nature is far more
                  constrained, and collect descriptions of "human
                  universals" (See Brown's 1991},
  author =	 {Hanson, Robin},
  langid =	 {english},
  timestamp =	 {2023-06-14 15:12:51 (GMT)},
  title =	 {Can humans be the {FORTRAN} of creatures?},
  url =
                  {https://www.overcomingbias.com/p/how-plastic-are-peoplehtml},
  urldate =	 {2023-06-14},
}
```

Inserting this reference by invoking `citar-insert-reference` results in 

> [1]R. Hanson, “Can humans be the fortran of creatures?” https://www.overcomingbias.com/p/how-plastic-are-peoplehtml (accessed Jun. 14, 2023).

Whereas exporting a file that cites that work via `org-md-export-to-markdown` will show it in the "bibliography" section as

> R. Hanson, “Can humans be the FORTRAN of creatures?” https://www.overcomingbias.com/p/how-plastic-are-peoplehtml (accessed Jun. 14, 2023).

I have used the IEEE csl citation style in this case, but the issue occurs with all the styles I tried.

bdarcus:
Are you using the default formatter, or the [citeproc-el one](https://github.com/emacs-citar/citar/blob/main/citar-citeproc.el)?

I realize it's not documented in the README (PR welcome), but I suspect that's it; either you aren't using the citeproc formatter, or you're using a different style for that?

benthamite:
Thanks for the quick reply.

[...]

PR #865, opened 2025-07-01
feat: Add option to preserve TeX markup in cache

benthamite:
This pull request attempts to address the problem reported in #784. **NB: the PR is not ready to be merged.** Please read below for a description of the required upstream changes.

The PR introduces a new user option, `citar-cache-preserve-markup`, to control whether TeX markup is preserved in bibliography fields when populating the Citar cache. This helps address issues with incorrect capitalization in formatted references, particularly when using `citeproc-el`.

#### Problem

When using `citar-insert-reference` with `citeproc-el`, titles and other fields may not be capitalized correctly. This is because `citar`'s caching mechanism, which uses `parsebib-parse`, strips TeX markup by default. For example, a title like `{Can humans be the {FORTRAN} of creatures?}` becomes `"Can humans be the fortran of creatures?"`, losing the intended capitalization of "FORTRAN".

#### Solution

This change introduces a new boolean custom variable, `citar-cache-preserve-markup`.

-   When `nil` (the default), `citar` maintains the existing behavior of stripping TeX markup.
-   When `t`, `citar` preserves TeX markup by passing `:display nil` to `parsebib-parse`.

The cache invalidation logic has also been updated. The cache will now be rebuilt if the value of `citar-cache-preserve-markup` changes, ensuring that the cached entries reflect the user's current setting.

#### Known Issues

Preserving markup across all fields can have side effects on downstream processors like `citeproc-el`, which may not expect markup in certain fields (e.g., `date` or `author`). This can lead to parsing errors or incorrect formatting for those fields.

This change provides a way to solve the capitalization issue, but a full solution requires improvements to the parsers in `citeproc-el`. Since I don’t understand the inner workings of this other package, I am not able to implement those changes myself.

### DIFF
--- a/citar-cache.el
+++ b/citar-cache.el
@@ -21,6 +21,7 @@
 (declare-function citar--prepend-candidate-citekey "citar" (citekey candidate))
 
 (defvar citar-ellipsis)
+(defvar citar-cache-preserve-markup)
 
 ;;; Variables:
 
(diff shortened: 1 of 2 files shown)


bdarcus:
I'll take a look at it soon, but note that the CI check is failing currently, and you can convert this to a draft to indicate it's "not ready" status.

PR #869, opened 2025-10-24
chore: (if|when)-let macros obsolete; use -let* counterparts

ashton314:
Emacs 31 makes `if-let` and `when-let` macros obsolete. I'm getting warning messages whenever I boot up Emacs about this. This commit switches `if-let` → `if-let*` and `when-let` → `when-let*`. The warning goes away with this change.

Also, fixed a typo in CONTRIBUTING.org.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -8,7 +8,7 @@
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 ;; Version: 1.4.0
 ;; Homepage: https://github.com/emacs-citar/citar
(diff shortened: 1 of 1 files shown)


roshanshariff:
I can't check right now, but does depending on the compat package help? I would hope it defines those starred macros for older Emacs versions as well.

ashton314:
@bdarcus Added `compat`. I'm not sure what version to use—I'm just following what minad has in his vertico package.

Issue #868, opened 2025-10-23
[FR] Omit out certain BibTeX fields when exporting/inserting BibTeX entries


arifer612:
## Problem Statement

When exporting/inserting BibTeX entries from my main bibliography file to another file (for example a `references.bib` file in a working project), I want the ability to not export all BibTeX fields of the selected entries. For example, suppose I have a BibTeX entry that looks like:
```bibtex
@book{2025-smith-hello-world,
  author = {John Smith},
  title = {Hello World},
  DATE_ADDED = {Sat May 10 12:38:02 2025},
  file = {/home/user/documents/2025-smith-hello-world},
  notes = {John Smith is a weird person},
  keywords = {big words},
  copyright = {CC BY-NC-ND},
  journal = {Journal of Hello Worlds},
  volume = 1,
  number = 1,
  page = {1 -- 100},
}
```

I want to export this to a working project and not have to include the `keywords`, `copyright`, `DATE_ADDED`, `file`, and `notes` fields as 
1. they will not be parsed by LaTeX, 
2. they are not relevant to the project files, and
3. if I am collaborating with others, I really I do not want my collaborators to see what funny notes I may have written.

The current situation is that whenever I use `citar-insert-bibtex`, I would have to check every newly inserted entry to see if it contains these unnecessary fields and delete them if present -- this is a chore that gets very repetitive and laborious the larger the exported bibliography becomes. However, the bright side is that this can be easily automated. In fact, this feature already exists in the `bibtex-completion` package.

## Proposed solution

What I propose is to mirror the functionality that exists in `bibtex-completion` to solve the issue. In particular, a custom variable `bibtex-completion-no-export-fields` that contains a list of BibTeX fields that will be ignored when exporting a BibTeX entry.

## Working solution

I have implemented this in [a branch of my fork](https://github.com/arifer612/citar/tree/feat/no-export-fields), and it works exactly as intended. For the given example above, what needs to be done is to simply include the following snippet in the dotemacs (if using `use-package`):
```emacs-lisp
(use-package citar
  ...
  :custom
  (citar-no-export-fields '("file" "DATE_ADDED" "copyright" "keywords" "notes"))
  ...
  )
```

I would like to know whether this feature request would be relevant to other `citar` users. I would also like to know if the section where I have placed the new custom variable is appropriate.

If all is good and there are users who would find this useful, I intend to submit this feature branch as a pull request.

bdarcus:
[...]

PR #872, opened 2025-11-06
feat: New citar-no-export-fields variable

arifer612:
Refer to #868

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -1681,6 +1681,12 @@ including the citekeys, is maintained in Zotero with Better BibTeX."
             (dolist (bib-file bibtex-files)
               (insert-file-contents bib-file))
             (bibtex-search-entry citekey)
+            (dolist (field citar-bibtex-no-export-fields)
+              (let ((position (bibtex-search-forward-field
+                               field t)))
+                (when position
+                  (delete-region (caar position)
+                                 (car (last position))))))
             (let ((beg (bibtex-beginning-of-entry))
                   (end (bibtex-end-of-entry)))
               (buffer-substring-no-properties beg end)))))


arifer612:
The relevant error messages are:
```
citar.el:1691:25: Error: the function ‘bibtex-beginning-of-entry’ is not known to be defined.
citar.el:1686:32: Error: the function ‘bibtex-search-forward-field’ is not known to be defined.
citar.el:1681:14: Error: the function ‘bibtex-set-dialect’ is not known to be defined.
While building ‘citar-citeproc.elc’: Failed to byte-compile ‘citar.el’
```

The issue seems to lie in the fact that `citar--insert-bibtex` uses the above three functions from `bibtex.el`. An easy fix should be to place a `(require 'bibtex)` in the preamble of `citar.el`; I will check if that's the correct solution.

I have also noticed that this issue has also plagued the CI compilation check before (case in point: https://github.com/emacs-citar/citar/actions/runs/18788444033/job/53614083105?pr=869#step:6:19).

arifer612:
Now that #873 has been merged, this PR is ready to be reviewed again.

citar.el
```diff
@@ -345,6 +345,11 @@ always prompt to select."
                (function-item citar-open-notes)
                (function-item citar-open-note))))
 
+(defcustom citar-no-export-fields nil
```
roshanshariff: Could we rename this to `citar-bibtex-no-export-fields` please? Citar supports multiple bibliography formats and I think we should move towards splitting each of them into its own subpackage, which would be `citar-bibtex` in this case. We can do the refactoring later, but we should keep that in mind for customization options so we don't have to rename and alias/obsolete them in the future.

roshanshariff (changes requested):
I just suggested renaming the customization option, but otherwise this looks good to me!

citar.el
arifer612: Not an issue, I've made the change as suggested!

bdarcus: Can you make sure to also align the commit message(s)?

roshanshariff: And also change the commit title and message to match the new variable name.

arifer612: Right, forgot about that. Fixed!

bdarcus:
Thanks again!

PR #879, opened 2026-03-02
Fix dangling braces

### LINKED ISSUE #876: Setting `citar-bibtex-no-export-fields` changes placement of closing curly brackets
**Describe the bug**
Setting `citar-bibtex-no-export-fields` changes the placement of the closing curly brackets

**To Reproduce**
With the following `init.el` (`citar-bibtex-no-export-fields` customized):
```
(require 'package)
(setq package-archives '(("elpa" . "https://elpa.gnu.org/packages/")
                         ("melpa" . "https://melpa.org/packages/")
                         ("org" . "https://orgmode.org/elpa/")))
(package-initialize)
(unless package-archive-contents (package-refresh-contents))

(require 'use-package-ensure)
(setopt use-package-always-ensure t)

(use-package citar
  :after tex
  :custom
  (citar-bibliography '("~/zotero/all.bib")))

(use-package tex
  :ensure auctex
  :mode
  ("\\.tex\\'" . latex-mode)
  :custom
  (citar-bibtex-no-export-fields '("abstract" "copyright" "file" "keywords" "langid" "urldate")))
```

`citar-bibtex-inserts` inserts (closing `}` in the `isbn` line):
```
@incollection{asdf,
  title = {lorem ipsum},
  shorttitle = {dolor},
  booktitle = {sit},
  author = {amet, consectetur},
  year = 2026,
  pages = {42--420},
  publisher = {adipiscing},
  address = {elit},
  doi = {42/420},
  isbn = {42/420}}
```

With the following `init.el` (`citar-bibtex-no-export-fields` not customized):
```
(require 'package)
(setq package-archives '(("elpa" . "https://elpa.gnu.org/packages/")
                         ("melpa" . "https://melpa.org/packages/")
                         ("org" . "https://orgmode.org/elpa/")))
(package-initialize)
(unless package-archive-contents (package-refresh-contents))

(require 'use-package-ensure)
(setopt use-package-always-ensure t)

(use-package citar
  :after tex
  :custom
  (citar-bibliography '("~/zotero/all.bib")))

(use-package tex
  :ensure auctex
  :mode
  ("\\.tex\\'" . latex-mode))
```

`citar-bibtex-inserts` inserts (closing `}` in its own line):
```
@incollection{asdf,
  title = {lorem ipsum},
  shorttitle = {dolor},
  booktitle = {sit},
  author = {amet, consectetur},
  year = 2026,
  pages = {42--420},
  publisher = {adipiscing},
  address = {elit},
  doi = {42/420},
  isbn = {42/420}
}
```

aikrahguzar:
Closes #876 

Also a performance improvement in the exporter: we insert the bibliographies into a buffer once instead of once for each citekey.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -1664,43 +1664,47 @@ including the citekeys, is maintained in Zotero with Better BibTeX."
    (concat "zotero://select/items/@" citekey)))
 
 ;;;###autoload
-(defun citar-insert-bibtex (citekeys)
-  "Insert bibliographic entry associated with the CITEKEYS."
+(defun citar-insert-bibtex (citekeys &optional bib-files)
+  "Insert bibliographic entries associated with the CITEKEYS in current buffer.
+Entries are searched among BIB-FILES. By default BIBFILES includes global
+bib files as well as bib files local to the current document."
   (interactive (list (citar-select-refs)))
-  (let ((bibtex-files
-         (citar--bibliography-files))
+  (let ((bib-files
+         (or bib-files (citar--bibliography-files)))
         (buffer (current-buffer)))
     (with-temp-buffer
       (bibtex-set-dialect)
-      (dolist (bib-file bibtex-files)
+      (dolist (bib-file bib-files)
         (insert-file-contents bib-file))
       (dolist (citekey citekeys)
-        (bibtex-search-entry citekey)
-        (dolist (field citar-bibtex-no-export-fields)
-          (let ((position (bibtex-search-forward-field
-                           field t)))
-            (when position
-              (delete-region (caar position) (nth 2 position)))))
-        (let ((beg (bibtex-beginning-of-entry))
-              (end (bibtex-end-of-entry)))
-          (unless (eq beg end)
-            (goto-char end)
-            (insert "\n\n")
-            (insert-into-buffer buffer beg (point))))))))
+        (when (bibtex-search-entry citekey)
+          (dolist (field citar-bibtex-no-export-fields)
+            (let ((position (bibtex-search-forward-field
+                             field t)))
+              (when position
+                (delete-region (caar position) (nth 2 position)))))
+          (let ((beg (bibtex-beginning-of-entry))
+                (end (bibtex-end-of-entry)))
+            (unless (eq beg end)
+              (goto-char end)
+              (insert "\n\n")
+              (insert-into-buffer buffer beg (point)))))))))
 
 ;;;###autoload
 (defun citar-export-local-bib-file ()
   "Create a new bibliography file from citations in current buffer.
 
 The file is titled \"local-bib\", given the same extension as
-the first entry in `citar-bibliography', and created in the same
-directory as current buffer."
+the first entry in `citar-bibliography', and created in the
+`default-directory'."
   (interactive)
   (let* ((citekeys (citar--major-mode-function 'list-keys #'ignore))
-         (ext (file-name-extension (car citar-bibliography)))
-         (file (format "%slocal-bib.%s" (file-name-directory buffer-file-name) ext)))
+         (bib-files (citar--bibliography-files))
+         (ext (file-name-extension (or (car-safe citar-bibliography)
+                                       citar-bibliography)))
+         (file (expand-file-name (format "local-bib.%s" ext))))
     (with-temp-file file
-      (citar-insert-bibtex citekeys))))
+      (citar-insert-bibtex citekeys bib-files))))
 
 ;;;###autoload
 (defun citar-insert-citation (citekeys &optional arg)
--- a/citar.el
+++ b/citar.el
@@ -1686,7 +1686,7 @@ including the citekeys, is maintained in Zotero with Better BibTeX."
                                field t)))
                 (when position
                   (delete-region (caar position)
-                                 (car (last position))))))
+                                 (nth 2 position)))))
             (let ((beg (bibtex-beginning-of-entry))
                   (end (bibtex-end-of-entry)))
               (buffer-substring-no-properties beg end)))))
--- a/citar.el
+++ b/citar.el
@@ -1668,30 +1668,26 @@ including the citekeys, is maintained in Zotero with Better BibTeX."
 (defun citar-insert-bibtex (citekeys)
   "Insert bibliographic entry associated with the CITEKEYS."
   (interactive (list (citar-select-refs)))
-  (dolist (citekey citekeys)
-    (citar--insert-bibtex citekey)))
-
-(defun citar--insert-bibtex (citekey)
-  "Insert the bibtex entry for CITEKEY at point."
-  (let* ((bibtex-files
-          (citar--bibliography-files))
-         (entry
-          (with-temp-buffer
-            (bibtex-set-dialect)
-            (dolist (bib-file bibtex-files)
-              (insert-file-contents bib-file))
-            (bibtex-search-entry citekey)
-            (dolist (field citar-bibtex-no-export-fields)
-              (let ((position (bibtex-search-forward-field
-                               field t)))
-                (when position
-                  (delete-region (caar position)
-                                 (nth 2 position)))))
-            (let ((beg (bibtex-beginning-of-entry))
-                  (end (bibtex-end-of-entry)))
-              (buffer-substring-no-properties beg end)))))
-    (unless (equal entry "")
-      (insert entry "\n\n"))))
+  (let ((bibtex-files
+         (citar--bibliography-files))
+        (buffer (current-buffer)))
+    (with-temp-buffer
+      (bibtex-set-dialect)
+      (dolist (bib-file bibtex-files)
+        (insert-file-contents bib-file))
+      (dolist (citekey citekeys)
+        (bibtex-search-entry citekey)
+        (dolist (field citar-bibtex-no-export-fields)
+          (let ((position (bibtex-search-forward-field
+                           field t)))
+            (when position
+              (delete-region (caar position) (nth 2 position)))))
+        (let ((beg (bibtex-beginning-of-entry))
+              (end (bibtex-end-of-entry)))
+          (unless (eq beg end)
+            (goto-char end)
+            (insert "\n\n")
+            (insert-into-buffer buffer beg (point))))))))
 
 ;;;###autoload
 (defun citar-export-local-bib-file ()
@@ -1705,8 +1701,7 @@ directory as current buffer."
          (ext (file-name-extension (car citar-bibliography)))
          (file (format "%slocal-bib.%s" (file-name-directory buffer-file-name) ext)))
     (with-temp-file file
-      (dolist (citekey citekeys)
-        (citar--insert-bibtex citekey)))))
+      (citar-insert-bibtex citekeys))))
 
 ;;;###autoload
 (defun citar-insert-citation (citekeys &optional arg)
--- a/citar.el
+++ b/citar.el
@@ -87,6 +87,13 @@ buffer.")
   :group 'citar
   :type '(repeat file))
 
+(defcustom citar-exported-bib-file-name "local-bib"
+  "The file name (without extension) used by `citar-export-local-bib-file'.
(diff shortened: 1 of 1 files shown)


roshanshariff (approved):
LGTM!

havarddj:
This is a really great feature, I'm looking forward to it being merged! I have one request though: could you make the file name an optional argument to `citar-export-local-bib-file`, and fall back to the usual "local-bib.(ext)" if none is provided? Or alternatively trigger a `completing-read` if called interactively.

My use case is this: I have a global .bib file which contains thousands of .bib keys (exported automatically via Zotero), but when I'm writing papers with collaborators, it's convenient to have a smaller bib file local to that repository (so my collaborators can more easily edit it). But my collaborators usually expect the bib file to be named "references.bib".

bdarcus:
> I wouldn't worry too much about tests failing on the snapshot version of Emacs. Those tests are meant as a heads-up of breaking changes in Emacs master we need to be aware of. But they do fail inexplicably sometimes, and that should only be treated as a blocker after Emacs master has been branched for the next release.

Maybe we shouldn't run the snapshot with warnings as errors? 

Either way, I agree it's not worth holding up merging. 

Though consider the suggestion from @havarddj?

aikrahguzar:
> This is a really great feature, I'm looking forward to it being merged! I have one request though: could you make the file name an optional argument to `citar-export-local-bib-file`, and fall back to the usual "local-bib.(ext)" if none is provided? Or alternatively trigger a `completing-read` if called interactively.

This makes sense. Hardcoded name is not ideal. Optional arg is straightforward so we should add that and I think interactively we can prompt with a prefix arg.

Beyond that should the default be made customizable? I think that makes sense but I don't use this feature so not sure. But that seem to be the best fit with @havarddj's workflow.

havarddj:
Hi, sorry for the delay. I think having a customizable variable would be very helpful for my use-case, perhaps something like `citar-export-local-bib-file-name`.

aikrahguzar:
@havarddj I pushed a new commit, please try that.

aikrahguzar:
@roshanshariff @bdarcus tests are now failing because Embark can't be installed since it has started requiring Emacs 29. What should we do about that?

bdarcus:
> @roshanshariff @bdarcus tests are now failing because Embark can't be installed since it has started requiring Emacs 29. What should we do about that?

Maybe we should follow Embark (and Vertico)? 

The alternative is changing the CI.

havarddj:
> @havarddj I pushed a new commit, please try that.

Hey @aikrahguzar! 

I couldn't quite get it working, it doesn't seem to export all the references in the document. The smallest working example I could make was to add a file `test.tex` in the `test/` directory, with the following contents:
```latex
\documentclass[11pt]{amsart}

\usepackage[
   backend=bibtex,
   style=alphabetic,
   natbib=true,
   url=false, 
   doi=true,
   eprint=false,
   backref=true
   ]{biblatex}
\addbibresource{test.bib}

\begin{document}
This is citation. \cite{downes-74}
This is another citation. \cite{mann-68}

\printbibliography
\end{document}
```
This compiles correctly (at least on my machine).
Running `M-x citar-export-local-bib-file` then creates a file `local-bib.bib`, but it's empty. In a more complex project, it would sometimes only add a couple of the bib entries, but (usually) not all. So perhaps there's a problem with `citar-insert-bibtex`? 

I loaded the local version of `citar` by cloning the PR and running `package-install-from-buffer` inside `citar.el` (and I've confirmed the version of `citar-export-local-bib-file` is correct, since it has `file` as an optional argument). I also ran `M-: (format "%s" (citar--major-mode-function 'list-keys #'ignore))`, which returns `"(mann-68 downes-74)"`. Unrelatedly, it would be nice if this list was sorted alphabetically, so the bib keys in the local bibliography are sorted as well!

aikrahguzar:
> Hey @aikrahguzar!
> 
> I couldn't quite get it working, it doesn't seem to export all the references in the document. The smallest working example I could make was to add a file `test.tex` in the `test/` directory, with the following contents:
> 
> ```
> \documentclass[11pt]{amsart}
> 
> \usepackage[
>    backend=bibtex,
>    style=alphabetic,
>    natbib=true,
>    url=false, 
>    doi=true,
>    eprint=false,
>    backref=true
>    ]{biblatex}
> \addbibresource{test.bib}
> 
> \begin{document}
> This is citation. \cite{downes-74}
> This is another citation. \cite{mann-68}
> 
> \printbibliography
> \end{document}
> ```
> 
> This compiles correctly (at least on my machine). Running `M-x citar-export-local-bib-file` then creates a file `local-bib.bib`, but it's empty. In a more complex project, it would sometimes only add a couple of the bib entries, but (usually) not all. So perhaps there's a problem with `citar-insert-bibtex`?

I can't reproduce this. I see both entries in the `local-bib.bib`. Make sure that you call `M-x reftex-parse-all` before the export. Otherwise, `reftex` wouldn't report the `test.bib` file.

> Unrelatedly, it would be nice if this list was sorted alphabetically, so the bib keys in the local bibliography are sorted as well!

Sorting, makes sense.

havarddj:
Thanks, running `reftex-parse-all` does fix it for me! Works like a charm now.

havarddj:
Out of curiosity, is there a reason why `citar-insert-bibtex` uses `bibtex-search-entry` under the hood but `citar-open-entry-in-file` uses `parsebib`? I was a little confused why it wouldn't export entries with `@online`, but that seems to be because `bibtex.el` is strict about what it considers an entry, and why for the other function parsebib is used: `https://github.com/joostkremers/parsebib/discussions/23`.

aikrahguzar:
I have now made the failure to find an entry loud: a message will be emitted mentioning missing the entry along with the bib files that were searched. Am empty list of bibliographies will be an error.

Sorting is also implemented now.

havarddj:
Awesome, thank you very much!

bigodel:
Hi, coming over from #885, where @aikrahguzar pointed to this PR due to some overlap.

For context: #885 adds a configurable name for the file produced by `citar-export-local-bib-file` via a new `defcustom` that accepts `string`, `function` or `nil`, plus a couple of helpers (`citar-local-bibtex-base-name` deriving a basename from the buffer's bibliography declaration, `citar--local-bibliography-files` extracting per-buffer local bib lookup), a function rename for clarity, and tests. Significant overlap with the configurability piece of this PR.

My suggestion is to consider splitting scope:

- **#879**: keep focused on the dangling-braces bug (its primary motivation, ready to land).
- **#885**: track the configurable-local-export-bib work (`defcustom` shape, helpers, tests, naming).

The brace fix could merge cleanly without being held up on UX/API discussion, and the larger conversation moves to a smaller follow-up PR.

If you'd rather keep everything together, happy to close #885 and move the pieces here as suggested changes. See [this comment](https://github.com/emacs-citar/citar/pull/885#issuecomment-4336789898) with some changes from #885 I'd like to see retained.

aikrahguzar:
@bigodel, how much work would it be to rebase #885 on top of this? I can merge this and then we can move to #885. The problem with this splitting this is that is also fixes the issues raised in https://github.com/emacs-citar/citar/issues/834 so it is really the title of the pr which is inaccurate. It has become a reworking of local bib export.

bigodel:
> @bigodel, how much work would it be to rebase #885 on top of this? I can merge this and then we can move to #885. 

I'm happy to rebase #885 on top of this once it merges.

One small ask: if a release is being cut on the back of this PR, would you consider also folding in the suggested renames, `citar-export-local-bib-file` -> `citar-export-local-bibtex-file` and `citar-exported-bib-file-name` -> `citar-exported-bibtex-base-name`. The new name reflects the fact that the exported file will be in BibTeX format (regardless of the extension) and that the variable's file name will only be used for its base name, as the extension is derived.

If a release isn't imminent, no urgency, I'm happy carrying the rename in #885 once I rebase. Either way works.

aikrahguzar:
@bigodel I have done the suggested renames, please try. @havarddj heads up about the renames. I will merge this pr soon (sorry for the long delay), unless there are objections.

aikrahguzar:
I was using a calling convention for `sort` that is new `Eamcs 30.2` and that caused a test failure on `Emacs 29.2`. I have fixed this. Please test if everything still works.

citar.el
```diff
@@ -1665,48 +1671,59 @@ including the citekeys, is maintained in Zotero with Better BibTeX."
...
+                (with-current-buffer target
+                  (insert-buffer-substring-no-properties source beg end)
+                  (insert "\n\n"))))
+          (message "Citkey %s not found among %S" citekey bibfiles))))))
 
 ;;;###autoload
-(defun citar-export-local-bib-file ()
-  "Create a new bibliography file from citations in current buffer.
+(defun citar-export-local-bibtex-file (&optional file)
+  "Create a new bibliography FILE file from citations in current buffer.
```
roshanshariff:
```suggestion
  "Create a new bibliography file from citations in current buffer.
```

citar.el
```diff
@@ -1665,48 +1671,59 @@ including the citekeys, is maintained in Zotero with Better BibTeX."
...
 
-The file is titled \"local-bib\", given the same extension as
-the first entry in `citar-bibliography', and created in the same
-directory as current buffer."
+By default, FILE is created in `default-directory' based on the value of
+`citar-exported-bibtex-base-name' with extension determined by the
+bibliographies of current buffer. If `citar-exported-bibtex-base-name' is
+nil or extension can't be determined, user is prompted for a filename."
```
roshanshariff:
```suggestion
The new bibliography will be named FILE, which if specified must
be in an existing directory.  If FILE is nil, then the bibliography
will be created in `default-directory' with base name
`citar-exported-bibtex-base-name' and the extension of the first
bibliography associated with the current buffer.  If
`citar-exported-bibtex-base-name' is nil or extension can't be
determined, prompt the user for a filename."
```

citar.el
```diff
@@ -1665,48 +1671,59 @@ including the citekeys, is maintained in Zotero with Better BibTeX."
...
   (interactive)
-  (let* ((citekeys (citar--major-mode-function 'list-keys #'ignore))
-         (ext (file-name-extension (car citar-bibliography)))
-         (file (format "%slocal-bib.%s" (file-name-directory buffer-file-name) ext)))
+  (let* ((citekeys (sort (citar--major-mode-function 'list-keys #'ignore)
+                         #'string-lessp))
+         (bibfiles (citar--bibliography-files))
+         (file (expand-file-name
+                (or file
+                    (if-let* ((citar-exported-bibtex-base-name)
```
roshanshariff, line 1720:
```suggestion
                    (if-let* ((_ citar-exported-bibtex-base-name)
```
I believe it's recommended for clarity to use the `_` "ignored binding" instead `if-let* ((VARNAME))`.

citar.el
```diff
@@ -1572,7 +1578,7 @@ specifying TYPE."
 
 ;;;###autoload
 (defun citar-attach-files (citekey-or-citekeys)
-  "Attach library file associated with CITEKEY-OR-CITEKEYS to outgoing MIME message."
+  "Attach library file containing CITEKEY-OR-CITEKEYS to outgoing MIME message."
```
roshanshariff: I think "associated with" was more clear than "containing", which implies the file (say a PDF) actually contains a citekey rather than just being associated with it.

roshanshariff (changes requested):
Thanks, LGTM! Just a few small stylistic and doc comments.

citar.el
aikrahguzar: From what I remember, I changed this to placate the `checkdoc` which was complaining about the long documentation line.

aikrahguzar: I guess an alternative is to change `library file` to just `file`. That also brings the docstring to under 80 characters.

roshanshariff: Yeah, just `file` should be fine.

citar.el
roshanshariff: Never mind, I could've sworn I saw that recommendation in the docs somewhere but I can't find it again.

PR #882, opened 2026-03-27
fix: prevent stale cache after failed bibliography parse

fpatz:
When citar-cache--update-bibliography re-parses a bibliography, it previously called (clrhash entries) before (parsebib-parse), destroying the existing valid entries. If parsebib-parse then errored partway through (e.g., because an external tool like Zotero Better BibTeX was mid-write when citar read the file), the entries hash table was left partially filled. Since the bib object was already in the cache from a prior successful run, the corrupted state persisted. And because props were never updated (the line after the failed parse), subsequent cache checks saw "no changes needed", so the stale partial entries were never refreshed—even after the bib file was fully written.

Fix by parsing into a temporary hash table first, and only replacing the cached entries on success. If parsebib-parse errors, the cache retains its previous valid entries.

### DIFF
--- a/citar-cache.el
+++ b/citar-cache.el
@@ -230,17 +230,12 @@ its return value can be reused.
 After updating, the `props' slot of BIB is set to PROPS."
   (let* ((filename (citar-cache--bibliography-filename bib))
          (props (or props (citar-cache--get-bibliography-props filename)))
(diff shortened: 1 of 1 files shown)


roshanshariff:
Merged, thanks!

PR #885, opened 2026-04-28
feat: Customizable basename for citar-export-local-bib-file

bigodel:
## Summary

Makes `citar-export-local-bibtex-file` (renamed from `citar-export-local-bib-file`) flexible:

- Takes a `FILENAME` argument; with prefix arg, prompts the user.
- Without prefix arg, derives the basename from a new user option `citar-export-local-bibtex-base-name` — string, function, or `nil` for a mode-aware default.
- The `nil` choice falls back to a new helper `citar-local-bibtex-base-name`, which derives a basename from the buffer's bibliography declaration (Org `#+bibliography:`, TeX `\bibliography{}`) via the new private `citar--local-bibliography-files`.

## Motivation

The hardcoded `local-bib.<ext>` output is opaque and conflicts with workflows where a project's bibliography file has a meaningful name (e.g., `references.bib`). Letting users choose the basename, or derive it from the buffer's own bibliography declaration, gives users much more flexibility when generating a local BibTeX file from citations.

The rename clarifies that the function writes BibTeX, not a generic "bib" artifact.

## Backwards compatibility

- Default value of the user option (`"local-bib"`) preserves prior behavior.
- Old function name retained via `define-obsolete-function-alias`.

## Public API additions

- `citar-export-local-bibtex-base-name` (defcustom)
- `citar-local-bibtex-base-name` (helper)
- `citar-export-local-bibtex-file` (renamed function)

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -1762,6 +1762,13 @@ buffer's directory in typical usage)."
       (dolist (citekey citekeys)
         (citar--insert-bibtex citekey)))))
 
+(define-obsolete-function-alias
+  'citar-export-local-bib-file
+  'citar-export-local-bibtex-file
+  "1.5"
+  "Renamed for clarity: the function writes BibTeX, not a generic
+\"bib file\".")
+
 ;;;###autoload
 (defun citar-insert-citation (citekeys &optional arg)
   "Insert citation for the CITEKEYS.
--- a/citar.el
+++ b/citar.el
@@ -350,6 +350,24 @@ always prompt to select."
   :group 'citar
   :type '(repeat string))
 
+(defcustom citar-export-local-bibtex-base-name "local-bib"
+  "Basename (without extension) for `citar-export-local-bibtex-file'.
+
+The value may be:
+
+  - a string — used as the literal basename;
+  - a function — called with no arguments, expected to return a
+    basename string;
+  - nil — defer to `citar-local-bibtex-base-name'.
+
+The file extension is taken from the first entry of
+`citar-bibliography'."
+  :type '(choice
+          (const    :tag "Mode-aware default (`citar-local-bibtex-base-name')" nil)
+          (string   :tag "Literal basename")
+          (function :tag "Function returning basename"))
+  :group 'citar)
+
 ;;;; File, note, and URL handling
 
 (defcustom citar-open-resources '(:files :links :notes :create-notes)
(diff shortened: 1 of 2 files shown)


aikrahguzar:
This is part of https://github.com/emacs-citar/citar/pull/879 (title doesn't make sense anymore), please see the discussion there and review the changes.

bigodel:
> This is part of #879 (title doesn't make sense anymore), please see the discussion there and review the changes.

Thanks for the pointer to #879, I'd dismissed it because of the title.

Agreed it has significant overlap and should land first. There are a few things from this PR I'd still like to retain in some form:

- Extending the defcustom type to accept a function (called with no args, returns a basename). On `nil`, prompt the user, but the prompt should live in the `interactive` declaration, with `filename` as a required argument so programmatic callers always pass it explicitly. A prefix argument could also force a prompt as an ad-hoc override.

- Rename `citar-exported-bib-file-name` -> `citar-exported-bibtex-base-name` (emphasizing that the extension does not matter and will be dervied), and `citar-export-local-bib-file` -> `citar-export-local-bibtex-file`. Without the explicit "bibtex" the names read like they could also produce CSL bibliographies.

- The `citar-local-bibtex-base-name` helper that derives a basename from the buffer's bibliography declaration (Org `#+bibliography:`, TeX `\bibliography{}`). Plug-and-play as a function value of the `defcustom`.

- The extracted `citar--local-bibliography-files` helper for per-buffer local bib lookup, especially if we agree on keeping `citar-local-bibtex-base-name`.

- Tests covering the option dispatch and the helpers.

citar.el
```diff
@@ -1693,21 +1721,54 @@ including the citekeys, is maintained in Zotero with Better BibTeX."
...
+
+The file extension is taken from the first entry of
+`citar-bibliography'; if FILENAME already carries an extension, it is
+replaced.  The file is created in `default-directory' (the current
+buffer's directory in typical usage)."
+  (interactive
+   (list
+    (cond
+      (current-prefix-arg
+       (read-file-name "File name: "))
```
bigodel, line 1752: My suggestion for prompting on `citar-exported-bibtex-file-name` being `nil`.

```suggestion
      ((or current-prefix-arg (not citar-exported-bibtex-file-name))
       (read-file-name "File name: " nil "some-sensible-default-or-error"))
```

Issue #765, opened 2023-03-25
Add error handling for citar-bibliography


bdarcus:
Need to check the file(s) exist before trying to load them.

```elisp
(defun citar-file--files-exist-p (files)
  "Check each of a list of FILES exists."
  (dolist (file files)
     (unless (file-exists-p file)
       (user-error "Could not find file: %s" file))))
```

PR #886, opened 2026-04-28
refactor: Skip missing per-buffer local bibliography files

bigodel:
## Summary

`citar--bibliography-files` now treats missing per-buffer local bibliography files as a soft case (silently filtered) instead of an error. Globals (`citar-bibliography`) retain the previous strict existence check.

## Motivation

The pre-flight existence check (introduced in #765) conflates two cases:

- **Global config**: paths in `citar-bibliography` are user-configured up front. A missing global typically indicates a typo and should fail fast.
- **Per-buffer locals**: paths declared via mechanisms like Org `#+bibliography:` or TeX `\bibliography{}`. These may legitimately point to files that haven't been written yet — e.g. a project-local bib materialized after writing citations via `citar-export-local-bibtex-file`.

Treating both equally meant inserting a citation into an Org buffer with `#+bibliography: project.bib` errored out before `project.bib` existed, blocking the natural pre-export workflow.

## Changes

- New helpers `citar--global-bibliography-files` (validates) and `citar--local-bibliography-files` (does not validate).
- Rewrite `citar--bibliography-files` to use them.
- Remove `citar-file--normalize-paths` and `citar-file--files-exist-p` (no longer used).
- Tests for the four cases (global only, missing global errors, missing local filtered, both included).

## Backwards compatibility

- Globals still validated — config typos still caught.
- Locals previously errored on missing; now silently filtered.

### DIFF
--- a/citar-file.el
+++ b/citar-file.el
@@ -84,23 +84,6 @@ separator that does not otherwise occur in citation keys."
 (defvar citar-library-file-extensions)
 (defvar citar-note-format-function)
 
-;;;; Convenience functions for files and paths
-
-(defun citar-file--normalize-paths (file-paths)
-  "Return a list of FILE-PATHS normalized with truename."
-  ;; REVIEW why hassle with this; just require a list?
-  (let ((paths (if (stringp file-paths)
-                   (list file-paths)
-                 file-paths)))
-  (citar-file--files-exist-p paths)
-  (delete-dups (mapcar #'file-truename paths))))
-
-(defun citar-file--files-exist-p (files)
-  "Check each of a list of FILES exists."
-  (dolist (file files)
-     (unless (file-exists-p file)
-       (user-error "Cannot find file: %s" file))))
-
 ;;;; Parsing file fields
 
 (defun citar-file--parser-default (file-field)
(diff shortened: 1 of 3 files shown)


aikrahguzar:
I am not sure this is a good idea. The missing files might be due to a typo etc. Silently ignoring the unavailability seems problematic. I think the problem this is trying to solve can be worked around by just commenting out the line for the bibliography.

bigodel:
> I am not sure this is a good idea. The missing files might be due to a typo etc. Silently ignoring the unavailability seems problematic.

Fair point on typo masking. I based the idea of ignoring on LaTeX's compilation process, which will produce an output even if the bibliography file does not exist. But to be fair to my initial idea, even LaTeX's compilation produces warnings and errors in stderr and log files.

Would you consider this change if I added a `display-warning` for missing files?

> I think the problem this is trying to solve can be worked around by just commenting out the line for the bibliography.

The comment-out workaround misses the use case driving this PR (a follow-up to #885). I want `citar-exported-bib-file-name` set to a function that reads the buffer's declared local bibliography (e.g. `#+bibliography: ./project.bib`) and uses that path as the export target, the buffer becomes the single source of truth for both lookup and write.

Issue #802, opened 2023-09-25
Custom `C-i` bindings do not play well with `citar-open` and `RET`


dschrempf:
**Describe the bug**
In `citar-open`, pressing `RET` results in an error `<C-i> is undefined`.

**To Reproduce**
1. Set
```elisp
  (define-key input-decode-map (kbd "C-i") [C-i])
```
2. Bind `<C-i>`, for example using `general`:
```elisp
  (general-mmap "<C-i>" #'better-jumper-jump-forward)
```
3. Run `citar-open` and press `RET`.

**Expected behavior**
The entry is opened.

**Screenshots**

**Emacs version:**
29.1

**Context**
- Most likely related to https://github.com/emacs-citar/citar/issues/707.
- Setting `(setq citar--multiple-setup (cons "<tab>"  "RET")` is a workaround (but in my opinion it should not be necessary).
- Could be a problem of my setup (I try to distinguish between `C-i` and `<tab>`

bdarcus:
Just in case, OS and (if relevant) distro?

dschrempf:
Thanks for having a look!

+ nix-shell -p nix-info --run "nix-info -m"
 - system: `"x86_64-linux"`
 - host os: `Linux 6.1.54, NixOS, 23.11 (Tapir), 23.11.20230919.5ba549e`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.17.0`
 - nixpkgs: `/nix/store/cz973qsjldbw4x7fx0rwcvhr9k645vz2-source`
 
 I use XMonad.

bdarcus:
When you get a chance,  @aikrahguzar, any input?

aikrahguzar:
This is exactly #707 and I don't know of any completely clean solution for that since it is a workaround for a limitation of completion ui. Changing the default value of `citar--multiple-setup` to use `<tab>` is the only option I can think of but I don't know if there are any unintended consequences.

bdarcus:
> I don't know of any completely clean solution for that since it is a workaround for a limitation of completion ui. Changing the default value of `citar--multiple-setup` to use `<tab>` is the only option I can think of but I don't know if there are any unintended consequences.

@dschrempf - thoughts?

Doesn't seem like there's much we can do about this?

dschrempf:
Using the workaround is an option for me, but I do not know if this situation is desired.

[...]

PR #888, opened 2026-08-03
fix: break on configs that separate `C-i` from `TAB`

anuvyklack:
Relate to #802 and [hel#31](https://github.com/helheim-emacs/hel/issues/31)

There is a common trick to distinguish `C-i` from `Tab` on GUI Emacs via `input-decode-map`. This is possible, because Emacs GUI reports physical Tab key as the symbol `tab` and `Ctrl+i` as the character 9. The same for Enter and Esc.

Currently, `citar--multiple-exit` replays the selection key by pushing the raw character `?\t` into `unread-command-events` which is then decoded as `C-i`, which nothing binds:

```
  1. RET on a candidate
       ▼
  2. citar--setup-multiple-keymap bound RET -> citar--multiple-exit
       ▼
  3. (setq unread-command-events (listify-key-sequence (kbd "TAB")))
     ...(kbd "TAB") is the single character 9, not the symbol `tab'
       ▼
  4. Emacs re-reads 9 -> input-decode-map -> [C-i]
       ▼
  5. citar's composed keymap binds TAB; nothing binds C-i
       ▼
  6. "<C-i> is undefined"
```

This patch replaces char `9` with symbol `tab` in `unread-command-events`. On GUI frames, it will be handled correctly and avoid the collision. On TTY, it will be translated back to 9 anyway.

### DIFF
--- a/citar.el
+++ b/citar.el
@@ -752,7 +752,13 @@ is used for exiting the minibuffer during completing read.")
 (defun citar--multiple-exit ()
   "Exit with the currently selected candidates."
   (interactive)
-  (setq unread-command-events (listify-key-sequence (kbd (car citar--multiple-setup)))))
+  (setq unread-command-events
+        (mapcar (lambda (event)
+                  (alist-get event '((?\t . tab)
+                                     (?\r . return)
+                                     (?\e . escape))
+                             event))
+                (listify-key-sequence (kbd (car citar--multiple-setup))))))
 
 (defun citar--setup-multiple-keymap ()
   "Make a keymap suitable for `citar--select-multiple'."


aikrahguzar:
This just looks like a more involved way of setting `citar--multiple-setup` to `("<tab>" . "<ret>")`. Or am I missing something?

anuvyklack:
This will break citar on a TTY.
  
The Emacs `key` -> `command` translation works as follows:

```
                  event
                    │
                    ▼
             input-decode-map
                    │
                    ▼
            lookup in keymaps ◀──────────┐
             │             │              │
             │ found       │ not found    │ found
             ▼             ▼              │
     run the command    function-key-map ─┘
                           │
                           │ not found
                           ▼
                          fail
```
  
On replay, citar pushes an event onto `unread-command-events`, and Emacs reads it as if it came from the keyboard — so it goes through this whole pipeline again.
  
Our `9` -> symbol `C-i` rule lives in `input-decode-map`.

The Emacs default `tab` -> `9` rule lives in `function-key-map`, which is consulted only after a failed lookup. It applies on GUI frames, where the `Tab` key sends the symbol `tab`. On a TTY the key already sends `9`, so nothing needs translating.

With your proposal, `(setq citar--multiple-setup '("<tab>" . "<ret>"))`, citar binds `tab` instead of `9`:

- On GUI:
  ```
  event: tab
    ▼
  lookup: tab -> match
  ```

- On TTY:
  ```
  event: 9
    ▼
  lookup: 9 -> not found (map binds tab)
    ▼
  function-key-map: no 9 -> tab rule (only tab -> 9)
    ▼
  fail
  ```

Also, `<ret>` is not a key at all:
- `"RET"` -> 13
- `"<return>"` -> symbol `return`

`function-key-map` has no entry for `ret` symbol, it knows `tab`, `return` and `escape`.

PR #889, opened 2026-08-17
feat(citeproc): allow user to customize preferred locale

mbatson:
Hi,

This PR proposes to add a user option `citar-citeproc-csl-preferred-locale`, and replace the hardcoded value of "en-US" passed as arg `LOC` to `citeproc-create` by `citar-citeproc-format-reference` with it.

This allows the user to set a different locale for formatting references with `citar-citeproc` if the selected CSL style does not set its own default. Previously in this situation the user was stuck with en-US localized references.

However, as the default value of `citar-citeproc-csl-preferred-locale` is set to "en-US", the same as the previously hardcoded value, this commit makes no actual change to the behavior of `citar-citeproc-format-reference` if the user makes no customization.

### DIFF
--- a/citar-citeproc.el
+++ b/citar-citeproc.el
@@ -33,6 +33,12 @@
 ;;  CSL style can also be set by calling 'citar-insert-reference' or
 ;;  'citar-copy-reference' with a prefix-argument.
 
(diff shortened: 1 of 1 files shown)


mbatson:
Force pushed an amendment to the commit message.

